PR #21, opened 2015-12-08
Rails

drownedout:
This is the basic rails app.


dcaf_case__management/Gemfile
```diff
@@ -0,0 +1,21 @@
+source 'https://rubygems.org'
+
+gem 'rails', '4.2.1'
+gem 'sqlite3'
+gem 'sass-rails', '~> 5.0'
+gem 'uglifier', '>= 1.3.0'
+gem 'coffee-rails', '~> 4.1.0'
+gem 'jquery-rails'
+gem 'turbolinks'
```
colinxfleming: I'm not a huge fan of turbolinks (possibly because I'm bad at javascript and it breaks js when you're bad at it). I'll make an issue on whether to remove and we can discuss.

dcaf_case__management/Gemfile
```diff
@@ -0,0 +1,21 @@
+source 'https://rubygems.org'
+
+gem 'rails', '4.2.1'
```
colinxfleming: We'll want to bump rails version, though it's not esp urgent. Will make an issue for this.

dcaf_case__management/app/views/case/index.html.erb
```diff
@@ -0,0 +1,97 @@
+<div class="container">
```
colinxfleming: YES. Absolutely love that you started to work in the early designs. Thanks **SO** much for doing that.

colinxfleming:
Documenting the steps I took to test this branch, so we can add it to the readme.

* git checkout -b drownedout-rails master
* git pull https://github.com/drownedout/dcaf_case_management.git rails 
* cd dcaf_case__manager
* bundle install 
* rake db:migrate
* rails s

and some fine looking views were right there in front of me! 

Nice work @drownedout ! Will merge this weekend.

PR #22, opened 2015-12-08
Added new wireframe files, deleted old ones

mebates:


colinxfleming:
Absolutely love the way these are coming along. Thanks @mebates !

Issue #23, opened 2015-12-08
more extensive readme - DCAF 101


mebates:
We need to be able to quickly explain DCAF's workflow to new team members. We'll use the User Flow #5 but need more of a checklist model of things to cover

colinxfleming:
@lwaldsc @NerdyGirl537 if you've got capacity can you tackle a draft of this with me over the next week or so? I might enlist someone else from the CM working group as well.

colinxfleming:
After a discussion with @mebates it seems like a good idea to turn this into a wiki page. I popped it into https://github.com/colinxfleming/dcaf_case_management/wiki/DCAF-101. 

Leaving open for the time being until we have a chance to discuss.

mebates:
Looks like a good start - kind of confusing with all of the abbreviations; maybe we can take a plain language pass at it.

colinxfleming:
@mebates good call on abbreviations and the need for a plain language pass. I'll see if I can take a whack at this in the off week, I think having this primer would really speed up onboarding.

lwaldsc:
Yeah. This was def just a quick outline draft. I can mess around slash flesh it out as well.

PR #30, opened 2015-12-22
Create cm-workflow

lwaldsc:
Create rough explanation of case manager workflow, needs from patients and case managers, and data needed


mebates:
where will this live?

colinxfleming:
@mebates I think either some separate readme or the wiki would work, but I'd honestly leave it up to you. (This was a stab at #23, so wherever you think is a good place for that is cool with me.)

colinxfleming:
@lwaldsc  talked to mollie and we're going to throw this in the wiki. Going to close without merging and put your work in a wiki page. Thanks so much for getting this going.

colinxfleming:
Put in https://github.com/colinxfleming/dcaf_case_management/wiki/DCAF-101 ; closing PR.

PR #32, opened 2016-01-03
Dockerize

colinxfleming:
Adds `Dockerfile` and `docker-compose.yml`, so we can containerize the dependencies and offer a somewhat simpler option. 

That said, I am nobody's idea of a devops person, so this is a pretty approximate stab. `$ docker-compose build && docker-compose up` seems to fire up both containers without issue, allows me to visit the rails app, create and retrieve a User record, etc. (Though I also tried to pop into `rails c` and couldn't, so idk what the issue is there.)

Anyone on the team have some more experience with devops matters than I do and want to review? @Kevin-Wei maybe? It's not urgent so we can let it hang for awhile without issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM rails:4.2.5
+MAINTAINER Colin Fleming <c3flemin@gmail.com> 
+
+# install packages
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev
+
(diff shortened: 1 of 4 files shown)


mchelen:
This works nicely for me, and I confirmed that editing local files resulted in live changes to the server :+1:

colinxfleming:
Thanks a ton for the review, @mchelen ! This had been hanging out for a little while so I'm very glad to close it up.

PR #38, opened 2016-01-12
Mongo Database Models for notes, pledges, cases, users, clinics, patients

drownedout:
Completed Mongo database.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,7 +6,10 @@ class User
          :recoverable, :rememberable, :trackable, :validatable
 
   # Non-devise generated
-  field :name, type: String
+  field :first_name, type: String
+  field :last_name, type: String
+  field :line, type: String
+  field :role, type: String
 
   ## Database authenticatable
   field :email,              type: String, default: ""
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -6,7 +6,10 @@ class User
          :recoverable, :rememberable, :trackable, :validatable
 
   # Non-devise generated
-  field :name, type: String
+  field :first_name, type: String
+  field :last_name, type: String
+  field :line, type: String
```
colinxfleming: re: line as a user attribute, I want to flag that this raises a UX question of how we want to handle DCAF being split into multiple geographies; does this attribute assign a CM to a line, or mark the last line they were on, or something else? 

But that's down the road a ways, so I think this is good for the time being, and mongo's flexible enough that we can change it down the road.

app/models/user.rb
```diff
@@ -6,7 +6,10 @@ class User
...
+  field :role, type: String
```
colinxfleming: Devise and Mongo user models don't play perfectly nice together, it turns out,  and you have to specifically whitelist controller attributes. I'll handle in a separate commit.

colinxfleming:
Looking good! This is fabulous work and it's extremely exciting to be building off of it. 

Thanks much @drownedout and @kevin-wei !

Issue #42, opened 2016-01-18
Data-dismiss on success messages doesn't work


colinxfleming:
![image](https://cloud.githubusercontent.com/assets/3866868/12379235/03ce1c3a-bd24-11e5-9685-9124e9b99591.png)

When you click that X on the right side of the green 'Signed in successfully' box, the message is supposed to dismiss, but doesn't. 

The HTML for this is in `app/views/layouts/_messages.html.erb`; I suspect it's a JS error, maybe the Bootstrap JS library not being loaded.

PR #52, opened 2016-01-26
add bootstrap javascript

RAGarza8:
Add bootstrap javascript into application.js to fix issue with sign in/sign out messages not disappearing when "x" is pressed


colinxfleming:
Looks great. Thanks for tracking this down. 

Fixes #42 .

PR #75, opened 2016-02-09
Patients New (form, controller)

drownedout:

PR #76, opened 2016-02-09
Updating source design files

mebates:
Wireframe .ai and .pdf along with latest .pngs for InVision


colinxfleming:
These are gorgeous. Thanks much @mebates and @adinneen and @lwaldsc  and @NerdyGirl537 ! 

The core layouts for adding a new patient look like they're fairly static so I think we can start getting our frontend stub from #75 to match.

Issue #71, opened 2016-02-07
CONTACT: Add a form for logging contact


colinxfleming:
Create a basic form for creating a contact -- see the call modal at https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983265 . Bonus points for putting it in the modal!

colinxfleming:
Leaving open for now -- partial located in `app/views/calls/_new_call.html.erb`. Could use a bootstrap fixer upper to match the invision design.

colinxfleming:
Resolved in 8512820

Issue #70, opened 2016-02-07
CONTACT: Add a representation of contact attempt to the data model


colinxfleming:
If I'm a case manager and I make a call, regardless of how it turns out, I want to track that -- how do we represent that in mongo? Is it a new object similar to `Note`, or is there something inherent in the `Note` object that we can use? 

My inclination is an object similar to `Note` but could go either way on this. The technical requirement here is that we be able to log timestamps of calls, what came out of them. 

Part of the question might be whether or not it's standard practice for a CM to take a note of 'attempted call, no response' for cases that have been going on. If yes, then probably we turn it into a `Note` and have all call attempts logged as notes for sure -- if not, could go either way.

colinxfleming:
Fixed via #78

Issue #47, opened 2016-01-23
Model: Create some seed data


colinxfleming:
Populate `db/seeds.rb` with some sample data that we can use for experimenting in a development environment. Should include:

* A completed case
* A half-finished case
* A fresh case 
* An abandoned case

Will work on this with @NerdyGirl537 and @lwaldsc to make sure this is consistent with spreadsheet patterns. 

(Gold medal version: FactoryGirl testing fixtures that also match these patterns.)

colinxfleming:
Discussed with @lwaldsc and she won the prize on this -- we'll pair on this to build out some test data.

colinxfleming:
Prelim seed data is constructed in #78 and we should continue to iterate as we incorporate new models into the build. Closing.

PR #78, opened 2016-02-11
Log a call along with its status

ajohnson052:
This creates new model/controller for calls and a modal that allows a user to log a new call.

In order to move ahead with this, I created some simple seed data.

Also, (and perhaps most importantly) notes.rb is renamed to note.rb!


app/views/calls/_new_call.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="modal" id="new_call">
```
colinxfleming: +1 for making this a partial off the back!

app/models/note.rb
```diff
@@ -5,5 +5,5 @@ class Note
 
 	field :notes, type: String
```
colinxfleming: thanks for fixing this!

app/views/cases/index.html.erb
```diff
@@ -76,24 +76,19 @@
...
-          <td>202-345-6677</td>
-          <td>Sue F</td>
-          <td>15 weeks</td>
+          <td><%= c.patient.primary_phone %></td>
+          <td><%= c.patient.name %></td>
+          <td><%= "#{(DateTime.now - c.last_menstrual_period_time).to_i/7} weeks" %></td>
           <td>(sample case notes. something or other. detail)</td>
-          <td><span class="glyphicon glyphicon-chevron-right" aria-hidden="true"></span></td>
+          <td><%= link_to "Call", "#new_call", data: {toggle: "modal"} %></td>
+          <%= render partial: 'calls/new_call', locals: {c: c} %>
```
colinxfleming: I think the core idea here is really strong and this works very smoothly! Nice work!

A thing I noticed when clicking thru is that this always brings up the info for Patient 2, because the id for every modal is the same. If possible, do you mind tweaking this so the proper modals show up for a given individual?

Two different approaches:

a) You have a `#new_call` id assigned to the modal partial; incorporate in case ids to the `link_to` and the id of the modal (so it links to `#new_call_4` and hits a modal with id `#new_call_4`, for example).

b) Slightly more challenging and javascript intensive: Limit the page to one hidden modal at the bottom, and when you hit, replace the `#modal-body` content. This way, if there are ever a ton of urgent cases or calls this keeps the HTML on the page more lean.

config/routes.rb
```diff
@@ -1,7 +1,11 @@
 Rails.application.routes.draw do
-  authenticate :user do 
+  authenticate :user do
     root to: 'cases#index', as: :authenticated_root
-    resources :cases
+    resources :cases do
+      member do
+        resources :calls, only: [ :create ]
```
colinxfleming: :+1:

db/seeds.rb
```diff
@@ -5,3 +5,21 @@
 #
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
+Call.destroy_all
+Case.destroy_all
+Patient.destroy_all
+
+10.times do |i|
+  Patient.create({name: "Patient #{i}", primary_phone: "123-123-1234"})
```
colinxfleming: It rules that you started the seed file!

What do you think about mixing up the primary phone number a little bit? Generally when we get a new patient we first check their phone number to make sure they aren't in our system already, so phone is kinda considered a unique key for the users. (Let me know if that doesn't make sense! A look at some sample DCAF data from the excel sheet might be helpful here.)

tl;dr: What do you think about making `primary_phone` be "123-123-123#{i}" so each of the 10 sample patients here has a unique phone number?

.byebug_history
```diff
@@ -0,0 +1,17 @@
+quit
```
colinxfleming: I imagine we should probably leave this file out of the merge. If you don't mind, would you remove this from the PR and pop in `.byebug_history` to the `.gitignore` file?

colinxfleming:
Fantastic work, @ajohnson052 ! I left a few thoughts as I went through it, let me know what you think.

colinxfleming:
Fixes #71 
Fixes #70 
Fixes #47

ajohnson052:
I was having some trouble with replacing the body of a single modal for new calls, so I went ahead with adding unique IDs. If this is cluttering up the html too much we can work on a smarter solution, but at least the basic functionality is there now. I'm glad that you caught my mistake!

colinxfleming:
@ajohnson052 don't fret about it! JS modal population is totally the kind of thing that we can put in a parking lot for the time being, and get back to when we're out of MVP. 

Thanks very much for finishing this off! I'll get this reviewed and merged later tonight.

colinxfleming:
image

Thanks so much for seeing this thru, @ajohnson052 !

Issue #59, opened 2016-02-05
Add civic.json


stvnrlly:
See http://open.dc.gov/civic.json/ for more information.

colinxfleming:
Info we'll need/want to fill this out: 

* Page url 
* Any partner info (include Mollie, Andrea, other contributors?) 
* License

stvnrlly:
Note that only `name`, `status`, and `tags` are required fields. I don't think this project has a homepage (though the repo is fine) and partner info would traditionally mean just DCAF (with the rest of us working on behalf of Code for DC). And of course the license will follow quickly after #58 :smile:.

colinxfleming:
Resolved in #81

PR #81, opened 2016-02-14
Civicjson

colinxfleming:
First pass at `civic.json`. 

Open questions about whether or not @mebates and @adinneen would like to be represented on this as co-leads, and what license to use (which will be resolved in a separate PR). 

@stvnrlly can you spotcheck?


colinxfleming:
Fixes #59

stvnrlly:
- [ ] `"MD,"` should be `"MD",` to make valid JSON
- [ ] the `license` field should be blank until one is selected
- [ ] `status` is probably more "Alpha"
- [ ] `contact` should probably be a team representative rather than the partner (maybe swap it with yourself in the `partner` field?)

colinxfleming:
Made @stvnrlly 's suggested changes and bumped the co-leads into the partners section. Merging.

PR #83, opened 2016-02-16
Minitest

ajohnson052:
Hey Colin, thanks again for the help yesterday! A few things came up as I was going through the rest of the call controller. It looks like we had been saving calls without statuses in our first test. Now I am validating by status in the call model so the test needed updated. There’s also one test that doesn’t make much sense the way I’m approaching it, but I’ve included inline comments to explain the problem. Feel free to come back with lots of comments! This isn't quite ready to merge, but I'm just assuming that a PR is still the easiest way to give/get feedback.


colinxfleming:
Thanks a ton for your hard work on this, Alexis! I should be able to review
this today or tmrw.

app/controllers/calls_controller.rb
```diff
@@ -5,6 +5,9 @@ def create
     @call = c.calls.new(status: params[:status])
 		if @call.save
 			redirect_to root_path
+    else
+      flash[:alert] = "Call failed to save! Please submit the call again."
```
colinxfleming: You totally have the right idea here -- tests are a great opportunity to refactor and add stuff like this, so if something goes wrong you have a better sense of why. (At least, that's my own philosophy of testing!)

app/models/call.rb
```diff
@@ -6,4 +6,6 @@ class Call
 	field :status, type: String
 	embedded_in :case
 
+	validates_presence_of :status
```
colinxfleming: :+1:

test/controllers/calls_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    #   assert_includes Case.find(@case).calls, Case.find(@case).calls.last
+    # end
+
+    it 'should redirect to the root path afterwards' do
+      post :create, status: @call.status, id: @case
+      assert_redirected_to root_path
+    end
+
+    it 'should not save if status is blank for some reason' do
+      assert_difference 'Case.find(@case).calls.count', 0 do
```
colinxfleming: perf!

test/controllers/calls_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  describe 'create method' do
+    before do
+      @call = attributes_for :call
+    end
+
+    it 'should create and save a new call' do
+      assert_difference 'Case.find(@case).calls.count', 1 do
+        post :create, status: @call.status, id: @case
```
colinxfleming: You have the core ideas of this dead on. Way to go!

Before I merge I want to make sure that the post parameters are working the way I think they are -- rails used to be a real bear about testing parameters like you use `status` here, and I want to just double check, since we'll probably use this as a model for testing other controllers as well.

test/controllers/calls_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    it 'should not save if status is blank for some reason' do
+      assert_difference 'Case.find(@case).calls.count', 0 do
+        post :create, status: nil, id: @case
+      end
+    end
+
+    it 'should reject other attributes besides status' do
+      post :create, status: @call.status, other: "extraneous attribute", id: @case
```
colinxfleming: Can you add an `assert_no_difference` block to this as well? We want to be sure that having extra stuff prevents the object from saving, to prevent sketchy people from doing sketchy stuff.

test/controllers/calls_controller_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    it 'should create and save a new call' do
+      assert_difference 'Case.find(@case).calls.count', 1 do
+        post :create, status: @call.status, id: @case
+      end
+      assert_response :redirect
+    end
+
+    # If we have to look up the call through a case this test seems redundant
+    # I.e. I think the previous test also confirms the association
+    # it 'should associate the new call with a case' do
```
colinxfleming: I agree that this is a little redundant as written (it's much more relevant in postgres / relational models), but we do want to test that the controller is saving the object to the DB the way we think it is. Maybe we remix this to check equality of the status, and that the parent case ids are the same? (I think keep the `assert_includes` line as well, great call on that.)

test/controllers/calls_controller_test.rb
colinxfleming: (You might have to add strong params to get this to work! Check out @drownedout 's `patient_params` in the patients controller for an example, and LMK if you have any trouble getting it to behave.)

colinxfleming:
This is really great work, @ajohnson052 . Nice job! 

I'll do a little wrenching on it tomorrow night to make sure things are behaving right, and I had a few comments and suggestions if you've got the opportunity this week, but these are great. (If you don't have the time I'm happy to merge this and knock the rest of them out this weekend.)

test/controllers/calls_controller_test.rb
ajohnson052: After adding 'validates_presence_of :status' to the call model, this test wasn't working with call: @call, but I think I've found the problem! @call was being defined in the setup and the before action. I took it out of the setup and changed the post method params back to what we originally had.

test/controllers/calls_controller_test.rb
ajohnson052: Do the strong params really prevent it from saving, or just prevent the extra attributes from being accepted? I have added strong params and I'm still struggling with this one

test/controllers/calls_controller_test.rb
ajohnson052: Can you think of a better way to write the assert_includes statement? As is I think it would always pass if @case has any calls at all

colinxfleming: After a little bit of toying around, I think you're right. Let's leave this block out, since it's redundant and confusing.

test/controllers/calls_controller_test.rb
colinxfleming: After some review with @ajohnson052 there's some strong params demonry going on with the parameters hash. Going to create an issue for resolution later.

colinxfleming:
@ajohnson052 :clap:

PR #91, opened 2016-02-21
sendgrid setup

colinxfleming:
Does heroku setup for sendgrid, so we can send emails and etc.


colinxfleming:
We can send password reset emails now! And invitation emails, once we get that figured out all the way. 

Fixes #127 also.

PR #112, opened 2016-02-23
renaming case -> pregnancy_case, added linter

Kevin-Wei:


app/models/pregnancy.rb
```diff
@@ -7,9 +7,9 @@ class Case
 
 	field :line, type: String #DC, MD, VA
 	field :language, type: String
-	field :case_id, type: String
+	field :p_case_id, type: String
```
colinxfleming: @Kevin-Wei I think we can switch these up to `pregnancy_id` and `pregnancy_status`, right? Or would it be better to prepend the `p` here?

app/policies/pregnancy_policy.rb
```diff
@@ -1,5 +1,5 @@
-class CasePolicy < ApplicationPolicy
+class PregnancyPolicy < ApplicationPolicy
```
colinxfleming: Note for me: We should strip pundit for now.

test/controllers/calls_controller_test.rb
```diff
@@ -4,7 +4,7 @@ class CallsControllerTest < ActionController::TestCase
   def setup
     @user = build :user
     sign_in @user
-    @case = create :case
+    @prengancy = create :pregnancy
```
colinxfleming: @Kevin-Wei  can you fix up the spelling here?

test/factories/call.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :call do
+    pregnancy
```
colinxfleming: good call on this

lib/tasks/factory_girl.rake
```diff
@@ -0,0 +1,15 @@
+namespace :factory_girl do
```
colinxfleming: @Kevin-Wei I like where your head's at with this. Let me mull over the best way to work this in.

colinxfleming:
Thanks for this, @Kevin-Wei ! Made a few comments and suggestions. 

Once those are resolved I'll merge with the lint rake task as is. Wonder if we should modify the rakefile to have it run on `rake test` as part of the regular suite, rather than a separate task? Can do that in a separate PR though.

app/models/pregnancy.rb
Kevin-Wei: @colinxfleming either is fine. Just throwing out the idea. whatever you think is better for consistency.

app/policies/pregnancy_policy.rb
Kevin-Wei: I'd be all for adding it back when it's needed

test/controllers/calls_controller_test.rb
Kevin-Wei: @colinxfleming 

This, I can do

app/models/pregnancy.rb
colinxfleming: @Kevin-Wei let's switch it to `pregnancy_id` and `pregnancy_status`, since I don't think there's another instance of `p_{whatever}` anywhere at the moment.

app/policies/pregnancy_policy.rb
colinxfleming: Oh definitely agree on adding it back when we need to set permissioning! I think it'll be a smidge easier to just start fresh rather than figure out what's already there.

lib/tasks/factory_girl.rake
Kevin-Wei: @colinxfleming Automatically running the linter usually frowned upon because DB whacking can get out of hand as factory/models grow in complexity.

I'd recommend having the CI run the linter once as part of the build before `rake test` is invoked and recommend people run the linter rake task only when they make any changes to the FactoryGirl factories.

Something Fancy like Guard could automate this but that's probably too much overhead for now.

colinxfleming: @Kevin-Wei copy that. So you're recommending it just be a CI task rather than a user task it sounds like. When we merge I'll add an issue so that we remember to add it to Circle's tasks.

Adding Guard sounds like a good candidate for backlog once we've hit MVP. Agree that it's too heavy for where we are at now.

Kevin-Wei:
@colinxfleming hey I kinda let this float around a little too long, but the sooner you can merge this in the better? I settled on PregnancyCase as the resource. That seems like a good compromise?

colinxfleming:
Sorry, that's on me! Pregnancy was good, I think just those few attributes (:p_case_id or something) were confusing. 

Since this is my miscommunication I'll merge this tonight (since it's a major change and we need it in sooner) and fix on my end a little after.

colinxfleming:
Thanks a ton for seeing this thru @Kevin-Wei !

PR #113, opened 2016-02-23
Case Index Views

drownedout:
View changes.

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -3,10 +3,17 @@
 */
 @import "bootstrap-sprockets";
 @import "bootstrap";
+@import url(https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,200,200italic,300,300italic,400italic,600);
 
 $dcaf-color: #5D3E8D;
 
 // DCAF color scheme button
+
+html {
+  font-family: 'Source Sans Pro', sans-serif;
+}
+
+
 .btn-primary {
   @extend .btn-primary;
   background-color: $dcaf-color;
@@ -49,7 +56,7 @@ img {
 // to make views framework-neutral
 main {
   @extend .container;
-  background-color: #eee;
+  background-color: #ffffff;
   padding-bottom: 80px;
   width: 100%;
   margin-top: 51px; // accommodate the navbar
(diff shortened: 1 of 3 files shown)


mebates:
Translation: @drownedout made a bunch of great style changes to the landing page! YAY. He's going to work on getting Source Sans Pro loaded into the app as well; every bit of type so far should be in Source Sans Pro. Thanks for your awesome work!

colinxfleming:
Thanks for this @drownedout @mebates ! Will review changes tomorrow.

app/views/layouts/application.html.erb
```diff
@@ -7,6 +7,7 @@
     <%= stylesheet_link_tag 'application', media: 'all' %>
     <%= javascript_include_tag 'application' %>
     <%= csrf_meta_tags %>
+    <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,200,200italic,300,300italic,400italic,600' rel='stylesheet' type='text/css'>
```
colinxfleming: @drownedout Can we get this moved into the asset pipeline rather than linking to it directly? Would prefer that this live in, and get loaded from, `vendor/assets/stylesheets` rather than a live http link.

app/assets/stylesheets/application.css.scss
```diff
@@ -3,10 +3,17 @@
 */
 @import "bootstrap-sprockets";
 @import "bootstrap";
+@import url(https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,200,200italic,300,300italic,400italic,600);
```
colinxfleming: Similar deal -- would prefer this get loaded in the asset pipeline rather than the live internet. HMU on slack if you need a hand with this.

colinxfleming:
Thanks a bunch @drownedout ! I had a few suggested alterations so we don't load assets from the live internet -- let me know if you need a hand making them, or won't have the time.

colinxfleming:
Also note that this touches a decent amount of the stuff that @ajohnson052 was working on in that dashboard, so we may need to spend some time reconciling over the week / Monday.

colinxfleming:
ALSO: This looks really nice! Has a really pleasant look and feel to it.

Issue #109, opened 2016-02-22
CALL LIST: Extract urgent cases loop into a partial


colinxfleming:
From `app/views/cases/index.html.erb`: 

```
        <% @urgent_cases.each do |c| %>
        <tr>
          <td><%= c.patient.primary_phone %></td>
          <td><%= c.patient.name %></td>
          <td><%= "#{(DateTime.now - c.last_menstrual_period_time).to_i/7} weeks" %></td>
          <td>(sample case notes. something or other. detail)</td>
          <td><%= link_to "Call", "#call-#{c.patient.primary_phone}", data: {toggle: "modal"} %></td>
          <%= render partial: 'calls/new_call', locals: {c: c} %>
        </tr>
        <% end %>
```

We're going to wind up using this in a few different places. Let's extract this into its own partial so that we can more easily append rows to our call log.

ajohnson052:
In PR 114, the table_content holds this block and takes a cases as a local variable

Issue #103, opened 2016-02-22
LOOKUP: Create partial for existing patient record display


colinxfleming:
When a record shows up, we should have a clean way of displaying it with a partial. We should create a partial that matches the one in InVision, so we can go to that person and check on their info, or add them to our call list.

See https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983263 for the invision prototype design.

Issue #102, opened 2016-02-22
LOOKUP: Create search route for patients


colinxfleming:
No point in saving these patient records if we can't pull them back up! Let's create a route in cases that lets us retrieve patients who have a phone number or secondary phone number that matches a patient's.

ajohnson052:
PR 114 should take care of this

Issue #108, opened 2016-02-22
CALL LIST: Call list should stay persistent when going back and forth between patient info and dashboard view


colinxfleming:
Case managers shouldn't need to rebuild their call list if they go to edit a patient's information; the call list should stay persistent until they log out or until some time passes without action.

colinxfleming:
@ajohnson052 approached this in a very thoughtful way and this is fixed in #114 because of it.

PR #114, opened 2016-02-23
Search functionality

ajohnson052:
Search results are now being populated with jQuery!

This also includes a method for adding and removing cases to and from the call list, but we will need to rework that part. 1) In the routes I originally included a user id in case we needed the option for a user (e.g. a manager) creating call lists for other users. It sounds like this isn't necessary. 2) Right now multiple users can add the same case to their call lists. 3) It needs some AJAX action.


colinxfleming:
Heroic! Thanks @ajohnson052 ! I'll get this reviewed tomorrow. 

For 1), you're absolutely correct that we won't need to worry about this. CMs can and do flag particular cases for each other, and the Urgent functionality takes care of it the rest of the way, since those all show up at the bottom. 

For 2) This is dead on as well. One of the major problems we're trying to solve is case managers not being able to look at a case at the same time, so we definitely want users to be able to futz around with the same case simultaneously. 

It looks like the CI check is coming back red -- we'll want to resolve that before merging.

Fabulous work! Thanks for your hustle on this.

ajohnson052:
My mistake on the testing! I forgot to run all of the new tests once I merged in master. I just had to sign in a user for the CasesControllerTest, and things appear to be running smoothly now. Looking forward to your comments!

colinxfleming:
CI is coming back clean now! Thanks for fixing!

Taking a look at this now.

app/controllers/cases_controller.rb
```diff
@@ -4,4 +4,20 @@ def index
     @p_case = @patient.cases.build
     @urgent_cases = Case.where(urgent_flag: true)
   end
+
+  def search
+    name_match = Patient.where(name: params[:search])
+    primary_match = Patient.where(primary_phone: params[:search])
+    secondary_match = Patient.where(secondary_phone: params[:search])
+    patients = name_match | primary_match | secondary_match
```
colinxfleming: Had no idea that `|` was a concatenation operator for arrays! Learn something new every day. Thanks for showing me this!

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,17 @@
+class UsersController < ApplicationController
+
+  def add_case
```
colinxfleming: This corresponds to creating call lists for others, right? Can we remove?

app/models/case.rb
```diff
@@ -30,6 +30,7 @@ class Case
 	field :urgent_flag, type: Boolean
 
 	belongs_to :patient
+	belongs_to :user
```
colinxfleming: probably strip this relationship as well?

app/controllers/users_controller.rb
colinxfleming: (Let me know if I'm wrong and we're keeping track of call list for an individual user with this method! Or if we can adapt it to keep call list in session)

app/views/cases/_search_results.html.erb
```diff
@@ -0,0 +1,11 @@
+<h4>Search Results</h4>
```
colinxfleming: :100: for making this a partial off the bat!

app/views/cases/index.html.erb
```diff
@@ -41,28 +37,9 @@
...
-          <td><span class="glyphicon glyphicon-chevron-right" aria-hidden="true"></span></td>
-        </tr>
-        <tr>
-          <td>202-345-6677</td>
-          <td>Patty W</td>
-          <td>9 weeks</td>
-          <td>(sample case notes. something or other. detail)</td>
-          <td><span class="glyphicon glyphicon-chevron-right" aria-hidden="true"></span></td>
-        </tr>
+        <%= render partial: 'table_content', locals: {cases: current_user.cases} %>
```
colinxfleming: Ahhhh I see what's up here. Sorry for commenting before understanding! If this is tied to `current_user` it's solid.

config/routes.rb
```diff
@@ -10,4 +10,7 @@
   end
   root :to => redirect('/users/sign_in')
   devise_for :users
+  patch 'users/:user_id/add_case/:id', to: 'users#add_case'
```
colinxfleming: Can you move these additional routes into the `authenticate :user do` block? Want to signify that they require being authenticated to do any of them.

colinxfleming:
@ajohnson052 this is really great work. Nicely done! This is further along than where I thought it was -- really impressed that you got this out in only two hours. 

I think the only real change is moving the new routes into the `authenticate` block. Other than that, it looks good, works on my machine, and passes tests.

Fixes #103 
Fixes #102 
Fixes #109 

Partially implements #106 
Partially implements #108

ajohnson052:
@colinxfleming I think addresses the authenticate user problem for add/remove calls. Let me know if there's anything I missed!

colinxfleming:
@ajohnson052  thank you so much for your hustle on this (again)! A+ work.

PR #117, opened 2016-02-25
Set up continuous deployment

colinxfleming:
Made some adjustments in CircleCI and to the `circle.yml` file. Let's see if they work!


colinxfleming:
(It works!)

PR #128, opened 2016-03-06
Add Search route tests

### LINKED ISSUE #104: LOOKUP: Create tests for search route
We should test the search route to ensure that it's working, and that it's only returning what it's supposed to, and nothing too weird.

### LINKED ISSUE #105: LOOKUP: Create integration test for record lookup and display
Record lookup is a fairly core function, so we should start testing it early. Let's make sure that searching a phone number of an existing record brings up the right record and displays its information.

colinxfleming:
Adds integration and unit tests for the search route, and does a little refactoring. 

Notable refactors/improvements: 
* Sets format of '/search' to JS only, to make the integration tests not cry
* Moves search controller action stuff to a Patient object method, and adds unit tests for same

Fixes #104 
Fixes #105

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -13,10 +13,6 @@ class Patient
     name_matches = Patient.where name: name_or_phone_str
     primary_matches = Patient.where primary_phone: name_or_phone_str
     secondary_matches = Patient.where secondary_phone: name_or_phone_str
-    results = []
-    (name_matches | primary_matches | secondary_matches).each do |match|
-      results << match
-    end
-    results
+    (name_matches | primary_matches | secondary_matches)
   end
 end
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -10,9 +10,13 @@ class Patient
   validates_presence_of :name, :primary_phone
 
   def self.search(name_or_phone_str) # TODO optimize
-    name_match = Patient.where name: name_or_phone_str
-    primary_match = Patient.where primary_phone: name_or_phone_str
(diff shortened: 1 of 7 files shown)


app/models/patient.rb
```diff
@@ -8,4 +8,15 @@ class Patient
 	has_many :pregnancies
 
   validates_presence_of :name, :primary_phone
+
+  def self.search(name_or_phone_str) # TODO optimize
+    name_matches = Patient.where name: name_or_phone_str
+    primary_matches = Patient.where primary_phone: name_or_phone_str
+    secondary_matches = Patient.where secondary_phone: name_or_phone_str
+    results = []
```
ajohnson052: It makes a lot of sense to move this logic to the model. Good call! 

The only suggestion I have is that we avoid looping through the results array in the model. We could take out lines 17-21 and simply define results on line 16 ```results = name_matches | primary_matches | secondary_matches```

test/models/pregnancy_test.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    it 'should find multiple patients if there are multiple' do 
+      assert_equal Patient.search('123-456-6789').count, 2
+    end
+
+    it 'should be able to find based on secondary phones too' do 
+      assert_equal Patient.search('999-999-9999').count, 1
+    end
+  end
+end
```
ajohnson052: Awesome job building out all of these tests!

app/models/patient.rb
colinxfleming: You're dead on here. If we're not doing anything with results, we can just eschew the results entirely and return the concat of those three arrays.

Issue #121, opened 2016-03-01
Call log modal broken by JS


colinxfleming:
Sad to say, the modal JS stopped working when the search bar went in. 

Need to fix so we can continue to log calls, but not a huge deal.

ajohnson052:
PR #129 addresses this, but it seems strange that it broke when the search bar went in. Makes me wonder if there's something else wrong...

colinxfleming:
There was some hardcore wrenching on the partials in my integration tests PR to dry everything up, so I think likely that's what screwed it up. At any rate, really appreciate your fixing it.

ajohnson052:
Sure thing!

PR #129, opened 2016-03-08
Hotfix - Call modal

ajohnson052:
The modal identifier in the new call link wasn't lining up with the modals so the link didn't do anything. Now it brings up a new call corresponding with the patient!


colinxfleming:
@ajohnson052 Thank you a ton for fixing this so quick! I'll write integration tests for call logging over the weekend so I don't accidentally break things in the future (or will at least be able to figure out exactly why I broke things).

colinxfleming:
Also I'm super glad this was a straightforward fix! I was worried that I'd really screwed something up when I added the jquery, haha.

ajohnson052:
Glad to help! It’s only fair that I make these fixes since I didn’t write my own integration tests…

colinxfleming:
Fixes #121

Issue #95, opened 2016-02-22
PT INFO: Create partial for patient information subsection


colinxfleming:
Create a partial for the patient information subsection. (Just the right side of the form!) 

See https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983267 for the designs.

charleshuang80:
making this one and https://github.com/colinxfleming/dcaf_case_management/issues/94 one big form, and then we can style and break out into partials. Does that work?

colinxfleming:
Settling in on this now. @charleshuang80 built a super solid foundation out so it shouldn't take very long.

Issue #94, opened 2016-02-22
PT INFO: Create partial for abortion information


colinxfleming:
Create partial for abortion information subsection. (Just the right side form.) 

See https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983267 for relevant design/prototype.

drownedout:
@colinxfleming I can take care of the partials

colinxfleming:
Going to put the nail in this now.

Issue #92, opened 2016-02-22
PT INFO: Create edit/update routes for Pregnancy


colinxfleming:
Create edit and update routes for Pregnancy, with a view for the edit route, so that we can recall and edit a case's information. 

This case should be populated by separate partials for individual sections, instead of one master file. See other issues in this sprint.

See https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983267 for relevant design/prototype.

colinxfleming:
Fixed in #130

PR #130, opened 2016-03-10
Start edit pregnancy form

charleshuang80:
Started the edit pregnancy form with patient and clinic information. Will continue on the rest of the form.


app/models/pregnancy.rb
```diff
@@ -3,39 +3,46 @@ class Pregnancy
 	include Mongoid::Timestamps
 	include Mongoid::History
 
-	field :voicemail_ok, type: Boolean
+	# relationships
```
colinxfleming: Nailed it. Thanks a ton for this reorg!

app/views/pregnancies/edit.html.erb
```diff
@@ -0,0 +1,59 @@
+<!-- Patient Info -->
+<%= bootstrap_form_for @pregnancy do |preg| %>
```
colinxfleming: A++++ view writin'

colinxfleming:
Looks like this fixes the following: 

#92 

Gets us most of the way there on: 

#95 
#94

charleshuang80:
Also, sorry, but you should take out the `<% console %>` on line 56 if you want to see how it will look or for tests (it is for while I've been working on it).

Issue #93, opened 2016-02-22
PT INFO: Create partial for top bar / core case information


colinxfleming:
Create a partial for the top bar of the design - The name, phone number, LMP, Status, Appt Date portion. 

See https://projects.invisionapp.com/share/6757W6WFJ#/screens/115983267 for relevant design/prototype.

charleshuang80:
@colinxfleming I'll start on the partials

colinxfleming:
I think the core of this is here and we can break the rest of it (appointment date as datepicker, last menstrual period as human readable thing) up into discrete issues. Will tag those issues in this and close.

colinxfleming:
Datepicker: #132 
LMP display: #141

Closing as good to go!

PR #131, opened 2016-03-10
Add Patient Information to Pregnancy edit form

charleshuang80:
-Added links to left hand side
-Added patient information

still need to:
-add special circumstances checkbox
-style/format form to use bootstrap grid columns better

### DIFF
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -28,6 +28,7 @@ class Pregnancy
 	field :age, type: Integer
 	field :city, type: String
 	field :state, type: String #ennumeration?
+	field :zip, type: String
 	field :county, type: String
 	field :race_ethnicity, type: String
 	field :employment_status, type: String
--- a/app/views/pregnancies/edit.html.erb
+++ b/app/views/pregnancies/edit.html.erb
@@ -25,10 +25,23 @@
 
   <section class="abortion-info">
     <section class="abortion-left-menu col-md-4">
-
+      <!-- links to hook into forms on right side -->
+      <a href="#">Abortion Information</a>
+      <br/>
+      <br/>
+      <a href="#">Patient Information</a>
+      <br/>
+      <br/>
+      <a href="#">Notes</a>
+      <br/>
+      <br/>
+      <a href="#">Call log</a>
+      <br/>
+      <br/>
+      <button class="btn btn-primary">Submit Pledge</button>
     </section>
     <section class="abortion-main-info col-md-8">
-      <div class="abortion-info">
+      <div class="abortion-info col-md-8">
         <h2>Abortion Information</h2>
 
         <div class="info-form-left col-md-6">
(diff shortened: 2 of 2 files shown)


app/models/pregnancy.rb
```diff
@@ -28,6 +28,7 @@ class Pregnancy
 	field :age, type: Integer
 	field :city, type: String
 	field :state, type: String #ennumeration?
+	field :zip, type: String
```
colinxfleming: +1 on zip being a string and not an int!

colinxfleming:
Looks good on first pass! Will take an in-depth look tonight, but can tell you that any suggested changes will wind up being pretty marginal stuff.

colinxfleming:
Thanks a ton for your hard work on this, as always!

app/views/pregnancies/edit.html.erb
```diff
@@ -50,10 +63,48 @@
           </p>
         </div>
       </div>
+      <div class="patient-info col-md-8">
```
colinxfleming: Nice, these are going to be super easy to extract into partials.

colinxfleming:
Looks like this takes care of the bulk of #93, #94, and #95, yeah? We'll still need to arrange, extract into partials, set up toggling, and make the update route work. (And tests!)

colinxfleming:
I'm going to leave those three issues open until they're in their own partials, but this will make it pretty trivial.

Thanks a ton for tackling this @charleshuang80 !

charleshuang80:
Yeah, I think that it has most of those partials tickets covered, but not all is completed yet. A few thoughts:
1. There are definitely still parts that are missing from each of those sections. (There are also sections missing according to the wireframes (many fewer fields so likely easier)
2. None of the fields have actually been hooked into the update method for the controller yet. I would be wary of breaking them up into the partials until that has been completed and we know it works. And then it would require some care to make sure the right locals are passed in but shouldn't be too difficult.
3. They are thus far one giant form. If we are breaking up the view into partials, we might want to think through if it's possible to break it into smaller forms.
4. Unless/until the partials will be used in other places, I'm not sure I see the value of breaking up the view/form. But that's just my opinion, it's likely you have a bigger picture view of how it'll be useful.

colinxfleming:
Going down the line:

1. Copy that. I think you're right to flag this. My thinking is that we can get it pretty close but it will wind up being a matter for @lwaldsc and @NerdyGirl537 to figure out and get exactly right. 
2. I hear you on the update method not being hooked up yet, and on the merit of taking this slow. If worse comes to worse we can develop it as a monolith and then break into partials as an enhancement. But I'd like to wind up with partials for sure.  
3. I think you're right to flag that there are parts of this we can separate from general information (notes, for example, might be a great candidate for this). But let's see how it shakes out first and make a decision on that down the road. 
4. My reasoning for wanting to break them up is ease in finding what to edit and keeping filesize down -- there's going to be a lot going on in the views, and there's going to be enough tweaking and editing that we want to keep things as simple as possible. KISS is my guiding principle here, followed closely by single responsibility.

PR #136, opened 2016-03-12
break edit form into partials

colinxfleming:
Breaks the edit view into a series of partials with locals, for more straightforward and isolated editing as we get everything in agreement. 

@charleshuang80 what do you think? Any feedback? Slightly mods some of the variable names and slightly changes some sections/divs/classes/ids, but nothing too heavy.

### DIFF
--- a/app/views/pregnancies/_abortion_information.html.erb
+++ b/app/views/pregnancies/_abortion_information.html.erb
@@ -1,0 +1,22 @@
+<section id="abortion_section" class="abortion-info col-md-8">
+  <h2>Abortion Information</h2>
+
+  <div class="info-form-left col-md-6">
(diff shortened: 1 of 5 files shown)


colinxfleming:
Merging since this is a blocker.

Forward momentum on #93 #94 #95

charleshuang80:
looks pretty good! as long as it's loading properly should be good, but I am going to sync and check it out now. will let you know or do a PR about any issues I come across.

colinxfleming:
I stubbed a couple of things in the interest of getting the core sprint
stuff out, but it looks great and was SUPER easy to break up, which I think
is a real testament to yr writing a very clean view.

PR #164, opened 2016-03-15
Building out the Call Log

ajohnson052:
This builds off of Daniel's work on the call log from last night. (I actually pulled down his pull request so all of his work should be reflected here.) It populates the log and allows the user to toggle between viewing all calls or just the ten most recent.

I feel like it's a little weird that the toggle method is reached through a post request, but Minitest was throwing a CORS error for the get request...


colinxfleming:
CORS errors have been the bane of my existence lately so I totally feel you. 

And also, gold medal for adding tests! $$$

Love this! I'll plan to review in depth on Thursday.

app/views/pregnancies/_call_log.html.erb
```diff
@@ -0,0 +1,16 @@
+<section id="call_log" class="call-info tab-pane active">
```
colinxfleming: Sorry, I'm responsible for the merge conflict here! We should go with yours -- please  resolve merge conflicts in favor of what's in here.

app/views/pregnancies/_call_log.html.erb
```diff
@@ -0,0 +1,16 @@
...
+		<h1 class="border-bottom title"> Call Log</h1>
+		<table class="table border-side">
+		<tr>
+			<th>Date</th>
+			<th>Time</th>
+			<th>Result</th>
+			<th>user</th>
+		</tr>
+		<tbody id="call-log-content" >
+			<%= render 'call_log_content', calls: @pregnancy.calls.order("created_at DESC").take(10) %>
```
colinxfleming: What do you think about changing this `take` to a `limit`? I think if we're only using the first ten calls on load, no point in grabbing all of them.

colinxfleming: Actually, I think this raises a bigger question. Discussing in the main, rather than inline comments.

colinxfleming:
@ajohnson052 I was thinking about this a little -- do we think that the calls are going to change while a patient's pregnancy info is being edited here, or that CMs will need to refresh call logs regularly over the course of talking to a patient? If not, maybe it makes more sense to edit the call log content partial into something like this instead of a separate route:

```
<tbody id="call-log-content" >
  <%= render 'call_log_content_recent', calls: @pregnancy.calls.order("created_at DESC").limit(10) %>
  <%= render 'call_log_content_all', calls: @pregnancy.calls.order("created_at DESC") %>
</tbody>
```

Welcome your thoughts here.

test/controllers/pregnancies_controller_test.rb
```diff
@@ -56,9 +56,17 @@ class PregnanciesControllerTest < ActionController::TestCase
...
     end
 
-    it 'should redirect if record does not exist' do 
+    it 'should redirect if record does not exist' do
       patch :update, id: 'notanactualid', pregnancy: @payload
       assert_redirected_to root_path
     end
   end
+
+  describe 'toggle_call_log method' do
```
colinxfleming: tests :100: :100: :100:

colinxfleming:
@ajohnson052 and I had some backchannel back and forth and agreed that it was best to avoid adding a new route if we could. So what if we did something like the following for the time being:

* Strip the new route and toggle controller action
* For now, remove the ability to toggle back and forth between last ten and all calls
* Instead, display something similar to how you currently have it on page load -- just the most recent ten -- under the call log section

I suspect there's a super clean way to display the full list of calls, and that it'll come down to us in a beam of light in sprint 4, haha. For now I'm comfortable with MVPing it. 

@ajohnson052 what are your thoughts? I think I'm generally a little gunshy about custom routes (with your `add_pregnancy` and `remove_pregnancy` being notable exceptions!) but, again, am open to suggestion otherwise. 

Sorry for dragging this out!

app/views/pregnancies/_call_log.html.erb
```diff
@@ -3,14 +3,29 @@
...
+				<tr>
+					<td><%= call.created_at.getlocal.strftime("%-m/%-d") %></td>
+					<td><%= call.created_at.getlocal.strftime("%l:%M %P") %></td>
+					<td><%= call.status %></td>
+					<td><%= "user" %></td>
+				</tr>
+			<% end %>
+			<% @pregnancy.old_calls.each do |call| %>
+				<tr class="old-calls hidden">
+					<td><%= call.created_at.getlocal.strftime("%-m/%-d") %></td>
```
colinxfleming: this reminds me, we'll want to figure out a unified way to display datestamps and timestamps. Going to create an issue for this and probably develop a convenience method for it.

app/views/pregnancies/_call_log.html.erb
```diff
@@ -3,14 +3,29 @@
...
+			<th>Time</th>
 			<th>Result</th>
 			<th>user</th>
 		</tr>
+		<div class="btn btn-primary" id="toggle-call-log">View all calls</div>
 		<tbody>
-			<td><%= pregnancy.initial_call_date %></td>
-			<td><%= pregnancy.status %></td>
-			<td><%= pregnancy.user %></td>
+			<% @pregnancy.recent_calls.each do |call| %>
```
colinxfleming: Minor change, but for consistency -- `pregnancy` should be available within the partial as a local variable -- any chance I could get you to use the local variable instance of the instance variable here?

app/assets/javascripts/pregnancies.coffee
```diff
@@ -1,3 +1,8 @@
 # Place all the behaviors and hooks related to the matching controller here.
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
+
+$(document).on "click", "#toggle-call-log", ->
+  $(".old-calls").toggleClass("hidden")
+  html = if $(".old-calls").hasClass("hidden") then "View all calls" else "Limit list"
```
colinxfleming: Worth noting that this toggle the calls on, but not sure it toggles back? We should file as an issue for now

app/views/pregnancies/_call_log.html.erb
colinxfleming: I think there's a way we can refine this to be one db call but I think this is cool for now. Let's file as an enhancement under issues.

app/models/pregnancy.rb
```diff
@@ -54,4 +54,12 @@ def self.most_recent
     order('created_at DESC').limit(1).first
   end
 
+	def recent_calls
+		calls.order("created_at DESC").limit(10)
+	end
+
+	def old_calls
```
colinxfleming: I think we can optimize this a little with the postgres cmd / activerecord method `offset` -- something like `calls.order('created_at desc').offset(10)` should work fab for this!

app/views/pregnancies/_call_log.html.erb
```diff
@@ -3,14 +3,29 @@
 		<h1 class="border-bottom title"> Call Log</h1>
 		<table class="table border-side">
 		<tr>
-			<th>Time Stamp</th>
+			<th>Date</th>
+			<th>Time</th>
 			<th>Result</th>
 			<th>user</th>
 		</tr>
+		<div class="btn btn-primary" id="toggle-call-log">View all calls</div>
```
colinxfleming: As an enhancement, let's make it so this button only shows up if there are 10+ calls (e.g. if there are more calls to show, and toggle on and off). I think this might be a good easy PR for someone to get comfy with the codebase so let's leave it hanging for now.

colinxfleming:
:100: :100: :100: :100: 

Thanks a ton for your hustle on this as always, @ajohnson052 !

PR #172, opened 2016-03-20
Add call creation

### LINKED ISSUE #155: Make sure all calls are tied to particular users
Do what it says on the label -- every call should be traced back to a specific user.

colinxfleming:
What started as tying calls to user became so much more. Adds a slew of controller and model tests for call, 

* Requires that a call be tied to a user
* Adjusts controller to fit that behavior
* Adds tests on the model and controller levels to confirm most of the call controller and object behaviors
* Adds a user method to get a call's creator
* Other assorted tidying

Fixes #155


colinxfleming:
Oh word! Thx for flagging @Kevin-Wei -- I'll create an issue for the backlog.

PR #199, opened 2016-03-29
fix #178

### LINKED ISSUE #178: Only display full call list toggle if there are 10+ calls
In `app/views/pregnancies/_call_log.html.erb`, there's a toggle button that displays the full set of calls instead of just limiting to the most recent 10. In the event that there aren't ten calls yet, let's make it not show up.

mchelen:


colinxfleming:
Going to leave this open for now per discussion in #178, does that sound good to you @mchelen ?

mchelen:
- Updated PR with suggestions from #178 
- Fixed logic for checking call count
- Moved button to bottom of list for better UX

colinxfleming:
@mchelen this works perfectly, thanks so much for sticking with it! Would you mind doing me a solid and resolving the merge conflict in favor of the h1 with the `border-bottom title` class value, and then I'm going to merge this up!

:100:

mchelen:
Ok I resolved the conflict and successfully merged in `master` branch, hope that's the right way to address it.

colinxfleming:
That's dead on. Thanks a ton!

Issue #195, opened 2016-03-28
Spin front-end design improvements into their own issues


mebates:
Creating a checklist of design improvements for CSS-inclined friends to tick off over the next few weeks.
You can always refer to the [native .ai files](https://github.com/colinxfleming/dcaf_case_management/blob/master/_design/DCAFwireframe020816.ai) or the [Vivify clickable prototype](https://projects.invisionapp.com/d/main#/projects/5385941) if you don't know what I'm talking about! 

**Everywhere**
- [x] Remove black line around purple buttons
- [x] Load Source Sans Pro and make it the default font for the site
- [x] Base font size should be 17 px, not 14 px
- [x] Font size for all input boxes should be 17 px
- [ ] It would generally be cool if we could implement a standard type hierarchy for H1, H2, H3 etc across the site. If you're interested in this task, let's chat!
- [x] Use sentence case across the site (i.e. "This is an item label" not "This Is An Item Label")

**Dashboard**
- [x] Widen width of name / phone number entry box
- [x] Lowercase "R" in "Search results" heading
- [x] 'Call' phone icon should be a white phone in a blue circle
- [x] 'Call' phone icon should be a white phone in a blue circle

**Patient detail page**
- [ ] Selected menu item in the left nav should match specs in InVision (purple text) Remove blue highlight. 
- [ ] Add right-facing caret to the left of the selected menu item.
- [x] 'Submit pledge' button should be BIG (super button) and should have more blank space above it, separating it from the left nav menu items. 
- [x] Add lines into the page separating top entry, left nav and main right window. 
- [x] Black bar at the top should be gray, type should be black

rebeccaestes:
test

mebates:
Details on the type hierarchy. The names I'm using may not align with the names you'll use in the HTML. All type is Source Sans Pro. (Sizes cribbed from here: https://cfpb.github.io/design-manual/identity/typography.html#web-hierarchy) 

Display: 48 px / 60 px, 300 (Light)
H1: 34 px / 42 px, 400 (Regular)
H2: 26 px / 32 px, 400 (Regular)
H3: 22 px / 28 px, 400 (Regular)
H4: 14 px, 700 (bold), uppercase, 1 px letter-spacing

Body copy: 17 px / 26 px (1.5 em), 400 (Regular)

mebates:
Great work @rebeccaestes !!!

colinxfleming:
#195 resolves the following: 
[...]

PR #201, opened 2016-03-29
front-end issues from #195 about half done

rebeccaestes:


colinxfleming:
This is fantastic work and looks gorgeous, thanks @rebeccaestes !

This code fails the tests for pretty silly reasons, just the capitalization of 'Results' in Search Results. You can fix on lines 25 and 37 of `test/integration/record_lookup_test.rb` by just changing `assert has_text? 'Search Results'` to `assert has_text? 'Search results'` with a lowercase `r` on both lines.

Issue #188, opened 2016-03-27
NOTES: Create a notes controller with a create and update route


colinxfleming:
So that we can save notes for pregnancies, let's create a controller that lets us create and update notes for a pregnancy.

Should have routing similar to `Calls`.

PR #204, opened 2016-03-29
add notes controller

ajohnson052:
This adds a basic create function and create/update routes. I'm still messing around with the form itself and trying to figure out how to avoid nesting forms without hurting the UX.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -1,0 +1,27 @@
+class NotesController < ApplicationController
+  before_action :find_pregnancy only: [:create]
+
+  def create
+    @note = @pregnancy.notes.new(note_params)
+    if @call.save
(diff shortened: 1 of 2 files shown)


app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,27 @@
+class NotesController < ApplicationController
+  before_action :find_pregnancy only: [:create]
```
colinxfleming: I think this might be missing a comma between the `find_pregnancy` and `only`? Small quibble though!

colinxfleming:
Looks good to me, @ajohnson052 ! After that comma issue is set I'll be good to merge. 

Let me know if I can be of assistance on the form stuff! 

Taps #188

ajohnson052:
@colinxfleming Should be all set now! Good catch!

PR #205, opened 2016-03-29
README formatting tweaks

valeriecodes:
There wasn't an issue for this, just some things I noticed when going through the readme.
Turns instructions into bulleted lists instead of code blocks, removes `$` from beginning of terminal commands for easier copy/pasting


colinxfleming:
Our readme is badly in need of a bump for sure! I'll make sure to add that as an issue for the off week, since the project and its organization has changed plenty since I wrote it. 

Thanks for making these enhancements, @valeriecodes ! Glad to have you on board.

Issue #193, opened 2016-03-27
NOTES: Create integration test for note creation


colinxfleming:

colinxfleming:
(This is underway in #209)

edwardpark:
Ops accidentally closed the PR but i'll still volunteer to tackle this one

PR #209, opened 2016-04-01
 Create integration test for note creation 

edwardpark:
i'll clean up the commit and force it over for a new pr but wanted to see for additional feedback for just happy path test to get this setup?

### DIFF
--- a/app/views/pregnancies/_call_log.html.erb
+++ b/app/views/pregnancies/_call_log.html.erb
@@ -1,6 +1,6 @@
 <section id="call_log" class="call-info tab-pane">
 	<div class="margin-bottom col-md-12">
-		<h2>Call log</h2>
+		<h1 class="border-bottom title">Call Log</h1>
 		<table class="table border-side">
 		<tr>
 			<th>Date</th>
@@ -8,7 +8,6 @@
 			<th>Result</th>
 			<th>CM</th>
 		</tr>
-		<div class="btn btn-primary" id="toggle-call-log">View all calls</div>
 		<tbody>
 			<% pregnancy.recent_calls.each do |call| %>
 				<tr>
@@ -27,6 +26,9 @@
 				</tr>
 			<% end %>
 		</tbody>
-            </table>
-	</div>
+  </table>
+		<% if pregnancy.old_calls.count > 0 %>
+      <div id="toggle-call-log">View all calls</div>
+    <% end %>
+  </div>
 </section>


colinxfleming:
Hey Edward, glad to have you! Thanks very much for your pr! 

I'll review over the weekend and offer feedback/suggestions.

test/integration/note_test.rb
```diff
@@ -0,0 +1,39 @@
...
+        click_link 'Notes'
+        
+        within('#notes')do
+            assert_text 'Notes' #confirm notes header is visible 
+            assert has_button? 'Submit'
+        end
+    end 
+    
+    it 'should let you add a new pregnancy note' do
+        fill_in 'note_body', with: 'Sample Note Body'
```
colinxfleming: For better attribute clarity we've renamed this `full_text` -- just a heads up!

test/integration/note_test.rb
```diff
@@ -0,0 +1,39 @@
...
+        end 
+        assert_text 'Sample Note Body'
+    end 
+
+    it 'should let you edit a pregnancy note' do
+        click_on 'edit'
+        fill_in 'note_body', with: 'Sample Note Body Edit'
+        click_on 'Submit'
+        
+        assert_text 'Sample Note Body Edit'
```
colinxfleming: I'd consider reloading the page as well to ensure that it stays persistent between pageloads.

test/integration/note_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    @pregnancy = create :pregnancy, appointment_date: nil, patient: @patient
+    visit edit_pregnancy_path(@pregnancy)
+  end
+ 
+  describe 'add patient pregnancy notes' do
+    it 'should display a pregnancy notes form and current notes' do
+        click_link 'Notes'
+        
+        within('#notes')do
+            assert_text 'Notes' #confirm notes header is visible
```
colinxfleming: I expect that the exact attributes we're looking for will get remixed a little when the page gets set up but I think your instinct here is right on!

test/integration/note_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test_helper'
+
+class PregnancyNoteTest < ActionDispatch::IntegrationTest
```
colinxfleming: Any chance you could do me a solid and give this a slightly more descriptive filename, and match the class name? Something like `note_creation_test.rb` so we can tell exactly what actions the integration test is looking at, just by the filename.

colinxfleming:
@edwardpark looks good! Obv we don't have this quite set up yet (the frontend side of this will be our major focus on Wednesday at the meeting) but I'm super excited that you've done us the solid of kickstarting it! 

Had a few suggestions that you can tackle at your leisure or on Wednesday. 

Also, make sure you say hi if you're there on Wednesday! Not sure I've had the pleasure of meeting you yet!

edwardpark:
@colinxfleming thanks! i'll setup and update these changes , I won't be able to actually make it this Wednesday ( out of town for a wedding ) but i'll be sure to stop by the next one after though.

colinxfleming:
Sounds good! I expect that by then we'll have the form implemented so we may have a couple more edits, but they'll probably be pretty straightforward and trivial. 

Looking forward to meeting you! Thanks again for your work!

colinxfleming:
(Note to self: Don't merge yet, but fixes #193)

PR #216, opened 2016-04-06
document step to load sample db in docker

mchelen:


colinxfleming:
:100: :100: :100: 

@mchelen I'll load all of these after we kick off tonight. Thanks a ton for making these edits, really appreciate your eyes!

PR #221, opened 2016-04-07
Seed calls

mchelen:


db/seeds.rb
```diff
@@ -24,5 +24,13 @@
   else
     flag = false
   end
-  patient.pregnancies.create({last_menstrual_period_time: DateTime.new(2016,1,1), urgent_flag: flag})
+  pregnancy = patient.pregnancies.create({last_menstrual_period_time: DateTime.new(2016,1,1), urgent_flag: flag})
+  8.times do
+    pregnancy.calls.create({status: 'Left voicemail', creating_user_id: user.id})
+  end
+  if(patient.name[-1, 1].to_i == 0) then
```
colinxfleming: Not terribly clear what this is doing -- looks like it's adding five calls to Patient 0, but I had to read it a few times to figure it out. 

Can we make it look a little more 'ruby'? We could either rewrite the `if` statement to be more clear about when it's executing the block (something like `if patient.name == 'Patient 0'`) or pop it out of the block entirely and add a line to the effect of `10.times do Patient.find(name: 'Patient 0').pregnancy.calls.create ...`. 

Sorry to be nitpicky about this!

colinxfleming:
Lookin' good @mchelen ! Had a very slight suggestion for a clarity change, LMK if you have any questions. 

Thanks for knocking this out, really appreciate your hard work on making setup easier!

db/seeds.rb
mchelen: Totally agree, this version was the result of some last minute copy paste.

I tried to follow your suggestion about separating it from the block like this:
```ruby
10.times do
  Patient.find(name: 'Patient 0').pregnancy.calls.create({status: 'Reached patient', creating_user_id: user.id})
end
```
But I don't think that works because 1 patient can have multiple pregnancies.

So I used the `if patient.name == 'Patient 0'` approach and that seems to work. Let me know if you think that looks good from a readability standpoint or if there is any better alternative.

db/seeds.rb
```diff
@@ -24,5 +24,13 @@
...
     flag = false
   end
-  patient.pregnancies.create({last_menstrual_period_time: DateTime.new(2016,1,1), urgent_flag: flag})
+  pregnancy = patient.pregnancies.create({last_menstrual_period_time: DateTime.new(2016,1,1), urgent_flag: flag})
+  5.times do
+    pregnancy.calls.create({status: 'Left voicemail', creating_user_id: user.id})
+  end
+  if patient.name == 'Patient 0'
+    10.times do
+      pregnancy.calls.create({status: 'Reached patient', creating_user_id: user.id})
```
colinxfleming: I'm currently getting an error when running `rake db:seed` with this: 

```
$ rake db:seed
rake aborted!
Mongoid::Errors::UnknownAttribute: 
message:
  Attempted to set a value for 'creating_user_id' which is not allowed on the model Call.
```

I think we can resolve if we change both the `creating_user_id`s to `created_by`s, like so: 

`pregnancy.calls.create({status: 'Left voicemail', created_by: user.id})`

@mchelen thoughts?

mchelen: Ah ok, yeah that did the trick. It looks like the `Call` model was updated after my fork was made, any way we could add an automated test for `rake db:seed` so that CI will catch it when this happens?

mchelen:
I merged master in so I could test with the updated code, let me know if there is anything I should do to clean this up.

db/seeds.rb
colinxfleming: I think this is a good idea. Adding it to the issues log.

colinxfleming:
Looks great. Thanks for pushing this over the finish line @mchelen !

Issue #244, opened 2016-04-17
STATUS: Add method to infer status to pregnancy model


colinxfleming:
@adinneen did some great work with @lwaldsc in #202 to hash out Statuses. We should build this into the Pregnancy model. 

In a nutshell, the statuses we want to capture (and the text output) are, in descending order of priority (e.g Resolved Without DCAF should trump everything :

* No 'Reached Patient' calls on record (No Contact Made)
* Has been reached, but doesn't have an appointment date (Needs Appointment)
* Has an appointment, but doesn't have a pledge marked as sent (Fundraising)
* Has a pledge marked as sent (Sent Pledge)
* Has a pledge marked as paid (Pledge Paid) 
* Resolved Without DCAF attribute is true (Resolved Without DCAF)

We don't have pledges wired up yet so we can't do that yet, so let's just do placeholders for that for now. We may also need to add the Resolved Without DCAF attribute to the Pregnancy model.

colinxfleming:
Resolved in #252

PR #252, opened 2016-04-19
infer status

ajohnson052:
This PR hasn't really been fully built out, but it's a start on Issue #244

Concerns
* Right now the sent field on pledges is a DateTime and the sent_by is a string. If this is the case then pledge_status?(sent) won't work, but I think it would be nice to keep the boolean on sent and make another field for the DateTime (I'm not sure why it isn't configured in the same way as paid so I didn't mess with it)
* pledge_status? just looks for any pledge that has been sent/paid. Do we want this or should it be looking at the last pedge for the pregnancy? Realy not sure how multiple pledges work
* This doesn't actually update the status field of the pregnancy. I was originally setting self.status instead of just the local variable, but that wasn't quite working how I wanted it to.
* The long string of elsifs is just kinda gross

Sorry for the novel!


colinxfleming:
No apology necessary for the long explanation! Love it! 

For pledge status, I think it's worth mulling over how to represent this before we finish this off - does it look like just checking for a DCAF pledge marked sent in an array of pledges (since we want to capture other org's pledges too)? My gut says that we should first decide whether or not DCAF hard/soft pledges are different than pledges we want to keep track of from other organizations, and then build around that. But again, that's probably the next sprint, so we can totally punt on that decision for now, and just have a couple commented out lines.

app/models/pregnancy.rb
```diff
@@ -67,4 +68,34 @@ def recent_calls
   def old_calls
     calls.order('created_at DESC').offset(10)
   end
+
+  def contact_made?
+    self.calls.each do |call|
```
colinxfleming: If memory serves, you might be able to omit the `self` reference here -- I think Ruby is smart enough to know to check calls on the given object.

app/models/pregnancy.rb
```diff
@@ -67,4 +68,34 @@ def recent_calls
...
+  end
+
+  def pledge_status?(status)
+    self.pledges.each do |pledge|
+      return true if pledge[status]
+    end
+    false
+  end
+
+  def infer_status
```
colinxfleming: Since we're basically going to be using this in place of `status`, let's get rid of the entire `status` field and replace it with this. 

In other words -- Can we rename this method to `status`, and delete the model attribute defined above?

ajohnson052:
@colinxfleming makes sense to comment out the pledge portion until we have a better idea what we're doing! I can also just delete it if you'd prefer.

colinxfleming:
@ajohnson052 I'd say keep it in for now, just so we can keep it on our collective radar. 

Rudeness from the test suite! I'll take a look at this. Looks like it's something simple since only one test is eating it.

colinxfleming:
Small but significant! One of our integration tests caught an issue that was preventing things from saving on the page, and the reason for that is that there's still a status field which is no longer a model-writeable attribute.

Super straightforward fix -- just need to comment out this in line 17 of `app/views/pregnancies/_patient_dashboard.html.erb`:

```
<%= f.text_field :status, label: 'Status:' %>
```

After that we're all good for merge! Thanks much @ajohnson052 !

ajohnson052:
Well, in the process of fixing that one little test I realized that the whole method was essentially written backwards... But now there are tests! So I think it actually good to go now. Since there were significant changes you might want to give the whole thing one last look though.

In particular, I don’t know if we want to display the inferred status in  place of the input field as I had it.

P.S. thanks for the note about running bundle update - that solved a lot of my problems :)

app/models/pregnancy.rb
```diff
@@ -67,4 +67,35 @@ def recent_calls
...
+
+  # def pledge_status?(status)
+  #   pledges.each do |pledge|
+  #     return true if pledge[status]
+  #   end
+  #   false
+  # end
+
+  def status
+    status =
```
colinxfleming: I think this is right on -- very readable and straightforward. I think you could even remove the variable setting on L86 if you want to simplify further and just have the strings.

app/views/pregnancies/_patient_dashboard.html.erb
```diff
@@ -14,7 +14,8 @@
 </div>
 <div class="col-md-4">
   <!-- case status should be c.select -->
-  <%= f.text_field :status, label: 'Status:' %>
   <!-- Appointment Date (not in Case model) -->
   <%= f.datetime_picker :appointment_date, label: 'Appointment date:', format: '%Y-%m-%d' %>
+  <label class="control-label" for="pregnancy_status">Status:</label>
```
colinxfleming: :100:

test/models/pregnancy_test.rb
```diff
@@ -40,12 +40,40 @@ class PregnancyTest < ActiveSupport::TestCase
...
 
-    it 'should have accessible userstamp methods' do 
+    it 'should have accessible userstamp methods' do
       assert @pregnancy.respond_to? :created_by
       assert @pregnancy.created_by
     end
-  end  
+  end
+
+  describe 'status method' do
```
colinxfleming: YUSSSSSSS tests

test/models/pregnancy_test.rb
```diff
@@ -40,12 +40,40 @@ class PregnancyTest < ActiveSupport::TestCase
...
+    it 'should default to "No Contact Made"' do
+      assert_equal 'No Contact Made', @pregnancy.status
+    end
+    it 'should update to "Needs Appointment" once patient has been reached' do
+      create :call, pregnancy: @pregnancy
```
colinxfleming: Two suggestions here:

* Add a Left Message call and confirm that the status stays at No Contact Made, so we can be sure that the existence of a call in general isn't updating it to Needs Appointment
* Clarify that the call here on L57 is a 'Reached patient' call; e.g. `create :call, pregnancy: @pregnancy, status: 'Reached patient'` so we can tell just by looking at the test what's going on, rather than popping into the factory

colinxfleming:
🎊  tests 🎊

colinxfleming:
@ajohnson052 this is really great, thank you! I had a suggestion for the unit tests (a little excessive safety, maybe) and a few readability suggestions, but this is really great stuff. Nice work!

Looks like there's a merge conflict (probably a Gemfile.lock I'd guess) -- should be a straightforward fixer upper. Let me know if we're not so lucky.

colinxfleming:
$$$$$$$$$$$$$

PR #270, opened 2016-05-03
Pledge Controller 

### LINKED ISSUE #260: PLEDGE: Create pledge controller with create and update routes
Nothing fancy or hard -- just rig up a quick pledge controller with `create` and `update` routes. 

The `notes_controller` is probably useful as a model here -- should be structured and oriented in a similar way.

Nico-ehs:
Added Controller and tests.

Fixes #260


app/controllers/pledge_controller.rb
```diff
@@ -0,0 +1,40 @@
+class PledgeController < ApplicationController
+  before_action :find_pregnancy, only: [:create]
+  before_action :find_pledge, only: [:update]
+
+  def create
+    @pledge = @pregnancy.pledges.new(pledge_params)
+    @pledge.created_by = current_user
+    @pledge.save
```
colinxfleming: I think you can strip this line -- `@pledge.save` will run as part of the conditional just below this, so no need to do this twice. Thoughts?

app/controllers/pledge_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      respond_to { |format| format.js }
+    else
+      head :bad_request
+    end
+  end
+
+  private
+
+  def pledge_params
+    params.require(:pledge).permit(:pledge_type, :amount, :other_pledge_identifier,
```
colinxfleming: :100:

spring-1000/bdadf250a96d13a0db6c0dfeb0e14538.pid
```diff
@@ -0,0 +1 @@
+34157
```
colinxfleming: Any idea what this is? We've had a spring directory but idk what spring-1000 is.

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+  before do
+    @user = create :user
+    sign_in @user
+    @patient = create :patient
+    @pregnancy = create :pregnancy, patient: @patient
+  end
+  
+  describe 'create method' do
+    before do
+      @pledge = create :pledge,
```
colinxfleming: Can we use `attributes_for :pledge` instead of `create :pledge` here? That way, we'll have the attributes we want to post without creating duplicate objects in the db.

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+                pledge_type: "Sample Pledge", 
+                amount: 5, 
+                other_pledge_identifier: "none", 
+                sent: DateTime.current, 
+                sent_by: @user, 
+                paid: False,
+                paid_date: DateTime.current
+      post :create, pregnancy_id: @pregnancy.id, pledge: @pledge, format: :js
```
colinxfleming: The create method isn't ajax (yet), so the `format: :js` will bomb here. Can you do me a solid and strip that?

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    it 'should create and save a new pledge' do
+      assert_difference 'Pregnancy.find(@pregnancy).pledges.count', 1 do
+        post :create, pregnancy_id: @pregnancy.id, pledge: @pledge, format: :js
```
colinxfleming: Same deal, this `format: :js` will bomb.

colinxfleming:
Hey @Nico-ehs , a couple notes:

a) Rails is kind of a bear about its conventions -- model names tend to be singular, and controllers tend to be plural. So we'll need to pluralize `pledge` to `pledges`. Probably the easiest way to do this is to run `rails g controller pledges`, move the relevant code over, and then run `rails d controller pledge` (singular) to remove the files generated with the singular. If you'd rather just change files that are already there rather than run the generator, I'll mark those up in this comment thread too just in case. (PS don't stress about this, as every rails developer ever will be able to tell you about the time they burned an afternoon trying to figure out why nothing was working, only to find out they called a controller `user` instead of `users`. Welcome to the club! :100:)

b) Looks like we need a route set up for these calls! You can set that up by adding `resources :pledges, only: [ :create, :update ]` after line 8 in `config/routes.rb`.

app/assets/javascripts/pledge.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
colinxfleming: rename to `pledges.coffee`

app/assets/stylesheets/pledge.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the pledge controller here.
```
colinxfleming: rename to `pledges.scss`

app/controllers/pledge_controller.rb
```diff
@@ -0,0 +1,40 @@
+class PledgeController < ApplicationController
```
colinxfleming: rename file to `pledges_controller.rb`; change the class name here to `PledgesController` with an S

app/helpers/pledge_helper.rb
```diff
@@ -0,0 +1,2 @@
+module PledgeHelper
```
colinxfleming: rename file `pledges_helper.rb`; rename module to `PledgesHelper`

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
+require 'test_helper'
+
+class PledgeControllerTest < ActionController::TestCase
```
colinxfleming: rename class to `PledgesControllerTest`

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+  
+  describe 'create method' do
+    before do
+      @pledge = create :pledge,
+                pledge_type: "Sample Pledge", 
+                amount: 5, 
+                other_pledge_identifier: "none", 
+                sent: DateTime.current, 
+                sent_by: @user, 
+                paid: False,
```
colinxfleming: I see you come from python! An annoying ruby-ism is that `true` and `false` are both lowercase -- probably change this line to `paid: false` or it'll look for an object called `False` and eat mud.

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  describe 'update method' do
+    before do
+      @pledge = create :pledge, pregnancy: @pregnancy, pledge_type: 'Original Pledge'
+      @pledge_edits = {pledge: {pledge_type: 'Edited Pledge'}}
+      patch :update, pregnancy_id: @pregnancy, id: @pledge, pledge: @pledge_edits, format: :js
+      @pledge.reload
+    end
+
+    it 'should render the correct template' do
```
colinxfleming: Let's comment this test out for now until we start wiring together the UI component of this.

test/controllers/pledge_controller_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      assert_redirected_to edit_pregnancy_path(@pregnancy)
+    end
+
+    it 'should log the creating user' do
+      assert_equal Pregnancy.find(@pregnancy).pledges.last.created_by, @user
+    end
+
+  end
+
+  describe 'update method' do
```
colinxfleming: Until we have a better sense of what updating a pledge looks and feels like, I think we can just comment out tests for this and create a backlog issue for this. Thoughts?

colinxfleming:
All marked up! Sorry for the delay @Nico-ehs ! 

Short version of changes: 
a) make a few routing changes
b) rerun the generator, or pluralize a couple files and classes by hand 
c) make the create action assume that it's a synchronous request, rather than an ajax
d) punt on the update controller stuff until we have a better sense of what that looks like

Let me know if you run into any issues. Thanks a ton bud!

colinxfleming:
Hey @Nico-ehs , pinging you on this. I'll take it up if I don't hear anything by Friday and finish 'er off, so we can make sure your hard work is represented when we meet next wk. $$

Nico-ehs:
Thanks for the tips. I updated the naming and commented out the test controller along with removing save and format.js.

colinxfleming:
Thanks much for this, @Nico-ehs !

PR #271, opened 2016-05-03
Restrict phone fields to a single number #225

### LINKED ISSUE #225: Restrict phone fields to a single number
@NerdyGirl537 and @lwaldsc flagged that the phone number fields can take unlimited lengths, which could lead to unwanted

We should fix by limiting the length of the field in the relevant model, and add a unit test to confirm the validation.

ewiggins:

* add length validations(12 characters) to the patient.rb primary and secondary phone fields

* Fixes #225

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -9,9 +9,12 @@ class Patient
   validates_presence_of :name, :primary_phone
 
   field :name, type: String # strip
-  field :primary_phone, type: String , length{ :maximum => 12}# validate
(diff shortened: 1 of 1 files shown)


colinxfleming:
Hey @ewiggins  -- reviewed this and it does exactly what it's supposed to, we just haven't really worked in good error messages on the controller/view side. I'm going to merge and close it out, and add a separate issue to render form errors for the backlog. 

Glad we worked through this! Thanks a ton for sticking with it!

PR #273, opened 2016-05-07
update controller, and js for adding patient

### LINKED ISSUE #265: PT CREATE: Wire up dashboards#search to return the create a patient partial if there are zero results
Right now, the create a patient partial is sitting in `app/views/patients/_new_patient.html.erb` but isn't hooked into anything. Let's adjust the `DashboardsController#search` method to return that partial if a search returns zero results. 

We'll also want an integration test for this, and a controller test for this.

### LINKED ISSUE #266: PT CREATE: PatientsController#create should create both a patient and a pregnancy from the payload
Our data model associates a pregnancy with a patient object. As such, let's alter `PatientsController#create` so that it creates a Patient and then creates a dependent Pregnancy object as well. 

We'll also want to create an integration test for this, and a controller test for this.

ajohnson052:
This PR combines patient creation with pregnancy creation and renders the partial for a new patient/pregnancy upon returning no search results.

I had difficulty with strong params in the Patient controller so I moved the logic to the pregnancies controller. If this seems weird, I'd be happy to reevaluate. As it is, all of our other logic in happening via pregnancies and we always create the two in tandem so I think it makes some sense.

* Fixes #266
* Fixes #265


app/controllers/dashboards_controller.rb
```diff
@@ -11,6 +9,8 @@ def search
     patients.each do |patient|
       @results << patient.pregnancies.most_recent
     end
+    @patient = Patient.new
```
colinxfleming: Not quite sure I understand this change, @ajohnson052 can you clarify?

app/models/pregnancy.rb
```diff
@@ -13,6 +13,8 @@ class Pregnancy
   embeds_many :calls
   has_one :clinic
 
+  validates_associated :patient, on: :create
```
colinxfleming: I like where yr head's at here. Good call

colinxfleming:
I've been thinking for awhile about whether we should remix the object model to focus on patients rather than pregnancies, as I think it's pretty confusing and doesn't make natural sense the way we have it set up now. Right now it's really inhospitable for people new to the codebase ('patients and pregnancies relate how, again?') and I feel like, on the tech side, we constantly have to be like 'so what's in pregnancy again?'

(Plus, 'A case manager has many pregnancies' is probably a statement that will draw some funny looks.) 

I 100% think you're on the right track by merging the two.

config/routes.rb
```diff
@@ -3,11 +3,10 @@
     root to: 'dashboards#index', as: :authenticated_root
     get 'dashboard', to: 'dashboards#index', as: 'dashboard'
     post 'search', to: 'dashboards#search', defaults: { format: :js }
-    resources :pregnancies, only: [ :edit, :update ] do
+    resources :pregnancies, only: [ :create, :edit, :update ] do
```
colinxfleming: Maybe this should look like so, for clarity's sake?

```
resources :patients do
  resources :pregnancies do
    resources :calls
    resources :notes
```

colinxfleming:
Hey @ajohnson052 , thanks a ton for doing this work. I think I'm going to want to mull this one over a little bit, but left a few notes. I want to avoid yak-shaving too much as we close this out, but also want to make sure we're making the right architecture decisions. Would welcome your feedback and thoughts.

app/controllers/dashboards_controller.rb
ajohnson052: Since we're calling the _new_pregnancy partial from within the javascript now, the instance variables from the dashboard index weren't accessible. However, I can switch back to using #build instead of #new.

config/routes.rb
ajohnson052: I see what you're saying, and I started to go through  with this change, but after switching the path names, e.g. edit_pregnancy_path to edit_patient_pregnancy_path, I realized that we'll also have to pass in an additional variable with each call to the path, i.e. edit_pregnancy_path(@pregnancy) becomes edit_patient_pregnancy_path(@patient, @pregnancy). 

Passing around those extra instance variables makes things a little messier, and we might be looking at things like edit_patient_pregnancy_path(@pregnancy.patient, @pregnancy) in a spot like pregnancies#update. Do you think it's still the smarter approach? I really can't decide which is better.

app/controllers/dashboards_controller.rb
colinxfleming: Makes sense! `new` is fine if you want to keep this as is.

config/routes.rb
colinxfleming: I totally hear you about the messiness, but I think I'm okay with this. After doing a lot of thinking I landed on 'patient and pregnancy should be separate objects, with a patient having a max of one active pregnancy at a time, and any number of inactive pregnancies' as (imo!) the best way to represent the way these pieces fit together. So we don't really have pregnancies existing in isolation in practice.

(Since this body being unearthed is pretty much my fault, I'm happy to make this routing change and associated controller/view changes, haha.)

colinxfleming: Let's do this: I'll handle this change in a separate pull request. For now, let's roll with what you've got.

colinxfleming:
Hey @ajohnson052 , the tl;dr is that I'm good to merge if you are -- I think this is an appropriate thing to get a working version in on, and then gradually improve as time goes on and we get past the point of initial deployment. Thanks a ton for your hard work on this, and for bearing with me as I mull this shit over, haha.

colinxfleming:
@ajohnson052 want to resolve merge conflicts on this and we'll get it in?

colinxfleming:
This does what it says on the label! Great work @ajohnson052 !

I'm going to merge because I want to get this in, but there are a few associated issues we're going to want to iron out here that I think require CM input:

* We'll want LMP to be either a datepicker or a weeks/months thing, based on what our CM leads think would be better. 
* Should we just assume that `initial_call_date` is date of entry and take it out of the form, or should we keep it separate? (Maybe we should kill this field in favor of just using `created_at`?)

colinxfleming:
(On second thought we should do the weeks months thing. The only real open question here is how to represent initial call.)

PR #281, opened 2016-05-20
Creating Concerns for Models

ewiggins:

Added a concern to the concerns folder app>>concerns>> Auditable.rb
Added call for concern to the top of each model
(etc)

Concerns
I need help with the code. I think I'm not able to test ( no connection to the server) because of how I wrote the concern or added it to the models. Not sure what it is. @colinxfleming Maybe you can take a look and get my pointed in the right direction.

As far as the connection, I've tried Mongod, rake db:seed, bundle install, installing Mongo again. Maybe we can take a look on the 24th at Code for DC.


colinxfleming:
Hey @ewiggins ! We'll try and figure out your connection issues next week; in the meantime I'll mark up your code and offer some feedback.

Thanks much for putting in the work on this!

app/models/concerns/Auditable.rb
```diff
@@ -0,0 +1,10 @@
+# Concern Audiable for Models
+Module Auditable.rb
```
colinxfleming: You can lowercase the word `Module`, and take off the `rb` at the end -- 

`module` is the reserved word, and if it's capitalized, ruby will probably go nuts thinking it's its own object.

app/models/concerns/Auditable.rb
```diff
@@ -0,0 +1,10 @@
+# Concern Audiable for Models
```
colinxfleming: Do me a solid and lowercase this filename? Should be `auditable.rb`, just for convention's sake.

app/models/clinic.rb
```diff
@@ -1,8 +1,10 @@
-class Clinic
+class Clinic < Active Record ::Base
+  include Auditable
```
colinxfleming: 💯

app/models/call.rb
```diff
@@ -1,8 +1,9 @@
-class Call
+ class Call < Active Record ::Base
```
colinxfleming: Two things here: 

* I'm not super sure we need to make our model class here descend from ActiveRecord but I might be wrong -- what do the docs say? 

* You nailed the inheritance syntax, but we'll want to change the spacing just a little bit since the name of the class is `ActiveRecord` -- if it turns out we do need to explicitly make these descend from ActiveRecord, let's do `class Call < ActiveRecord::Base`. 

Let me know if this doesn't make sense!

app/models/concerns/Auditable.rb
```diff
@@ -0,0 +1,10 @@
...
+Module Auditable.rb
+ extend ActiveSupport :: Concern
```
colinxfleming: Similar deal, you can pop out the spaces here and have it read `extend ActiveSupport::Concern`.

app/models/call.rb
ewiggins: @colinxfleming What do you mean by decend from ActiveRecord? Do you mean do we need to include a class model < ActiveRecord::Base at the beginning of each model? I saw it that way in the examples.
"And now using Concerns , your Event model reduces to

class Event < ActiveRecord::Base    
    include Commentable
    include Attendable
end"

If you meant something else let me know! I'm not too hip to the Rudy lingo yet :)

PR #284, opened 2016-05-21
Fix static date in Dashboard overview with dynamic date range helper method

### LINKED ISSUE #256: DASH: hook up dashboard 'week of' to current date
Dashboard overview has a 'Week of' title that is currently static. Should be hooked up to reflect actual week of the current date. A dashboard helper method seems like one good option.

sco-bo:
* Creates a helper method in `dashboards_helper` to display the current 'week range' in the dashboard overview. 
* Generates a test for the method, which is currently passing

* Fixes #256

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -1,2 +1,9 @@
 module DashboardsHelper
+
+  def week_range
+    today = DateTime.now
+    week_start = today.beginning_of_week(:monday).strftime("%B %d")
+    week_end = today.end_of_week(:monday).strftime("%d")
+    "#{week_start} - #{week_end}"
+  end
 end
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -3,8 +3,8 @@ module DashboardsHelper
   def week_range(date: DateTime.now, start_day: :monday)
     week_start = date.beginning_of_week(start_day)
     week_end = date.end_of_week(start_day)
-    week_start_string = week_start.strftime("%B %d")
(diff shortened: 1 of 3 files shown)


sco-bo:
Shoot! One situation that I totally didn't take into account is when a particular week spans 2 different months. This PR unfortunately wouldn't address that. Let me know how you want to move forward. Feel free to reject this PR then I can make the necessary changes on my end. Apologies for the error!

app/helpers/dashboards_helper.rb
```diff
@@ -1,2 +1,10 @@
 module DashboardsHelper
+
+  def week_range(date: DateTime.now, start_day: :monday)
```
colinxfleming: 💯

colinxfleming:
Thanks a ton for the work here @sco-bo ! I'll pull down locally tmrw to confirm behavior, but it looks dead on to me.

sco-bo:
You bet @colinxfleming! Really happy to be able to contribute ;)

test/helpers/dashboards_helper_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+
+class DashboardsHelperTest < ActionView::TestCase
```
colinxfleming: We should test the helper methods separately in the pregnancies model too! Adding an issue -- not related to this, just making a note that this is a pattern we should follow.

colinxfleming:
Declaring victory and merging. Thanks a bunch @sco-bo !

sco-bo:
Awesome @colinxfleming ! You bet, thanks!

PR #286, opened 2016-05-24
Bug fixes on Issue 179

### LINKED ISSUE #179: New Call Modal failing to save 'I Reached the Patient' calls
The new call modal isn't working properly all the time. So far I seem to have issues when I try to go back to the dashboard index from the edit pregnancy page by clicking on "DCAF Case Manager" from the nav bar. If I refresh the page it works fine.

drownedout:
Fixed 179: I inspected the button element in the console and noticed that the button did not have the post method associated with it even though the other links did.The problem was that the button was a "button_to" rather than a "link_to" . I swapped it out and it now works.

Fixes #179


colinxfleming:
This looks good to me and works on my machine, thanks for working thru this @drownedout ! 

I'm going to hold on the merge for a few days due to the `Gemfile.lock` changes, which I'm a little worried might cause some conflicts in other PRs. But there aren't any additional changes we need to do and we should merge when this round of PRs settles down a bit. 

Great work!

Kevin-Wei:
Maybe we should specify all our gems so we don't get too many gemlock repo changes

colinxfleming:
I'm starting to lean that way to be honest! This might be something I
tackle in the off week or this weekend.

PR #288, opened 2016-05-25
Add default of current date to new pregnancy create form - issue #279

### LINKED ISSUE #279: New pregnancy create form: Change default to current date
Rolling into one issue since these both came out of @ajohnson052 's pull request in #273  -- 

Here's our form as it stands now -- 

screen shot 2016-05-19 at 10 38 31 am

An open question here: Is there a meaningful difference between initial call date and the day we actually log them in our system? If there is, or if initial call date ever shifts, then we should keep them separate, but otherwise we can probably use the `created_at` timestamp. 

@lwaldsc @NerdyGirl537 can you weigh in on your preferences here from a CM perspective? Should we keep `initial_call_date` as a field or can we infer it based on when we enter them into our system?

Nico-ehs:
Added date default of current date to new pregnancy create form.

Fixes #279

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -11,6 +11,7 @@ class DashboardsController < ApplicationController
     end
     @patient = Patient.new
     @pregnancy = @patient.pregnancies.new
+    @date=Time.now.to_date
 
     respond_to { |format| format.js }
   end
--- a/app/views/pregnancies/_new_pregnancy.html.erb
+++ b/app/views/pregnancies/_new_pregnancy.html.erb
@@ -14,7 +14,7 @@
     <%= f.text_field :last_menstrual_period_weeks, label: "LMP" %>
   </div>
   <div class="col-md-2">
-    <%= f.date_field :initial_call_date, label: "Initial Call Date" %>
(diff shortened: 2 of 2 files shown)


app/controllers/dashboards_controller.rb
```diff
@@ -11,6 +11,7 @@ def search
     end
     @patient = Patient.new
     @pregnancy = @patient.pregnancies.new
+    @date=Time.now.to_date
```
Kevin-Wei: Date.today maybe?

colinxfleming: @Kevin-Wei I think the current Rubocop standard is `Time.zone.today`. I'm inclined to go with that, what do you think @Nico-ehs ?

colinxfleming: @Nico-ehs what do you think about renaming this variable to something like `@today`? It's not really a generic date but today's date, so I think that might be a little easier to read on the fly. What do you think?

Kevin-Wei: Rubocop to the rescue

@today sounds pretty good

colinxfleming:
Renamed this PR to reflect what it does in plain language, to make it a touch less confusing. 

@Nico-ehs seeking yr thoughts on the comment thread, but once we have that figured out I think we're all clear!

PR #293, opened 2016-05-25
add helper methods for recently completed calls. make adding/removing…

### LINKED ISSUE #241: CALL LIST: Add attribute to pregnancy to indicate whether they've been called in session
So that a CM can tell if they've called someone, we need to be able to fill out the call list partial as detailed in #240. 

Here's a minimal implementation: Let's say that if: 
* someone has a call from the `current_user` 
* that is less than 8 hours old, 
* AND they would otherwise be in the call list, 

then they should show up in Calls Completed. 

We should accomplish this with an inferred field on the Pregnancy or Patient model.

charleshuang80:

* helper methods for recently called and not recently called
* consolidate add/remove calls response js.erb files so both tables will be updated
* make the controller response work with the new js.erb file

* Fixes #241

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,5 +1,5 @@
 class UsersController < ApplicationController
-  before_action :get_user_and_pregnancy
+  before_action :get_user_and_pregnancies
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
 
   def add_pregnancy
@@ -18,8 +18,9 @@ class UsersController < ApplicationController
 
   private
 
-  def get_user_and_pregnancy
+  def get_user_and_pregnancies
     @pregnancy = Pregnancy.find params[:id]
     @user = User.find params[:user_id]
+    @urgent_pregnancies = Pregnancy.where(urgent_flag: true)
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,12 +4,16 @@ class UsersController < ApplicationController
 
   def add_pregnancy
     @user.pregnancies << @pregnancy
-    respond_to { |format| format.js }
+    respond_to do |format|
+      format.js { render template: 'users/refresh_pregnancies.js.erb', layout: false }
+    end
   end
 
   def remove_pregnancy
     @user.pregnancies.delete @pregnancy
-    respond_to { |format| format.js }
+    respond_to do |format|
+      format.js { render template: 'users/refresh_pregnancies.js.erb', layout: false }
+    end
   end
 
   private
--- a/app/helpers/pregnancies_helper.rb
+++ b/app/helpers/pregnancies_helper.rb
@@ -6,4 +6,26 @@ module PregnanciesHelper
   def days_options
     (0..6).map { |i| ["#{i} days", i] }
   end
+
+  # ticket 241 recently called criteria:
+  # someone has a call from the current_user
+  # that is less than 8 hours old,
+  # AND they would otherwise be in the call list
+
+  def call_list_pregnancies
+    current_user.pregnancies.reject { |p| recently_called_by_user?(p) }
+  end
+
+  def recently_called_pregnancies
+    current_user.pregnancies.select { |p| recently_called_by_user?(p) }
+  end
+
+  def recently_called_by_user?(pregnancy)
+    pregnancy.calls.any? { |c| c.created_by_id == current_user.id && recent?(c) }
+  end
+
+  def recent?(call)
+    time_of_call = call.updated_at
+    time_of_call > 8.hours.ago ? true : false
+  end
 end
--- a/app/helpers/pregnancies_helper.rb
+++ b/app/helpers/pregnancies_helper.rb
@@ -6,26 +6,4 @@ module PregnanciesHelper
   def days_options
     (0..6).map { |i| ["#{i} days", i] }
   end
-
-  # ticket 241 recently called criteria:
-  # someone has a call from the current_user
-  # that is less than 8 hours old,
-  # AND they would otherwise be in the call list
-
-  def call_list_pregnancies
-    current_user.pregnancies.reject { |p| recently_called_by_user?(p) }
-  end
-
-  def recently_called_pregnancies
-    current_user.pregnancies.select { |p| recently_called_by_user?(p) }
-  end
-
-  def recently_called_by_user?(pregnancy)
-    pregnancy.calls.any? { |c| c.created_by_id == current_user.id && recent?(c) }
-  end
-
-  def recent?(call)
-    time_of_call = call.updated_at
-    time_of_call > 8.hours.ago ? true : false
-  end
 end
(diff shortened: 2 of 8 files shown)


colinxfleming:
Will be tag-teaming review of this with @ajohnson052 next week!

app/controllers/users_controller.rb
```diff
@@ -4,12 +4,16 @@ class UsersController < ApplicationController
 
   def add_pregnancy
     @user.pregnancies << @pregnancy
-    respond_to { |format| format.js }
+    respond_to do |format|
```
colinxfleming: Great call here -- I think this sidesteps a lot of JQuery problems we'd have otherwise.

app/helpers/pregnancies_helper.rb
```diff
@@ -6,4 +6,26 @@ def weeks_options
   def days_options
     (0..6).map { |i| ["#{i} days", i] }
   end
+
+  # ticket 241 recently called criteria:
+  # someone has a call from the current_user
+  # that is less than 8 hours old,
+  # AND they would otherwise be in the call list
+
+  def call_list_pregnancies
```
colinxfleming: It might make more sense to make these model methods and test them as model methods -- what do you think? Like, this might be a great `User` method.

app/helpers/pregnancies_helper.rb
```diff
@@ -6,4 +6,26 @@ def weeks_options
...
+
+  def recently_called_pregnancies
+    current_user.pregnancies.select { |p| recently_called_by_user?(p) }
+  end
+
+  def recently_called_by_user?(pregnancy)
+    pregnancy.calls.any? { |c| c.created_by_id == current_user.id && recent?(c) }
+  end
+
+  def recent?(call)
```
colinxfleming: This makes a lot of sense to me as a method on the `Call` model -- thoughts about moving it over?

app/helpers/pregnancies_helper.rb
```diff
@@ -6,4 +6,26 @@ def weeks_options
...
+  # ticket 241 recently called criteria:
+  # someone has a call from the current_user
+  # that is less than 8 hours old,
+  # AND they would otherwise be in the call list
+
+  def call_list_pregnancies
+    current_user.pregnancies.reject { |p| recently_called_by_user?(p) }
+  end
+
+  def recently_called_pregnancies
```
colinxfleming: Maybe move this over to the `User` method too?

app/helpers/pregnancies_helper.rb
```diff
@@ -6,4 +6,26 @@ def weeks_options
...
+    current_user.pregnancies.select { |p| recently_called_by_user?(p) }
+  end
+
+  def recently_called_by_user?(pregnancy)
```
colinxfleming: Not sure what model this should live with, what do you think @charleshuang80 ?

app/views/dashboards/index.html.erb
```diff
@@ -2,8 +2,8 @@
   <div class="col-md-12">
     <%= render partial: 'dashboards/overview' %>
     <%= render partial: 'dashboards/search_form' %>
-    <%= render partial: 'dashboards/table_content', locals: { title: 'Your call list', table_type: 'call_list', pregnancies: current_user.pregnancies } %>
-    <%= render partial: 'dashboards/table_content', locals: { title: 'Your completed calls', table_type: 'completed_calls', pregnancies: current_user.pregnancies } %>
+    <%= render partial: 'dashboards/table_content', locals: { title: 'Your call list', table_type: 'call_list', pregnancies: call_list_pregnancies } %>
```
colinxfleming: If we change those helper methods to model methods, it'll necessitate changing this around -- just a flag, just in case.

colinxfleming:
@charleshuang80 this is really fabulous work. You've taken some really intensely complicated business logic and built it into really elegant, simple, and readable ruby code. Thanks a ton for handling this!

I had a few suggestions w/r/t placement of some of the stuff in `PregnancyHelper`, but once we figure out how to best handle that let's merge this.

charleshuang80:
@colinxfleming, if you think the methods will work and logically be better placed in the various models, I am fine with it and I'll defer to how you think it should be set up. Here is a little bit of my thought process on why I kept them where they are now:

- I think the `recent?` method is good to move into the `Call` model. (I'll add a comment there also)
- For the other methods, I'll plead a little bit of inexperience and not enough time to check thoroughly. My use of model methods has been with that model's own attributes and not through associations so I wasn't sure if these methods would work (and maybe work well in testing) in the various models as we're dealing with multiple associations.
- If the methods will work fine in the models and it doesn't contradict any best practices then I am not worried about how the `_table_content` partial and the other methods will need to be tweaked, that should be easy.

So basically I knew that everything would be available in the view helpers, and wasn't totally confident (and didn't have time to check) what information would be available if we used model methods, so that is where I put everything. I have nothing against moving the methods as long as they will function as we intend. If you have an idea of how you want the methods to be dispersed then I have no objection, but if you want to discuss further just let me know. Thanks!

app/helpers/pregnancies_helper.rb
charleshuang80: I think this method is totally fine to move. Have added a comment that shares my thoughts on the others.

colinxfleming:
Hey @charleshuang80 -- my reasoning basically boils down to two things. (And to clarify, they'll totally work either way!)

The first ties into how I understand how helpers fit into rails' architecture. I understand Rails helpers to be used for things that don't have anything to do with data transformation, such as wrapping something up nice and clean for use in a view, or doing some straightforward transformation in a controller. See, for example, your work in `PregnanciesHelper#weeks_options` -- not really tied to a model, just something we use in a view. (Here's a stack overflow thread discussing something similar - http://stackoverflow.com/questions/5019794/when-to-use-helpers-vs-model )

The other part is how we'll actually use them in practice. We'll always associate those call list methods with a User, and it's totally feasible that we might want to someday look at someone else's call list who isn't currently logged in, so why not abstract it away from just `current_user` and into the User object itself? Same sort of principle for moving over the `recent?` method into the call object.

Let me know what you think! I'm totally open to being wrong here, and you should totally tell me if this doesn't make sense.

charleshuang80:
@colinxfleming I think this should be in line with what we discussed. I also got the urgent pregnancies list to refresh with the other lists (so the add/remove will toggle). Had one or two minor thoughts as I was finalizing this but we can discuss later. Let me know what you think!

app/controllers/users_controller.rb
```diff
@@ -1,21 +1,26 @@
...
+    end
   end
 
   private
 
-  def get_user_and_pregnancy
+  def get_user_and_pregnancies
     @pregnancy = Pregnancy.find params[:id]
     @user = User.find params[:user_id]
+    @urgent_pregnancies = Pregnancy.where(urgent_flag: true)
```
colinxfleming: This makes a lot of sense. Good call on this.

colinxfleming:
This looks good to me! Thanks again for your hard hustle on this and for pushing it over the finish line!

app/controllers/users_controller.rb
charleshuang80: Didn't have it before, which was what kept throwing errors when I tried to add it into the `refresh_pregnancies` view before.

PR #295, opened 2016-05-25
Allow multiple users to have a pregnancy on their call list #228

### LINKED ISSUE #228: Issues with adding patients to a call list from urgent pregnancies
@lwaldsc and @NerdyGirl537 had some issues getting patients to add to the call list properly, especially from the urgent pregnancies form. Here are the specific issues they ran into and steps to troubleshoot from. 

* Adding pt from call search does not add, only replaces:
  URL: dashboard
  Thing I don't like: Adding from the urgent cases does not properly populate the call log
  Thing I would like better: Adding from either urgent cases or search properly appends to the call log

* Call list is not maintained when returning to main screen from PT screen from link at top left:
  URL: edit page (to dashboard)
  Thing I don't like: When return to call list from PT info, urgent case has fallen off call list
  Thing I would like better: Keep urgent calls on call list when returning to main screen

Kevin-Wei:

* many to many -ifies user and pregnancy
* fix controller test
* makes pledge factory valid

* Fixes #228

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,7 +14,7 @@ class User
           # :timeoutable
 
   # Relationships
-  has_many :pregnancies
+  has_and_belongs_to_many :pregnancies, inverse_of: :users
 
   # Fields
   # Non-devise generated
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -17,6 +17,7 @@
     <tbody>
       <div id="<%= table_type %>_content">
         <% pregnancies.each do |p| %>
+        <% current_user.reload %>
           <tr class="pregnancy-data" id="<%= p.patient.id %>">
             <td><%= p.patient.primary_phone %></td>
             <td><%= link_to p.patient.name, edit_pregnancy_path(p) %></td>
@@ -31,7 +32,7 @@
             <% end %>
 
             <td><%= link_to "<span class='glyphicon glyphicon-earphone' aria-hidden='true'></span><span class='sr-only'>Call</span>".html_safe, "#call-#{p.patient.primary_phone}", data: {toggle: "modal"} %></td>
-            
+
             <%= render partial: 'calls/new_call', locals: {p: p} %>
           </tr>
         <% end %>
--- a/test/factories/pledge.rb
+++ b/test/factories/pledge.rb
@@ -3,8 +3,6 @@ FactoryGirl.define do
     pregnancy
     pledge_type 'Soft'
     amount 100
-    after :build do |pledge, eval|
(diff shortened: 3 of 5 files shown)


app/models/user.rb
```diff
@@ -14,7 +14,7 @@ class User
           # :timeoutable
 
   # Relationships
-  has_many :pregnancies
+  has_and_belongs_to_many :pregnancies, inverse_of: :users
```
colinxfleming: 💯

app/views/dashboards/_table_content.html.erb
```diff
@@ -17,6 +17,7 @@
     <tbody>
       <div id="<%= table_type %>_content">
         <% pregnancies.each do |p| %>
+        <% current_user.reload %>
```
colinxfleming: If it's necessary to reload, can we bump this over to the controller? I'm a little wary of doing this in the view, I think. Thoughts?

Also, this might tap up against some of what @charleshuang80 is doing in #293 -- see his work on `app/views/users/refresh_pregnancies.js.erb` for what's coming down the pike here.

test/factories/pledge.rb
```diff
@@ -3,5 +3,8 @@
     pregnancy
     pledge_type 'Soft'
     amount 100
+    after :build do |pledge, eval|
```
colinxfleming: Two quick things about this: 

a) not quite sure how this applies to the rest of the PR. Was it making tests fail?
b) We've generally been doing this in the form of `created_by { FactoryGirl.create(:user) }` -- see the `Patient` factory, for example. If we need to keep it, what do you think about swapping it over to use the standard format?

colinxfleming:
Confirming that this resolves the bug described in #228, where a patient could only be on one person's call list at a time. Thanks much @Kevin-Wei !

I had a few questions / suggestions about a couple of things in the PR (a reload in a view and a change to the pledge factory), but on the whole, looks good to me. 

Thanks a ton for taking care of this!

Kevin-Wei:
I just run the linter before PR's so as not to submit invalid ones. 

as for partial reload. Charles solution looks better. Without the user reload the partial table doesn't toggle the "add" "remove" correctly. since mongo nonsense is hardcore and the block can't detect the changes in the array until the DB is called again. There is probably a much better solution out there somewhere.

colinxfleming:
Thanks for fixing up the pledge factory, appreciate it. 

And thanks for your thoughts on the partial reload. What are yr thoughts about removing the reload from the view and holding out for @charleshuang80 's PR to get merged?

colinxfleming:
Come to think of it, that's a minor enough change that I think we can just merge and I'll fix manually. Merging now.

Issue #263, opened 2016-05-01
PLEDGE: Controller tests for pledges_controller


colinxfleming:
Let's make sure our controller is rocksolid by writing tests for it. Again, we can use the notes controller tests as a guide here. 

Depends on #260 being completed.

colinxfleming:
Resolved in #263 by @ashlynnpai , closing!

PR #296, opened 2016-05-26
Add controller tests for pledges_controller #263

ashlynnpai:

* The uncommented tests are passing
* The commented tests are failing but I put a note on each test as to why it's failing

* Controller tests for pledges_controller #263

Sorry the whitespace and Gemfile.lock got messy. If it's too messy to read let me know.


test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+  before do
+    @user = create :user
+    sign_in @user
+    @patient = create :patient
+    @pregnancy = create :pregnancy, patient: @patient
+  end
+
+  describe 'create method' do
+    before do
+      @pledge = attributes_for :pledge
```
colinxfleming: 💯  great call on this!

test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+
+#	NOTE: :pledge is getting passed in as a string
```
colinxfleming: I think we can just remove this block that's commented out, since we're using `attributes_for :pledge` instead, which is severely cleaner. Thoughts?

test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+      post :create, pregnancy_id: @pregnancy.id, pledge: @pledge
+    end
+
+    it 'should create and save a new pledge' do
+      assert_difference 'Pregnancy.find(@pregnancy).pledges.count', 1 do
+        post :create, pregnancy_id: @pregnancy.id, pledge: @pledge
+      end
+    end
+
+  # NOTE: it's redirecting so status code 302 (controller redirects to edit_pregnancy_path)
```
colinxfleming: You are right on leaving these out for now.

test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+#	NOTE: I created a blank template for this test	
+    it 'should render the correct template' do
+      assert_template 'pledges/update'
+    end
+
+    it 'should respond with success' do
+      assert_response :success
+    end
+
+  # NOTE: should work, but the call to reload is causing the factory attribute to reload,
```
colinxfleming: INTERESTING. It sounds like there's something about the update method that's preventing it from actually saving -- we're running nearly identical test code for the notes controller and it seems to be working okay. 

Let me take a deeper look tomorrow.

test/factories/pledge.rb
```diff
@@ -3,5 +3,6 @@
     pregnancy
     pledge_type 'Soft'
     amount 100
+    created_by { FactoryGirl.create(:user) }
```
colinxfleming: 💯

colinxfleming:
@ashlynnpai you rule for pushing this over the finish line! I'm going to take a deeper look at some of the problems you ran into tomorrow, and we'll see if we can't get it hashed out. 

Thanks a ton!

test/controllers/pledges_controller_test.rb
ashlynnpai: Yes, I wasn't sure if you wanted those specific values so I left them in. But if you don't need them.

test/controllers/pledges_controller_test.rb
colinxfleming: I think I found our sticking point! The problem child seems to be line 57, where we're building the update hash. I'll leave comments there.

test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+
+    it 'should log the creating user' do
+      assert_equal Pregnancy.find(@pregnancy).pledges.last.created_by, @user
+    end
+  end		
+
+  describe 'update method' do		
+    before do
+      @pledge = create :pledge, pregnancy: @pregnancy, pledge_type: 'Original Pledge'
+      @pledge_edits = {pledge: {pledge_type: 'Edited Pledge'}}
```
colinxfleming: Here's our sticking point! 

We have the `@pledge_edits` var as a nested hash here, right? Let's look at how we pass that in the `patch` command: 

`patch :update, pregnancy_id: @pregnancy, id: @pledge, pledge: @pledge_edits, format: :js`

So in effect, it feeds a payload that looks like this: ` { pledge: { pledge: { pledge_type: 'Edited Pledge' } } }`. Note the additional level of nesting -- in this, Rails is looking for an attribute called `pledge` on the pledge object, seeing that it doesn't exist, and just swallowing the rest of it. 

**In short, you should be able to make these tests pass by changing this line to `@pledge_edits = { pledge_type: 'Edited Pledge' }`.**

test/controllers/pledges_controller_test.rb
```diff
@@ -1,89 +1,99 @@
...
+  #       assert_equal @pledge.history_tracks.count, 2
+  #       @changes = @pledge.history_tracks.last
+  #       assert_equal @changes.modified[:updated_by_id], @user.id
+  #       assert_equal @changes.modified[:pledge_type], 'Edited Pledge'
+  #     end
+
+  # NOTE: there is no model validatation (presence) for pledge type so having nil pledge type does not cause save to fail, 
+  # in contrast there is a model validation for note :full_text		
+  #     it 'should refuse to save pledge type to blank' do
+  #       [nil, ''].each do |bad_text|
```
colinxfleming: This test will probably fail as is, I don't think we're validating presence on pledge type in the model yet. You should feel free to make this change to get this test to pass as written. Let me know if you have any questions about that!

colinxfleming:
Hey Ashlynn, I did that digging and I think I've found a solution. Give it a shot and let me know what you think?

ashlynnpai:
Thanks for solving the pledge update bug! Incorporated your latest comments.

colinxfleming:
:100: !

colinxfleming:
This looks good to me! Thanks so much for sticking with this @ashlynnpai !

PR #313, opened 2016-05-30
adjust so that docker-compose run web rake test works

colinxfleming:
Fix a small stupid config bug that was making the test suite eat mud when you tried to run it from within `docker-compose`.


colinxfleming:
(Now we just need to add phantomjs to the rig.)

PR #317, opened 2016-05-31
Move files to Vendor/Assets #134

### LINKED ISSUE #134: Move external assets into vendor/assets/
We currently have three external assets that we should move into the vendor directory. They are as follows:

* - [ ]  app/assets/stylesheets/application.css.scss: `@import url(https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,200,200italic,300,300italic,400italic,600);` (should be moved into stylesheets so it's accessible)
* - [ ] app/views/layouts/application.html.erb: `<link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,200,200italic,300,300italic,400italic,600' rel='stylesheet' type='text/css'>`
* - x app/assets/javascripts/bootstrap_toggle.js and app/assets/stylesheets/bootstrap_toggle.js

Should be moved to their respective directories in the `vendor` directory, and then should confirm that they're still loading as normal.

ewiggins:
* I moved files from app assets into vendor/assets.

* Fixes #134


vendor/assets/stylesheets/bootstrap_toggle.css
```diff
@@ -25,4 +25,4 @@
 .toggle-off.btn-sm{padding-left:20px}
 .toggle.btn-xs{min-width:35px;min-height:22px}
 .toggle-on.btn-xs{padding-right:12px}
-.toggle-off.btn-xs{padding-left:12px}
```
ewiggins: @colinxfleming I'm not sure why the ci/circleci isn't accepting the move. I removed the extra lines. Do you see anything else wrong? Github says the move would merge just fine.

colinxfleming: I think this was just some other tests that were in master failing - I wrote a couple bad integration tests that randomly fail, which is rude. You can probably fix by pulling down changes in master here -- try `git fetch upstream`.

vendor/assets/javascripts/bootstrap_toggle.js
```diff
@@ -176,5 +176,4 @@
     $checkbox.bootstrapToggle('toggle')
     e.preventDefault()
   })
-
-}(jQuery);
\ No newline at end of file
+}(jQuery);
```
colinxfleming: :100:

colinxfleming:
Looking nice, @ewiggins !

Can you do me a solid and move the `application` files back to where they were, though? `application.css.scss` should live in `app/assets/stylesheets/application.css.scss` and `application.html.erb` should live in `app/views/layouts/application.html.erb`. 

After that we should be all clear!

colinxfleming:
For context, the reason here is in http://guides.rubyonrails.org/asset_pipeline.html under section 2.2.1: 

```
vendor/assets is for assets that are owned by outside entities, such as code for JavaScript plugins and CSS frameworks. Keep in mind that third party code with references to other files also processed by the asset Pipeline (images, stylesheets, etc.), will need to be rewritten to use helpers like asset_path.
```

PR #337, opened 2016-06-04
added LMP field to new pregnancies, and hid placeholder label

rekhers:
New Pregnancies now has the second LMP field, both places where LMP is now have hidden labels over the day field.

### DIFF
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -10,6 +10,12 @@ $font-size: 17px;
   padding: 15px 0;
 }
 
+.lmp-2{
+  label{
+    visibility: hidden;
+  }
+}
+
 #menu {
   border-right: 1px solid black;
   height: 70vh;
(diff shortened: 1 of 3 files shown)


app/views/pregnancies/_new_pregnancy.html.erb
```diff
@@ -13,6 +13,9 @@
   <div class="col-md-2">
     <%= f.text_field :last_menstrual_period_weeks, label: "LMP" %>
   </div>
+   <div class="col-md-2 lmp-2">
```
colinxfleming: 💯

app/assets/stylesheets/pregnancies.scss
```diff
@@ -10,6 +10,12 @@ $font-size: 17px;
   padding: 15px 0;
 }
 
+.lmp-2{
```
colinxfleming: Should we adjust this class name to reflect what it's actually doing? Maybe `hide-label`?

colinxfleming:
Yussss. Thank you @rekhers ! 

I had a small suggested change re: css class tagging, but I think this is all clear after that.

PR #339, opened 2016-06-04
set rails version to 4.2.6 and bundle update

### LINKED ISSUE #300: Bump rails version to 4.2.6

colinxfleming:

Brakeman scans revealed that our version of rails had a cross site scripting vulnerability in it; we're hence bumping to 4.2.6. 

I was a little concerned that this was going to bust something, but the test suite (including integration tests) still passes and some basic smoke tests all come back clean. 

* Bumps rails version to 4.2.6

* Fixes #300


colinxfleming:
Note that this should make the CI in #301 pass.

PR #344, opened 2016-06-06
Add ruby-audit and bundler-audit to CI

### LINKED ISSUE #169: Set Ruby version to 2.2.4
Heroku is starting to yell at me about how they're depreciating 2.0.0, and I don't want this to go by the wayside. We shouldn't resolve now to avoid having to lock people into specific ruby versions, but we should resolve before the MVP goes live.

### LINKED ISSUE #340: Additional security gems to install
https://github.com/civisanalytics/ruby_audit
https://github.com/rubysec/bundler-audit
https://github.com/rubysec/ruby-advisory-db

colinxfleming:

* Adds `ruby-audit` and `bundler-audit` to keep tabs on known vulnerabilities
* Adds them to CI, to fail builds if there are known vulnerabilities
* Bumps the rails version in the dockerfile
* Sets the ruby version in gemfile

* Fixes #340 
* Fixes #169


colinxfleming:
Note to self that I'll need to make sure everything builds in docker-compose before merge. 

Also -- since I'm not going to be at the meeting on Thursday, I'm going to hold off on merging this just in case it causes weird problems or people have difficulty bumping their ruby versions.

colinxfleming:
Planning to merge this after #369 #366 #364 #347 get merged.

PR #346, opened 2016-06-08
Remove up and down arrows from date field

### LINKED ISSUE #309: Remove datepicker up-down arrows and the clear X from initial call date field
From @mebates in https://github.com/colinxfleming/dcaf_case_management/issues/195#issuecomment-221441155 -- this is kinda uggs and cutting off the actual date:

image

Space is already kind of at a premium there, so let's take the clear `x` and the up-down arrows out of that form field. May involve changing the field datatype in the view or making some other adjustments. 

The code is in `app/views/pregnancies/_new_pregnancy.html.erb`.

sco-bo:

Really only one commit here. The first was actually me updating my fork...the lazy way, oops sorry! I made changes to the date_field 'field' and tweaked the sass date type styling on that date_field

* removes clear button in date_field field
* tweaks sass to remove up-down arrows in date_field

* Fixes #309


app/assets/stylesheets/application.css.scss
```diff
@@ -154,3 +154,12 @@ section {
   @extend .btn-primary;
   @extend .btn-xs;
 }
+
+// Search result 
+// shell styling
+.form-group input {
+  &[type="date"]::-webkit-inner-spin-button {
```
colinxfleming: :100: to you for figuring this out!

app/views/pregnancies/_new_pregnancy.html.erb
```diff
@@ -22,7 +22,7 @@
   </div>
 
   <div class="col-md-2">
-    <%= f.date_field :initial_call_date, label: "Initial Call Date", value: @today %>
+    <%= f.date_field :initial_call_date, label: "Initial Call Date", value: @today,  required: "required" %>
```
colinxfleming: The `required: 'required'` thing is interesting here! Is this a pattern we should use for the name and phone fields too?

colinxfleming: Actually, come to think of it, this touches on an open question we have on form validation that I'd rather not take action on at this time. Let's punt on this for now.

colinxfleming:
Thanks a ton for this, @sco-bo ! Really appreciate your working out that CSS black magic here!

app/views/pregnancies/_new_pregnancy.html.erb
sco-bo: @colinxfleming ya looking at it more closely I should have made it `required: true` as that's a little more readable. Didn't think of that last night but in the future that's an option.

Issue #327, opened 2016-06-04
Set up SSL certificate


colinxfleming:
@stvnrlly suggested using letsencrypt as an SSL solution -- setting this up to make sure getting SSL set up is on our radar, and marking as admin until we figure out what the discrete steps are. 

Putting this in the parking lot for now as well: https://github.com/lgromanowski/letsencrypt-plugin

cmgiven:
Since you're using Docker Compose, a really easy way to get this done is: https://github.com/DanielDent/docker-nginx-ssl-proxy

It's an HTTPS proxy that takes care of obtaining and renewing LE certs. It sets the `x-forwarded-proto` header, so `force_ssl` in Rails should work fine. Maybe there are issues with using a proxy that I'm unaware of, though (I've never tried this with a Rails setup).

cmgiven:
Example docker-compose file: https://github.com/cmgiven/justgrimes/blob/master/docker-compose.yml

colinxfleming:
Changing this to a beginner backend task. 

We need to alter our app to use SSL. This has three steps:

* Go thru https://github.com/DanielDent/docker-nginx-ssl-proxy and set up SSL
* Edit config files in rails to set `force_ssl` to true (this should be easily googleable)
* Tweak our `docker-compose` file to incorporate the changes above, using @cmgiven 's as an example

Kevin-Wei:
PR includes to way to enable it sitewide

colinxfleming:
Took care of this heroku-side this evening. In the interest of not beating on it, I just went with the SSL endpoint option within heroku ($20) and the SSLEndpoint module ($15).

After we hit MVP we can revisit for a cheaper option, but I'm comfortable throwing money at it for a month or two.

PR #364, opened 2016-06-09
force ssl in config

Kevin-Wei:

* force ssl in production

* Deals with #327  Rails side

see notes http://www.eq8.eu/blogs/14-config-force_ssl-is-different-than-controller-force_ssl
(can be done controller side if not needed for whole site)


colinxfleming:
Thanks a ton for this switch @Kevin-Wei ! 

I'll merge once we have the docker/domain config adjustments made as well.

colinxfleming:
Going to wind up rolling SSL stuff into some other docker stuff, so just getting this in on the sooner side. Thanks for making this tweak @Kevin-Wei !

PR #373, opened 2016-06-17
Add Checkbox and attempt at bootstrap toggle 

### LINKED ISSUE #365: Need 'Voicemail OK?' check box somewhere on the patient page
@dfelsent helped me realize that we don't have a 'Voicemail OK?' check box any where on the patient info pages. Can we add the check mark in on the Patient Info page? 

image

sco-bo:

Added a checkbox to the patient_information partial and had a go at trying to replace the checkboxes with bootstrap toggle

* Adds checkbox to patient_information 
* Mocks up potential code for replacing bootstrap checkboxes with bootstrap toggles on the patient information and search pages.

* Fixes #365

### DIFF
--- a/app/views/pregnancies/_new_pregnancy.html.erb
+++ b/app/views/pregnancies/_new_pregnancy.html.erb
@@ -27,8 +27,16 @@
 
   <div class="col-sm-2">
     <%= f.form_group :voicemail_ok, label: { text: 'Voicemail OK?' } do %>
-    <%= f.check_box :voicemail_ok, label: 'Yes' %> <!-- TODO add bootstrap toggle -->
+      <%= f.check_box :voicemail_ok, label: 'Yes' %> 
     <% end %>
+    <!-- This is one attempt at the bootstrap toggle -->
+    <%# f.form_group :voicemail_ok, label: { text: 'Voicemail OK?' }, data: {toggle: 'toggle', on: 'Yes', off: 'No' } do %>
+      <%# f.check_box :voicemail_ok, label: 'Yes' %>
+    <%# end %>
+    <!-- Here is another attempt, toggle not displaying at all -->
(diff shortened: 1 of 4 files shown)


colinxfleming:
Thanks for this @sco-bo ! I'll put the nail in the coffin this weekend.

sco-bo:
Sounds good @colinxfleming. Looking forward to learning the best way to to approach this

app/views/pregnancies/_new_pregnancy.html.erb
```diff
@@ -27,8 +27,16 @@
 
   <div class="col-sm-2">
     <%= f.form_group :voicemail_ok, label: { text: 'Voicemail OK?' } do %>
-    <%= f.check_box :voicemail_ok, label: 'Yes' %> <!-- TODO add bootstrap toggle -->
+      <%= f.check_box :voicemail_ok, label: 'Yes' %> 
     <% end %>
+    <!-- This is one attempt at the bootstrap toggle -->
+    <%# f.form_group :voicemail_ok, label: { text: 'Voicemail OK?' }, data: {toggle: 'toggle', on: 'Yes', off: 'No' } do %>
```
colinxfleming: hey! So I think what's going on here can be best summarized as: javascript is awful. 

When you pop in new HTML via ajax, javascript doesn't necessarily execute on it unless you specifically tell it to. So the page loads, the bootstrap toggle script fires, and then it just sits there. It continues sitting there even as new data-toggle elements get added. 

So basically, the fix here is to add `<script>$('[data-toggle="toggle"]').bootstrapToggle();</script>` to the bottom of the new pregnancy partial, which will tell the check boxes with data-toggle to be toggle switches. 

I have lost what feels like years of my life to this very problem! Sorry you ran into it.

In conclusion, javascript :(

Issue #277, opened 2016-05-09
Models: Turn auditability into a concern, instead of having it separately in every model


colinxfleming:
Here is an opportunity to use the Rails feature `concerns` for someone interested in taking a stab!

In just about every one of our models, we have the following code:

```
  # History and auditing
  track_history on: fields.keys + [:updated_by_id],
                version_field: :version,
                track_create: true,
                track_update: true,
                track_destroy: true
  mongoid_userstamp user_model: 'User'
```

Rails has a feature in Rails 4 called `Concerns` which are basically helpers for models. Here's an SO thread about them: http://stackoverflow.com/questions/14541823/how-to-use-concerns-in-rails-4

In a nutshell, we want the code above popped into a Concern called `Auditable`, and instead of having the same seven lines of code in every model, we want `includes Auditable` to do the same thing.

This will keep us from driving ourselves crazy if we ever have to change any of that code.

ewiggins:
@colinxfleming I'm going to work on this! Seems like a fun mini project for the larger project :) p.s. I just took a GitHub tutorial so I'm a pretty much a pro. lol 💯

colinxfleming:
Awesome, thanks Elisheba! Let me know if you run into any snags or need a
hand with the documentation.

ewiggins:
@colinxfleming I give up on this. Maybe some other smart beginner can figure it out :) I'm going to work on something else.

colinxfleming:
@ewiggins I'm gonna take yr PR and run with it! Sorry that this turned out to be way more of a pain in the ass than I thought it was going to be.

colinxfleming:
Turns out concerns are a lying demon hellchild. Closing as a wontfix.

ewiggins:
I know right! I don't usually quit things, but I was starting to hate concerns with a passion!

PR #392, opened 2016-06-22
Creating Concerns for Models #277

ewiggins:

* Adds the audible.rb to concerns
*  calls concern in each model

* Fixes#277


colinxfleming:
@ewiggins I'm really sorry for sending you on this wild goose chase. If it's any consolation I just spent an hour beating on it and couldn't get it to work right either. Looks like this is something we might just have to revisit if we ever get to rails 5. 

I'm closing the associated issue as a wontfix.

PR #420, opened 2016-06-25
Fix most CI false flag failures

### LINKED ISSUE #375: Rewrite call logging integration test so it doesn't randomly fail
Our capybara tests are occasionally failing randomly. I suspect this is either an issue in my writing sketch integration tests or with PhantomJS just being a jerk sometimes for no reason. Making a note to investigate.

colinxfleming:

Turns out minitest is faster than the webdriver, so tests are randomly failing once in awhile.

Here is a blog post about the general principle about why they're failing - https://nulogy.com/who-we-are/company-blog/articles/sleep-is-for-the-weak/

* Fixes integration tests to wait for elements instead of just pushing ahead, always pushing ahead

* Bumps #375

### DIFF
--- a/app/views/pregnancies/_notes.html.erb
+++ b/app/views/pregnancies/_notes.html.erb
@@ -4,23 +4,23 @@
 			<div class="info-form-left col-sm-12">
 				<h2>Notes</h2>
 				<%= bootstrap_form_for [pregnancy, note] do |f| %>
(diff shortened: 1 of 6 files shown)


colinxfleming:
The integration test for logging calls is still consistently being a jerk, but we seem to be out of the woods on the other random failures. I'm going to merge in the interest of solving 90% of the problem.

Issue #403, opened 2016-06-24
Edit changelog entries to appear in reverse chronological order


eheintzelman:
Change log is in chronological order, whereas Call Log is in reverse chron order.  Can we make these consistent?

How about making the Change Log reverse chron.

colinxfleming:
This is likely a pretty straightforward change in `app/views/pregnancies/_change_log.html.erb`! Marking as beginner friendly, since I think it's a good entrypoint into our codebase. 

Compare to `app/views/pregnancies/_notes.html.erb` for an example of the reverse chron.

mebates:
+1. Thanks @eheintzelman

KatSDC:
I have a solution for this! Just tying up loose ends and then I'll submit the PR.

PR #436, opened 2016-06-28
Make search cast a wider net #318

### LINKED ISSUE #318: Make search cast a wider net
Right now, our search code casts a pretty strict net, requiring someone to enter in exactly what's in database to retrieve a record. 

Here are a few ways I'd like to loosen it before release:

* Make it hyphen agnostic (e.g. when searching for '111-222-3333', it'll find the record if '1112223333' is entered into the form too
* Make it case insensitive (e.g. 'susie everyteen' will return the record for Susie Everyteen)

KatSDC:

I made the search both case insensitive for the name field and formatting agnostic for the phone number fields.

* app/models/patient.rb

* Fixes #318


app/models/patient.rb
```diff
@@ -32,10 +32,16 @@ class Patient
...
   # Methods
+
+
   def self.search(name_or_phone_str) # TODO: optimize
-    name_matches = Patient.where name: name_or_phone_str
-    primary_matches = Patient.where primary_phone: name_or_phone_str
-    secondary_matches = Patient.where secondary_phone: name_or_phone_str
-    (name_matches | primary_matches | secondary_matches)
+    clean_phone = name_or_phone_str.gsub(/\D/, '')
+    phone = clean_phone[0..2] + '-' + clean_phone[3..5] + '-' + clean_phone[6..10]
```
colinxfleming: @KatSDC What do you think about renaming `phone` to `formatted_phone`? My inclination is to err on the side of being as clear as possible, and just calling it `phone` is a little vague. Thoughts?

app/models/patient.rb
```diff
@@ -32,10 +32,16 @@ class Patient
...
+    begin
+      name_matches = Patient.where ({name: /^#{Regexp.escape(name_or_phone_str)}$/i})
```
colinxfleming: For style/consistency's sake, can we remove the `({` and `})` around the hash here? It'll work either way, just want to keep it tidy I think!

KatSDC: @colinxfleming Happy to make these changes! I also found once I upgraded rails that search was throwing an error unless I converted these parts (ex: clean_phone[0..2]) to a string with the to_s method. Additionally I have a solution for the chronological order of the change log for issue #403 - should I submit that along with this item or would you prefer that in its own PR?

colinxfleming:
Looks like tests were being a little rude to us last night for a reason -- if you send it a name-looking thing, it'll return an error. Minor fix though!

Try it in ruby:

```ruby
def search(name_or_phone_str)
  clean_phone = name_or_phone_str.gsub(/\D/, '')
  puts clean_phone
  puts clean_phone[0..2] + '-' + clean_phone[3..5] + '-' + clean_phone[6..10]
end

puts search('colin fleming')
```

That'll yield an error, because when we clean up the string, it's reducing it to something that's not a phone, and then dying because there's nothing at `clean_phone[3..5]`. 

Here are two ways to fix it, I defer to you on which one you prefer:

* write a conditional function to check that the length is appropriate before formatting it as a phone (something like: `phone = if clean_phone.length == 10 then clean_phone[0..2]... else name_or_phone_str`
* Use string interpolation (fancy way to say `"#{variable}"`) instead, which automatically runs `#to_s` on everything you feed it. So instead of `phone = clean_phone[0..2] + '-' + ...` you could do `phone = "#{clean_phone[0..2]}-#{clean_phone[3..5]}..."`

Let me know if I can clarify any of this. And thank you a ton for getting this moving, it's going to make @lwaldsc 's life a lot easier in two weeks!

app/models/patient.rb
colinxfleming: Please submit the #403 thing in its own PR! The smaller the pull request the better, is my philosophy.

colinxfleming: (And I see that you beat me to the thing I posted about below! Long-winded explanation of the problem and how to solve it is below, let me know if you have any questions or if I can unstick anything.)

colinxfleming:
Thanks for the quick action @KatSDC ! Glad to have you on the team.

Issue #388, opened 2016-06-20
Turn off 'Remember me' option on sign in


colinxfleming:
Nah.

colinxfleming:
Fixed in #444

Issue #379, opened 2016-06-20
Turn on Devise::Timeoutable


colinxfleming:
For an extra layer of security, we should make it so our sessions time out. We can do this using stuff from within our current devise install, I believe. Let's set at 2 hours initially, unless @lwaldsc or @NerdyGirl537 think we can reduce it to below that. 

Action items:
* Add `:timeoutable` to user model
* Figure out how to set the timeout at 2 hours, then set it at 2 hours

colinxfleming:
Closed in #445

PR #437, opened 2016-06-28
Rate-limit login form #381

### LINKED ISSUE #381: Rate-limit login form
Right now (AFAIK) we don't rate-limit our login form, which means that you could feasibly just brute-force a password and login. We should implement some kind of rate-limiting that makes you cool off for 15 minutes after a few failed logins. 

I'm sure this is something devise does, either with an extension or with something else. 

(For all I know, this is already turned on! but we should confirm that's the case if so.)

ashlynnpai:

* configure lockable in initializers/devise.rb for account lock for 15 mins after 5 failed logins
* enable lockable module in user.rb
* enable lockable fields in user.rb

* Fixes #381


app/models/user.rb
```diff
@@ -8,9 +8,9 @@ class User
           :recoverable,
           :rememberable,
           :trackable,
-          :validatable
+          :validatable,
+          :lockable
```
colinxfleming: 💯

bin/rails
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
colinxfleming: Can you put this block back in? Would like to keep `spring` in if we can!

bin/rake
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
colinxfleming: Same deal -- can you pop `spring` back in?

config/initializers/devise.rb
```diff
@@ -158,7 +158,7 @@
   # Defines which strategy will be used to lock an account.
   # :failed_attempts = Locks an account after a number of failed attempts to sign in.
   # :none            = No lock strategy. You should handle locking by yourself.
-  # config.lock_strategy = :failed_attempts
+  config.lock_strategy = :failed_attempts
```
colinxfleming: yussss

config/puma.rb
```diff
@@ -11,5 +11,7 @@
 on_worker_boot do
   # Worker specific setup for Rails 4.1+
   # See: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot
-  ActiveRecord::Base.establish_connection
+  ActiveSupport.on_load(:active_record) do
```
colinxfleming: You can probably revert this change too -- I switched us back to `unicorn` because puma was giving me some unrelated guff.

colinxfleming:
Love this, thank you so much for your work here @ashlynnpai ! Happy you got to enjoy Mongo's best feature, beating the tar out of the models on the fly. 💯 

Had a few comments, and it looks like there are some merge conflicts to resolve, but I think after that we can declare victory and merge this up!

ashlynnpai:
Please let me know if I cleaned up the merge conflicts okay.

colinxfleming:
Looks good! Thanks for your quick action @ashlynnpai !

colinxfleming:
(The CI is failing because of a lousy integration test I wrote; this doesn't affect actual logging calls functionality.)

colinxfleming:
@ashlynnpai we have a couple of similar issues to this (wrenching on devise configs) in #388 (disabling `:rememberable`) and #379 (turning on `:timeoutable`). Do you want to take a stab at those to play around with devise a little more?

ashlynnpai:
I'll take a look.

PR #442, opened 2016-06-30
Ajaxify pregnancy edit view and note creation

### LINKED ISSUE #308: Create notes with ajax instead of synchronously
* **What screen are you on?** Pregnancy edit page, specifically the notes partial

* **What are you trying to do?** Create a new note

* **What should be happening?** It should create a new note with javascript, and not reload the page

* **What's happening instead?** It's reloading the page.

The stuff to edit is in `NotesController#create`, specifically L9-14. We'll want to make that look more like the update method and wire it to handle an ajax request. We'll also want to adjust the notes form in `app/views/pregnancies/_notes.html.erb` to handle the remote call.

### LINKED ISSUE #404: Pregnancy edit form should save via ajax
(TK technical requirements)

colinxfleming:

What started as just changing a controller to respond to js turned into an overhaul of the pregnancy edit form structure. 

TK: Fixing tests. But I wanted to get the PR in early so team ux knew this was coming down the pike, and so nobody else burned time on those issues.

Also TK: Adjusting `notes/create.js.erb` to append or reload the partial.

* Makes every pregnancy partial have its own form, to avoid form crosstalk
* Makes those forms submit on change
* Moves those forms and controller to be ajaxy
* Adjusts notes form and controller to be ajaxy too

* Fixes #308 
* Fixes #404

### DIFF
--- a/app/assets/javascripts/pregnancies.coffee
+++ b/app/assets/javascripts/pregnancies.coffee
@@ -2,7 +2,11 @@
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
 
+
 $(document).on "click", "#toggle-call-log", ->
   $(".old-calls").toggleClass("hidden")
   html = if $(".old-calls").hasClass("hidden") then "View all calls" else "Limit list"
   $("#toggle-call-log").html(html)
+
+$(document).on "change", ".edit_pregnancy", ->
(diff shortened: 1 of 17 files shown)


colinxfleming:
Got everything to play nice, finally. A few of the known problem children integration tests are being jerks, but the functionality's all there, so I'm going to merge rather than continue to cause trouble.

PR #443, opened 2016-06-30
Fix issues #400 and #399, removing 'add/remove' links from Completed …

### LINKED ISSUE #399: Urgent cases list:  question about add/remove
What do folks think about the use of add/remove for the urgent cases list?  I'm concerned that a user could interpret this as add/remove from the urgent cases list, not the call list as intended.    

@mebates @lwaldsc @NerdyGirl537 your thoughts?

### LINKED ISSUE #400: Strip `add/remove` from the Completed Calls section

sco-bo:

Made a small change to `_table_content.html.erb`, removing the 'add/remove' links to both 'Completed Calls' and 'Urgent Cases'

* Removes 'Add/Remove' links from 'Completed Calls' View
* Removes 'Add/Remove' links from 'Urgent Cases' View

* Fixes #400 
* Fixes #399


colinxfleming:
Hey @sco-bo ! I think this takes a slightly heavier approach. That table content partial gets used for the call list, completed calls, and search results as well -- see `app/views/dashboards/index.html.erb` -- and we want to be able to keep the `add/remove` in the call list or search results.

What are yr thoughts about restoring those changes and adding some conditional logic? If you glance at `app/views/dashboards/index` you'll see that the `title` local gets passed in, I'd bet you could rig up something where the add/remove only gets populated with certain values. 

Let me know if I can clarify further here -- sorry for the sketch information in the issue, that's my fault!

sco-bo:
Oh I see, my bad. I'll take a look and try to throw something up this weekend :)

sco-bo:
@colinxfleming thank you! That clarification totally helps and makes sense. So if I understand correctly, the add/remove links should only be removed from the 'Completed Calls' and 'Urgent Cases' tables?

colinxfleming:
Nailed it! My reasoning is that by the time a call is completed there's no
sense in taking it off the call list. Similarly, the Urgent Cases is its
own thing that exists kind of separately, where interaction with the call
list isn't really relevant.

I should flag -- let's make sure the `td` tags stay in there no matter
what, so it doesn't break the table formatting.

Thanks a ton for knocking this out, bud!

sco-bo:
Hey @colinxfleming! So I have an implementation of this which puts in an empty td tag. Check out the screenshot below though; because there's no placeholder, the table formatting is thrown off a little. Any suggestions on how to make the cell the same size if there's no content? I will continue to look around for a potential solution. Thanks!
screen shot 2016-07-06 at 9 08 45 am

colinxfleming:
@sco-bo my inclination is to get this in and handle styling in a separate
PR, handled by someone who rules at CSS or bootstrap styling. Your thoughts?

sco-bo:
@colinxfleming ok that works for me

test/integration/urgent_cases_test.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+
+class UrgentCasesTest < ActionDispatch::IntegrationTest
```
colinxfleming: 💯 tests 💯

colinxfleming:
Thanks for pushing this over the finish line and for adding tests!

sco-bo:
You bet!

PR #444, opened 2016-06-30
Turn off 'Remember me' option on sign in #388

ashlynnpai:

Turn off rememberable module and fields in user.rb

* Turn off rememberable module and fields in user.rb
* Left the rememberable settings in devise.rb untouched 

* Turn off 'Remember me' option on sign in #388

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,10 +6,10 @@ class User
   devise  :database_authenticatable,
           :registerable,
           :recoverable,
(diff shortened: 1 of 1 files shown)


colinxfleming:
Heroic! 

Pretty sure the test failing here is just a test I didn't write well, so most likely we can merge. I'll rebuild the CI just to be on the safe side. 

Thanks so much for handling @ashlynnpai !

PR #445, opened 2016-06-30
Turn on Devise::Timeoutable #379

ashlynnpai:

Configure Devise timeoutable.

* add timeoutable module to user.rb
* set timeout to 2 hours in devise.rb

* Turn on Devise::Timeoutable #379

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -152,7 +152,7 @@ Devise.setup do |config|
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
   # time the user will be asked for credentials again. Default is 30 minutes.
-  # config.timeout_in = 30.minutes
+   config.timeout_in = 2.hours
 
   # ==> Configuration for :lockable
   # Defines which strategy will be used to lock an account.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,10 +6,10 @@ class User
   devise  :database_authenticatable,
           :registerable,
           :recoverable,
-          :rememberable,
           :trackable,
           :validatable,
           :lockable
+  # :rememberable
   # :confirmable
   # :timeoutable
 
@@ -31,7 +31,7 @@ class User
   field :reset_password_sent_at, type: Time
 
   ## Rememberable
-  field :remember_created_at, type: Time
+  # field :remember_created_at, type: Time
 
   ## Trackable
   field :sign_in_count,      type: Integer, default: 0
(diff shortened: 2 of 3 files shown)


config/initializers/devise.rb
```diff
@@ -152,7 +152,7 @@
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
   # time the user will be asked for credentials again. Default is 30 minutes.
-  # config.timeout_in = 30.minutes
+   config.timeout_in = 2.hours
```
colinxfleming: 💯 good call on setting this high!

colinxfleming:
Code looks good, @ashlynnpai ! Looks like it's making a bunch of tests mass-fail though. Can you still log in okay if you do this from a local server, etc? (You might have to run `rake db:seed` to reset db info first.)

ashlynnpai:
@colinxfleming I thought I had pulled the test errors from upstream but I could be wrong. I will take another look in a couple of hours.

ashlynnpai:
@colinxfleming I had to install phantom.js to run the test suite properly. I'm now getting the one failing test (the timecop freeze for 8 hours). I can reach the log in page and log in fine. I do see all those tests failing on CI but unable to reproduce.
calllistfail

colinxfleming:
That's nuts! Let me try rebuilding the CI, hang tight.

colinxfleming:
I'm getting the same results when I rerun on CI, and when I run them locally. Weeeeird.

@ashlynnpai I don't think this is you, I think this is something stupid with either Timecop or Capybara and how it interacts with Timeoutable. I'm going to let this hang for a min while I get this sorted out, but this PR should be fine as is.

ashlynnpai:
@colinxfleming 

I tried adding `Capybara.reset_sessions!` to the call list test and I got three passing test suites in a row with only the timecop test failing. I will also try commenting out the timecop test and removing the `reset_sessions`.

with reset session

ashlynnpai:
@colinxfleming Commenting out the Timecop test and also the reset_sessions code, I have three passing runs of the test suite. 

remove timecop test

colinxfleming:
So am I correct in understanding that the timecop test is making pretty
much everything blow up?

ashlynnpai:
Those are the results I'm getting, unless you add  `after { Capybara.reset_sessions! }`. If you have time to reproduce I would like to find out if you get the same result.

ashlynnpai:
I see trees of green

colinxfleming:
@ashlynnpai you are amazing for figuring this out! Thank you so much for diagnosing this problem with our test suite -- I'm going to create an issue to fix that timecop test, and merge this in. 

💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯 💯

PR #470, opened 2016-07-03
Install codesake-dawn gem 

### LINKED ISSUE #390: Install codesake-dawn gem
In the interest of catching more security stuff, we should install and implement the `codesake-dawn` gem into our testing process!

The steps here are roughly as follows: 
* Install the gem
* Figure out how to run it from the command line (probably something like `bundle exec codesake-dawn`)
* Add a command to `circle.yml` so that the command returns an error if `codesake-dawn` flags anything. Feel free to use the convention in the file (e.g. `gem install --no-rdoc...`) This will make it so that code with security warnings doesn't get deployed. 

Tagging as beginner friendly since this is a cool way to find out more about continuous integration in our environment!

ewiggins:

* Added codesnake gem to gem file
*  added command line code for it to run in circle.yml
* (etc)

* Fixes #390


ewiggins:
Soo this code is supposed to fail circleci if there are vulnerabilities, and there are. So if it fails circleci it's good?

ewiggins:
Where code for codesnake came from. https://github.com/thesp0nge/dawnscanner

circle.yml
```diff
@@ -13,4 +13,5 @@ test:
     - gem install --no-rdoc brakeman; bundle exec brakeman --exit-on-warn .
     - gem install --no-rdoc ruby_audit; bundle exec ruby-audit check
     - gem install --no-rdoc bundler-audit; bundle-audit update; bundle-audit check
+    - gem install --no-rdoc dawnscanner
```
colinxfleming: $$$

Once we have it downloaded in CI, we'll want to run it. Can you add a `bundle exec dawn --exit-on-warn .` to this, so the build runs it and fails if there's anything sketch?

Gemfile
```diff
@@ -58,4 +58,8 @@ end
 group :production do
   gem 'unicorn'
   gem 'rails_12factor'
+
+group :development do
```
colinxfleming: There's another `development` group in the Gemfile, around line 26. Want to just chuck the dawnscanner gem in there for cleanliness' sake?

circle.yml
colinxfleming: (on the same line, separated by a semicolon! so the full line should be:

`- gem install --no-rdoc dawnscanner; bundle exec dawn ...` etc)

colinxfleming:
Hi Elisheba! Had a couple comments, looking good. 

It looks like there's some leftover cruft from the auditable PR, but I'll resolve that over here once we have the `dawnscanner` related stuff squared away.

ewiggins:
@colinxfleming Ok. Let me work through removing the last commits. If I don't get it by Wednesday, then you can remove them from circleci. :) I want to try and do it myself from the command line.

ewiggins:
Rebase didn't work -http://www-cs-students.stanford.edu/~blynn/gitmagic/ch05.html#_8230_and_then_some

circle.yml
```diff
@@ -13,4 +13,5 @@ test:
     - gem install --no-rdoc brakeman; bundle exec brakeman --exit-on-warn .
     - gem install --no-rdoc ruby_audit; bundle exec ruby-audit check
     - gem install --no-rdoc bundler-audit; bundle-audit update; bundle-audit check
+    - gem install --no-rdoc dawnscanner; bundle exec dawn --exit-on-warn
```
colinxfleming: 💯

PR #474, opened 2016-07-04
Add new patients to call list after creation, populate search result with info

### LINKED ISSUE #397: Build call list:  automatically populate search result with name/phone number
After searching for a patient by name or phone number, the user gets prompted to create a new patient if the search produced no results.  Could the form field for name/phone number get automatically populated based on the search input (i.e., either name or phone number)?

### LINKED ISSUE #398: Make new patients show up in the call list after creation
After creating a new patient using the build call list function, the new patient does not show up in "your call list."

colinxfleming:

This PR adds a few case manager conveniences that are really, really, really annoying. Specifically, the new pregnancy form autopopulates the search params, and after you create a patient they get added to the call list.

* Moves `add_pregnancy` and `remove_pregnancy` to be model methods
* Tweaks dashboards controller to pass along name or phone
* Calls `User.add_pregnancy` within `PregnanciesController#create`
* Adds tests! Glorious tests

* Fixes #397 
* Fixes #398

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -6,13 +6,24 @@ class DashboardsController < ApplicationController
   def search
     patients = Patient.search params[:search]
     @results = []
-    patients.each do |patient|
-      @results << patient.pregnancies.most_recent
-    end
-    @patient = Patient.new
-    @pregnancy = @patient.pregnancies.new
+    patients.each { |patient| @results << patient.pregnancies.most_recent }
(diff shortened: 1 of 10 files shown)


colinxfleming:
Last thing I have: Can you do me a solid adjust the sample user in `db/seeds.rb` so that it has a valid password, and that the password is correctly logged into the terminal at the bottom? Want to make sure we can still flatten and reset our data for QA purposes.

Issue #475, opened 2016-07-04
Nice to have: Write a script to import spreadsheet data into the system


colinxfleming:
Since Lisa will be picking up the work of the person who came before her, we'll want to give her a little bit of seed data so she can have a more realistic experience of what using the system is like in a live-fire situation. 

Let's plan on writing some kind of script (and checking it into version control so we can iterate on it as we continue to roll out) that will let us take a spreadsheet and dump it into the system.

NerdyGirl537:
@colinxfleming @lwaldsc if you don't have time to get to the script I'll help input the data by hand.

colinxfleming:
@NerdyGirl537 It's a problem we're going to have a few times in the next few months, so I want to make sure we make some forward progress on it. If you're coming tomorrow I might stick you with a dev and have you pair on it to get it going.

colinxfleming:
Untagging deploy since this has the potential to be a very messy operation, sad to say.

colinxfleming:
I'm marking this backlog for two reasons:
* it's not necessary for release to DC
* right now I think we should be focusing on features that bring CM happiness.

colinxfleming:
Overtaken by events, now. Closing.

PR #483, opened 2016-07-07
WIP - Excel import functionality

MiroFurtado:
From #475-- this implements the excel import. 
Updated seeds.rb script to allow importing data from a csv file.
Model csv file can be found at db/patients.csv
Usage for set up is "rake db:seed FILENAME" or just "rake db:seed" to use the previously implemented functionality.

### DIFF
--- a/db/patients.csv
+++ b/db/patients.csv
@@ -1,0 +1,4 @@
+Name,primary_phone,initial_call_date,urgent_flag,last_menstrual_period_weeks,last_menstrual_period_days,left_voicemail_num,reached_patient_num
+Patient McPatient,202-363-1226,"Sat, 27 Oct 2012",TRUE,2,1,1,0
+Jane Doe,202-385-3902,"Sat, 27 Oct 2012",FALSE,3,2,0,2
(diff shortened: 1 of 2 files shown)


colinxfleming:
Thanks a ton for getting this started up! I'm gonna let this hang until tomorrow.

MiroFurtado:
Sure! It might be helpful to know exactly what sort of information we want to be importing from excel/what the spreadsheet would look like. We should also put together a spreadsheet with more model data in it.

NerdyGirl537:
Thoughts on import: We store city and state in the same cell.  They may or may not be separated by a comma.  Sometimes, there are multiple phone numbers in that cell.  The number of people in household isn't standardized -- sometimes it is a number; sometimes more details.

colinxfleming:
Increasingly, I'm coming to think that this is a thing we do for the VA/MD release or a wider release. We have a limited time between now and release, and I think we should focus that on CM features rather than programmatic data backlog entry, a problem at which we can throw nontechnical person power. 

Sorry for giving you bad guidance on this @MiroFurtado ! Let's have this PR continue to sit for the time being in case we want to pick it up after August 22.

PR #487, opened 2016-07-07
Pregnancy menu styling

### LINKED ISSUE #303: Remove blue highlight, add right-facing caret, increase font-size in pregnancies/_menu.html.erb
From @mebates in #195 - 

Currently, in the pregnancy edit view, you can see a full blue highlight over `Abortion Information`:

screen shot 2016-05-28 at 5 57 21 pm

We should adjust this to match the designs, which feature a purple caret instead and no blue highlight: 

screen shot 2016-05-28 at 5 59 50 pm

(See also https://github.com/colinxfleming/dcaf_case_management/blob/master/_design/DCAFwireframe020816.pdf ).

Probably this involves busting into `pregnancies.scss` and editing the `.abortion-left-menu.li.active` styling.

mchelen:
- increase menu & submit button font size
- change active menu background color
 i couldn't find the css being used for `.nav-pills > li.active > a` so i had to use `!important`
- add glyphicon
 implemented with only css, could also be done with an html `<span>` 

fixes #303

### DIFF
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -61,6 +61,10 @@ $font-size: 17px;
       
     }
   }
+  a.submit-btn {
+    font-size: 22px;
+  }
+
 }
 
 .border-side {
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -31,6 +31,8 @@ $font-size: 17px;
 }
 
 .abortion-left-menu {
+  font-size: 22px;
+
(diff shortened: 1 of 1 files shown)


colinxfleming:
💯 thank you so much for knocking this out @mchelen ! 

I'm going to spin out an issue from this to fix the `important`s but that is most definitely a later problem. Really appreciate your putting this together.

colinxfleming:
THIS LOOKS SO GOOD

screen shot 2016-07-09 at 2 17 23 am

colinxfleming:
Styling tag spun into #496

Issue #459, opened 2016-07-02
Pledge confirmation should specify what state user is in 


colinxfleming:
From #405 - 

> 6 - Pledge confirmation should specify what state the [patient] is in. [1 user] 

Let's add this to the pledges modal in `app/views/pregnancies/_pledge`. Nothing fancy.

NerdyGirl537:
In past, this has been part of the patient identifier: First Letter of State and last 5 of phone number (ex. D4-0258 for DC caller).

PR #489, opened 2016-07-07
From issue #459, pledge now specifies which state user is in.

MiroFurtado:
From #459, pledge specifies patient state. Also fixed some problematically coded tests for special circumstance input and LMP days.


colinxfleming:
@MiroFurtado not sure the entire PR made it in here, just looks like test edits -- can you push up any other changes you've made?

MiroFurtado:
@colinxfleming Yeah, forgot to commit the other part and then got swamped by other things, I'll do it by tonight.

colinxfleming:
Thanks much for this, @MiroFurtado ! 

Can you do me a few solids, and then we can get this in? 

a) Resolve the merge conflicts
b) it looks like there are a few files with `~`s at the end -- can you remove these? Looks like they're exact dupes of a few files, not sure what they're artifacts of.

An additional question for @lwaldsc and @NerdyGirl537 -- do we need state as in state of residence, or do we really need line?

PR #492, opened 2016-07-07
adds parameters to be filtered from loggin

### LINKED ISSUE #382: Filter out patient PII in system logs
DCAF works with a lot of sensitive information, some of which we should definitely not be logging. We can edit by adding fields to `config/initializers/filter_parameter_logging.rb`.

Let's plan on filtering anything that looks like kinda sensitive information. To start: 

* Patient: Name, primary phone, secondary person, secondary phone
* Pregnancy: Appointment date

Whoever takes on this ticket should also look at fields within the patient and pregnancy models and see if there's anything else that looks like it might pop. (The question you should ask yourself is: 'If I were a real garbage human being, what's the worst I could do with this information?' If it's something nasty, then we should filter it.)

CamLatimer:

I included params to filter on inside /config/initializers/filter_parameter_logging.rb, anything thought could be sensitive.  In issue 382, Colin mentioned filtering on the pregnancy and patient model, but I also thought the addresses from the clinic model might be good to add here too, as well as the full_text attribute from the notes model in case people add names and other info inside their notes. 

* Fixes #382

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,11 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+
+
+
+
+
+
+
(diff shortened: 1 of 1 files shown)


colinxfleming:
Thanks a ton for taking care of this, @CamLatimer !

PR #498, opened 2016-07-09
Turn off registration but keep edit

### LINKED ISSUE #497: Turn off account registration

colinxfleming:

Had to yakshave this a bit but got it going. 

**Until we figure out an appropriate way to handle in release 2 (probably administrator invitation), we'll need to create new users from the rails console.**

* Removes non-password-editing routes from devise
* Turns off frontend user signup
* Fixes tests

* Fixes #497

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,8 +9,7 @@ class User
           :trackable,
           :validatable,
           :lockable,
-          :timeoutable,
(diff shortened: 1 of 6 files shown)


colinxfleming:
Note that this was the last piece before spinning up the prod db, so that's next.

PR #519, opened 2016-07-17
Css load order

### LINKED ISSUE #496: Remove !important tags from pregnancy menu styling
We brute forced this through for MVP, but it shouldn't always be brute forced. A nice piece of code hospitality for someone who wants to take it on.

See `app/assets/stylesheets/pregnancies.scss` -- the `a, a:hover` tags contain an `!important` to force the color and background color, because we couldn't get the inheritance to chill out. The task here is to remove the important tags and recreate in a way that doesn't rely on the important tags. (HTML span tags? Figuring out where the inheritance is coming from? Up to you!)

mchelen:

- load custom css after bootstrap

* custom css was being overridden by bootstrap css because bootstrap was being loaded afterward
* this changes the css load order to move custom css later on
* allows `!important` to be removed

* Fixes #496

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -1,6 +1,7 @@
 /*
  *= require rails_bootstrap_forms
  *= require_tree ../../../vendor/assets/stylesheets/.
+ *= require_self
  *= require_tree .
 */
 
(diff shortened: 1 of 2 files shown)


colinxfleming:
Hero status! Thanks so much @mchelen !

PR #535, opened 2016-07-23
Add mark as resolved to views

### LINKED ISSUE #528: Mark patients as resolved without help from dcaf

colinxfleming:

Adds @ashlynnpai 's resolved without assistance from dcaf work to the views; alters integration tests appropriately as well.

I didn't have a better place to put it, so I tossed it as a checkbox under the `Abortion information` section. 

Screenshot: 

image

* Fixes #528


mebates:
Once they check that box, should we freeze the record somehow?

PR #545, opened 2016-07-26
Rehab seedfile

### LINKED ISSUE #543: Add call list in db seed
The default seeded db should contain entries for some of:
Your call list
Your completed calls

related to #539 #520

colinxfleming:

@mchelen wisely pointed out that

This pull request makes the following changes to the seedfile:
* Adds a second user, creates everything with the second user
* Adds two patients to call list and one to completed calls
* Adds a note, for good measure

Spits out the following after login:

image

@mchelen look good to you? 

* Fixes #543

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -3,13 +3,22 @@ Pregnancy.destroy_all
 Patient.destroy_all
 User.destroy_all
 
-user = User.create name: 'testuser', email: 'test@test.com', password: 'password', password_confirmation: 'password'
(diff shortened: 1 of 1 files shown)


mchelen:
Would it help for testing to have a few more calls in the call list? When doing reordering right now with 2 calls, there are only 2 possible orders they can be in.

colinxfleming:
@mchelen sure, I'll do that now (straight to master, life on the edge!)

PR #546, opened 2016-07-26
Search phone pattern

### LINKED ISSUE #512: Make search function work for partial phone numbers
As a reminder to @colinxfleming we need to make the search work for partial phone numbers.

colinxfleming:

Overhauls the search to look for phone or name patterns. Or, at least it will when I finish it. 

* Search on name or phone pattern (eventually)
* Stores phone as just number, no hyphens
* Adds a before validation hook that cleans up phone number, so we don't have to change the way we interact with phone
* Alters some tests

* Fixes #512 (or at least it will)


colinxfleming:
Tagging in @KatSDC for review and feedback. The major change to be aware of is that this now saves phone numbers as a string with no dashes, so we don't have to cheat our way through formatting phone numbers anymore. 

The reason tests are failing is that searching on an empty string is returning every non-nil answer. So `Patient.search 'Name'`, when it searches on phone, is returning all non-nil entries. Not sure what the right way to handle this is, welcome your thoughts @KatSDC . Thanks for your eyes!

colinxfleming:
We should be up to green! This also adds `primary_phone_display` and `other_phone_display` as options, which format the appropriate phone if it's not empty. Tests pass and we're looking good, just waiting for @KatSDC 's eyes.

colinxfleming:
I'm satisfied that this is working as intended, and the line that's not covered is a fallback, so I think this is good to merge. @lwaldsc consider your wish for partial phone search granted.

PR #547, opened 2016-07-26
Filter :search parameter from system logs

### LINKED ISSUE #529: Filter :search in system logs
Let's make sure to filter out `:search` from application logs, in addition to PII in our actual model fields. 

See #382 for the first pass at this -- should be as simple as adding another field.

benjamincrane:
I completed some work on Case Manager that's ready for review!

* Adds ", :search" to the list of filtered parameters fields in system logs

* Fixes #529


colinxfleming:
💯 

Thanks for taking care of this, bud!

PR #558, opened 2016-08-02
adds jQuery to close popups

### LINKED ISSUE #530: Dashboard: Last note doesn't disappear when you click away
In the dashboard view, last note doesn’t disappear when you click away -- you have to actively click on the + trigger, meaning you can create a cascading mess of note popups. We’ll want to add some kind of js trigger to close those note popups on clickaway.

jeffreyeaton76:
This adds some jQuery to dismiss the popups you see when you click a note in Dashboard view. Now when you click anywhere that is not in the active popup the popup is dismissed. This also prevents there from being more than one popup open at a time. 

* jQuery added to app/assets/javascripts/application.js
* It works but I'm too noobie to write a test for it

* Fixes #530

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -19,7 +19,7 @@
 //= require_tree ../../../vendor/assets/javascripts/.
 
 $(document).ready(function(){
-  // TODO migrate to pledges.coffee
+	// TODO migrate to pledges.coffee
 	$('.pledge_submit').click(function () {
 		$('.pledge_modal_screen1').hide();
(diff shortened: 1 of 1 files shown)


colinxfleming:
This looks excellent, thank you so much for this @jeffreyeaton76 ! Really appreciate your work on this! 

Works on my machine -- merging.

colinxfleming:
(PS I think this is minor enough that we don't need to write a test for it, since it's not preventing anyone from using a feature. So no stress on that front!)

PR #559, opened 2016-08-02
[WIP] Add drag and drop ui

### LINKED ISSUE #520: Make call list drag and droppable
During the pilot, @lwaldsc mentioned that she'd find it helpful to be able to rearrange her call list on the fly, rather than have them display in the order that they were added to her call list.

Open questions: 

* What's the use case for this? How do CMs organize their calls, and how should we reflect that in the app?
* What should we make them sortable on? Name, phone number, LMP, everything in the section?
* The possibility of doing this drag and drop was raised also -- is there something we get out of that which we don't get out of oneclick sorting?

mchelen:
- Call list can be reordered by drag & drop
- Generate an ordered array of row ID's to send to endpoint

* Fixes #520


app/assets/javascripts/calls.coffee
```diff
@@ -1,3 +1,40 @@
...
+        parent = $('#'+item_id).parent()
+        children = parent.children()
+        order = []
+        for child, value of children
+          if value.id? then order.push(value.id)
+        console.log(order)
+
+        $.ajax(
+          type: 'POST'
+          url: '/something/reorder_pregnancies'
```
mchelen: This is the last remaining part, setting the proper URL here and the post data structure.

app/assets/javascripts/calls.coffee
```diff
@@ -1,3 +1,40 @@
...
+          dataType: 'json'
+          data: { order }
+        )
```
mchelen: This is the last remaining part, setting the proper URL here and the post data structure.

colinxfleming: This looks slamming, thanks bud!

I was at work a little later than I anticipated today. I'll start plugging
our respective prs into each other tomorrow.

colinxfleming:
@mchelen I merged this branch into my other track, it's looking good with hardcoded ids. Have some further issues to sort out before we declare victory, but this is so, so, so awesome.

PR #563, opened 2016-08-05
Modal fix issue 506

### LINKED ISSUE #506: Pledge modal dismisses if you miss a button
Missing a button causes the pledge modal in the edit view to dismiss. Not sure why. Annoying. Needs further investigation.

drownedout:
I think this had to do with Bootstrap. When you clicked the "modal" class it would close by default.

So I added another div that covered the "modal" div - blocking the user from clicking it.

Fixes #506


mchelen:
#506

colinxfleming:
Thanks for handling this, @drownedout ! This was driving me up the wall.

Issue #118, opened 2016-03-01
Call list cases should time out after 8 hours


colinxfleming:
From discussions with @lwaldsc about call list depopulation behavior: 

* A case manager should begin their night with an empty call list 
* Any calls from the previous evening should time out after a certain amount of time 
* We can't do this based on user session because someone might have to restart and that would wipe away their call list

colinxfleming:
This came up during the pilot, so I'm promoting it to the front for release. I also want to spec out exactly how it should behave before we start programming on it. Tagging in @lwaldsc, @eheintzelman , @SabrinaRitacco   and @NerdyGirl537  about best way to implement here. 

Here are two system behavior options to get us started:
* Can we safely say that every day at like 4am, we should clear out everyone's call lists? (This is probably the path of least resistance.)
* We could also set it up so that a call on the call list timed out after 12 hours; eg. a patient added to the call list at 7pm would be removed at 7am.

colinxfleming:
topping for @lwaldsc @NerdyGirl537 @SabrinaRitacco @eheintzelman . I'm inclined to say that at 4am we run a scheduled system job that clears out every single call list no matter what - what do you all think?

mebates:
Makes sense to me.

NerdyGirl537:
What about looking at status?  If I haven't reached a PT, I want to keep them in the call list.  If they disappear, I might not remember who I didn't reach.  For the rest, 4 AM seems reasonable.

colinxfleming:
I think that's a useful insight from @NerdyGirl537 . I think we can turn this into a real task now, as follows:

* Add a `clear_call_list` method to the user model that iterates through every call in `user.pregnancies` and removes it if the patient hasn't been reached yet
* Create a rake task that calls the 'clear_call_list' on every user (see https://devcenter.heroku.com/articles/scheduler#defining-tasks for an example of how we'll use this)
* Flag for @colinxfleming when this is done and we'll make it a scheduled task via heroku's scheduler

mebates:
Hey guys - this issue is a bit confusing. Can you summarize where we landed on the rules about clearing calls (in plain language)? Need the info for reporting discussion... Thanks!

colinxfleming:
The logic here shook out as follows:

We remove you from a user's call list at 4am nightly if all of the following are true:
* you called the patient
* and reached them
* within the past 8 hours

mebates:
[...]

PR #570, opened 2016-08-07
rake task for clearing call list

ajohnson052:
* create a clear_call_list method on User
* create a clear_call_lists rake task
* create a STATUSES hash on Pregnancy

* Bumps #118

Worth noting: right now recent calls are those made within the last 8 hours. If we want to run this task at 4am, we may want to make some adjustments

### DIFF
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -6,6 +6,15 @@ class Pregnancy
   include Mongoid::Userstamp
   include LastMenstrualPeriodHelper
 
+  STATUSES = {
+    no_contact: 'No Contact Made',
+    needs_appt: 'Needs Appointment',
+    fundraising: 'Fundraising',
+    pledge_sent: 'Pledge Sent',
+    pledge_paid: 'Pledge Paid',
+    resolved: 'Resolved Without DCAF'
+  }
+
   # Relationships
   belongs_to :patient
   has_and_belongs_to_many :users, inverse_of: :pregnancies
@@ -104,17 +113,17 @@ class Pregnancy
 
   def status
     if resolved_without_dcaf?
-      'Resolved Without DCAF'
+      STATUSES[:resolved]
     # elsif pledge_status?(:paid)
-    #   status = "Pledge Paid"
+    #   STATUSES[:pledge_paid]
     elsif pledge_sent?
-      'Pledge sent'
+      STATUSES[:pledge_sent]
     elsif appointment_date
-      'Fundraising'
+      STATUSES[:fundraising]
     elsif contact_made?
-      'Needs Appointment'
+      STATUSES[:needs_appt]
     else
-      'No Contact Made'
+      STATUSES[:no_contact]
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -72,15 +72,11 @@ class User
   # AND they would otherwise be in the call list
 
   def recently_called_pregnancies
-    pregnancies.select { |p| recently_called?(p) }
+    pregnancies.select { |p| recently_called_by_user?(p) }
   end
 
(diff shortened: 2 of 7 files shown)


app/models/pregnancy.rb
```diff
@@ -5,6 +5,15 @@ class Pregnancy
   include Mongoid::Userstamp
   include LastMenstrualPeriodHelper
 
+  STATUSES = {
```
colinxfleming: Love this refactor

app/models/user.rb
```diff
@@ -80,4 +80,16 @@ def remove_pregnancy(pregnancy)
...
+
+  def clear_call_list
+    pregnancies.each do |p|
+      pregnancies.delete(p) if recently_reached?(p)
+    end
+  end
+
+  private
+
+  def recently_reached?(preg)
```
colinxfleming: what do you think about renaming this to `recently_reached_by_user`, for clarity's sake? Thoughts?

test/models/pregnancy_test.rb
```diff
@@ -57,35 +57,35 @@ class PregnancyTest < ActiveSupport::TestCase
 
     describe 'status method' do
       it 'should default to "No Contact Made" when a pregnancy has no calls' do
-        assert_equal 'No Contact Made', @pregnancy.status
+        assert_equal Pregnancy::STATUSES[:no_contact], @pregnancy.status
```
colinxfleming: Bumping all the tests to reflect your refactor is positively heroic. Thanks a ton for taking care of that!

colinxfleming:
Thanks @ajohnson052 , this is really outstanding! To hook it up, I'd just add `rake clear_call_lists` as a scheduled task, no? 

I think your flagging that `recent?` not really suiting our needs for this at 8 hours is a really on point . Right now the other place we're using that is in determining what calls wind up in the Completed Calls section, and which ones wind up in the regular Call List section. We have a few options here: 

* Take out the `recent?` as a condition to complete call lists, and clear them out if they've been reached at all ever
* Adjust `recent?` to be longer, like, 16h or so, which will also affect how long people stay in 'Completed Calls'. Might be a problem for CMs who make calls late and then make a few extra calls in the morning?
* Shard the `recent?` into a separate method so that we can dictate timeout behavior specifically for this one. For this, we could define it to be a week or something broad, so that people don't disappear from your call list for the length of your shift unless you've logged a Reached Patient call

I think I lean toward the first or third. @NerdyGirl537 @lwaldsc would love for you all to weigh in here.

app/models/user.rb
ajohnson052: That would be fine, but I was modeling it after recently_called? so I would want to change both to maintain consistency. 

To that effect, do you think these should be public or private methods? They should probably both be one or the other.

colinxfleming: I think making them both private methods is right on, since we don't really call them anywhere else, no?

colinxfleming: Also a +1 for changing the name of them both to `*_by_user` I think! My instinct is that `recently_called/reached` could mean that either someone @ dcaf generally called them also, so I think I'd err on the side of specific. Yr thoughts @ajohnson052 ?

colinxfleming:
Is this ready to rumble @ajohnson052 ?

colinxfleming:
Is this ready to rumble @ajohnson052 ? Looks good to my eyes but wanted to get an all clear from you first!

ajohnson052:
Yes, please merge. My comment about the 8 hour limit on recent calls is the only thing I wanted to flag.

colinxfleming:
💯 thank you so much @ajohnson052 ! I'm going to punt on how to do recent for now, since it's a UX decision, and merge this. A++ work!

Issue #415, opened 2016-06-24
Autoemail on email address and password changes


colinxfleming:
If not, we should! Let's make sure this is the case, so that users are alerted to account tampering.

colinxfleming:
Devise doesn't do this out of the box, but we can backlog and keep on deck for release 2.

Here's a set of directions to implement - https://github.com/plataformatec/devise/wiki/Notify-users-via-email-when-their-passwords-change

benjamincrane:
Looking at this one tonight.

PR #571, opened 2016-08-09
WIP - Autoemail on email address and password changes

benjamincrane:

Tried to get email notifications to be sent on password change.

*where applicable, follows steps from https://github.com/plataformatec/devise/wiki/Notify-users-via-email-when-their-passwords-change
*adds letter_opener gem

* Is in the vague direction of a fix for #415


colinxfleming:
hey bud! Good news, you're like SUPER close on this. I'll leave a couple marks around, but the short of it is this:

* Go ahead and revert the changes in `application.rb` and `development.rb` -- if memory serves, when an email sends, you can see what actually got sent (or whether anything got sent at all) by checking the rails server logs, no?
* I think you're going to want to do the `after_update` hook in the user model -- looks like it's missing right now and I think it's the last step of hooking this up. After that I'm reasonably sure this is going to work, and when you change a password, you'll see it in the user logs!

config/application.rb
```diff
@@ -9,6 +9,9 @@
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
+
+require 'letter_opener' if Rails.env.development?
```
colinxfleming: Go ahead and ditch this -- Rails knows to load everything in the Gemfile based on groups on app load, so so you're doing this twice by having letter opener in the `development` group in the actual Gemfile.

config/environments/development.rb
```diff
@@ -14,9 +14,11 @@
   config.action_controller.perform_caching = false
   config.serve_static_files                = false
 
-  # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
+  # Do care if the mailer can't send.
+  config.action_mailer.delivery_method = :letter_opener
```
colinxfleming: Go ahead and revert changes to this file as well -- we can make this work without raising delivery errors or using `letter_opener`.

Gemfile
```diff
@@ -29,6 +29,7 @@ group :development do
   gem 'brakeman', require: false
   gem 'ruby_audit', require: false
   gem 'bundler-audit', require: false
+  gem "letter_opener", :git => 'git://github.com/cavi21/letter_opener.git'
```
colinxfleming: Not sure we need this, so go ahead and ditch if you're in agreement.

test/mailers/user_mailer_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_helper'
+
+class UserMailerTest < ActionMailer::TestCase
+  test "password_changed" do
+    mail = UserMailer.password_changed
```
colinxfleming: `password_changed` takes an argument of user id, right?

app/models/user.rb
```diff
@@ -1,4 +1,5 @@
 class User
+
   include Mongoid::Document
   include Mongoid::Userstamp::User
```
colinxfleming: Chuck a `after_update :send_password_change_email, if: :needs_password_change_email?` in here (per https://github.com/plataformatec/devise/wiki/Notify-users-via-email-when-their-passwords-change ) and this should be in business!

colinxfleming:
Marked up -- @benjamincrane please do reach out if you have any questions or hit any snags! This is super close and I'm looking forward to having it in production when we launch.

colinxfleming:
Hey @benjamincrane , how is this looking?

benjamincrane:
I made a lot of the changes you suggested and based on the logs in my version an email was generated after I updated the password. I want to figure out a way to test it using my email address (if that's possible) but that may not be necessary.

config/environments/development.rb
```diff
@@ -14,9 +14,10 @@
   config.action_controller.perform_caching = false
   config.serve_static_files                = false
 
-  # Don't care if the mailer can't send.
+  # Do care if the mailer can't send.
```
colinxfleming: Would you mind reverting this back to don't, since `raise_delivery_errors` is still set to false? Don't want us to be rooting around in here six months from now and be like 'wtf'.

colinxfleming:
I'm reasonably sure that this will work as is from looking at it (and also if you've got an email posting to logs), so I'll branch off of your work here and deploy it to our staging server and test it there tomorrow night. Thanks a ton for pushing that part of this over!

test/mailers/user_mailer_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_helper'
+
+class UserMailerTest < ActionMailer::TestCase
+  test "password_changed" do
```
colinxfleming: kind of a minor thing, but can you do me a solid and follow the style of the other tests? e.g. `it 'should do a thing' do`.

test/mailers/user_mailer_test.rb
```diff
@@ -0,0 +1,12 @@
...
+    mail = UserMailer.password_changed(id)
```
colinxfleming: Without looking, I'm guessing the test is choking because `id` doesn't refer to anything. I'd suggest creating a user object (something like `user = create :user`) and then passing `user into the `password_changed` method. 

You're pretty much nailing the rest of the test!

app/models/user.rb
```diff
@@ -131,4 +134,12 @@ def recently_reached_by_user?(preg)
   def recently_called_by_user?(preg)
     preg.calls.any? { |call| call.created_by_id == id && call.recent? }
   end
+
+  def needs_password_change_email?
+    encrypted_password_changed? && persisted?
+  end
+
+  def send_password_change_email
+    UserMailer.password_changed(id).deliver
```
colinxfleming: Can you swap this `deliver` to `deliver_now`? (One less thing we'll have to fix when we migrate to rails 5!)

benjamincrane: I saw the "deprecated" warnings and changed it in my current local copy. I've just been struggling with that user id part this whole time so I haven't pushed it.

benjamincrane:
Well, that's kind of a weird error. I'm not sure how "Expected: 2" is correct when the test has 1 on line 45.

`      # confirm email got sent
assert_equal Devise.mailer.deliveries.count, 1`

colinxfleming:
I pushed this to staging, changed the testuser info to my email and reset the password, and looky here: 

image

I'm making a couple business-y edits (mostly to the text) in a separate branch, but this is gold. Thanks a ton for staying on this @benjamincrane , this is a substantial app security improvement you've whipped up here!

colinxfleming:
A quick PS on the confirm email got sent item -- that's an example of not great test writing by me! It used to be that we didn't really send other emails, but now we are, so there's not a guarantee that it'll only send one email ever during the entire test suite run. 

I've replaced it with an `assert_difference` which does something similar but scopes it much better. This was a good flag!

PR #584, opened 2016-08-17
Call modal for other contact 

### LINKED ISSUE #527: Represent Other Contact information in call modal
How should we represent the fields `other contact name` / `other contact phone` / `other contact relationship` in the call modal?

mchelen:

* Shows other contact name, relationship, and number if it exists
* Only checks if the other number exists. Should we check if name & relationship are blank? #585 

* Fixes #527

* We should add "other contact" info to DB seed #586


colinxfleming:
💯 thanks a ton @mchelen ! I'll review this over the next few days.

mebates:
Can you drop a screenshot of it in here, please?

mchelen:
image

app/views/calls/_new_call.html.erb
```diff
@@ -6,6 +6,15 @@
           <button type="button" class="close close-terms" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
           <h4 class="modal-title calls-request">Call <b><%= p.patient.name %></b> now:</h4>
           <br>
+          <% if !p.patient.other_contact.blank? && !p.patient.other_phone.blank? %>
+          <h4 class="modal-title calls-request calls-other-contact">
+            <%= p.patient.other_contact %>
+            <% if !p.patient.other_contact_relationship.blank? %> (<%= p.patient.other_contact_relationship %>) <% end %>
```
colinxfleming: rails-y style feedback -- the standard style guide (which we follow only loosely, but still) tends to favor positive truthy checks. You might think about rewriting this to use `present?` if you've got a sec, though it'll def be fine either way. 

(Also it's a nice opportunity to check out some of the stuff buried in activesupport -- http://api.rubyonrails.org/classes/Object.html#method-i-present-3F )

colinxfleming:
@mchelen I had the tiniest of style feedback, but this looks good to me. 

@mebates look good from your end? 

No data model changes, so I'll plan on merging on Saturday night or Sunday morning.

app/views/calls/_new_call.html.erb
mchelen: Oh yeah I'm happy to do that, I did a quick google to see how to check if a variable is blank and this is what it came up with :D

Issue #526, opened 2016-07-18
Represent patient notes in the call modal


colinxfleming:
@lwaldsc spent some time clicking into the pregnancy view before making a call, to look at the notes for an individual person. Should we find a way to include the full set of notes for a patient in the call modal?

colinxfleming:
@lwaldsc @eheintzelman 

We both think having the full set of notes in the modal is somewhat impractical and has the potential to get very messy very quickly

Could we have a Call PT button within the actual pregnancy file at the top of the left hand tabs menu? Mirroring the Send Pledge button at the bottom.

Then a CM could pick a PT on the list, read notes, and be able to call from inside the pregnancy view.

NerdyGirl537:
I think the full set of notes would be overwhelming/unwieldy on the main page.  This could be a training opportunity as well, as once the PT has been added to the call queue, the CM can just return to the main menu and do the call.

colinxfleming:
Sounds like the next step here is team UX insight. Tagging in @mebates

mebates:
I think both would be helpful.
1) A button within the pregnancy file to place a call. I think this makes sense on the "Call Log" section
2) A portion of the last note in the call modal - perhaps it could be limited by character numbers so you don't see a full novel, but maybe the first few lines with an ellipsis? That way it gives you a general idea of what is going on.

colinxfleming:
Sounds like a plan. I think @mebates is right on.

Action items here are: 

- [ ] Add a button to pop up a call modal in the `_call_log` partial. Should behave exactly like the one on the dashboard for now. 
- [ ] Add the first 400 characters of the most recent note to the new call modal underneath the links to log a call. See the note helper in `dashboards_helper` for some code you can adapt. 

tagging in frontend.

colinxfleming:
Feature is CM happiness related; tagging deploy

jeffreyeaton76:
I'm going to take a crack at this.

PR #588, opened 2016-08-17
Call log notes

jeffreyeaton76:
This adds notes to the call modal. It also adds the call modal to the Call Log partial BUT the notes are not rendering here. They ARE rendering when you access the call modal from the Dashboard. 

screen shot 2016-08-16 at 8 54 21 pm

* Relates to #526


colinxfleming:
Thanks much @jeffreyeaton76 ! I'll keep an eye on this in case you have any changes, and plan on getting it in NLT Saturday evening.

Issue #434, opened 2016-06-27
'Special circumstances' should be checkboxes, not a dropdown


mebates:
Patients can have more than one 'Special Circumstance' at a time. For MVP, you could eliminate the 'Other' plus text field for now and add it later. 

![image](https://cloud.githubusercontent.com/assets/6223901/16399170/132c0120-3c9e-11e6-86c3-02bfdf49ae02.png)

colinxfleming:
I'll revert this to a plain text field for the time being. 

In the long term, this probably involves setting up an associated special circumstance model with the options, and then doing a has_many :through. Or at least that's how you do it in postgres, might be a little different in Mongo. 

Backlogging for the time being.

eheintzelman:
The UX team would like to know if we can prioritize this so that the data is clean.  @mebates would especially like this!

colinxfleming:
Acknowledged; I'll see if I can spin something up tomorrow.

colinxfleming:
We should have this in place by showtime, marking deploy.

ashlynnpai:
Taking a stab at this.

colinxfleming:
Fixed by @ashlynnpai in #591

PR #591, opened 2016-08-17
'Special circumstances' should be checkboxes #434

ashlynnpai:
I tried to do this by just changing field type to array but I'm now wrangling with the form syntax, sticking labels on the checkboxes.

See for example code: http://stackoverflow.com/questions/21568957/using-checkboxes-to-submit-values-to-an-array

### DIFF
--- a/app/views/pregnancies/_patient_information.html.erb
+++ b/app/views/pregnancies/_patient_information.html.erb
@@ -78,8 +78,11 @@
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
           <div>       
-            <%= f.label :special_circumstances, "special circumstances" %>    
-            <%= check_box_tag('pregnancy[special_circumstances][]', "Fetal anomaly", false, class: "checkbox") %>
+            <%= f.label :special_circumstances, 'Special circumstances' %>    
+            <%= check_box_tag('pregnancy[special_circumstances][]', 'Fetal anomaly', false, class: 'checkbox') %>
+            <%= check_box_tag('pregnancy[special_circumstances][]', 'Rape', false, class: 'checkbox') %>
+            <%= check_box_tag('pregnancy[special_circumstances][]', 'Homelessness', false, class: 'checkbox') %>
+            <%= check_box_tag('pregnancy[special_circumstances][]', 'Incest', false, class: 'checkbox') %>
           </div>
         </div>
       </div>
--- a/app/views/pregnancies/_patient_information.html.erb
+++ b/app/views/pregnancies/_patient_information.html.erb
@@ -78,11 +78,51 @@
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
           <div>       
-            <%= f.label :special_circumstances, 'Special circumstances' %>    
-            <%= check_box_tag('pregnancy[special_circumstances][]', 'Fetal anomaly', false, class: 'checkbox') %>
-            <%= check_box_tag('pregnancy[special_circumstances][]', 'Rape', false, class: 'checkbox') %>
-            <%= check_box_tag('pregnancy[special_circumstances][]', 'Homelessness', false, class: 'checkbox') %>
-            <%= check_box_tag('pregnancy[special_circumstances][]', 'Incest', false, class: 'checkbox') %>
+            <%= f.label :special_circumstances, 'Special circumstances' %>   
+            <br />
+            <div>
+              <div class="col-sm-6">
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Fetal anomaly', 
+                            @pregnancy.special_circumstances.include?('Fetal anomaly')) %>
+                  <%= label_tag('Fetal anomaly') %>
+                </div>
+                <div>
+                <%= check_box_tag('pregnancy[special_circumstances][]', 'Homelessness', 
+                          @pregnancy.special_circumstances.include?('Homelessness')) %>
+                <%= label_tag('Homelessness') %>
+                </div>
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Domestic Violence', 
+                            @pregnancy.special_circumstances.include?('Domestic Violence')) %>  
+                  <%= label_tag('Domestic Violence') %>
+                </div>
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Other', 
+                            @pregnancy.special_circumstances.include?('Other')) %>
+                  <%= label_tag('Other') %>
+                </div>  
+                <%= text_field_tag 'pregnancy[special_circumstances][]', @pregnancy.special_circumstances.include?('Other'), 
+                          :placeholder => 'Details' %>            
+              </div>
+              <div class="col-sm-6">
+                <div>
+                    <%= check_box_tag('pregnancy[special_circumstances][]', 'Rape', 
+                              @pregnancy.special_circumstances.include?('Rape')) %>
+                    <%= label_tag('Rape') %>
+                </div>  
+                <div>
+                    <%= check_box_tag('pregnancy[special_circumstances][]', 'Incest', 
+                              @pregnancy.special_circumstances.include?('Incest')) %>
+                    <%= label_tag('Incest') %>
+                </div>
+                <div>
+                    <%= check_box_tag('pregnancy[special_circumstances][]', 'Prison', 
+                              @pregnancy.special_circumstances.include?('Prison')) %>
+                    <%= label_tag('Prison') %>
+                </div>  
+              </div>  
+            </div>  
           </div>
         </div>
       </div>
--- a/app/views/pregnancies/_patient_information.html.erb
+++ b/app/views/pregnancies/_patient_information.html.erb
@@ -77,47 +77,28 @@
                        options_for_select(referred_by_options,
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
-          <div>       
-            <%= f.label :special_circumstances, 'Special circumstances' %>   
-            <br />
-            <div>
-              <div class="col-sm-6">
-                <div>
-                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Fetal anomaly', 
-                            @pregnancy.special_circumstances.include?('Fetal anomaly')) %>
-                  <%= label_tag('Fetal anomaly') %>
-                </div>
-                <div>
-                <%= check_box_tag('pregnancy[special_circumstances][]', 'Homelessness', 
-                          @pregnancy.special_circumstances.include?('Homelessness')) %>
-                <%= label_tag('Homelessness') %>
-                </div>
-                <div>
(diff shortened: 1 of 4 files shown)


colinxfleming:
Thanks for this, @ashlynnpai ! I'll keep an eye on this for the next few days -- if you could have everything in this PR by Friday afternoon that would be clutch -- I'm planning on having all the DB changes in by Saturday morning, so we can spend Saturday and Sunday entering in data. 

Super glad you got to the bottom of this! 💯

ashlynnpai:
I can't find the right form syntax to get this data persisted. Think I'm going to have to scrap this and create a new model for special circumstances, unless you have a better idea.

ashlynnpai:
The view code is fugly without a model. The checkbox data is persisting but the text input for Other is not wired up properly. I was thinking that if any text were inputted it would create a hash with 'Other' as the key. If 'Other' was in the array, it would display the hash value in the input field, and otherwise it would display the placeholder.  After hacking around a bit I couldn't seem to get that. Not sure if this was even the best approach.

special_circumstances

app/views/pregnancies/_patient_information.html.erb
```diff
@@ -77,7 +77,53 @@
...
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
-          <%= f.text_field :special_circumstances, autocomplete: 'Off' %>
+          <div>       
+            <%= f.label :special_circumstances, 'Special circumstances' %>   
+            <br />
+            <div>
+              <div class="col-sm-6">
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Fetal anomaly',
```
colinxfleming: I think this is outstanding work and is super close! 

One of the few nonstandard tools we use is the `rails_bootstrap_forms` gem, which makes stuff like this a bit easier. I'd check out this https://github.com/bootstrap-ruby/rails-bootstrap-forms#checkboxes-and-radios and the section on `form_group`. You might try something like:

```ruby
f.form_group :special_circumstances, label: 'Special circumstances' do
  div class='col-sm-6'
    f.check_box 'Homelessness', checked: pregnancy.special_circumstances.include?('Homelessness')
```

etc. that'll take care a lot of the kind of pain-in-the-ass styling for you.

app/views/pregnancies/_patient_information.html.erb
```diff
@@ -77,7 +77,53 @@
...
+                          @pregnancy.special_circumstances.include?('Homelessness')) %>
+                <%= label_tag('Homelessness') %>
+                </div>
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Domestic Violence', 
+                            @pregnancy.special_circumstances.include?('Domestic Violence')) %>  
+                  <%= label_tag('Domestic Violence') %>
+                </div>
+                <div>
+                  <%= check_box_tag('pregnancy[special_circumstances][]', 'Other',
```
colinxfleming: I think for an MVP we can put the `other` in the backlog, and just leave it up to CMs to take good notes. I think this is a pretty important chunk and I'd rather have most of this than none of it!

colinxfleming:
Hey @ashlynnpai -- noted this in the comments but want to reiterate that this is really solid work. I was hoping we'd be able to avoid adding a model 

Had some suggestions as to how to de-fuglify it, and recommend just dropping the `other` part for now.

colinxfleming:
oops, hit enter to soon. 

I think after that it should be good to go, no?

ashlynnpai:
@colinxfleming Thanks for your comments. Will work on the bootstrap form stuff tonight. Do you want the Other checkbox removed or just the input field?

colinxfleming:
Let's strip both the check box and text field for now I think

ashlynnpai:
@colinxfleming  

Using f.check_box the data doesn't save. I'm not sure that f.check_box can be used to push items into an array. There is no special_circumstance model. We can put special_circumstances into its own model, but I'm not sure if I would be able to get that coded up today. 

https://github.com/colinxfleming/dcaf_case_management/pull/591/commits/dcdfdde77ee38ed1af2a9a2a7bdea66f0a0b86f1

colinxfleming:
Let's try and avoid the separate model if at all possible.

Go ahead and revert this to using the rails standard checkbox if the
bootstrap forms gem isn't playing nice. I'll take a glance at this tonight
if I can

ashlynnpai:
@colinxfleming The current commit  7f81b2cb1ba0f5a80841ca3f5fb46caf3968f862 should be functional even if fugly. The bootstrap form doesn't seem to wrap the check_box_tag form helper.

colinxfleming:
Hey @ashlynnpai ! I dug through the code and it looks like there's a way to get this to work with bootstrap forms gem out of the box. Check out the `check_box` method itself in the source code:

https://github.com/bootstrap-ruby/rails-bootstrap-forms/blob/7254f987cf7790f0abad12b777ea16ed656d0920/lib/bootstrap_form/form_builder.rb#L111

You'll see the `check_box_with_bootstrap` (aliased to `check_box`) method accepts arguments as follows: `name, options = {}, checked_value = "1", unchecked_value = "0", &block`. 

I think the only reason that the bootstrap form didn't seem to save was that it was saving as a 0 and 1, instead of the string value. But that seems to be literally the only reason! So something like this ought to work:

`f.check_box :special_circumstances, { label: 'Fetal anomaly',  name: 'pregnancy[special_circumstances][]', checked: pregnancy.special_circumstances.include?('Fetal anomaly') }, 'Fetal anomaly', ''`

Want to reiterate that it's amazing that you worked through how to do this without a model! Very impressed and grateful, this is for sure a much better way to do it than hacking together an adhoc model and adding some app bloat. A++++ work!

ashlynnpai:
@colinxfleming That's awesome. Are you going to make the change or did you want me to put your code in my branch?

colinxfleming:
Give it a shot if you have time, otherwise I'll do this tonight. Want to
get everything involving a model change in asap so Jen can start entering
data

ashlynnpai:
It seems to be working!! Thanks!!

Any ideas on how to resolve this error for the test?

Capybara::Ambiguous: Ambiguous match, found 6 elements matching checkbox "Prison"

form group refactor

ashlynnpai:
wizard

colinxfleming:
the wizard is you, bud! I have some mandatory fun with coworkers, but I'll
check this out afterward. Do you feel like this is in a good place?

ashlynnpai:
It works on manual testing.  The integration test throws an error.

app/views/pregnancies/_patient_information.html.erb
```diff
@@ -77,8 +77,28 @@
                        options_for_select(referred_by_options,
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
-          <%= f.text_field :special_circumstances, autocomplete: 'Off' %>
-        </div>
+         
+          
+          <%= f.label :special_circumstances, 'Special circumstances' %>
```
colinxfleming: Can go ahead and nuke this line

app/views/pregnancies/_patient_information.html.erb
```diff
@@ -77,8 +77,28 @@
...
                                           pregnancy.referred_by),
                                           autocomplete: 'off' %>
-          <%= f.text_field :special_circumstances, autocomplete: 'Off' %>
-        </div>
+         
+          
+          <%= f.label :special_circumstances, 'Special circumstances' %> 
+          <%= f.form_group :special_circumstances do %>
+            <div class="col-sm-6">
+              <%= f.check_box :special_circumstances, { label: 'Fetal anomaly', name: 'pregnancy[special_circumstances][]',
```
colinxfleming: Here's how this is showing up in practice -- they all seem to have similar matching labels - 

```
<div class="checkbox">
  <label for="pregnancy_special_circumstances">
    <input name="pregnancy[special_circumstances][]" type="hidden" value="">
    <input name="pregnancy[special_circumstances][]" type="checkbox" value="Fetal anomaly" id="pregnancy_special_circumstances"> Fetal anomaly
  </label>
</div>
```

So I think what's going on here is that capybara is clicking the associated label, but then getting tripped up because there are six of the same label. (You can kinda see the weirdness if you try clicking on one of them that isn't fetal anomaly -- it'll check or uncheck fetal anomaly regardless.) So I think the challenge here is figuring out how to override the `label_for`.

You might try digging around the bootstrap rails gem  issues -- this looks kinda similar at first glance https://github.com/bootstrap-ruby/rails-bootstrap-forms/issues/178

I'll dig into this with you in a bit.

app/views/pregnancies/_patient_information.html.erb
ashlynnpai: Special circumstances won't have any heading. I tried to put the label in the form group line and the styling went wonky.

app/views/pregnancies/_patient_information.html.erb
ashlynnpai: Awesome tip.

ashlynnpai:
Integration test is passing, labels are inactive, data is saving. Is this solution okay?

How do you want the special circumstances section labelled?

colinxfleming:
💯💯

I think if we can get the labels to correspond to their particular checkbox
that is awesome, but I'm not going to lose sleep over it if we have to roll
with it as is. I'll check the code in a minute

test/integration/update_patient_info_test.rb
```diff
@@ -90,7 +90,9 @@ class UpdatePatientInfoTest < ActionDispatch::IntegrationTest
       find('#pregnancy_household_size_children').select '3'
       find('#pregnancy_insurance').select 'Other state Medicaid'
       find('#pregnancy_referred_by').select 'Other abortion fund'
-      fill_in 'Special circumstances', with: 'Stuff'
+      #fill_in 'Special circumstances', with: 'Stuff'
```
colinxfleming: Go ahead and remove these depreciated parts of the test -- we've got it working now!

app/views/pregnancies/_patient_information.html.erb
colinxfleming: Copy that!

colinxfleming:
Code looks excellent to me. Cool if you want to take a stab at getting the labels to play nice, otherwise we can file it to the backlog and that's also great with me. The important, extremely valuable thing is being able to persist special circumstances in this valuable. LMK what works best for you, since it's already 10pm on a Friday!

ashlynnpai:
So label_for doesn't match the id anymore. I'm not sure how to access label_for, except maybe to add a separate label tag as we had before.

colinxfleming:
@ashlynnpai copy that. Let's call this good to go, I think we're at the point of diminishing returns. Will you leave a TODO in the code about what's not matching up so we can dig into it later, and then let's declare victory and merge?

colinxfleming:
Want to reiterate that I'm very impressed that you tackled this so well, and am deeply appreciative! Am going to get an extra few hours of sleep this weekend because of your hard hustling here!

mebates:
Thanks @ashlynnpai !!! 🎉  💜

PR #594, opened 2016-08-19
Corrected Phone and name box sizes on New patient partial

Barry-malone:
Added new sizing to text box fields on new patient partial, form layout slightly changed.
please let me know if I need to fix anything and I will take care of it asap.

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -159,6 +159,10 @@ section {
   @extend .btn-xs;
 }
 
+.btn-space {
+    margin-left: 5px;
+}
+
 // Search result
 // shell styling
 .form-group input {
@@ -166,4 +170,5 @@ section {
     -webkit-appearance: none;
     display: none;
   }
+
 }
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -160,7 +160,7 @@ section {
 }
 
 .btn-space {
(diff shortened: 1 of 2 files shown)


colinxfleming:
oh man, thank you so much! I'll review this tonight.

Barry-malone:
It is my pleasure! your project is awesome, and really well structured :+1:  :)

colinxfleming:
Thanks @Barry-malone !

Here's how this renders for me -- 

image

It looks like this renders a little crooked -- seems to overflow on the right. I suspect this is because of the `container` div. But you took exactly the right path here -- ratcheting up the initial div. 

I looked at the designs out of curiosity and they seem to omit initial call date. I mostly added that in as a temporary stopgap and never took it out, so I think it's probably about time to remove it -- in practice, initial call day is always going to be that day. 

So I think in summary I have four small action points:
* switch all the `col-md`s to be `col-sm`s, since that's what we're using elsewhere in the app
* remove that container div
* remove the `initial_call_date` field from this form, which I think we'll start just inferring from patient creation timestamps after @NerdyGirl537 hand enters data this weekend 
* Redo the `col-sm-X` values so that everything fits (defer to you on what looks best, but maybe 4-3-3-2 would work?)

If you have a sec to handle before Sunday that'd be killer, otherwise I'll be able to knock it out on Sunday. 

Thanks so much for cranking out this pr!

colinxfleming:
Sorry about the lag on my end, @Barry-malone !

PR #590, opened 2016-08-17
Time out urgent pregnancies

charleshuang80:
I rule and have completed some work on Case Manager that's ready for review!

This pull request makes the following changes:
* pulls a pregnancy's audit trails (`history_tracks`) and checks to see if the `urgent_flag` has been `true` for more than 6 days
* adds a class method for Pregnancy to change all pregnancies that are not still urgent to have `urgent_flag = false`

(If there are changes to the views, please include a screenshot so we know what to look for!)

It relates to the following issue #s: 
* Fixes #119

PR #598, opened 2016-08-20
Time out urgent pregnancies edits

### LINKED ISSUE #119: Time out urgent pregnancies after five days
(note: starting to create issues not associated with sprints for backlog issues. hope that's okay @adinneen , let me know if I should do this different)

@lwaldsc flagged that there's a scenario where urgent cases shoud probably eventually fade out, so that we avoid having a big backlog that nobody ever cleans up except like one superstar. (see also "the time the voicemail didn't get cleaned out for like a month", says @lwaldsc .)

So we have an open question of how to handle this. Should we have people time out after a week or two of being urgent and not being able to reach a patient, or not have them time out at all, or some other behavior? 

cc @NerdyGirl537

colinxfleming:

Minor edits to some outstanding work by @charleshuang80 in #590 -- time out urgent pregnancies after five days

* Fixes #119

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -1,6 +1,6 @@
 class DashboardsController < ApplicationController
   def index
-    @urgent_pregnancies = Pregnancy.where(urgent_flag: true)
+    @urgent_pregnancies = Pregnancy.urgent_pregnancies
   end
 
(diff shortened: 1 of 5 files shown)


colinxfleming:
Made TODOs for later testing. Merging!

PR #600, opened 2016-08-20
Single pregnancy

### LINKED ISSUE #595: Convert pregnancies to has_one

colinxfleming:

TLDR: Patient-has-many-pregnancies as a way of keeping track of patients who come back is much more of a huge pain than it's worth, so this punts on a whole class of validation and data storage problems in favor of simplicity. We can fix any reporting duplication with group bys, like god intended.

We should take advantage of this in a few other ways, like changing the has to an embed and fixing the routing, in a separate PR. 

BIG s/o to my dog integration tests for confirming that everything still works after this change!

* Converts `Patient has many pregnancies` into a `patient has one pregnancy` 
* Fixes tests and seedfile

* Fixes #595 
* Bumps #Y

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -6,10 +6,10 @@ class DashboardsController < ApplicationController
   def search
     patients = Patient.search params[:search]
     @results = []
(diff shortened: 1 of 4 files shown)


colinxfleming:
Here goes nothing!

Issue #579, opened 2016-08-14
Add multiple pregnancies functionality


colinxfleming:
quick brain dump: 
* either write a complex scope where a patient can only have one active pregnancy at a time OR
* treat patients in complete iso from each other and have them not show up in search except in some TBD admin mode after they're marked complete

Leaning toward B to increase safety/privacy and because of the fact that we can answer this in reporting with a fairly straightforward analytics function

colinxfleming:
601 goes toward the latter, just need to add some unique-on-phone validator

colinxfleming:
We went with B, object model-wise.

PR #601, opened 2016-08-20
WIP Remix object model

### LINKED ISSUE #278: Routing change: Patient > Pregnancy > Calls, Notes, Pledges
To accurately reflect our object relationships, we should reorient our routing as follows:

```
resources :patients do
  resources :pregnancies do
    resources :calls
    resources :notes
```

This is almost certain to wind up being a pretty time intensive lift so let's backlog it for the time being.

Raised in #273

colinxfleming:

😬 

This makes the central unit of the app the patient instead of the pregnancy, and digs us out of the substantial technical debt we've accumulated because I made a dumb decision about models a few weeks in. Mostly I am excited to no longer have to explain why pregnancies instead of patients have calls. Behold this, my penance. 

* centers everything on patient instead of pregnancy

Fixes #278 
Bumps #579


colinxfleming:
Surprise to nobody: this turned into a helluva yakshave. But controllers and models are converted!

Incidentally I'm extremely grateful we wrote a very rigorous test suite.

colinxfleming:
I think we're in business.

colinxfleming:
Not a real test failure. Merging.

PR #603, opened 2016-08-20
Fix nightly cleanup

### LINKED ISSUE #118: Call list cases should time out after 8 hours

colinxfleming:

Adjusts the `nightly_cleanup` methods to reflect patients being at the center of the app. 
* Fixes #118

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -105,7 +105,7 @@ class Patient
   end
 
   def self.trim_urgent_patients
-    patient.each do |patient|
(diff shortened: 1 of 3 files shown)


colinxfleming:
Codecov failure seems to be false metal. merging.

PR #604, opened 2016-08-20
WIP fix name params for clinic

colinxfleming:
Jenn identified an issue where when updating clinic names, the patient name would be set to the name of the clinic but the clinic won't update

Problem is that on the second submit and submits thereafter, clinic id is getting passed into the primary payload which is in turn making rails think it needs to set other stuff. I think. 

Whipping up a WIP to keep notes in


colinxfleming:
Game plan is to roll this out and then run the following to mass update entered patients:

```
Patient.each do |patient|
  Patient.collection.update_one({name: patient.name}, '$set' => { clinic_name: patient.clinic.name } ) if patient.clinic
end
```

PR #610, opened 2016-08-21
Display notes in call modal, add call option to call log on edit view

### LINKED ISSUE #526: Represent patient notes in the call modal

colinxfleming:

Picks up @jeffreyeaton76 's heroic work in #588 

screen shot 2016-08-21 at 2 14 58 pm
screen shot 2016-08-21 at 2 15 10 pm

* Fixes #526

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -11,9 +11,9 @@ module DashboardsHelper
     "#{week_start_string} - #{week_end_string}"
   end
 
-  def plus_sign_glyphicon(text)
-    return nil unless text.length > 41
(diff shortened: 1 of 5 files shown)


mebates:
Can we make some adjustments to the "record new call" function on the Call Log? I would prefer that it separate from the words "Call Log", potentially by aligning to the right side of the table. We can also just use a standard link style (17 px) and skip the icon since it seems to be a bit problematic with the size.

colinxfleming:
Spun this request into #667

PR #611, opened 2016-08-21
fix pledge modal button by attaching event handlers to document

### LINKED ISSUE #597: Fix javascript error that is causing the pledge form to not pop up

mchelen:

* make sure the pledge modal next button works every time, without needing to reload

* Fixes #597


colinxfleming:
Confirm that this does what it says on the label and that @mchelen does in fact rule! Merging.

Issue #656, opened 2016-10-05
Undo 'remove from call list'


colinxfleming:
From #614 : At one point during an early shift, a case manager accidentally removed someone from her call list and she had to go back into the system to restore it. Suggested that some kind of safety valve would be helpful.

Obviously something way back in our product backlog that might be resolved by some kind of activity feed. But want to have this represented somehow.

PR #664, opened 2016-10-06
added confirmation dialog

itsWill:

When removing a patient from the call list we now ask for confirmation. If that is what the user wants through a pop up dialogue. It looks like this:

screen shot 2016-10-05 at 9 29 39 pm

Fixes https://github.com/colinxfleming/dcaf_case_management/issues/656

@colinxfleming

### DIFF
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -33,7 +33,7 @@
             <td class="blank-td"></td>
           <% else %>
             <% if current_user.patients.include?(patient) %>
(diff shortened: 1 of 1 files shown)


mebates:
Can we add "...from your call list:" to the end of the phrase? So it doesn't sound like you're removing them from the system.

colinxfleming:
My dude @itsWill ! Thanks so much for making this enhancement, and for following up on @mebates ' feedback so quickly! Really appreciate your work on this front -- this is going to save our users a lot of headache, I think. 

Was able to recreate and confirm behavior -- merging.

PR #668, opened 2016-10-14
Add patient to top of call order

### LINKED ISSUE #653: Add call to call list should always add that patient to the top of the call list
A case manager pointed out that adding people to the call list doesn't always seem to do the same thing -- sometimes people just show up in the middle of it, which can be a little disorienting. 

Right now we're saving call list order to a field in user, and then I think we're re-loading it. We should change the behavior such that using the `add_patient` method always adds that patient to the front of the array. 

Relevant methods are the `User#call_order` field, and `User#add_patient` / `User#reorder_call_list`

Stems from the omnibus in #614 .

pmbrent:
I rule and have completed some work on Case Manager that's ready for review! You guys rule too; keep up the good work!

Always adds new patients to the top of a User's `call_list` and also resolves an ambiguity when reordering patients: Patients present in the `call_order` will now take priority over unsorted patients in the `call_list`. (Before, "unordered" patients were sneaking in right after the first patient in the `call_order`.)

It relates to the following issue #s: Fixes #653


colinxfleming:
Thanks much for your work on this @pmbrent , and for doing the digging to find us! 

I'm trying to catch up on sleep tonight, but I'll give this a review tomorrow or early in the weekend.

pmbrent:
@colinxfleming No rush! It looks like I may have broken some tests, so I'll take another look tonight.

pmbrent:
@colinxfleming I've updated the two tests which appear to be affected by the new `add_patient` ordering changes. I'm not sure why `drag_and_drop_test` is unhappy; there's a comment on the test that indicates it may have already been a problem.

colinxfleming:
Thanks much for the followup @pmbrent ! Using `call_order.length` is a particularly inspired call, I think. And a HUGE thanks for updating the tests as you go -- now that we're using this in production, keeping the level of QA high is super crucial. 

There's a remaining issue I'd like to sort out if we can: I'm trying it out on a local, and it's following intended behavior on page reload, so the call list field is updating correctly after page refresh. However, on adding a new patient to the call list, it looks like it's reordering by name, like so: 
Original call list, with call list order 7, 2, 4:
screen shot 2016-10-15 at 2 13 06 pm

Call list after adding 9 but before reloading: 
screen shot 2016-10-15 at 2 13 17 pm

Call list after reloading, lookin' good:
screen shot 2016-10-15 at 2 13 27 pm

So in a nutshell, I think the core of this is working great, but for some reason when the call list is reloaded it's ordering it improperly. Without looking, I suspect that this is due to the code in `views/patients/refresh_patient.js.erb` reloading the wrong thing, but I'm not positive.

Do you have the capacity to sort out this lingering issue? If not I'm happy to sort it out before we declare victory here. 

Thanks a ton, and thanks again for finding us!

pmbrent:
Hi @colinxfleming, I don't recall seeing that issue but can certainly look into it. However, it might be the end of the week before I get to it. Thanks for letting me know.

colinxfleming:
No worries about the timing -- thanks again for tackling this issue, it's a
huge help!

pmbrent:
I also see the reloading issue; it looks like it happens whenever the Add or Remove link in a patient row is used. (The entire list is temporarily resorted alphabetically, but `call_order` is correctly used when the page is refreshed.) I'll look into it.

pmbrent:
@colinxfleming Looks like you were right; `views/patients/refresh_patient.js.erb` was using `call_list_patients` instead of `ordered_patients`.

colinxfleming:
@pmbrent just pulled this down and it looks like it's working great! (Sorry for the bad tests making CI fail initially!) 

Want to let you know that I really appreciate your sticking with this issue and pushing it over the line, and our CMs will too. A huge priority of this project is case manager happiness, and the call list is a big part of that -- improvements like this which make it more reliable are so key. I also really appreciate your taking the time to tighten bolts and fix tests as you went, so that this is in good maintainable shape going forward. Thank you so much!

PR #669, opened 2016-10-14
Create a rake task for updating user role data from string to integer

### LINKED ISSUE #533: What procedure should we use to rename fields in the live db?
It seems like when we change a db model field, existing objects stop saving until you run `rake db:seed` again. That's not a tenable situation after we start storing actual production data in the system.

What are best practices around this? Is there a gem we can install, or some rake task we can write which will automatically update? 

@Kevin-Wei especially interested in yr thoughts here, since I imagine this is a problem you've run into before.

pcofilada:

I created a rake task for updating User role data from String to Integer. In order to change the role field totally you just need to run the rake task first which is `rake user:update_role` then after that change the field role type to Integer.

* User model role field data

* Fixes #533

### DIFF
--- a/lib/tasks/user.rake
+++ b/lib/tasks/user.rake
@@ -1,0 +1,21 @@
+namespace :user do
+  desc "update role field production data"
+  task update_role: :environment do
+    User.all.each do |user|
+      enum_role = check_role(user.role)
+      user.update(role: enum_role)
+    end
+
+    puts 'User role updated'
+  end
+
+private
+  def check_role(role)
+    case role
+    when 'admin'
+      1
+    else
+      0
+    end
+  end
+end


lib/tasks/user.rake
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    puts 'User role updated'
+  end
+
+private
+  def check_role(role)
+    case role
+    when 'admin'
+      0
```
colinxfleming: I think we'll need to reverse this based on the enum pecking order in #670, but I see what you mean.

pcofilada:
@colinxfleming I already pushed an update on this branch. I just reverse 0 and 1  in order to make it inline with #670

colinxfleming:
@pcofilada thanks again for lending your mongoid expertise to this project -- what I really appreciate is that this gives us a model for similar need going forward, which we'll be able to adapt as we make other db adjustments. Really appreciate it -- thank you so much!

pcofilada:
@colinxfleming Sure no problem. I will look on other issues to do.

Issue #626, opened 2016-09-06
Design administrative support screens for user management


colinxfleming:
Right now I'm creating new accounts through the heroku console, which clearly is not tenable. We should turn account creation back on for people with administrative privileges.

mebates:
This sounds like a 'support' function. Any other support-type things we should consider? Def needs more UX thinking.

colinxfleming:
Google dump:
* http://wiki.summercode.com/rails_authentication_with_devise_and_cancan
* http://stackoverflow.com/questions/19017978/allowing-admins-to-add-users-with-devise
* http://stackoverflow.com/questions/23879648/devise-allow-only-admin-to-create-users

(There's a baked in devise form we can use for the immediate thing.)

I'm not sure about support type stuff, but I'm not super worried about it honestly. I think this kind of stuff will rise pretty organically.

colinxfleming:
After a good night of sleep (imagine that, after a good night of sleep I look at problems differently, who'd have thought), I think there might be some basic admin support of users we can roll up -- some kind of admin panel that's just a paginated view of users, their last login timestamp, and a button to send them a password reset email might be very helpful. What do you think @mebates ?

mebates:
Yes. A support screen or two. Needs work and thinking. Doing this at my job right now too.

colinxfleming:
Sounds like in practice this is probably two tasks:

a) Backend wise, wiring up `devise` and `pundit` libraries so that authorized users (e.g. CMDs, program admins such as myself and Jenn) can create users 
b) UX wise, figuring out what authorized users need to be able to handle this administrative component of the system.

The former we can start on now, since it's basically just 'wire up the devise route and install pundit'. I'll spin that off into its own issue and we can use this one for admin support pages. @mebates I'm super interested in what we can crib from your day job for this.

nimiro:
Hi, here are some draft wires for user account management based on what we discussed at the meetup last week: [DCAFwireframe101016_user_account_management.pdf](https://github.com/colinxfleming/dcaf_case_management/files/520274/DCAFwireframe101016_user_account_management.pdf)

The data columns definitely aren’t finalized, and I wasn’t sure how important it is for directors to be able to bulk reset passwords and/or lock accounts. So there are two different options for the table view - multi-select or inline buttons in each table row. But hopefully this is a useful start for getting feedback.

colinxfleming:
[...]

PR #670, opened 2016-10-15
Admin Create New Users

### LINKED ISSUE #649: Wire up Devise and Pundit to allow admins to create users
Right now I'm creating users by hand through the rails console, which means that if I get hit by a bus, no more new users, which is a problem. I'd like to be able to get to a point where any of our admin users can create accounts, so that we can turn this into DCAF's problem.

Here are the steps to that, as I see them:
* Change `User#role` to be an `enum` instead of a string (see `Patient#voicemail_preference` for an example of this using mongoid). Roles of `cm` and `admin` are fine for now
* Set up devise `registerable`
* Install the `pundit` gem and create a policy that allows only logged in users with the role `admin` to 
* GOLD MEDAL VERSION: Write an integration test ensuring that you can't use the registration form if you aren't logged in and an administrator

malively:

Allows admins to create new users.

* Changes the `User#role` to `enum`. I don't know if the current `role` is being utilized at all so I wasn't sure if creating a data migration task was necessary to populate the enum value. 
* Added a 'Created User' link to the Navigation Links only visible to admins
* Created a new user form. It was suggested to do it with Devise. After some research it seems like to me Devise doesn't handle this well since it is designed for people to create their own accounts.
* It was also suggested to use Pundit to ensure only admins can create users. It seemed simpler and just as effective to simply check if the `current_user` is an admin in the controller methods and views. 
* Added integration and controller tests to ensure only admins can create users

This PR has a lot of changes and a few assumptions made. Let me know if anything seems wrong or strange and I will be happy to fix it.

* Fixes #649

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -3,6 +3,27 @@ class UsersController < ApplicationController
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
 
+  def create
+    raise 'Permission Denied' unless current_user.admin?
+    @user = User.new(user_params)
+    if @user.save
+      flash[:success] = 'User created!'
+      redirect_to session.delete(:return_to)
+    else
+      # if validation errors, render creation page with error msgs
+      render 'new'
+    end
+  end
+
+  def new
+    unless current_user.admin?
+      render text: 'Permission Denied', status: :unauthorized
+      return
+    end
+    @user = User.new
+    session[:return_to] ||= request.referer
+  end
+
   def add_patient
     current_user.add_patient @patient
     respond_to do |format|
@@ -25,6 +46,10 @@ class UsersController < ApplicationController
   end
 
   private
+
+  def user_params
+    params.require(:user).permit(:name, :email, :password, :password_confirmation)
+  end
 
   def retrieve_patients
     @patient = Patient.find params[:id]
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -17,7 +17,7 @@ class UsersController < ApplicationController
 
   def new
     unless current_user.admin?
-      render text: 'Permission Denied', status: :unauthorized
+      redirect_to root_path
       return
     end
     @user = User.new
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -59,6 +59,7 @@ class User
   field :locked_at,       type: Time
 
   # Validations
+  # email presence validated through Devise
   validates :name, presence: true
   validate :secure_password
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,7 @@
 # Fields are all devise settings; most of the methods relate to call list mgmt.
 class User
   include Mongoid::Document
+  include Mongoid::Enum
   include Mongoid::Userstamp::User
 
   # Devise modules
@@ -25,7 +26,7 @@ class User
   # Non-devise generated
   field :name, type: String
   field :line, type: String
-  field :role, type: String
+  enum :role, [:cm, :admin]
   field :call_order, type: Array
 
   ## Database authenticatable
@@ -58,7 +59,7 @@ class User
   field :locked_at,       type: Time
 
   # Validations
-  validates :email, :name, presence: true
+  validates :name, presence: true
   validate :secure_password
 
   def secure_password
@@ -113,6 +114,10 @@ class User
       # TODO: reexamine this behavior in awhile
       patients.delete(p) if recently_reached_by_user?(p)
     end
+  end
+
+  def admin?
+    role == :admin
   end
 
   private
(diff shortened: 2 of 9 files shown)


app/models/user.rb
```diff
@@ -58,7 +59,7 @@ class User
   field :locked_at,       type: Time
 
   # Validations
-  validates :email, :name, presence: true
+  validates :name, presence: true
```
colinxfleming: Why the removal of email from validations here?

test/integration/create_user_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+class CreateUserTest < ActionDispatch::IntegrationTest
```
colinxfleming: 💯 💯 💯 💯 💯

app/controllers/users_controller.rb
```diff
@@ -3,6 +3,27 @@ class UsersController < ApplicationController
...
+      redirect_to session.delete(:return_to)
+    else
+      # if validation errors, render creation page with error msgs
+      render 'new'
+    end
+  end
+
+  def new
+    unless current_user.admin?
+      render text: 'Permission Denied', status: :unauthorized
```
colinxfleming: Can we set this as a redirect to root?

app/views/layouts/_navigation_links.html.erb
```diff
@@ -5,6 +5,9 @@
     <% unless current_page?(authenticated_root_url) %>
       <li><%= link_to 'Dashboard', authenticated_root_path %></li>
     <% end %>
+    <% if current_user.admin? %>
+      <li> <%= link_to 'Create User', new_user_url %></li>
```
colinxfleming: I ask because this is something that has bitten me in the behind a lot recently; can we use `new_user_path` over `new_user_url` here?

app/views/users/new.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class='authform'>
+  <h3>Create User</h3>
+    <%= form_for @user do |f| %>
```
colinxfleming: v nice! If you want to take a stab at a once over, you might consider using the rails bootstrap forms gem we use -- will let you simplify down to something like this:

```
bootstrap_form_for @user do |f|
  f.email_field :email, autofocus: true
  f.text_field :name
  ...
end
```

config/routes.rb
```diff
@@ -15,6 +15,7 @@
     patch 'users/reorder_call_list', to: 'users#reorder_call_list', as: 'reorder_call_list', defaults: { format: :js }
   end
   root :to => redirect('/users/sign_in')
+  resources :users, only: [:new, :create]
```
colinxfleming: For readability's sake, can we put this in the `authenticate` block if it doesn't bust up everything? (If it DOES make things break, we'll just get it later.)

app/controllers/users_controller.rb
```diff
@@ -3,6 +3,27 @@ class UsersController < ApplicationController
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
 
+  def create
+    raise 'Permission Denied' unless current_user.admin?
```
colinxfleming: Having it choke on someone not being an admin seems like an expected outcome here -- is it more appropriate to have this return unauthorized and halt? I can see an argument for it either way:

* The former is probably more the textbook pattern, which makes it more readable / maintainable / etc. 
* OTOH, raising an error like you're doing here makes it more likely to get picked up by a monitoring service that we use in production, thereby making it easier to figure out if something sketchy is going on and someone's trying to screw with this endpoint.

@malively What are your thoughts?

app/models/user.rb
malively: Devise has its own validates on user email so if you submit the form with an invalid email it would return the same email error twice. 
I can see if its possible to remove the validation on Devise side to make the User model more complete

app/controllers/users_controller.rb
malively: Since `UsersController#new` already deflects an unauthorized user I think raising an error is a better response for `UsersController#create` since something must be REALLY WRONG if a non-admin user is able to get through to the create route.

app/models/user.rb
colinxfleming: Learn something new every day! I'd say let's follow your pattern on this, then. Thanks for documenting.

app/controllers/users_controller.rb
colinxfleming: I think you're right about this. I might wind up wrapping that in its own error class, but I think you're right on.

malively:
@colinxfleming made some code changes to address some of your comments. Let me know if you have any other comments!!!

app/views/users/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class='authform'>
+  <h3>Create User</h3>
+    <%= bootstrap_form_for @user do |f| %>
+        <%= f.email_field :email %>
+        <%= f.text_field :name %>
```
colinxfleming: 💯 💯

colinxfleming:
CI is failing on a scrub test, so I'm going to merge this. Thank you so much @malively , this is excellent work and a MAJOR value-add for this project -- it's a major step toward letting the case management directors run this system autonomously. 

Very grateful that you've chosen to contribute your brainpower and time to this project, and stoked every time I see a PR with your name on it. Thanks again!

Will also go great with #669 !

mebates:
Hi! Can we see this in a screengrab? How does it relate to the mockups being worked on in #626 ?

Issue #629, opened 2016-09-13
Appointment date added to Dashboard information


lwaldsc:
Related to #510 

A few CMs have flagged they would like a PT's appointment date in the dashboard information so they can also prioritize PTs with close appointments 

![image](https://cloud.githubusercontent.com/assets/10578608/18487690/269479be-79c4-11e6-9356-499a236130d6.png)

mebates:
Cool! Should be easy to add, just squeeze the space we leave for comments.

colinxfleming:
Promoting this to active!

lwaldsc:
Fantastic!

colinxfleming:
Tagging as hacktoberfest. This entails going in to `app/views/_table_content.html.erb` and adding a table cell with `patient.appointment_date` (or maybe patient.pregnancy.appointment_date`, I forget which). This in turn will be a great help for our case managers as they prioritize calls. 

If necessary to get the squeezing right, you can adjust the note text length in the patient model, IIRC.

nickgnd:
Hi, 
I would like to work on this issue, I'll give you a PR asap :)

colinxfleming:
Fixed in #671

PR #671, opened 2016-10-16
Add PT's appointment date in the dashboard information

nickgnd:
* Add appointment date information to dashboard: call list, completed calls and urgent cases.
(with integration test)

* #629 

schermata 2016-10-16 alle 02 24 42

### DIFF
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -12,6 +12,7 @@
         <th>Name</th>
         <th>LMP</th>
         <th>Status</th>
+        <th>Appointment date</th>
         <th>Notes</th>
         <th></th>
         <th></th>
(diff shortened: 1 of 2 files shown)


colinxfleming:
@nickgnd thanks a ton for your work on this, and for reaching out and finding us! One of the things that our users have asked us for the most is this, so it's going to make for a lot of happy people. 

I'll pull down tomorrow before I merge for QA, but I don't see anything that should hold this up beyond that. 

Thanks again, and nice work! (And thank you for adding the test!)

nickgnd:
Hi @colinxfleming, As you notice, I added basic integration tests to the PR, but they check only the presence of the field `appointement_date` and doesn't care about other fields.
If you want I could add extra checks against the other table fields. Let me know :)

colinxfleming:
@nickgnd I think this is good to go as is -- thanks again, I really appreciate this, and our case managers are going to be able to put this to very good use.

PR #673, opened 2016-10-17
Clear Call List After No Login

### LINKED ISSUE #666: Create call list clearing rules for between shifts
From a conversation between @mebates and [the CM directors](https://github.com/colinxfleming/dcaf_case_management/issues/118#issuecomment-246886012): 

> So what came up tonight was the many patients who are 'dormant' for days, weeks, or forever. We don't want to encourage CMs to keep calling PTs who don't give DCAF a call back, and leaving them on the list for the whole shift might do that. And what if you come back a month later and they're still there? Or 9 months?

@NerdyGirl537 seconded this.

Two things here: 
a) finding some way to discourage people from calling patients over and over again when they don't hear back. There may or may not be a programmatic way to do this -- I suspect that the solution here is training.

b) starting with a fresh call list when you hop on a fresh shift. This I think is doable -- as part of the cleanup task, clear users' call lists entirely if they haven't logged in for two weeks or so. I don't imagine there are that many cases where people take multiple shifts like that. 

@NerdyGirl537 thoughts? 

Spun out from #118 .

malively:

Clears user's call list if they don't sign in for _2 weeks_

* Adds constant `TIME_BEFORE_INACTIVE` to `User` model
* Inside `User#clear_call_list` clears patient list if `User#last_sign_in_at` is 2 weeks or longer
* Test new code in the user model test file

* Fixes #666

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -61,6 +61,8 @@ class User
   validates :email, :name, presence: true
   validate :secure_password
 
+  TIME_BEFORE_INACTIVE = 2.weeks
+
   def secure_password
     return true if password.nil?
     pc = verify_password_complexity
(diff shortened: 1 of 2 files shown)


colinxfleming:
@malively you're a treasure. Thanks a ton for knocking this out, bud. Love how clean and readable it is!

I'll pull down and test locally tomorrow before merge, but I can't see anything that should hold this up -- looks great to me.

colinxfleming:
Pounding the button -- thanks again @malively !

Issue #674, opened 2016-10-19
Edit voicemail display text for Voicemail OK


colinxfleming:
Change from `Voicemail OK` to `Voicemail OK, ID OK`. Trivial change -- code is located in `dashboards_helper`. Likely involves a test change too. 

Stems from some user feedback in #614.

valeriecodes:
I'll take this.

colinxfleming:
hey @valeriecodes ! Good to see you, hope you've been well. 

@itsWill kindly knocked this out yesterday, so I think we're good on this front. Any chance I could direct you toward #448 if you're looking for a quick-ish one?

valeriecodes:
Sounds good, thanks @colinxfleming !

PR #688, opened 2016-10-20
modified text to VOICEMAIL OK, ID OK.

itsWill:

easy one line text change from VOICEMAIL OK, to VOICEMAIL OK, ID OK

@colinxfleming 

Fixes https://github.com/colinxfleming/dcaf_case_management/issues/674

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -15,7 +15,7 @@ module DashboardsHelper
   def voicemail_options
     enum_text = { not_specified: 'No instructions; no ID VM',
                   no: 'Do not leave a voicemail',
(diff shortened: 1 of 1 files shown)


colinxfleming:
Change renders cleanly. Thanks again @itsWill !

screen shot 2016-10-23 at 10 33 22 pm

mebates:
What does it look like on the Call List page?

colinxfleming:
Code on the call list page is unaffected, still looks like so:

screen shot 2016-10-26 at 12 52 16 am

mebates:
How about right below the modal? I was thinking about the first entry - looks like it's a dropdown now.

Issue #686, opened 2016-10-20
Build out draft user admin panel


colinxfleming:
One of the ongoing goals for this app is to make sure that we get it to a place where engineers need to maintain fewer non-code parts of it; to that end, @nimiro is doing some heroic work building out some beautiful designs. Check them out here: https://github.com/colinxfleming/dcaf_case_management/files/520274/DCAFwireframe101016_user_account_management.pdf

The steps here are as follows:
- Add a view and a route for `users/index`
- Code up the first page of that in hard HTML (no functionality needed yet)
- If possible, add a simple controller test to confirm that the route is accessible

Stems from #626

itsWill:
I'll take a stab at this!

mebates:
Which page of the mockup are we working from? Great work @nimiro !

mebates:
@itsWill - can you touch base with @nimiro before you start, and keep an eye on our convo in #626? Thanks!

itsWill:
@mebates I just picked the first page, I understand the frontend design is not ready yet so this mainly sets up the backend, we can then change the view according to the design you and @nimiro come up with once its decided.

colinxfleming:
Demoting from active development so we can focus on required stuff for Dec 1 -- thanks a ton to @itsWill for getting this process started!

colinxfleming:
Here are some super dope designs courtesy @nimiro --https://github.com/colinxfleming/dcaf_case_management/files/571543/DCAFwireframe110316_user_account_management.pdf

I'm ready to start moving on this, and I think it'd be a nice thing to get into place. Marking active.

PR #689, opened 2016-10-20
built out admin panel

itsWill:

Built out the admin panel according to https://github.com/colinxfleming/dcaf_case_management/issues/686.

Things todo:
- [ ] clean up inline css
- [ ] extract partials
- [ ] formate the table data correctly

But we can do these once we have more functionality implemented

screen shot 2016-10-20 at 5 54 34 pm

@colinxfleming


mebates:
Hey guys - this didn't get finished from a design perspective and what you built isn't what we decided on for design. Please touch base with @nimiro !

itsWill:
@mebates I just picked the first page, I understand the frontend design is not ready yet so this mainly sets up the backend, we can then change the view according to the design you and @nimiro come up with once its decided.

app/controllers/users_controller.rb
```diff
@@ -1,8 +1,13 @@
 # Additional user methods in parallel with Devise -- all pertaining to call list
 class UsersController < ApplicationController
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
+  before_action :admin_user, only: [:new, :index]
```
colinxfleming: Thanks for abstracting this! Love it

test/controllers/users_controller_test.rb
```diff
@@ -20,6 +20,22 @@ class UsersControllerTest < ActionController::TestCase
     end
   end
 
+  describe 'index method' do
```
colinxfleming: 💯 💯 💯 💯 💯  for tests!

colinxfleming:
@mebates like @itsWill noted, this rigs up the backend so that we can throw the actual designs in. Fully expect that we'll have another pull request or three before we're 100% good to go, but this PR (setting up routing and tests) means that the next person to take a stab will be able to hit the ground running when team ux has reached an endpoint in #626 . $$$$

mebates:
Great!

Issue #683, opened 2016-10-20
Build scratch call line splash page


colinxfleming:
[Here is a rough mockup of the user select screen.](https://cloud.githubusercontent.com/assets/17989540/18189002/c64c7ae0-7084-11e6-86ee-dec20293707e.png). The task here is to recreate this, except without the cancel button, per discussion in the UX thread #589.

Let's put this in a file in `app/views/dashboards/select_line.html.erb` for the time being, and add a get route for it. (And tests, if possible.) For now, just hard-code everything in HTML and skip any actual functionality. No need to have it actually do anything yet. 

This is fairly straightforward, so tagging hacktoberfest.

lomky:
I'll take a stab at this tonight.

lomky:
Rough version done and submitted!

PR #690, opened 2016-10-21
Line Select View Draft

lomky:
Added a rough line select view, route & associated test as spelled out in issue #683. It probably needs a pass from an actual frontend developer. :smile_cat: 

* adds the line select view & route
* adds a test for that route

dcaf line select draft

* Bumps #683

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -15,6 +15,9 @@ class DashboardsController < ApplicationController
     respond_to { |format| format.js }
   end
 
+  def lineselect
+  end
+
   private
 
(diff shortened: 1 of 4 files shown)


mebates:
Awesome. Thank you!

colinxfleming:
💯 

Thank you so much for seeing this thru @lomky ! Glad to have you on the project, and congrats on your first rails pull request iirc?

We'll be iterating on this until we're using it to set a session variable, and in turn using that to set the `line` attribute for patients. (DCAF divides its patients into different lines based on state to lighten the load on individual CMs, and to help them do less context switching since the laws and circumstances affecting patients can vary from state to state.)

Issue #675, opened 2016-10-19
New patient partial should display as the result of a search regardless of whether it returns results


colinxfleming:
Per the code in `dashboards_controller` and `views/dashboards/search.js.erb`, we currently pop up the form to create new patients if and only if there isn't a result. As a CM pointed out, since we have pattern matching and partial searches enabled, we should really be popping up the partial.

The task here is to:
- go into `search.js.erb` and edit it, such that regardless of whether the resultset is blank a user will still get the `patients/new_patient` partial. 
- Add a test confirming this behavior (you can piggyback on the code in `integration/record_lookup_test.rb`).

Stems from #614.

colinxfleming:
(Should get design to weigh in on this before we actively write code against it. I jumped the gun here.)

colinxfleming:
Closing as a wontfix

PR #701, opened 2016-10-26
Render new patient partial even if a search does not return results #675

ashlynnpai:
* edits search.js.erb to display new patient partial even if no search results were returned
* edits new patient partial to display different headings based on whether search results were returned
* add integration test

* New patient partial should display as the result of a search regardless of whether it returns results #675

no results returned
partial_match_search


ashlynnpai:
Didn't realize someone else grabbed this issue. Doh!

colinxfleming:
oh no @ashlynnpai !! Sorry, that's on me -- I'll make sure to do a good job tagging issues back when PRs for them spin up.

PR #704, opened 2016-10-27
Add Dynamic Call Lists

tingaloo:

Added ajax partial rendering for call lists. I think it would be nice for CMs to have instant feedback on their calling actions. It works for ALL calling actions other than "I reached the patient." Reaching the patient triggers a redirect.

Also fixes a bug where urgent patients list is completely wiped when one REMOVES a call.

You can test it on edit patient page > call log > record a call
And removing calls/calling actions on home page.

I will add functional tests in the future, but I am not familiar with minitests so it will take some time.

- AJAX partial rendering
- Fixes urgent patients bug on CALL REMOVE

fixes #704

### DIFF
--- a/app/views/calls/create.js.erb
+++ b/app/views/calls/create.js.erb
@@ -1,1 +1,16 @@
-$(".new-call").modal('hide');
+
+
+// Fires on call complete POST
+
+// TODO: write if statement for url ending in /edit/
+    $("#call_log").html("<%= escape_javascript(render partial: 'patients/call_log') %>")
+      .promise().done(function() {
+        $(".new-call").modal('hide')
+        $('body').removeClass('modal-open')
+        $('.modal-backdrop').remove()
+      });
+if (location.pathname == "/") {
+  $("#call_list").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your call list', table_type: 'call_list', patients: current_user.ordered_patients, sortable: true }) %>")
+  $("#completed_calls").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your completed calls', table_type: 'completed_calls', patients: current_user.recently_called_patients }) %>")
+  $("#urgent_patients").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Urgent cases', table_type: 'urgent_patients', patients: Patient.urgent_patients }) %>")
+}
--- a/app/views/calls/create.js.erb
+++ b/app/views/calls/create.js.erb
@@ -1,16 +1,18 @@
+if (location.pathname.includes("edit")) {
+  console.log("on call record")
+  $("#call_log").html("<%= escape_javascript(render partial: 'patients/call_log') %>")
+    .promise().done(function() {
+      $(".new-call").modal('hide')
+      $('body').removeClass('modal-open')
+      $('.modal-backdrop').remove()
+    });
+}
 
-
-// Fires on call complete POST
-
-// TODO: write if statement for url ending in /edit/
-    $("#call_log").html("<%= escape_javascript(render partial: 'patients/call_log') %>")
-      .promise().done(function() {
-        $(".new-call").modal('hide')
-        $('body').removeClass('modal-open')
-        $('.modal-backdrop').remove()
-      });
-if (location.pathname == "/") {
+if (location.pathname == Routes.authenticated_root_path) {
   $("#call_list").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your call list', table_type: 'call_list', patients: current_user.ordered_patients, sortable: true }) %>")
   $("#completed_calls").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your completed calls', table_type: 'completed_calls', patients: current_user.recently_called_patients }) %>")
   $("#urgent_patients").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Urgent cases', table_type: 'urgent_patients', patients: Patient.urgent_patients }) %>")
+  $(".new-call").modal('hide')
+  $('body').removeClass('modal-open')
+  $('.modal-backdrop').remove()
 }
--- a/app/views/calls/create.js.erb
+++ b/app/views/calls/create.js.erb
@@ -1,10 +1,14 @@
+var URL = location.pathname,
+    substring = "edit";
+
+if (URL.indexOf(substring) !== 1) {
   $("#call_log").html("<%= escape_javascript(render partial: 'patients/call_log') %>")
     .promise().done(function() {
       $(".new-call").modal('hide')
       $('body').removeClass('modal-open')
       $('.modal-backdrop').remove()
     });
-
+}
 
 if (location.pathname == Routes.authenticated_root_path) {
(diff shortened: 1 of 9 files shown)


app/views/calls/create.js.erb
```diff
@@ -1 +1,16 @@
-$(".new-call").modal('hide');
+
+
+// Fires on call complete POST
+
+// TODO: write if statement for url ending in /edit/
```
colinxfleming: Should we backlog this or are you expect to handle? If we're backlogging it, can you create an issue for it and remove the TODO?

app/views/calls/create.js.erb
```diff
@@ -1 +1,16 @@
...
+    $("#call_log").html("<%= escape_javascript(render partial: 'patients/call_log') %>")
```
colinxfleming: Flagging this for @mchelen since I know he's also wrenching on some modal code

app/views/patients/_call_log.html.erb
```diff
@@ -2,10 +2,10 @@
   <div class="margin-bottom col-sm-12">
     <h2>
       Call Log
-      <small><%= link_to "Record new call <span class='glyphicon glyphicon-earphone' aria-hidden='true'></span><span class='sr-only'>Call</span>".html_safe, "#call-#{patient.primary_phone_display}", data: {toggle: "modal"} %></small>
+      <small><%= link_to "Record new call <span class='glyphicon glyphicon-earphone' aria-hidden='true'></span><span class='sr-only'>Call</span>".html_safe, "#call-#{@patient.primary_phone_display}", data: {toggle: "modal"} %></small>
```
colinxfleming: Looks like the changes to this file are turning our local into an instance variable, and that's because if it's a local then the call chokes as follows.

```
MONGODB | localhost:27017 | dcaf_case_management_development.update | SUCCEEDED | 0.000512s
  Rendered patients/_call_log.html.erb (73.1ms)
  Rendered calls/create.js.erb (75.1ms)
Completed 500 Internal Server Error in 97ms

NameError - undefined local variable or method `patient' for #<#<Class:0x007f9b2aea7370>:0x007f9b2e1eaf98>
Did you mean?  @patient:
  app/views/patients/_call_log.html.erb:5:in `_app_views_patients__call_log_html_erb___1989415623033319292_70152230704360'

```

I'd prefer to have this as a local variable rather than an instance variable, but can live without it if it's a nontrivial change. What are your thoughts @tingaloo ?

app/views/users/refresh_patients.js.erb
```diff
@@ -1,3 +1,4 @@
 $('#call_list, #completed_calls, #urgent_patients').empty();
 $('#call_list').append('<%= escape_javascript(render partial: "dashboards/table_content", locals: { title: 'Your call list', table_type: 'call_list', patients: current_user.ordered_patients, sortable: true }) %>');
 $('#completed_calls').append('<%= escape_javascript(render partial: "dashboards/table_content", locals: { title: 'Your completed calls', table_type: 'completed_calls', patients: current_user.recently_called_patients }) %>');
+$("#urgent_patients").append('<%= escape_javascript( render partial: "dashboards/table_content", locals: { title: 'Urgent cases', table_type: 'urgent_patients', patients: Patient.urgent_patients }) %>')
```
colinxfleming: 👍

app/views/calls/create.js.erb
```diff
@@ -1 +1,16 @@
...
+      .promise().done(function() {
+        $(".new-call").modal('hide')
+        $('body').removeClass('modal-open')
+        $('.modal-backdrop').remove()
+      });
+if (location.pathname == "/") {
```
colinxfleming: In the spirit of rails, can we turn this into a `path` variable? Dropping in something like `<%= authenticated_root_path %>` instead of `/` as the url will help keep this code nimble.

colinxfleming:
Also if you're interested in adding a functional test, you can probably bolt something into `tests/integration/call_list_test` -- minitest has a pretty straightforward `assert`/`refute` system. We use the Capybara library, which gives us stuff like `has_text?`, `has_no_text?`, and `visit`. Here's a link to the cheat sheet -- https://gist.github.com/zhengjia/428105 -- let me know if I can be helpful on this front.

app/views/patients/_call_log.html.erb
tingaloo: I can live without it, I know no other approaches making it ajax friendly.

app/views/calls/create.js.erb
tingaloo: Included js-routes, it's a solid wrapper for handling ruby routes in JS.

app/views/patients/_call_log.html.erb
colinxfleming: Copy that.

colinxfleming:
@tingaloo looks like we're in good shape on this, by and large, and the remaining stuff is testing -- integration tests and tracking down these few test failures:

```
 FAIL["test_0001_should close the modal when clicking Left voicemail", #<Class:0x0000000658e818>, 32.84228811900539]
 test_0001_should close the modal when clicking Left voicemail#LoggingCallsTest::logging Left voicemail (32.84s)
        Expected false to be truthy.
        test/integration/logging_calls_test.rb:97:in `block (3 levels) in <class:LoggingCallsTest>'

 FAIL["test_0001_should close the modal when clicking Couldn't reach patient", #<Class:0x0000000658d1c0>, 85.16618454799755]
 test_0001_should close the modal when clicking Couldn't reach patient#LoggingCallsTest::logging Couldn't reach patient (85.17s)
        Expected false to be truthy.
        test/integration/logging_calls_test.rb:97:in `block (3 levels) in <class:LoggingCallsTest>'
```

Let me know if you need assistance with this or if I can be helpful at all.

tingaloo:
@colinxfleming Wow I thought I was going insane when those tests didn't pass. I was looking into whether or not the poltergeist driver didn't work with rendering partials, using capybara-webkit, and all sorts of random BS.

But the problem is that for some reason this line
```
if (location.pathname.includes("edit")) {
```
**breaks** everything.

It didn't make sense why it would in the first place. Either there is an interpretation conflict between JS and ruby OR some ES2015 functions don't work. BUT I tested a let variable and it didn't compile either. Dumb me didn't realize not every project has babel or some JS transpiler built in. I'll add that later on.

In the meantime, the old school comparison still works.
```
var URL = location.pathname,
    substring = "edit";

if (URL.indexOf(substring) !== 1) {
```

colinxfleming:
Oh man, punishing. Yeah, a tenet of this project is 'as many defaults as possible' in the interest of having other people be able dive in easier -- this is why the frontend is in plain old ERB rather than react or angular, for example -- so we're on coffeescript, the rails js default engine. Really sorry that bit you!

tingaloo:
@colinxfleming Tests pass locally but break on CircleCI, you can take a look at this [guide](https://circleci.com/docs/troubleshooting-guide/). I am unfamiliar with circleCI; not sure whether the fails are random or there are issues with capybara and ajax.

colinxfleming:
It looks like Dashboard integration tests L19, L24, and L29 are consistently choking which is weird -- usually the logging calls test and the update info integration tests are janky and this one is pretty solid. They do wind up passing half the time though. If I can get it to work we can call it clean -- I think it's probably just a test that needs to be better at waiting for ajax and not a problem with your code.

colinxfleming:
there we go, it's seeing things our way now. I'll add the dashboard test to the list of tests to check on.

@tingaloo is this good to go from your end?

tingaloo:
One test fails on CircleCI; it passes locally so I'm not sure how to debug it.

colinxfleming:
Thanks again @tingaloo !

PR #706, opened 2016-10-29
updated comment to be more readable and findable.

sergeimellow:


colinxfleming:
@sergeimellow I'm gonna close this out -- we don't currently have this model implemented, but I do appreciate the nudge in the right direction.

We organize our issues according to milestones -- I'd check out either an issue tagged beginner friendly or an issue in our active development milestone. 

Thanks for finding us!

PR #369, opened 2016-06-10
added some more stuff to the change log. Added helpers to make change…

Kevin-Wei:
… log tell yo who changed stuff

It relates to the following issue #s: 
* Attempt at #319 

we need some more helpers to clean up the way the thing looks. Lots of guid hiding needed here

PR #707, opened 2016-10-29
Better changelog

### LINKED ISSUE #319: Incorporate changelog for pregnancy and clinic models, in addition to patient model
Right now, the `change_log` partial in the pregnancy edit view is only really displaying changes made to the patient object, such as name and phone number. We should adjust so that changes to pregnancy and clinic objects are represented here as well. 

Full description TK.

### LINKED ISSUE #639: AuditTrail#modified_by_id not populating when an object is edited (MongoidHistory problems)
When I edit an existing patient or pregnancy, I expect that my user ID will save to the AuditTrail entry.

Instead, the AuditTrail object is being saved with `modified_by_id` being set to nil.

What I want is for `modified_by_id` to populate with my user ID when an object is modified and a new AuditTrail object is created.

colinxfleming:

Takes @Kevin-Wei 's work in #369 and starts pushing it forward again so that the changelog fixes are usable.

- Adds in pregnancy changes to changelog
- Adds some entrypoints for future development, so that continued work on this takes place in the audit trail model

TK: 
- <strike>Tests for model and helper</strike>
- <strike>Figure out why some BRs aren't getting escaped properly</strike>
- <strike>Actually get user to show up. Not sure what the deal here is exactly (userstamping it like we have other methods doesn't seem to be terribly cooperative), but we're working on it</strike>

- Fixes #319, eventually
- Fixes #639 

** Note: Run rename task on the AuditTrail collection after merging **


colinxfleming:
A few shaved yaks later I think this is in decent shape. @Kevin-Wei can I trouble you to review? 

Spits out a table like this. (The empty array in special circumstances is still screwing with it, but pregnancy changes are there, and adding more stuff to the shell should be easier for when we get around to implementing pledge objects.)

screen shot 2016-10-31 at 12 50 28 am

app/models/audit_trail.rb
```diff
@@ -1,18 +1,39 @@
...
+  include Mongoid::Userstamp
+  mongoid_userstamp user_model: 'User'
+
+  IRRELEVANT_FIELDS = %w(user_ids updated_by)
 
   # convenience methods for clean view display
-  def tracked_changes_fields
-    modified.keys.map(&:humanize)
+  def date_of_change
+    created_at.display_date
```
colinxfleming: I hear you. Decision from team design was that just date was fine, and it'll still sort in order of change.

colinxfleming:
I've stowed the TODO for a bit later. More value in having this ready to go than it does perfect. (Thanks @Kevin-Wei )

Issue #709, opened 2016-11-03
BUDGET: Create class method for determining money pledged in a given week


colinxfleming:
Our goal here: Take all patients whose appointment date is in a given week and calculate the DCAF pledge amount. Here's the rub: I don't know what the best way to do it is. 

Here are the requirements: We need to be able to filter to pledges in a given calendar week, and distinguish those between pledges that have been sent and those that haven't.

Here's a suggested path to this, but I'm amenable to another option:
* Create a class method on the `Patient` object that filters down to anyone with an appointment date in a given week and maps that collection of objects down to a hash like `{ pledged: 1000, sent: 500 }`.

We'll in turn use this to build a helper method that takes a hash as an input and spits out a nice looking budget bar.

lomky:
I'll take a stab at this :grinning: 

My only question is if it would be one function or two:

Are there other use cases that would want all the patients with appointments in the next week?
 - if so, then we could teach the `Patient` object a more generic function like `patients_with_upcoming_appointment` taking a duration parameter, and then maybe move the data extraction we're looking for into a helper.

Alternately if that is over-engineering it, we could always make that split later if we ran into another use case later.

colinxfleming:
Love the idea of abstracting it like that -- if this is something that
lends itself well to that, I'd say go for it!

On Thu, Nov 3, 2016 at 7:38 PM Kathryn Tipton <notifications@github.com>
wrote:

I'll take a stab at this 😀

My only question is if it would be one function or two:

Are there other use cases that would want all the patients with
appointments in the next week?

   - if so, then we could teach the Patient object a more generic function
   like patients_with_upcoming_appointment taking a duration parameter, and
   then maybe move the data extraction we're looking for into a helper.

Alternately if that is over-engineering it, we could always make that split
later if we ran into another use case later.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/colinxfleming/dcaf_case_management/issues/709#issuecomment-258306246>,
or mute the thread

lomky:
I'm not going to get this to a PR state tonight. I'll try to get it there this weekend!

lomky:
@colinxfleming I think I have found where this is sticking in my head. I'm used to a slightly different MVC pattern.

[...]

Issue #634, opened 2016-09-14
CMD Reporting MVP


mebates:
Real time reporting for the Case Manager Directors. 

Notes from user interviews: https://github.com/colinxfleming/dcaf_case_management/wiki/Data-Reporting-Management-notes

Need to work to define the MVP based on the data we're currently collecting, then perhaps a second release after we have full budgeting functionality. Go team!

mebates:
@mjrich check out our first hit at reporting needs! can datakind help?

mebates:
First stab at a very-very MVP "DCAF Data Center", based on data I know we can immediately display. Obviously the MD and VA values will be "null" until we start those call lines. We have a lot more reporting to display, so we'll probably have to work in some navigation options. 

![dcafwireframe091316_data center](https://cloud.githubusercontent.com/assets/6223901/18497755/21fcaa76-79ff-11e6-917a-a57bc1ef6d6b.png)

Patients Contacted = each unique patient to which one call was made, even if we didn't reach them
New patients Contacted = each new patient to which a first call was made, even if we didn't reach them

Do you like the name "Data Center"? Any other ideas? Other thoughts on the layout?

lwaldsc:
I love this so much!

Is there a way to add a CM's personal stats in there as well?

Something like DC/MD/VA/You

I know that's something a couple CMs have mentioned wanting to have and
something we've briefly mentioned in the past.

NerdyGirl537:
This looks great.  A quick question about the "new patients contacted": is that counting all the callers entered in the CM App?  If not, I'd love to see a summary of how many calls came in during the week.  

Also, I noted that this is for the CMDs.  Would it be available for the CMs?  

For naming, perhaps "Call Line Summary"?

And just a random note for realism: we usually send between 6 and 12 pledges a week.

lwaldsc:
Ahhhhhhhh I missed that it was for the CMDs only. I do think the CMs would
like a version of this though, including their own personal stats. Maybe
not as detailed? Not an urgent request, obviously, but something nice if
the coding isn't difficult.

colinxfleming:
Making a quick note before I forget to make sure we account for the info
that nnaf asks us for every year

colinxfleming:
[...]

PR #715, opened 2016-11-04
WIP Create patients contacted class method

### LINKED ISSUE #712: REPORTING: Create Patients contacted class method
Sourced from #634 : It's time to start working on some of the backend machinery so we can populate the components of the Data Center. The first step of this is to 

Here's what I'd like: 

* A class method `Patient.contacted_since(datetime)`, which takes a regular datetime or timestamp (e.g. `Time.now`) or a date as an argument
* Which filters to all patients with a call logged as `Reached patient` >= the `date` and counts them
* And which identifies how many of those patients have been marked `pledge_sent` this week
* And which outputs something like a hash as follows: 
```rb
{ 
  since: datetime, 
  contacts: (count of reached patients),
  first_contacts: (count of first time reached patients),
  pledges_sent: (count of pledges sent)
}
```

It is TOTALLY FINE if you get just a component of this done, e.g. naming the method and just getting all patients who have been reached -- developing infrastructure and leaving it for another to finish off is awesome!

Tests would be fabulous, but cool if it's not in the cards!

agkramer:
Let's start small here - submitting this pull request:

* adds name and comments for creating describe 'contacted_since method' in patient_test.rb
* adds name and comments for creating contacted_since(datetime) class method

* create patient contacted since class method

Fixes #712

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -144,6 +144,10 @@ class Patient
     false
   end
 
+  def contacted_since(datetime)
+    
+  end
+
   # Search-related stuff
   class << self
     # Case insensitive and phone number format agnostic!
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -112,19 +112,17 @@ class Patient
     #   pledges_sent: (count of pledges sent)
     # }
 
-    patient_reached = 0
-    patient_pledges_sent = 0
-
 
     Patient.each do |patient|
-      patient.calls.each do |call|
-        if call.reached? && call.updated_at >= datetime
-          reached_patient += 1
-        end
+      calls = patient.calls.select { |call| call.status == "Reached patient" && call.updated_at >= datetime }
+      unless calls.nil?
+        patients_reached << patient
       end
     end
 
-    return { :since => datetime, :contacts => reached_patient, :first_contacts => 20, :pledges_sent => 20 } #hard coding in first_contacts and pledges_sent for now
+    first_contact = patients_reached.select { |patient| patient.initial_call_date >= datetime }
+
+    return { :since => datetime, :contacts => patients_reached.length, :first_contacts => first_contact.length, :pledges_sent => 20 } #hard coding in first_contacts and pledges_sent for now
   end
 
   def recent_calls
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -100,6 +100,33 @@ class Patient
     end
   end
 
+  def self.contacted_since(datetime)
+    # 1) filters to all patients with a call logged as Reached patient >= the date and counts them
+    # 2) identifies how many of those patients have been marked pledge_sent this week
+
+    # returns hash
+    # {
+    #   since: datetime,
+    #   contacts: (count of patients reached)
+    #   first_contacts: (count of first time reached patients),
+    #   pledges_sent: (count of pledges sent)
+    # }
+
+    patient_reached = 0
+    patient_pledges_sent = 0
+
+
+    Patient.each do |patient|
+      patient.calls.each do |call|
+        if call.reached? && call.updated_at >= datetime
+          reached_patient += 1
+        end
+      end
+    end
+
+    return { :since => datetime, :contacts => reached_patient, :first_contacts => 20, :pledges_sent => 20 } #hard coding in first_contacts and pledges_sent for now
+  end
+
   def recent_calls
     calls.order('created_at DESC').limit(10)
   end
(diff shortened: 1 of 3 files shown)


npm-debug.log
```diff
@@ -0,0 +1,201 @@
+0 info it worked if it ends with ok
```
colinxfleming: Not quite clear why this is included -- if it's safe to remove, can you strip it?

app/models/patient.rb
```diff
@@ -144,6 +144,19 @@ def still_urgent?
     false
   end
 
+  def contacted_since(datetime)
```
colinxfleming: You'll want to make this a class method instead of an instance method -- was just discussing this in #709 as a matter of fact! Also check out http://culttt.com/2015/06/10/understanding-class-methods-verses-instance-methods-in-ruby/ for a fuller explanation.

npm-debug.log
agkramer: Will do.  I might have done this by accident when setting up.

agkramer:
alright @colinxfleming, I made some edits - take a look and let me know if I'm on the right track.  

Also, having some difficulty with understanding exactly what you want for:
* ```pledge_sent``` (total number of pledges sent this week? or pledges sent since ```datetime```?), and
* ```first_contact``` (is this a field somewhere, or am I adding another count for if this was the first time this person has been reached?).

colinxfleming:
hey bud! Answering your questions real quick -- 

a) Let's do all this (including `pledges_sent`)  as since Datetime, so we can be sure to make it really flexible and consumable in other places. For example, see #634 -- there are a few places we can use something similar to this.
b) for first contact, you can use the field `initial_call_date`.

app/models/patient.rb
```diff
@@ -100,6 +100,33 @@ def self.trim_urgent_patients
...
+    #   contacts: (count of patients reached)
+    #   first_contacts: (count of first time reached patients),
+    #   pledges_sent: (count of pledges sent)
+    # }
+
+    patient_reached = 0
+    patient_pledges_sent = 0
+
+
+    Patient.each do |patient|
```
colinxfleming: You might try out ruby's `select` here instead of iterating and incrementing -- probably a bit quicker to filter down rather than looking thru things!

colinxfleming:
@agkramer confirm that you're on the right track! Looking sharp. Let me know how everything is cooking.

agkramer:
@colinxfleming - I'm having a little trouble trying to understand using select to filter based on embedded data.  I'm writing this then thinking this is likely not the way to go about this.  Please advise.

```
  def self.contacted_since(datetime)
    # 1) filters to all patients with a call logged as Reached patient >= the date and counts them
    # 2) identifies how many of those patients have been marked pledge_sent this week

    # returns hash
    # {
    #   since: datetime,
    #   contacts: (count of patients reached)
    #   first_contacts: (count of first time reached patients),
    #   pledges_sent: (count of pledges sent)
    # }


    Patient.each do |patient|
      calls = patient.calls.select { |call| call.status == "Reached patient" && call.updated_at >= datetime }
      unless calls.nil?
        patients_reached << patient
      end
    end

    first_contact = patients_reached.select { |patient| patient.initial_call_date >= datetime }

    return { :since => datetime, :contacts => patients_reached.length, :first_contacts => first_contact.length, :pledges_sent => 20 } #hard coding in first_contacts and pledges_sent for now
  end
```

colinxfleming:
Hey @agkramer , can you push up to github so I can see exactly what's in the code?

colinxfleming:
(Also if you can get it to work another way, definitely go with it! Better to have this in than perfect)

app/models/patient.rb
```diff
@@ -100,6 +100,31 @@ def self.trim_urgent_patients
...
+    # returns hash
+    # {
+    #   since: datetime,
+    #   contacts: (count of patients reached)
+    #   first_contacts: (count of first time reached patients),
+    #   pledges_sent: (count of pledges sent)
+    # }
+
+
+    Patient.each do |patient|
```
colinxfleming: Swap this out with `all.each do |patient|` -- the `Patient` part of this is inferred from it being a class method.

app/models/patient.rb
```diff
@@ -100,6 +100,31 @@ def self.trim_urgent_patients
...
+      calls = patient.calls.select { |call| call.status == "Reached patient" && call.updated_at >= datetime }
+      unless calls.nil?
+        patients_reached << patient
```
colinxfleming: We aren't setting this `patients_reached` variable anywhere -- for now, can we declare it above the `each` thing?

PR #716, opened 2016-11-05
WIP - Budget bar helper!

### LINKED ISSUE #710: BUDGET: Create budget bar helper method
The task here is to develop a Rails helper method that we can use to take an input and spit out a budget bar.

Here are the technical requirements:

* Generate a helper called `budget_bars_helper`
* Assume an input as follows: a hash consisting of the following: `{ pledged: INT, sent: INT }`
* Ideally using the [bootstrap progress bar component](http://getbootstrap.com/components/#progress), have the output be a progress bar with the pledged in one color and sent in another color. For testing, compare to 1000. So assuming an input of `{ pledged: 500, sent: 250 }`, it should fill 75% of the budget bar, with 50% colored as pledged and 25% colored as sent, and 25% as left open.
* Assume that we're going to just drop it into a view like so:
```ruby
<div class="row">
  <%= budget_bar(@cash)%>
</div>
```
* Style roughly according to @mebates ' sketch here: 
image

Please flag if I can clarify anything!

colinxfleming:

Adds a budget bar helper method we can drop into the view. Method to populate this is in active development by another dev, but as long as we feed it a hash it'll work perfectly fine. Will tag in our UXers when it's good and ready!

* Adds code to generate progress bars! Just pop in a hash with keys `pledged` and `sent` and values of integers and out comes the progress bar.

TODOs: 
- [ ] Account for possibility of overflow
- [ ] Set a minimum width or some way to drop $$ info into the view like in the #710 picture
- [ ] Possibly adjust CSP to allow for inline style (consult with @mccabe615 about the goodness of this idea, otherwise we can whip up something in a stylesheet I'm sure)
- [ ] Add env vars to heroku
- [ ] Apply other styling as requested by team UX
- [ ] Test things!

SUPER MEGA PRELIM VERSION TO PROVE THAT THE CONTENT TAGS WORK:
screen shot 2016-11-04 at 10 32 51 pm

* Fixes #710


colinxfleming:
Other TODO: https://signalvnoise.com/posts/1108-what-belongs-in-a-helper-method (oops)

colinxfleming:
Feeling good about this now. This stitches in @lomky 's work building the expenditures methods. Note that we'll likely have to make some small adjustments there to get the time worked out right, but this pops the core in and gives us something to iterate on, and the space to handle stuff like overflow. 

Keeping as WIP for now but this is looking better.

colinxfleming:
From here, I think we can limit any edits to the Patient model's method, the budget bar helper, or the budget bar view, and we can iterate on this as it stands. Merging after I get builds to green.

Issue #711, opened 2016-11-03
PLEDGE VALIDATION: Validate Pregnancy#pledge_sent field in model


colinxfleming:
Here's a hugely helpful thing that lets us ensure we're always working with high quality data -- 

We need to validate the `pledge_sent` field to ensure that the following fields are set before we allow people to mark it true:

* Name
* DCAF soft pledge amount
* Clinic
* Appointment date

Would be AWESOME if you can also write unit tests confirming presence of each of the fields as well! 

Should be solvable with some good googling. HMU if there are snags!

colinxfleming:
(nb @ashlynnpai 's got this!)

PR #718, opened 2016-11-05
attempts at PLEDGE VALIDATION: Validate Pregnancy#pledge_sent field in model #711

ashlynnpai:
Code is not working.

I tried a couple of things:

* Conditional validations, which we talked about. Here's an example from Railscasts: http://railscasts.com/episodes/41-conditional-validations. If the condition is true, the validation on pledge_sent gets triggered, otherwise it gets bypassed.

* Using the same method with before_update.

I'm out of ideas (I think) for keeping the logic in the model instead of blocking the update in the controller. Do you see a problem in my code?

### DIFF
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -33,6 +33,17 @@ class Pregnancy
   # Validations
   validates :created_by,
             presence: true
+  validate :pledge_sent, :check_other_validations?, if: :updating_pledge_sent?
+
+  def updating_pledge_sent?
+    pledge_sent == true
+  end
+
+  def check_other_validations?
+    if dcaf_soft_pledge.blank?
+      errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank')
+    end
+  end
 
   # History and auditing
   track_history on: fields.keys + [:updated_by_id],
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -59,14 +59,9 @@ class Pregnancy
   end
 
   def check_other_validations?
-    if dcaf_soft_pledge.blank?
-      errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank')
-    elsif patient.name.blank?
-      errors.add(:pledge_sent, 'Patient name cannot be blank')
-    elsif patient.clinic_name.blank?
-      errors.add(:pledge_sent, 'Clinic name cannot be blank')
-    elsif patient.appointment_date.blank?
-      errors.add(:pledge_sent, 'Appointment date cannot be blank')
-    end
+    errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank') if dcaf_soft_pledge.blank?
+    errors.add(:pledge_sent, 'Patient name cannot be blank') if patient.name.blank?
+    errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic_name.blank?
+    errors.add(:pledge_sent, 'Appointment date cannot be blank') if patient.appointment_date.blank?
   end
 end
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -33,7 +33,7 @@ class Pregnancy
   # Validations
   validates :created_by,
             presence: true
-  validate :pledge_sent, :check_other_validations?, if: :updating_pledge_sent?
+  validate :pledge_sent, :check_other_validations, if: :updating_pledge_sent?
 
   # History and auditing
   track_history on: fields.keys + [:updated_by_id],
@@ -51,14 +51,14 @@ class Pregnancy
   #   end
   #   false
   # end
-  
+
   private
-  
+
   def updating_pledge_sent?
     pledge_sent == true
   end
 
-  def check_other_validations?
+  def check_other_validations
     errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank') if dcaf_soft_pledge.blank?
     errors.add(:pledge_sent, 'Patient name cannot be blank') if patient.name.blank?
     errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic_name.blank?
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -35,16 +35,6 @@ class Pregnancy
             presence: true
   validate :pledge_sent, :check_other_validations?, if: :updating_pledge_sent?
 
-  def updating_pledge_sent?
-    pledge_sent == true
-  end
-
-  def check_other_validations?
-    if dcaf_soft_pledge.blank?
-      errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank')
-    end
-  end
-
   # History and auditing
   track_history on: fields.keys + [:updated_by_id],
                 version_field: :version,
@@ -61,4 +51,22 @@ class Pregnancy
   #   end
   #   false
   # end
+  
+  private
+  
+  def updating_pledge_sent?
+    pledge_sent == true
+  end
+
+  def check_other_validations?
+    if dcaf_soft_pledge.blank?
+      errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank')
+    elsif patient.name.blank?
+      errors.add(:pledge_sent, 'Patient name cannot be blank')
+    elsif patient.clinic_name.blank?
+      errors.add(:pledge_sent, 'Clinic name cannot be blank')
+    elsif patient.appointment_date.blank?
+      errors.add(:pledge_sent, 'Appointment date cannot be blank')
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


colinxfleming:
You're SUPER close on this! Like, mega close. Digging in now!

app/models/pregnancy.rb
```diff
@@ -33,6 +33,15 @@ class Pregnancy
   # Validations
   validates :created_by,
             presence: true
+  before_update :check_other_validations?
+
+  def check_other_validations?
+    dcaf_soft_pledge.present? &&
+    Patient.first.name.present? &&
```
colinxfleming: Swap these out for lowercase `patient` and remove the `first`s -- it'll be smart enough to pull the associated object thru activerecord.

app/models/pregnancy.rb
```diff
@@ -33,6 +33,15 @@ class Pregnancy
...
+    Patient.first.appointment_date.present? &&
+    Patient.first.clinic.name.present?
```
colinxfleming: Swap the `clinic.name` here out for `clinic_name` -- we haven't implemented Clinic objects yet, which is something I should have done a better job of documenting! Sorry, that's on me.

test/models/pregnancy_test.rb
```diff
@@ -43,4 +43,23 @@ class PregnancyTest < ActiveSupport::TestCase
       assert @pregnancy.created_by
     end
   end
+
+  before do
+    @pregnancy.dcaf_soft_pledge = 500
+    @pt_1.appointment_date = DateTime.now + 14
+    @clinic = create :clinic
```
colinxfleming: Same deal as above -- go ahead and set `@pt_1.clinic_name` to some string or another. Doesn't matter what just as long as it's not nil.

test/models/pregnancy_test.rb
```diff
@@ -43,4 +43,23 @@ class PregnancyTest < ActiveSupport::TestCase
...
+  end
+
+  describe 'pledge_sent' do
+    it 'should update the pledge_sent field to true' do
+      @pregnancy.pledge_sent = true
+      assert_equal true, @pregnancy.pledge_sent
```
colinxfleming: I think this might be a trouble spot -- because this is just checking value and not validation, I think this is always going to return true. 

Think of the common pattern in controllers -- 

```
if @patient.save
  # do successful thing
else
  # do failure thing
end
```

The actual `save` method is going to return true or false based on whether the object is valid and saves, which will then execute whichever part of the conditional. We could use `valid?` here too. But the value will hang out as true until you actually check for validity, which `save` does.

For example:

```
 irb(main):001:0> a = Patient.first
=> #<Patient _id: 58164096311c8852aca80137 ... >
irb(main):002:0> a.primary_phone = 'ashlynn is awesome' # set this to a string
=> "ashlynn is awesome"
irb(main):003:0> a.primary_phone # still a string
=> "ashlynn is awesome"
irb(main):004:0> a.valid? # check for validity, which will stop it 
=> false
irb(main):005:0> a.primary_phone # and now it's gone!
=> ""
```

I'd recommend changing this to `assert @pregnancy.valid?`. Let me know if this makes sense!

test/models/pregnancy_test.rb
```diff
@@ -43,4 +43,23 @@ class PregnancyTest < ActiveSupport::TestCase
...
+    end
+
+    it 'should not update pledge_sent field to true if the DCAF soft pledge amount is nil' do
+      @pregnancy.dcaf_soft_pledge = nil
+      @pregnancy.pledge_sent = true
+      assert_equal nil, @pregnancy.pledge_sent
```
colinxfleming: Similar deal here -- `pledge_sent` is going to always return true here because we haven't actually checked for validity yet. Swap the assertion here out to `refute @pregnancy.valid?`

app/models/pregnancy.rb
```diff
@@ -33,6 +33,15 @@ class Pregnancy
   # Validations
   validates :created_by,
             presence: true
+  before_update :check_other_validations?
```
colinxfleming: Ahh you are so close! We should try and do this with validations rather than hooks if we can. Here's a suggested potential way that use the `check_other_validations?` method you wrote:

* Exclusions: http://guides.rubyonrails.org/active_record_validations.html#exclusion
Something like `exclusion: in: [ true ], unless: check_other_validations?` (In plain english: 'Value cannot be true, unless check other validations is also true.`

ashlynnpai:
I tried to get the integration test to pass by supplying additional data. The test still failed and the data seemed to intermittently persist to other tests, so just commented out.

ashlynnpai:
I should refactor the test data in the model tests when we figure out how to get this working.

test/integration/submit_pledge_test.rb
```diff
@@ -6,6 +6,9 @@ class SubmitPledgeTest < ActionDispatch::IntegrationTest
     @user = create :user
     @patient = create :patient
```
colinxfleming: FactoryGirl lets you create stuff and assign attributes in one line -- if you do `@patient = create :patient, clinic_name: 'Nice Clinic', appointment_date: DateTime.now + 14` and `@pregnancy = create :pregnancy, patient: @patient, dcaf_soft_pledge: 500` it'll create the objects and set those attributes while it does it. 

I got the test to pass using this! Maybe it wasn't saving as written?

ashlynnpai: I thought you could set attributes after creation. I'll take another look tonight but glad you got it to pass.

ashlynnpai: I see, there was a submit at the end. Thanks for catching that!!

ashlynnpai:
Thanks man!! Let me know if this works for you.

test/models/pregnancy_test.rb
```diff
@@ -43,4 +43,35 @@ class PregnancyTest < ActiveSupport::TestCase
...
+    @pregnancy.dcaf_soft_pledge = 500
+    @pt_1.clinic_name = 'Nice Clinic'
+    @pt_1.appointment_date = DateTime.now + 14    
+  end
+
+  describe 'pledge_sent validation' do
+    it 'should not validate pledge_sent if the DCAF soft pledge field is blank' do
+      @pregnancy.dcaf_soft_pledge = nil
+      @pregnancy.pledge_sent = true
+      refute @pregnancy.valid?
```
colinxfleming: Oh man this is great, A++++ work. Love how readable and clear this is. 

Can you confirm in these sets of tests that the right error message comes back too? I think that'll keep the test coverage at 100%, and we've got some modal stuff coming up that will consume that.

app/models/pregnancy.rb
```diff
@@ -50,4 +51,22 @@ class Pregnancy
...
   # end
+  
+  private
+  
+  def updating_pledge_sent?
+    pledge_sent == true
+  end
+
+  def check_other_validations?
+    if dcaf_soft_pledge.blank?
```
colinxfleming: If these aren't set, we want them to all spit out, yeah? What are your thoughts on turning these into a bunch of one-liners? e.g.:

```
errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank') if dcaf_soft_pledge.blank?
# ... and so on
```

app/models/pregnancy.rb
```diff
@@ -33,6 +33,7 @@ class Pregnancy
   # Validations
   validates :created_by,
             presence: true
+  validate :pledge_sent, :check_other_validations?, if: :updating_pledge_sent?
```
colinxfleming: I like this a lot -- validating only when changing it to true is a really creative and clever implementation! Nicely done 💯

app/models/pregnancy.rb
```diff
@@ -50,4 +51,22 @@ class Pregnancy
...
   #   false
   # end
+  
+  private
+  
+  def updating_pledge_sent?
+    pledge_sent == true
+  end
+
+  def check_other_validations?
```
colinxfleming: What are your thoughts about renaming this method? Usually having `?` at the end of a method indicates it either returns true or false, and this isn't really doing this in the current (very good and useful!) implementation. Instead, it's confirming that necessary information is saved so it can mark the pledge sent. Thoughts?

app/models/pregnancy.rb
ashlynnpai: Sure. Will work on it this week.

ashlynnpai:
I seem to be having problems again with the test data not resetting.

colinxfleming:
@ashlynnpai are you still hitting that snag? This looks pretty good to me, and the test failure is from one of our scrub tests.

colinxfleming:
This looks to me like we can merge it if you're feeling good about it @ashlynnpai ! Let me know!

ashlynnpai:
Yes, I ran the test suite three times. Looks good.

colinxfleming:
@ashlynnpai you are number one! Thank you so much for pushing this over the finish line and staying on this!

colinxfleming:
Fixes #711 🎉

PR #720, opened 2016-11-11
Improvements to Dockerfile, reducing layers, etc...

DarthHater:

A quick look at the Dockerfile today, I saw some room for improvement. Reducing the amount of layers created, removing apt-get cache, some DRY stuff. 

* Modify Dockerfile to do apt-get install all in one fell swoop
* Also modify Dockerfile to do apt-get update in same layer as install to prevent caching issues
* ENV variable for DCAF_DIR so it can be changed in one place rather than 4-5 (can't remember)
* Addition of cleaning temp files, apt-get clean to reduce image size

Best practices gleaned from:
* https://intercityup.com/blog/downsizing-docker-containers.html
* https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,33 @@
 FROM rails:4.2.6
 MAINTAINER Colin Fleming <c3flemin@gmail.com> 
 
+# configure environment variable
+ENV DCAF_DIR=/usr/src/app
+
 # install packages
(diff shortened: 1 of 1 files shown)


mchelen:
Looks good to me, based on a quick test.

colinxfleming:
🎉

PR #722, opened 2016-11-11
Dockerfile user change

DarthHater:

It's generally not the best idea to have root running your application. Switched to a specific user, dcaf. 

* Adds a dcaf user and group
* switches ownership of DCAF_DIR to dcaf user and group
* switches the Docker image to use the dcaf user


colinxfleming:
Looks like this is OBE due to the heroic work in #723 and #724; closing. ( @DarthHater let me know if I've got that wrong!)

PR #723, opened 2016-11-11
Docker ignore file

DarthHater:

Addition of a dockerignore file, prevents unnecessary files from making it into the image 

* Addition of barebones Dockerignore file, just ignoring .git, .gitignore, README.md for now


mchelen:
Seems like a good practice to use :+1:

PR #724, opened 2016-11-12
Migrate from rails to ruby Dockerfile

### LINKED ISSUE #721: Switch Dockerfile from rails to ruby
According to the rails official image, it will be deprecated on Dec 31st, 2016:

https://hub.docker.com/_/rails/

Advice is to switch to ruby, install rails, etc...

Some handy dandy instructions on how to do so: http://blog.kontena.io/dockerizing-ruby-application/

DarthHater:

Sooooo this switches the Dockerfile from rails (deprecated) to ruby, also switches to using alpine from debian, and trims down ruby quite a bit. End result is the image is half the previous size, should run faster, etc... This PR would supercede #720 #722 and #723  

Testing looked ok to me, someone with much more knowledge of the ecosystem should definitely examine since it's an OS change, etc...

* Switches to ruby:3.2.1-alpine for base image
* Rejumbling of packages installed
* Switches to using a dcaf user ( consistent with #722 )
* Adds an expose for port 3000
* Adds dockerignore file 

* Fixes #721

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.gitignore
+README.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,9 @@ FROM ruby:2.3.1-alpine
 MAINTAINER Colin Fleming <c3flemin@gmail.com> 
 
 # configure environment variable
-ENV DCAF_DIR=/usr/src/app
+ENV DCAF_DIR=/usr/src/app \
+    BUILD_DEPENDENCIES="build-base libxml2-dev libxslt-dev linux-headers" \
+    APP_DEPENDENCIES="nodejs"
 
 # get our gem house in order
 RUN mkdir -p ${DCAF_DIR}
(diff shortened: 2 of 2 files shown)


DarthHater:
Test failures on all of the commits, but they seem unrelated ¯\_(ツ)_/¯

colinxfleming:
Yeah, the test failures are 100% unrelated -- we have a couple scrub tests. 

Here's my output from `docker-compose run web` after doing `docker-compose build` -- looks like it chokes on a `tzinfo` error. @DarthHater can you replicate or is this just my machine acting weird? 

```
$ docker-compose run web 
Array values in the parameter to `Gem.paths=` are deprecated.
Please use a String or nil.
An Array ({"GEM_PATH"=>["/usr/local/bundle"]}) was passed in from bin/rails:3:in `load'
=> Booting WEBrick
=> Rails 4.2.7.1 application starting in development on http://0.0.0.0:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Exiting
/usr/local/bundle/gems/tzinfo-1.2.2/lib/tzinfo/data_source.rb:182:in `rescue in create_default_data_source': No source of timezone data could be found. (TZInfo::DataSourceNotFound)
Please refer to http://tzinfo.github.io/datasourcenotfound for help resolving this error.
# ... big ol' stack trace
```

colinxfleming:
Appears to be very straightforwardly resolvable by adding the `tzinfo-data` gem to the gemfile though. I'll spin that up now.

DarthHater:
It did not, @colinxfleming , I'm on OS X so maybe there's something native that can use (but that seems odd too). I can add this I think as an APK package if you want.

DarthHater:
Actually, I can't add it. For some reason there isn't a ruby-tzinfo package for Alpine 3.4 yet. Gem install probably best for now!

PR #725, opened 2016-11-12
Add tzdata-info to run rails server in docker

colinxfleming:

* Adds `tzinfo-data` gem to make `docker-compose run web` not choke on a timezone error
* Includes everything in #724 

@DarthHater @mchelen can you all take a look and confirm this is cool? This seemed like the least painful solution to the problem (other stuff involves stackoverflow or docker repo arguments over what wild workaround is the least unobtrusive).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.gitignore
+README.md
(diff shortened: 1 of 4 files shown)


DarthHater:
This looks fine to me. I don't have the issue locally but if it works for you then A+ +1

Issue #733, opened 2016-11-16
PLEDGE VALIDATION: Add errors to view


colinxfleming:
Known national treasure @ashlynnpai wrote up some validations which require

On the patient edit view load (or modal load or whatever), let's do the following:

* See what would happen if we did try and save a pledge (without actually saving it)
* If any errors come back, render them in the modal in `app/views/patients/_pledge.html.erb`

This will give our CMs a nudge in the right direction, keep our data quality high, and implement a notable quality of life improvement. 

This is the last big piece of #203 (besides testing), then we're in!

valeriecodes:
I'll take a stab at this!

valeriecodes:
It looks like the modals aren't currently using the pledge models or controllers. Should they be? If not, how should we check for errors?

colinxfleming:
Nah, keep it using the patients update endpoint like it is in the form.

As for method, I'm not sure if there is a Rails accepted way to do
something like 'try to validate under these conditions' - if so let's use
that, otherwise can just add some conditionals to the view, probably? What
do you think?

I'll make sure to depreciate the unused stuff this wknd. You're right to
flag this as it makes getting the lay of the land much harder. Sorry about
that!

PR #739, opened 2016-11-19
Add pledge errors to view

valeriecodes:
Addresses https://github.com/colinxfleming/dcaf_case_management/issues/733
Adds red text with error messages in case of invalid pledge.
Screenshot:
screen shot 2016-11-19 at 3 54 27 pm

I have a couple notes and questions, which I will add inline.


app/controllers/patients_controller.rb
```diff
@@ -20,6 +20,7 @@ def create
 
   def edit
     @note = @patient.notes.new
+    @pledge = Pledge.new(amount: @patient.pregnancy.dcaf_soft_pledge, patient: @patient, created_by: current_user, pledge_type: 'soft')
```
valeriecodes: Should the amount always be pulled from `dcaf_soft_pledge` and should the pledge type always be `soft`? The view currently references `dcaf_soft_pledge` as the amount

colinxfleming:
💯 💯 

I'll add an integration test tmrw and then merge. Thank you @valeriecodes

Issue #731, opened 2016-11-16
ACCOUNTANT: Add Fulfillment section to Patient edit view


colinxfleming:
This is a good opportunity to get a feel for how we have the patient view arranged. Here are the designs from @eheintzelman : https://cloud.githubusercontent.com/assets/17989540/19648468/5031e410-99d0-11e6-9331-76cac8a8ca50.png

For now, just recreate the panel on the right in the patients edit view. Most likely the steps here are as follows:
* Add a 'Fulfillment' section to `views/patients/edit.html.erb`, and `views/patients/_menu.html.erb` -- follow the standard
* Create a fulfillment partial; toss the skeleton of a form in it. Doesn't have to do anything yet, just stub it! (Remember that we use the rails bootstrap forms gem - feel free to liberally crib from other forms.)

Stems from #643

eheintzelman:
I got some feedback from the accountant making the design even simpler.  Here's an update. 

![pledge fulfillment v3 - landing](https://cloud.githubusercontent.com/assets/17989540/20367068/e9a5c99c-ac1b-11e6-89a4-61fe5a801e57.png)

mebates:
Looks great! I wonder if we could prevent the 'Pledge Fulfillment' tab from appearing or being clickable until after the pledge is sent. Good or bad UX?

colinxfleming:
Team dev can make that work either way, so whatever you think the good ux
solution is, haha. Either way we probably obscure it from regular CMs and
just show it to people with increased privileges, right?

ashlynnpai:
I'll take a stab.

eheintzelman:
Just talked to DCAF folks and yes it's fine to just obscure the pledge fulfillment from regular CMs, if that's easiest.

For the accountant/admin, I like Mollie's idea of only making the pledge fulfillment tab clickable once the pledge is sent.

ashlynnpai:
Can you tell me where the higher level privileges are defined? I see only admin_user.

colinxfleming:
I think `current_user.admin?` should work -- it should use the enum on the
user model, which is how we are giving people increased privilege. See some
of the users controller methods for an example of how we've been doing
this. (Sorry I can't be more specific, I'm on my phone!)

ashlynnpai:
Thanks, I didn't realize you were using admin for that.

ashlynnpai:
[...]

PR #741, opened 2016-11-20
ACCOUNTANT: Add Fulfillment section to Patient edit view #731

ashlynnpai:
* render fulfillment partial if admin and pledge has been sent
* show link if admin and make link clickable after pledge has been sent
* create fulfillment form hooked up to model
* create integration test to verify when link shows and when it becomes clickable

pledge_sent_true

Issues
* check weeks along field: is it showing the correct data?
* abortion care $ field: they need to enter an integer, not a string with "$", is this clear?
* boolean checkbox label: this is supposed to be bold by default 
* the disabled link is not lining up with the other links

pledge_sent_false

* currently using date_field because datepicker wasn't working. I'm not sure how to override the defaults and it might be easier to get datepicker working, if you want it exactly like in the mockup. Currently if you click or hover anywhere in the field a black arrow appears, if you click on that the calendar pops up, or you can click on the placeholder and the M/D/Y fields become active, then you can enter the date in directly.

date_field default

### DIFF
--- a/app/views/patients/_pledge_fulfillment.html.erb
+++ b/app/views/patients/_pledge_fulfillment.html.erb
@@ -9,26 +9,26 @@
 
       <div class="row">
         <div class="col-sm-12">
-          <%= f.fields_for patient.pregnancy do |pt| %>
-            <%= pt.form_group :resolved_without_dcaf do %>
-              <%= pt.check_box :resolved_without_dcaf, label: 'Pledge Fulfilled' %>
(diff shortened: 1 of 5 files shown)


app/views/patients/_pledge_fulfillment.html.erb
```diff
@@ -0,0 +1,36 @@
+<section id="pledge_fulfillment" class="abortion-info tab-pane active">
```
colinxfleming: Do me a solid and rename this to just `_fulfillment.html.erb`?

Issue #589, opened 2016-08-17
FRIDAY Establish multiple call lines on start


jayzmiles:
Based on - "As a CM, I need to be able to choose the call line I'm working on tonight so that I input patient's data into the correct line"

jayzmiles:
A couple questions/observations we ran into:

- Is there a situation where a user (in any role) would need to view information from multiple call lines at once?
- We may solve for this issue from the perspective of the CM and then work on adapting this feature to other roles? (higher-level views for DCAF leadership or others)

Please feel free to add anything I missed, @eheintzelman.

eheintzelman:
@lwaldsc @NerdyGirl537 We would like your input. We discuss at the Sept 4th meeting if that's easier.

Here are a few follow-ups to Jay's comment above, focusing on solving the permissions issues from the perspective of the CM:

1- Are there times, outside of working a line, when a CM would need access to the system?  If so, for what purpose? 
2- While working one line, are there situations when a CM would need to view or edit data for a different line?
3 - Are there situations when a CM would need to work multiple call lines?

NerdyGirl537:
@eheintzelman, the short answer on all three is currently no.  On the second item, there is a slight likelihood that a caller might select the wrong line and the CM not realize until after entering in the patient into the system.  That might require accessing the other line's data or some other way to get the patient's info.  

@jayzmiles, the CM directors and the treasurer may.  It's difficult to say without having the budget function built out and no reporting functions.  It might be better to ask the CM directors and treasure.

eheintzelman:
@NerdyGirl537 That certainly keeps things simple. What about the CMs handling the Spanish-only patients?

eheintzelman:
@colinxfleming Call line selection is ready for development.  It turned out to be quite simple for the CMs since there are not needs to access multiple lines at the same time or toggle between lines. We are not tackling roles other than CM that might benefit from the ability to view information from multiple lines at once in this user story. 

1. Add concept of 3 lines (in code).  DC Line, MD Line, VA Line.

2. Add screen allowing the user to select call line. 

Welcome to DCAF! (h2)
(blank space)
What line are you working tonight? (h3)
DC Line    MD Line   VA Line   (label/body; radio buttons; options are mutually exclusive) 
Cancel (color: red; otherwise standard button styling) Start (Standard button styling; justify right with cancel button)

3. Let the user know what line they are on in the header under "DCAF Case Manager." Options are DC Line, MD Line or VA LIne

[...]

PR #745, opened 2016-11-23
Add option to select from multiple lines on login

colinxfleming:

This sets up line selection on login. The way it works is as follows: 
* In the event that session isn't set, redirect to the line select path from any action in the dashboards controller
* In the event that session IS set, don't do anything
* The lineselect path just sets a form variable and redirects to authenticated root

Here's how it plays out: 

After login, the lineselect page:
image

From the dash, the `Current: XX` will take you back to the line selection path. 
image

@Kevin-Wei I'll be doing some associated work on this but it would rule if you could review and make sure this all makes sense! 

* Bumps #589

### DIFF
--- a/app/assets/javascripts/lines.coffee
+++ b/app/assets/javascripts/lines.coffee
@@ -1,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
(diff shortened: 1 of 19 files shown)


colinxfleming:
Next steps here before merge:

* Get a design review from @mebates

colinxfleming:
This is looking good and functional! I'll spin the design approval into its own issue.

mebates:
@colinxfleming - did you start a new issue for design approval? I would prefer if the button didn't include instructions. I think it can just say 'Start'. The question 'What line...?' is instruction enough.

colinxfleming:
@mebates I did not but I will. I think per discussion with jenn there are a
few pain points worth hammering out here

PR #752, opened 2016-11-26
[Don't merge, utility] - Data entry form

colinxfleming:

Adds a form matching the spreadsheet order for mass data entry, so we can punch in individual patients easier. 

TK testing and confirming that everything shows up right once entered. 

screen shot 2016-11-26 at 4 35 06 pm


colinxfleming:
I hand-tested this and it looks good except for adding notes, so I added a quick and dirty redirect once people are entered. 

Probably won't be worth merging but will be nice to have on hand for data entry.

NerdyGirl537:
@colinxfleming what instructions do I need to share with the volunteers for accessing this?

colinxfleming:
To close the loop on this, I've emailed @NerdyGirl537 separately with details for mass entry. I'll make a note when I've deployed this. 

Note for self: Looking increasingly like merging to master is the right thing to do, and then turning it off in a separate PR.

colinxfleming:
Drop of coverage isn't a serious problem due to this being adhoc. Ticket to close these routes once things are disabled is set. Merging.

Issue #727, opened 2016-11-13
ACCOUNTING: Make patient identifier searchable


colinxfleming:
Right now we have patient identifier as a method that's inferred from a patient's `line` and `primary_phone_no`. Our accountant primarily works with those identifiers, so we're going to want to incorporate that into our search functionality -- e.g. I should be able to enter `D3-1122` and have it return the patient with that identifier. 

Here are two options:
* persist as a db field that gets populated after save -- e.g. something like `after_save :update_patient_identifier, if: :primary_phone_or_line_changed?` plus associated model changes
* bolt something on to our existing search method that is smart enough to parse entering patient codes. If there's an easy way to do this, sweet -- if not let's not worry about it.

Either is perfectly fine with me. Developer's choice! 

This is requested by our accountant user and stems from needs in #643

DarthHater:
A few things on patient identifier, it's always a combination of those two fields? Is there a potential for them to be non unique (two patients could have the same identifier by accident because they share common fields that make it happen)?

Since it uses last five of phone number I could see it happening, if they shared line all you would need is two people in DC who have 11234 as their last five, right? Not the most common of occurrences for sure but it just feels weird to me allowing for such a case to occur. 

(If I'm overthinking this tell me) @colinxfleming @vtmeghan

colinxfleming:
@DarthHater just in case we never closed the loop on this -- I think the right way to do this is to ensure that the way we generate or infer a patient identifier is a standalone method, so we can pop it out with something better. I'll hit up DCAF's CMDs about this separately and we can spin that part of it into a separate method. Thanks for flagging this -- let me know if I can do anything else to unstick.

lwaldsc:
I definitely don't think we're married to using PT phone number based ID at all. I am 100% behind another unique identifier. 

As said, it may also help us for the few times there are different PTs with the same number. As long as we keep some sort of safeguard against multiple entries for one PT (not exactly we want to do here).

NerdyGirl537:
I think the issue here is referring to it as a "unique" identifier.  The idea of the identifier is to mask part of the PPI while providing enough information, when combined with the patient name, to make sure the correct pledge is being paid.  So think of it as a coded identifier rather than a unique one.  Thoughts?

DarthHater:
@NerdyGirl537 yeah, I get that. I'm referring to it as unique as if it's searchable, you could technically get two if you were using the identifier, potentially causing some UX issues, blah blah blah.

[...]

PR #754, opened 2016-11-27
WIP - Implement multiple lines

### LINKED ISSUE #589: FRIDAY Establish multiple call lines on start

colinxfleming:

Makes `current_line` mean something -- once a line is set in session, a CM's searches should only return individuals from that line, patients they create should be in that line, etc. This implements this logic in controllers and views, and refactors a couple things along the way.

s/o to the team for writing good, adaptable code. I was prepared for this to be absolutely hell on earth and pretty much every change we had to make turned out to be pretty straightforward.

* Changes line to the constant `LINES`, settable by `ENV['LINES']` and defaulting to `[:DC, :MD, :VA]`
* Adds scopes for different lines, in case we ever need them
* Remixes new patient form to have line be editable
* Alters `Patient#search` to accept a line scope

The only significant change to the views is in the new patient form partial. It's getting a little cramped and I think we kinda need line in there, though I defer to @mebates . Here's a scratch of expanded:

screen shot 2016-11-27 at 1 21 26 am

TODO:
- <strike> Allow users to edit line from patient edit view</strike> (jk)
- [x] Limit call lists to just people in the current line
- [x] Limit urgent lists to just people in the current line
* Add a whole load of tests for the following: 
- [x] Patient scopes, 
- [x] patient search line narrowing, 
- [x] integration test to confirm that searching on one line doesn't pick up patients from another line

One other thing: We can rig it up so that you can check a box to search all lines instead of just the current line. Not sure if this is needed or not. Would welcome thoughts about that from @NerdyGirl537 @lwaldsc @mebates .

* Fixes #589, at long last!


mebates:
I'm confused about why we would let them change the line for the patient. If you are logged in on the DC line and listening to calls from DC voicemail, shouldn't that be default? Do we need to even expose that anywhere? Any use case when you would need to change it @lwaldsc ?

eheintzelman:
We discussed this in issue #589.  See @NerdyGirl537 comments - she didn't think there was a common enough use case for switching lines to necessitate this functionality.

colinxfleming:
My bad for not reading the discussion closely -- thanks for linking, Eva.
I'll remove this before merge.

On Mon, Nov 28, 2016 at 10:14 AM Eva Heintzelman <notifications@github.com>
wrote:

We discussed this in issue #589
<https://github.com/colinxfleming/dcaf_case_management/issues/589>. See
@NerdyGirl537 <https://github.com/NerdyGirl537> comments - she didn't think
there was a common enough use case for switching lines to necessitate this
functionality.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/colinxfleming/dcaf_case_management/pull/754#issuecomment-263296626>,
or mute the thread

colinxfleming:
@mebates @eheintzelman a thing I was going to maybe punt on but maybe not -- should we add a box to the main dashboard search for 'Search all lines and not just your current line'? Use case here is probably mostly Spanish CMs, who bounce between lines to talk to Spanish speaking patients.

Defaulting to no, but can build it out if you think we should -- I think it's a pretty straightforward implementation, based on adding lines to the search.

NerdyGirl537:
Thoughts on the Spanish CMs: I'm not sure they work more than one call at a
time.  There aren't usually a large number of Spanish callers; my
experience is that maybe one a week on a line.  Not sure the number of
calls warrants building it out.  Thoughts?

lwaldsc:
Yeah. The only time we would need a pt to switch a line is if they say,
called the DC line and were actually a MD pt.

That being said, I don't hate the idea of a full database search if it
doesn't compromise security or search speed significantly. Maybe instead of
a dropdown box, we could have a pop up that says something like "This PT is
currently files under the [MD/DC/VA] line. Please contact the CM on duty on
that line."

Thoughts?

colinxfleming:
If full db search winds up being a thing, let's tackle it outside this PR -- @lwaldsc and @NerdyGirl537 have me convinced that it's not necessary for release 2. Thanks all for your thoughts!

colinxfleming:
While settling in to do some work on this tonight I realized that I was a bit confused about the UX part, so I want to make sure of a few things and document my assumptions here: 

* There should be no line edit select box on the patient edit view
* Their line should be set on patient create and not changed after that
* A CM should not be able to swap between DC, MD, and VA 'lines' without logging out and logging back in
* Urgent patients should be scoped to a line (e.g. when on DC, should only show DC urgents)
* A call list should also be scoped to a line; e.g. if you had a MD shift last week and a DC shift this week, then only DC patients should show up on your call list when you log in and set to DC 

These are the assumptions that I'll program around unless I hear otherwise from @mebates @eheintzelman @NerdyGirl537 @lwaldsc . thx all!

eheintzelman:
@colinxfleming Your assumptions jibe with my understanding.

app/models/patient.rb
```diff
@@ -65,6 +69,7 @@ class Patient
             :primary_phone,
             :initial_call_date,
             :created_by,
+            :line,
```
lomky: Does the addition of `:line` to the `validates` list make it a required field? If so, what happens to all the existing Patients in production without one?

app/models/user.rb
```diff
@@ -82,12 +82,14 @@ def secure_password
   # that is less than 8 hours old,
   # AND they would otherwise be in the call list
```
lomky: Should this comment be updated to mention the new call line criteria, or is that implicit in the "would otherwise be in the call list" ?

app/views/dashboards/_search_form.html.erb
```diff
@@ -5,6 +5,7 @@
   <%= bootstrap_form_tag url: '/search', remote: true, layout: :inline do |p| %>
     <%= p.text_field :search, placeholder: 'Name (Susan Smith) or Phone (555-555-5555)', hide_label: true, class: 'search_form_input' %>
     <%= p.submit 'Search', class: 'btn btn-primary' %>
+    <%#= p.check_box :all_lines, label: "Search all lines instead of just #{current_line}" %>
```
lomky: Consider noting the issue # with the discussion of why this line exists but is commented out?

app/controllers/dashboards_controller.rb
```diff
@@ -1,14 +1,18 @@
...
 
   def index
-    @urgent_patients = Patient.urgent_patients
+    @urgent_patients = Patient.urgent_patients(current_line)
     @expenditures = Patient.pledged_status_summary
   end
 
   def search
-    @results = Patient.search params[:search]
+    # if all lines selected, search from all lines instead of current line?
```
lomky: This comment is a little confusing, since all lines can't be selected atm.

app/controllers/dashboards_controller.rb
```diff
@@ -1,14 +1,18 @@
...
+
   before_action :pick_line_if_not_set, only: [:index, :search]
 
   def index
-    @urgent_patients = Patient.urgent_patients
+    @urgent_patients = Patient.urgent_patients(current_line)
     @expenditures = Patient.pledged_status_summary
   end
 
   def search
```
lomky: Keep an eye out for merge conflicts with the issue touching search & call lines (through the patient id's) #727

app/helpers/lines_helper.rb
```diff
@@ -1,15 +1,13 @@
 module LinesHelper
   def lines
-    ENV['LINES'].present? ? split_and_strip(ENV['LINES']) : %w(DC MD VA)
+    LINES
   end
 
-  def current_line
+  def current_line_display
```
lomky: Changing function names gives me the shivers. I see you got the tests all fixed up, so I'll grep through the code base a bit.

test/helpers/lines_helper_test.rb
```diff
@@ -5,15 +5,16 @@ class LinesHelperTest < ActionView::TestCase
 
   describe 'convenience methods' do
     it 'should spit out an array of available lines' do
-      assert_equal lines, %w(DC MD VA)
+      assert_equal lines, %w(DC MD VA).map(&:to_sym)
     end
 
     it 'should return current line' do
       assert_nil current_line
```
lomky: This test still passes, but it's worth noting it used to call a different function (what is now `current_line_display`) and thus is testing a different section of code. Maybe add another line `assert_nil current_line_display` to be thorough?

app/helpers/lines_helper.rb
lomky: My paranoia is satisfied that nothing will break. Smaller codebases are a nice change of pace!

test/integration/record_lookup_test.rb
```diff
@@ -49,6 +58,7 @@ class RecordLookupTest < ActionDispatch::IntegrationTest
 
       assert has_text? 'Search results'
       assert has_text? 'Susan Everyteen'
```
colinxfleming: You are exactly right and this is a GREAT flag. Thank you!

app/models/patient.rb
colinxfleming: A perk of using `enum` here is that rails refuses to let this ever be null, so the validation is more for clarity's sake than anything, and as a consequence we don't need to bulk update any records. (Way to go us!) Here's counts from the prod db verifying counts:

```
irb(main):001:0> Patient.count
=> 670
irb(main):002:0> Patient.where(_line: :DC).count
=> 670
irb(main):003:0> Patient.where(_line: :MD).count
=> 0
irb(main):004:0> Patient.where(_line: :VA).count
=> 0
irb(main):005:0> Patient.where(_line: nil).count
=> 0
```

app/models/user.rb
colinxfleming: I don't think a little extra clarity would hurt!

app/views/dashboards/_search_form.html.erb
colinxfleming: Just going to strip this, since it's not super complex and easy enough to redo from scratch. More confusing than helpful to keep it in. Great flag.

app/controllers/dashboards_controller.rb
colinxfleming: don't comment up your code at 2am, kids.

app/helpers/lines_helper.rb
colinxfleming: 💯

test/helpers/lines_helper_test.rb
colinxfleming: This is a great flag. Thank you for pointing this out!

colinxfleming:
thanks much @eheintzelman for confirming my assumptions and @lomky for the clutch PR review! The only frontend change remains the one screenshotted, breaking out the intake form a little bit and letting you assign line on that. team UX pleae flag if we should revert that. 

I'm a LITTLE concerned about people getting lost in the shuffle but I trust in our CM corps' skill. Still, I'm going to make a placeholder issue to catch up about the line system in January, just to be sure we aren't losing people in the cracks. 

I'm going to add one more test (integration for creating patients in two different lines and confirming they don't cross over) to satisfy my own neuroses and then this is good to merge.

lwaldsc:
Assumptions are correct, Colin.

colinxfleming:
Final integration test added, merging after this passes CI.

colinxfleming:
These tests are failing for inconsistent/fake reasons so I'm going to merge.

Issue #747, opened 2016-11-24
MULTIPLEDGE: Frontend changes to accept multiple pledges from other funds


colinxfleming:
Since the MD and VA lines require a lot of collaborative work with other funds, we need to have some clean notion of that represented easily. The scope of this issue is rigging up those changes in the patient edit view. 

![img_0463](https://cloud.githubusercontent.com/assets/6223901/19990799/e47216b6-a206-11e6-8080-3fdb1788435f.JPG)

Which follows the following list of changes:
1) move 'Resolved w/out..." to the right of Clinic details
2) Move 'Cost details' below
3) Abortion cost on left column
4) Right column: PT contribution, then DCAF, then NAF, then "+ Add other". (The first 3 should be from the Pregnancy object; the others will be pledge objects).
5) Add other chunk to include "Other", a dropdown, and money input box
6) Dropdown values = Baltimore Abortion Fund, Richmond Reproductive Freedom Project, Blue Ridge Abortion Assistance Fund, Tiller Fund (NNAF), Carolina Abortion Fund, Women's Medical Fund (Philadelphia), NYAAF (New York), Other
7) User is able to "add other" repeatedly

These also require some model changes I'm personally working on in a separate thread. For now, I think the really important things are making the stylistic changes and getting the 'Add Other' to work. 

Stems from #746

itsWill:
I'll take a stab at this one

colinxfleming:
Good to see you again, @itsWill !

We're in our final push and are flipping a switch on Sunday 12/4, can you be sure to make a pull request by Thursday? Something that's a work in progress is totally fine -- just want to make sure we don't lose any of your hard work! 

Thanks my dude!

Issue #746, opened 2016-11-24
Implement pledges from other funds (Multipledge)


colinxfleming:
Per @lwaldsc , the DC line doesn't see that much out-of-DCAF action, but the MD and VA lines often involve collaboration with other funds, so it's important that we implement the ability to represent N other pledges in a case, instead of how we're doing this now (which is pretty much just NAF and us). So it's important to have it implemented for showtime. 

This probably entails the following:

* implementing the pledge objects to represent contributions toward a case from other funds
* altering the view to shake out as follows:

![img_0463](https://cloud.githubusercontent.com/assets/6223901/19990799/e47216b6-a206-11e6-8080-3fdb1788435f.JPG)

Which follows the following list of changes:
1) move 'Resolved w/out..." to the right of Clinic details
2) Move 'Cost details' below
3) Abortion cost on left column
4) Right column: PT cont, then DCAF, then NAF, then "+ Add other"
5) Add other chunk to include "Other", a dropdown, and money input box
6) Dropdown values = Baltimore Abortion Fund, Richmond Reproductive Freedom Project, Blue Ridge Abortion Assistance Fund, Tiller Fund (NNAF), Carolina Abortion Fund, Women's Medical Fund (Philadelphia), NYAAF (New York), Other
7) User is able to "add other" repeatedly

Probably the two steps here are limited to just the backend and frontend changes (and also some test changes, of course). 

I'm going to personally rack the model changes since I have no idea how to explain the business logic. The frontend changes are probably just 'jquery comfort' and don't necessarily require any specialized knowledge.

Stems from #632

colinxfleming:
This is fully resolved with #788

PR #755, opened 2016-11-27
Multipledge models

colinxfleming:

One of the remaining confusing things about this app is that we (I) aren't doing a good job meaningfully distinguishing a DCAF pledge from an NNAF pledge from a BRAAF pledge from a BAF pledge. The former two are common scenarios, the latter are less common. To simplify this distinction, this renames the `Pledge` object to `ExternalPledge`, and does some other assorted prep for stringing it into the view. 

* Pledge is `ExternalPledge` now
* Adds `ExternalPledgesController#destroy`
* Edits some of the `ExternalPledge` attributes to significantly simplify the object model.

TODO: 
- [x] test new controller destroy method
- [x] test external_pledge default scope
- [x] fix external_pledges controller strong params
- [x] enforce uniqueness of pledge -- no more than one active pledge from a given source at a time

* Affects #746 , #747


colinxfleming:
🎉

PR #756, opened 2016-11-27
[WIP] change the view to accept multiple pledges #747

itsWill:

This changes the `_abortion_info.html.erb` partial to look like the changes recommended in #747

changes the view to look like the one #747 by reposissioning the input fields and adding a button to add other pledge sources

screen shot 2016-11-27 at 2 09 02 am
 
screen shot 2016-11-27 at 2 09 23 am

clicking on the add other button after selecting a field will add another input field with the name corresponding to the selection 

screen shot 2016-11-27 at 2 09 54 am

Issues:
 * The user can continuously add input fields even the same ones over and over. There are no checks to see which input fields have been generated thus far
*  The html looks correct but it needs to be properly wired to the back end so that the values are being routed to the correct controllers, methods, params etc..
* Need some tests
* Not entirely sure what the best practices for client side scripting are a good code review is needed here

I'm not sure this is the exact behaviour requested but I think it be useful to be able to clarify on slack.  So I kindly request an invite to join the slack channel my email is guilhermerpmansur@gmail.com 🙃 

#747

### DIFF
--- a/app/assets/javascripts/patients.coffee
+++ b/app/assets/javascripts/patients.coffee
@@ -9,3 +9,20 @@ $(document).on "click", "#toggle-call-log", ->
 
 $(document).on "change", ".edit_patient", ->
   $(this).submit()
+
+$(document).on "click", ".add-button", ->
+  value = $(".pledges-select").val()
+  if value == ""
+    alert("You have to select a value from the dropdown menu!")
+  else
+    html = """
+          <div class="form-group">
+            <label class="control-label"> #{value} </label>
+            <div class="input-group">
+              <span class="input-group-addon">$</span>
+              <input type="text" autocomplete="off" class="form-control">
+            </div>
+          </div>
+          """
+    $("#pledge-contributions").append(html)
+
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -56,4 +56,11 @@ module PatientsHelper
   def disable_continue?(patient)
     patient.pregnancy.pledge_info_present? ? 'disabled' : ''
   end
+
+  def pledges_options
+    [nil,"Baltimore Abortion Fund", "Richmond Reproductive Freedom Project",
+      "Blue Ridge Abortion Assistance Fund", "Tiller Fund (NNAF)",
+      "Carolina Abortion Fund", "Women's Medical Fund (Philadelphia)",
+      "NYAAF (New York)", "Other"]
+  end
 end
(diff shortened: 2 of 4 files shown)


colinxfleming:
My dude! Thank you! The Code for DC slack has an autoapproval if you head to their website -- go to http://codefordc.org/ and control-f 'Join' and that'll get you going in the right direction. 

I'll review this code in a sec.

app/views/patients/_abortion_information.html.erb
```diff
@@ -6,13 +6,19 @@
...
+      </div>
+
       <div class="row">
         <div class="col-sm-6">
           <div class="info-form-left">
             <%#= f.fields_for (patient.clinic || patient.build_clinic) do |cl| %>
-              <h3>Clinic details</h3>
                 <%= f.select :clinic_name, options_for_select(clinic_options, patient.clinic_name) %>
-                <%#= cl.select :name, options_for_select(clinic_options, patient.clinic.name), label: 'Clinic name', autocmomplete: 'off' %>
+                <%#= cl.select :name, options_for_select(clinic_options, patient.clinic.name), label: 'Clinic name', autocomplete: 'off' %>
```
colinxfleming: haha thank you for fixing this!

app/assets/javascripts/patients.coffee
```diff
@@ -9,3 +9,20 @@ $(document).on "click", "#toggle-call-log", ->
...
 $(document).on "change", ".edit_patient", ->
   $(this).submit()
+
+$(document).on "click", ".add-button", ->
+  value = $(".pledges-select").val()
+  if value == ""
+    alert("You have to select a value from the dropdown menu!")
+  else
+    html = """
+          <div class="form-group">
```
colinxfleming: Do you think there's a way we can turn this into an `html.erb` partial we can append? I think I'd prefer that to something hidden in the js assets, but I'm not married to it -- what are your thoughts?

app/helpers/patients_helper.rb
```diff
@@ -56,4 +56,11 @@ def clinic_options
   def disable_continue?(patient)
     patient.pregnancy.pledge_info_present? ? 'disabled' : ''
   end
+
+  def pledges_options
```
colinxfleming: 💯

colinxfleming:
@itsWill just merged #755 , so the model `ExternalPledge` is in place and ready for wrenching in here!

Issue #758, opened 2016-11-29
Alternate contact not populating in call pop up


lwaldsc:
Using the system tonight, alternate contacts were not populating in the call screen pop up. These were contacts with no alternate number.

colinxfleming:
Code here is in `calls_helper` and `views/calls/_new_call.html.erb` -- probably this is just a matter of a helper only rendering stuff if alt name and alt number are set. Sounds like the desired behavior is for presence of either one to make alt info show up in the call partial.

colinxfleming:
In bug report-ese, here's what this comes out to:
* Given that I have a patient who has a secondary contact name but not a secondary contact number,
* I expect that when I open a new call modal, I will see the secondary contact name;
* Instead, I don't see the secondary contact name.

Pretty small fix but Monday will be the first exposure most of our users have to this system, so if we can get this resolved by Sunday afternoon it would be clutch.

@valeriecodes if you've got capacity can you take a look at this? @lwaldsc can you keep an eye on this thread and help @valeriecodes out in case she needs help whipping up an example case?

lwaldsc:
Gotcha. Sounds good!

valeriecodes:
👍  Thanks @colinxfleming and @lwaldsc

PR #760, opened 2016-12-02
Render other contact's name on call modal even if no number

valeriecodes:
Resolves: https://github.com/colinxfleming/dcaf_case_management/issues/758

Here are screenshots of the main cases:
No secondary contact (unchanged)
screen shot 2016-12-01 at 9 21 51 pm

Secondary contact name but no number
screen shot 2016-12-01 at 9 25 35 pm

Secondary contact name and number
screen shot 2016-12-01 at 9 25 19 pm

Secondary contact number but no name
screen shot 2016-12-01 at 9 36 57 pm


colinxfleming:
@lwaldsc as the reporter of the bug in #758 can you approve?

lwaldsc:
The No Name but Other Contact Number should also display the secondary
number.

Other than that, the screenshots look fantastic!! Thanks, @valeriecodes

valeriecodes:
Thanks @lwaldsc . I updated the screenshots. Look good?

colinxfleming:
I think if there are remaining parts of this they're related to copy rather than code, so I'm going to hit go. Thanks again for digging in here @valeriecodes !!

PR #764, opened 2016-12-03
Add external pledge roll and form

### LINKED ISSUE #747: MULTIPLEDGE: Frontend changes to accept multiple pledges from other funds

colinxfleming:

Jumps off @itsWill 's outstanding work to railsify external pledges -- rolls them up via a partial, etc. (Seriously, thank you. Going to bed at a decent hour tonight because of your prep work here.)

IMPORTANT NOTE: The hacky route here I took may be biting us -- the abortion information partial seems to be jacked up as a result of this. Ajax submits to the pledge create route are choking because it thinks they're html for who even knows what reason, and I think it's making the update patient info integration test get confused too. 

As a stopgap, I have it sending an html request, which unfortunately reloads the page - e.g. it behaves more like patient create than it does notes create, which is what we want. I'll hack on this some over the weekend time permitting. 

* Adds partials for rendering ext pledge edit and new
* Hooks them into the pledges controller, our long 
* Fixes up tests

Screenshot:

screen shot 2016-12-03 at 2 24 56 am

* Fixes #747


colinxfleming:
Here goes nothin'.

mebates:
This is looking different from the design. We wanted an "+Add" action so you could add more than one.

NerdyGirl537:
From having used this as part of the data entry, you can add as many as you
need.

PR #772, opened 2016-12-07
Update Docker Image to use Ruby 2.3.3 tag

DarthHater:

I decided to update the Docker image to use the 2.3 tag instead of the 2.3.1 tag, this should keep it more current on patch updates to Ruby 2.3. In the future maybe we play with moving to 2.4, but I'd like to talk to @colinxfleming or others first to make sure that is fine.

One line change to the Dockerfile


colinxfleming:
May want to make a corresponding change in the gem file as well? I think
that's where we set the Ruby version in the app itself. Either way I think
this is a good idea for sure @darthhater

DarthHater:
@colinxfleming okie dokie, updated Gemfile, blah blah blah, and I repinned the Docker image to 2.3.3 since both should be kept in sync. rails s works for me :)

colinxfleming:
This looks good to me. @DarthHater if it's cool with you I'd like to merge toward the back end of December -- we have another hacknight and I'm planning on doing a bunch of code quality / readability edits while I'm in Michigan -- would love to roll this out as part of that. Thanks a ton for keeping on this my dude!

DarthHater:
@colinxfleming good by me, merge whenever, I'll try and keep an eye on stuff

DarthHater:
@colinxfleming bit past December (no rush). Only question is if I should look at Ruby 2.4.0 instead of 2.3.3, that might get us further ahead of the curve.

colinxfleming:
Let's stick with 2.3.3 for now, and wait until `ruby-audit` (which we run as part of CI) starts failing builds before we upgrade again. Keeping sharp is nice but we're well above the rails minimum for 5. 

I'll make an associated readme bump to tell people to install the proper version.

PR #776, opened 2016-12-08
Add "Other medical issue" #631

### LINKED ISSUE #631: Add "Other medical issue"
Need to add "Other Health issues" checkbox to the special circumstances

KatSDC:

I updated "Fetal anomaly" to "Fetal diagnosis" and added check boxes for "Other medical issue" and "Undocumented".

* Updated the special circumstances check boxes per user feedback

image

* Fixes #631

### DIFF
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -79,12 +79,14 @@
           <p style="font-weight:bold">Special circumstances</p>
           <%= f.form_group :special_circumstances do %>
             <div class="col-sm-6">
-              <%= f.check_box :special_circumstances, { label: 'Fetal anomaly', name: 'patient[special_circumstances][]',
-                              checked: patient.special_circumstances.include?('Fetal anomaly'), namespace: 'Fetal' }, 'Fetal anomaly', '' %>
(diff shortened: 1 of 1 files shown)


KatSDC:
Kevin said that if we need to update previous entries for "Fetal anomaly" we will have to write a script to run through and update those entries. Is this something we want to do? I'd be happy to tackle that as well if the answer is yes. I'd just need to discuss with you what the production environment is running on.

colinxfleming:
@KatSDC yeah, @Kevin-Wei is right, since we're storing this as an array of strings we'll want to bulk update it. If you want to dig into that I'd appreciate it -- you can probably whip something up using the `rails console` in a dev environment, if that makes sense -- but we can backlog it otherwise. 

Code looks good to me, I'll plan on having this deployed next Sunday during our regularly scheduled release window.

colinxfleming:
ps thank you!! This is something a couple of CMs have asked for recently, esp as we've moved onto all three lines, so this is very timely.

colinxfleming:
@KatSDC nice work up in here -- love how clean this turned out! I've spun out the task to update these entries in our prod db to #786.

PR #788, opened 2016-12-24
Initial work to get AJAX working on external pledges

### LINKED ISSUE #778: Switch External Pledges to use ajax
Due to time crunch, External Pledge creation (in edit pregnancy) was rolled out synchronously, and when you make a new one it reloads the page. I'd like to cut that functionality and instead have the new external pledge route run a `create.js.erb` that reloads the external pledges for that patient. Nonessential but nice to have and will make for a more seamless experience.

DarthHater:

This moves the initial work to use AJAX so the page doesn't fully reload. 

* Controller mod (borrowed approach from notes)
* Made a partial out of the external pledges list
* Updating create.js.erb to reload partials
* Moved some logic into the abortion info layout
* Updated tests

* Fixes #778

### DIFF
--- a/app/views/external_pledges/_external_pledges_table.html.erb
+++ b/app/views/external_pledges/_external_pledges_table.html.erb
@@ -1,5 +1,4 @@
-<div class="col-sm-6">
-  <div id="existing-external-pledges">
+<div id="existing-external-pledges">
     <%# slight hack - individual edit forms for each pledge %>
     <% external_pledges.each do |external_pledge| %>
       <%= render 'external_pledges/edit', patient: patient, external_pledge: external_pledge %>
--- a/app/views/external_pledges/_external_pledges_table.html.erb
+++ b/app/views/external_pledges/_external_pledges_table.html.erb
@@ -1,5 +1,4 @@
 <div id="existing-external-pledges">
-    <%# slight hack - individual edit forms for each pledge %>
     <% external_pledges.each do |external_pledge| %>
       <%= render 'external_pledges/edit', patient: patient, external_pledge: external_pledge %>
     <% end %>
--- a/app/views/external_pledges/_external_pledges_table.html.erb
+++ b/app/views/external_pledges/_external_pledges_table.html.erb
@@ -1,0 +1,7 @@
+<div class="col-sm-6">
+  <div id="existing-external-pledges">
+    <%# slight hack - individual edit forms for each pledge %>
+    <% external_pledges.each do |external_pledge| %>
+      <%= render 'external_pledges/edit', patient: patient, external_pledge: external_pledge %>
+    <% end %>
+</div>
--- a/app/views/external_pledges/create.js.erb
+++ b/app/views/external_pledges/create.js.erb
@@ -1,2 +1,2 @@
 $('#existing-external-pledges').empty().append('<%= escape_javascript(render partial: "external_pledges/external_pledges_table", locals: { patient: @patient, external_pledges: @patient.external_pledges.order_by('created_at desc') } ) %>');
-$('#new-external-pledge').empty().append('<%= escape_javascript(render partial: "external_pledges/new", locals: { patient: @patient, new_external_pledge: @new_external_pledge } ) %>');
+$('#new-external-pledge').empty().append('<%= escape_javascript(render partial: "external_pledges/new", locals: { patient: @patient, new_external_pledge: @patient.external_pledges.new } ) %>');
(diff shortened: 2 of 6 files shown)


app/views/patients/_abortion_information.html.erb
```diff
@@ -66,16 +66,14 @@
...
-          <%# slight hack - individual edit forms for each pledge %>
-          <% patient.external_pledges.each do |external_pledge| %>
-            <%= render 'external_pledges/edit', patient: patient, external_pledge: external_pledge %>
-          <% end %>
-        </div>
+        <%= render partial: 'external_pledges/external_pledges_table', locals: { patient: patient, external_pledges: patient.external_pledges.order_by('created_at desc') } %>
 
         <div id="new-external-pledge">
-          <%= render 'external_pledges/new', patient: patient, external_pledge: new_external_pledge %>
+          <%= bootstrap_form_for [patient, new_external_pledge], remote: true do |f| %>
```
colinxfleming: Thoughts about keeping this in the `external_pledges/_new` partial? I think it might be slightly more findable there

app/views/external_pledges/_external_pledges_table.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="col-sm-6">
```
colinxfleming: Moving this to its own partial is dead on 💯

app/views/external_pledges/create.js.erb
```diff
@@ -1 +1 @@
-// rerender #existing-external-pledges
\ No newline at end of file
+$('#external_pledge_amount').val('');
```
colinxfleming: I expect that you'll want to massage this just a bit. When you hit submit, on a successful thing, you're looking to do the following:

* clear the form (in this case, both the amount and the source -- probably you can just copy this line and change the id and be good to go)
* re-rack the external pledges table partial so that the new external pledge shows up.

In notes, which you've done a great job modeling this after, we do it like this: 

```
$('#notes_log').empty().append('<%= escape_javascript(render partial: "notes/notes_table", locals: { notes: @notes } ) %>');
```

That is -- we empty the div with existing external pledges in it, and then replace its contents with the partial, which is now accounting for the new entry.

app/views/patients/_abortion_information.html.erb
DarthHater: Should be possible, I'll give it a whirl.

app/views/external_pledges/create.js.erb
DarthHater: For some reason I can't get the external_pledge_amount to clear, no idea why. If you got an extra eye, go for it!

DarthHater: Also, maybe I should just rerender the partial sorta like how we do with the existing pledges? Figure that might actually be better since the dropdown list would need to lose a value?

colinxfleming:
That's a really excellent point about the drop down list needing to take
out options. Sounds good on re racking that partial, though let me know if
it winds up being a pain.

app/views/external_pledges/create.js.erb
```diff
@@ -1 +1,2 @@
-// rerender #existing-external-pledges
\ No newline at end of file
+$('#existing-external-pledges').empty().append('<%= escape_javascript(render partial: "external_pledges/external_pledges_table", locals: { patient: @patient, external_pledges: @patient.external_pledges.order_by('created_at desc') } ) %>');
+$('#new-external-pledge').empty().append('<%= escape_javascript(render partial: "external_pledges/new", locals: { patient: @patient, new_external_pledge: @new_external_pledge } ) %>');
```
colinxfleming: the new external pledge line here is what's causing the choke, it seems

colinxfleming: Probably what's going on here is that this instance variable is now a persisted object, so you'll want to flush this out with a new object. I recommend popping out `new_external_pledge: @new_external_pledge` with just simply `new_external_pledge: @patient.external_pledges.new`. So you'll get a line as follows:

```
$('#new-external-pledge').empty().append('<%= escape_javascript(render partial: "external_pledges/new", locals: { patient: @patient, new_external_pledge: @patient.external_pledges.new } ) %>');
```

colinxfleming: (Ah, I see -- the reason it's getting flushed out is because the `external_pledges_controller`, which doesn't have an instance variable for `@new_external_pledge`, is what's triggering the action. So rails isn't kicking over a new external pledge from the patients controller to the external pledges create controller. Makes sense, awkwardly.)

DarthHater:
@colinxfleming (or anyone else), tests adjusted and passing, current failure seems to be unrelated. I think this is ready to go unless you find some issue with it.

colinxfleming:
This looks solid to me! I'll take a final cursory pass tomorrow but I think we're good to roll this out. Thanks a ton @DarthHater and @lomky !

PR #796, opened 2017-01-02
Initial setup of Google SSO

### LINKED ISSUE #323: Investigate 2-factor auth implementations with Devise
Raised by @stvnrlly in slack -- is this a feasible thing to implement? Considerations off the top of my head:

a) cost (but only if it's really outrageously expensive, as DCAF has some money to burn on this and we're way under budget!)
b) support (we have a couple older volunteers who have a hard time with computers sometimes) 
c) implementation (devise extension? something else?)

I'm pro but would like thoughts from the dev squad team, @stvnrlly , @lwaldsc @NerdyGirl537 and at least one CM lead (Meghan?).

DarthHater:

I've begun work to start using Google SSO as a sign in option. 

* Adds the omniauth-google-oauth2 gem
* Implements controllers, model changes, etc... from the omniauth-google-oauth2 README
* Addition of two key ENV variables (manual setup needed for these, __GOOGLE_KEY__, __GOOGLE_SECRET__). More info here: https://richonrails.com/articles/google-authentication-in-ruby-on-rails (values need to be set to the values you get from Google) ) __BEST PRACTICE:__ add these to your .bash_profile or whatever so that they don't vanish into thin air
* Addition of cacert.pem to root to make up for https dealiemabob, directions here: https://richonrails.com/articles/google-authentication-in-ruby-on-rails (put into .gitignore to avoid having it in git, duh)
* Adding some tests, and some omniauth config for a mock user, so it can run on CI, etc...

Still needs:
* Ay caramba this thing looks ugly, someone should make it look way better
* Make a decision on if we want to hide the regular sign in past the dev environment, likely useful in dev
** Maybe make this controllable via an ENV variable too, in case others want to use the functionality but don't want or can't use Google SSO?
* Add description to README.md on how to set all this up for local dev
* Implement tests, a plethora of information is available here: https://github.com/omniauth/omniauth/wiki/Integration-Testing

* Fixes #323


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def google_oauth2
+      @user = User.from_omniauth(request.env["omniauth.auth"])
+
+      if @user.persisted?
+        flash[:notice] = I18n.t "devise.omniauth_callbacks.success", :kind => "Google"
+        sign_in_and_redirect @user, :event => :authentication
+      else
+        session["devise.google_data"] = request.env["omniauth.auth"]
+        redirect_to new_user_registration_url
```
colinxfleming: This is the failure condition, right? Can we change this to `redirect_to root_path`?

app/models/user.rb
```diff
@@ -77,6 +79,20 @@ def secure_password
     end
   end
 
+  def self.from_omniauth(access_token)
+    data = access_token.info
+    user = User.where(:email => data["email"]).first
+
+    # Uncomment the section below if you want users to be created if they don't exist
```
colinxfleming: Confirm that we don't want users to be created if they don't exist 💯 . Got this exactly right

app/controllers/users/omniauth_callbacks_controller.rb
DarthHater: This is copy pasta from their example. I think we can do anything, I can definitely change that.

app/models/user.rb
DarthHater: I'll remove the comment then, left it in case we wanted it.

config/routes.rb
```diff
@@ -21,9 +21,10 @@
     resources :lines, only: [:new, :create]
   end
   root :to => redirect('/users/sign_in')
-  devise_for :users, skip: [:registrations]
+  devise_for :users, :controllers => { :omniauth_callbacks => "users/omniauth_callbacks" }
+  # devise_for :users, skip: [:registrations]
```
colinxfleming: Are these routing changes essential or can we go back to skipping most of the registration stuff? I think we might be able to get away with `devise_for :users, controllers: { omniauth_callbacks: 'users/omniauth_callbacks' }, skip: [:registrations]`

config/routes.rb
```diff
@@ -21,9 +21,10 @@
...
   as :user do
-    get '/users/edit' => 'devise/registrations#edit', as: 'edit_user_registration'
-    put '/users' => 'devise/registrations#update', as: 'registration'
+    # get '/users/edit' => 'devise/registrations#edit', as: 'edit_user_registration'
```
colinxfleming: Would like to restore these too. We only need these two routes from registration.

test/omniauth_helper.rb
```diff
@@ -0,0 +1,18 @@
+# in test/helpers/omniauth_helper.rb
+
+OmniAuth.config.mock_auth[:google_oauth2] = {
```
colinxfleming: 🇺🇸

config/initializers/devise.rb
```diff
@@ -235,7 +235,7 @@
   # ==> OmniAuth
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
-  # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+  config.omniauth :google_oauth2, ENV['GOOGLE_KEY'], ENV['GOOGLE_SECRET'], {client_options: {ssl: {ca_file: Rails.root.join("cacert.pem").to_s}}}
```
colinxfleming: I *think* we can get rid of the client options hash here -- works on my machine without. How about you?

config/routes.rb
DarthHater: There was some error with this stuff not commented out. I think it can get added back in just wanted to make sure the landscape was settled before I did.

config/initializers/devise.rb
DarthHater: I think it will work if you are doing http, if you switch to https it might not? I don't know. Probably can get removed.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
+  def google_oauth2
+      @user = User.from_omniauth(request.env["omniauth.auth"])
+
+      if @user.persisted?
```
colinxfleming: Adding a note here: When logging in as a user not in the db this chokes. I suspect the answer is altering `User#from_omniauth` a bit so that it looks more like a `find_by email` call.

DarthHater: Yeah I was just looking at that. Probably what needs to be done.

config/initializers/devise.rb
colinxfleming: I'm going to push this up to staging (which uses https). Will test both ways (starting without)

colinxfleming:
Deploying staging yields the following error: 

screen shot 2017-01-07 at 3 22 05 pm

This is probably an argument for per instance API keys, no? e.g. different sets of keys for staging, dcaf-prod, hypothetical-other-fund-prod, etc.

DarthHater:
It's just a URL that needs added to the google app. I'll do that (it'll
need to be added for prod too)

DarthHater:
@colinxfleming added that URL to the app, should be good

app/controllers/users/omniauth_callbacks_controller.rb
colinxfleming: Here's a two point plan for fixing this: 

To prevent bad sign ins from erroring out:
```
# in this controller
              with: -> { redirect_to root_path }

# in user model
  def self.from_omniauth(access_token)
    data = access_token.info
    user = User.find_by email: data['email'] # this replaces the `where`. A little cleaner since we're unique on email

    user
  end

```

test/controllers/omniauth_callbacks_controller_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+require 'omniauth_helper'
+
+class OmniauthCallbacksControllerTest < ActionController::TestCase
+  before do 
+    request.env["devise.mapping"] = Devise.mappings[:user] 
+    request.env["omniauth.auth"] = OmniAuth.config.mock_auth[:google_oauth2] 
+  end
+
+  describe 'access top page' do
```
colinxfleming: We should throw this in an integration test rather than a controller test. The code is pretty much spot on, just the wrong spot for it! Will work with you on this.

colinxfleming:
Sure enough, works like a charm! Thanks much!

Here's a proposed rollout for this: At least for phase 1, let's keep password entry available and start migrating individual users over. Phase 2 can be turning off password based entry entirely, figuring out how to adjust our tests/onboarding/training, etc. I think that does a decent job of splitting this up.

Scenarios to test:
* Valid login with google, user exists in db
* Valid login with google, user does not exist in db

Units to test:
* User#from_omniauth
* UsersController#google_oauth2 (success and failure conditions)

I think that's it though, right?

PR #801, opened 2017-01-07
Quick ditty to show the environment

### LINKED ISSUE #792: Indicate environment in top navbar
During data entry, we had a near miss where someone almost punched some data into the staging environment. 

* **What are we trying to do?** Clearly indicate in the navbar what the current environment / heroku app is.

* **What feature or behavior is this required for?** Clearly marking what fund or environment 

* **How should we solve this issue?** Replace `DC Abortion Fund` in `app/views/layouts/_navigation.html.erb` with `ENV['FUND'] || 'DEVELOPMENT'`. (`ENV['FUND']` is a variable we have set up.)

DarthHater:

* Adds a ternary if to the _navigation.html.erb to output the rails env value if it isn't production

* Fixes #792

### DIFF
--- a/app/views/layouts/_navigation.html.erb
+++ b/app/views/layouts/_navigation.html.erb
@@ -7,7 +7,7 @@
         <span class="icon-bar"></span>
         <span class="icon-bar"></span>
       </button>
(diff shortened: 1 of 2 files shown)


colinxfleming:
Works for me! Thanks @DarthHater !

PR #803, opened 2017-01-10
Add validations to Add Patient Form

### LINKED ISSUE #784: Perform Validation on the Add Patient From Prior to Submit.
Perform front-end validation on the new patient form. It can be frustrating to fill out the form and lose all that information because an extra number has snuck into the phone number. Some HTML5 form elements already come with validation.

tingaloo:

Added simple form validators to the add new patient form. 

Phone number is required and must be 10 numbers
Name is required

screen shot 2017-01-09 at 6 48 51 pm
screen shot 2017-01-09 at 6 49 02 pm

* Fixes #784

### DIFF
--- a/app/views/patients/_new_patient.html.erb
+++ b/app/views/patients/_new_patient.html.erb
@@ -8,10 +8,10 @@
   <%= bootstrap_form_for patient do |f| %>
     <div class="row">
       <div class="col-sm-2">
-        <%= f.text_field :primary_phone, label: 'Phone', value: phone %>
+        <%= f.text_field :primary_phone, label: 'Phone', value: phone, required: true, pattern: '.{10}' %>
       </div>
       <div class="col-md-4">
-        <%= f.text_field :name, label: 'Name', value: name %>
+        <%= f.text_field :name, label: 'Name', value: name, required: true %>
       </div>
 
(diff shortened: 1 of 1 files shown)


app/views/patients/_new_patient.html.erb
```diff
@@ -8,10 +8,10 @@
   <%= bootstrap_form_for patient do |f| %>
     <div class="row">
       <div class="col-sm-2">
-        <%= f.text_field :primary_phone, label: 'Phone', value: phone %>
+        <%= f.text_field :primary_phone, label: 'Phone', value: phone, required: true, pattern: '[0-9]{10}' %>
```
colinxfleming: I had no idea you could use a regexp pattern but that's AWESOME.

Our CMs tend to enter phones in one of three ways:

* 1112223333
* 111-222-3333
* 111.222.3333

We have a prevalidation hook in the model that strips out any non-numbers, so entry in this way is totally cool. @tingaloo could you do me a solid and adjust this regexp so that it allows those three cases?

tingaloo: Yup! I didn't know html5 validators existed until recently. Anyways all three cases should pass.

colinxfleming:
Thanks a ton for making that change so quick, @tingaloo ! Merging and rolling out on Sunday.

PR #802, opened 2017-01-08
Tests and linting for SSO

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

Jumps off the incredible setup work by @DarthHater in #796 to set up google auth. Adjusts the seedfile a little, lints a little, and adds model and integration tests.

@DarthHater for your approval!

It relates to the following issue #s: 
* Fixes #323

PR #804, opened 2017-01-10
Quick tweaks to namespace the ENV variables

DarthHater:

Small tweaks to change the ENV variables from #796 and #802 to be namespaced

* Updates env variables in devise.rb
* Updates README.md to new variable names


colinxfleming:
Looks good to me! I've made these changes to the keys in staging and prod.

PR #807, opened 2017-01-10
Small change from ENV to ARG since these values are all build time

DarthHater:

Simply stated, the ENV variables in the Dockerfile can benefit from being build time ARG instead. No reason to persist them as the application does not need them post build. Also allows someone to override them (if need be) via the docker build command. 

* Switches ENV to ARG in Dockerfile, single lines them (no multiline for ARG it appears)


colinxfleming:
Lookin' good! Merging.

Issue #771, opened 2016-12-07
Write tests for data entry functionality


colinxfleming:
This is a good opportunity to learn some minitest, the ruby library we use for writing tests!

* **What are we trying to do?** QA our data entry form, `/data_entry`!

* **What feature or behavior is this required for?** We wrote a data entry function in a hurry for deployment; we want to have it tested so that we don't accidentally break it in a future small release. 

* **How should we solve this issue?**

We should write the following: 
* A controller test in patients that confirms that the methods `data_entry` works, and that `data_entry_create` works with both the success and failure conditions
* A new Capybara integration test that i) fills out the form successfully and confirms that the fields show up on the patient edit page it dumps you on; and ii) fills out the form unsuccessfully and confirms that any errors are on the page.

vincecampanale:
I'd be happy to take this one!

colinxfleming:
hey @vincecampanale , that would rule! Let me know if I can assist with anything as you dig in.

vincecampanale:
Hey @colinxfleming happy to help! I'm running into some setup issues. The command `rake db:seed` is returning this error: 

```
$ rake db:seed
rake aborted!
NoMethodError: undefined method `ruby_dirs=' for #<Quality::Config:0x007ff28311d050>
Did you mean?  ruby_files
/Users/vincecampanale/dev-stuff/open-source/dcaf_case_management/lib/tasks/quality.rake:8:in `block in <top (required)>'
/Users/vincecampanale/.rvm/gems/ruby-2.3.1/gems/quality-27.3.0/lib/quality/rake/task.rb:57:in `initialize'
/Users/vincecampanale/dev-stuff/open-source/dcaf_case_management/lib/tasks/quality.rake:3:in `new'
/Users/vincecampanale/dev-stuff/open-source/dcaf_case_management/lib/tasks/quality.rake:3:in `<top (required)>'
/Users/vincecampanale/.rvm/gems/ruby-2.3.1/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `load'
/Users/vincecampanale/.rvm/gems/ruby-2.3.1/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:268:in `block in load'
/Users/vincecampanale/.rvm/gems/ruby-2.3.1/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb:240:in `load_dependency'
[...]

PR #816, opened 2017-01-26
Data entry minitests

vincecampanale:

* Add new test in patients_controller_test.rb for the data_entry method in the patients controller
* Add new test in patients_controller_test.rb for the data_entry_create method in the patients controller
* Includes commenting out quality.rake due to deprecated gems

* Drafting tests for #771


test/controllers/patients_controller_test.rb
```diff
@@ -123,4 +123,30 @@ class PatientsControllerTest < ActionController::TestCase
       assert_redirected_to root_path
     end
   end
+
+  #confirm get :data_entry returns a success code
+  describe `data_entry method` do
```
colinxfleming: kind of a nit, but to stick with the style can we switch this out to single quotes instead of backticks?

test/controllers/patients_controller_test.rb
```diff
@@ -123,4 +123,30 @@ class PatientsControllerTest < ActionController::TestCase
...
+
+  # confirm sending a 'post' with a payload results in a new patient
+  describe `data_entry_create method` do
+    before do
+      @test_patient = attributes_for :patient, name: 'Test Patient'
+      @test_patient[:pregnancy] = attributes_for :pregnancy
+    end
+
+    it 'should create and save a new patient' do
+      assert_difference 'Patient.count', 1 do
```
colinxfleming: this is passing with flying colors, nicely done!

test/controllers/patients_controller_test.rb
```diff
@@ -123,4 +123,30 @@ class PatientsControllerTest < ActionController::TestCase
       assert_redirected_to root_path
     end
   end
+
+  #confirm get :data_entry returns a success code
+  describe `data_entry method` do
+    before do
```
colinxfleming: On this first test, we aren't doing anything with data, so you can cut this `before` block. You should keep it for the create method one though!

test/controllers/patients_controller_test.rb
```diff
@@ -123,4 +123,30 @@ class PatientsControllerTest < ActionController::TestCase
...
+      @new_patient = attributes_for :patient, name: 'Test Patient'
+      @new_patient[:pregnancy] = attributes_for :pregnancy
+    end
+
+    it 'should respond success on completion' do
+      assert_response do get :data_entry, :success
```
colinxfleming: Ah, I see what's up. You're super close -- let's take a look at the documentation for `assert_response` here -- http://guides.rubyonrails.org/testing.html#helpers-available-for-integration-tests -- check out section 5.2.1. The nut of it is here: 

```
  get "/articles/new"
  assert_response :success
```

the format here is slightly different but pretty close! Basically, you want to switch this into two statements; one that does a get, and then one that asserts it has a successful response.

vincecampanale:
@colinxfleming  Thanks for the feedback and help! Made a couple of the suggested changes - let me know what else needs to be done.

Issue #699, opened 2016-10-24
Implement Rate Throttling 


mccabe615:
The following package can implement various forms of rate throttling and banning on various functions. This is something worth implementing on functions such as login and forgot password.

https://github.com/kickstarter/rack-attack

colinxfleming:
We have this established in a limited way for login attempts thru devise, but I agree that there's a need to implement a general policy around it and that this is a good idea. 

From a skim of the docs I suspect that this is a pretty straightforward install -- marking hacktoberfest in case someone wants to take a stab. My inclination is to do this formally for login, forgot password, user create, and implement some of the best practices here: https://www.kickstarter.com/backing-and-hacking/rack-attack-protection-from-abusive-clients

Sound good to you @mccabe615 ?

mccabe615:
Awesome. I'd like to take a crack at this but not sure if I have time immediately.

colinxfleming:
This is a priority and I'd be stoked if someone knocked it out!

lomky:
Taking a stab at this tonight.

lomky:
@colinxfleming 

Here is what I've come up with from your input & the guides:

 - a standard throttle of `limit: 10, period: 10`
 - more severe throttle for these, maybe tighter restriction to `limit: 2, period: 2`?
    - `login`, `forgot password`, `user create`
  - maybe also restrict tries per email for these?
    - `login`, `forgot password` 

let me know your thoughts on the particular numbers and the email thought. 

Digging into the initial installation process now, hoping to get that PR in my end of hacknight.

lomky:
Oh, I just noticed they offer a fully written basic configuration. I'm going to hack at that a bit before trying anything custom.

lomky:
Initial functionality good-to-go. I'd like the weekend to get some tests in, but if you don't here from me by Monday feel free to merge.
Or merge if you make the call that it's good to go, but file me another ticket to write the tests please :D

lomky:
[Testing rack::attack](http://blog.hayleyanderson.us/2015/06/05/using-and-testing-rack-attack-to-improve-the-security-of-your-rails-app/)

PR #820, opened 2017-01-27
Ready For Merge: rack::attack throttling

lomky:
Implemented exact Example Configuration suggested by Kickstarter. 

Strongly suspect this isn't quite right yet (didn't check URLs for login, etc).

* Bumps #699


colinxfleming:
yusssssss

@lomky should I plan to branch off this and write tests, or do you have it handled?

DarthHater:
This is awesome, as per usual. In checking the rack-attack docs, I think you still need to set https://github.com/lomky/dcaf_case_management/blob/9045d1bd18d64cdf2b11a2b6745f45dbc733c5a9/config/environments/test.rb#L21 to true. Not sure what that will change but it is worth verifying it doesn't blow up your tests, etc...

lomky:
To document for posterity: Rack-Attack *seems* to suggest we need to [enable caching](https://github.com/kickstarter/rack-attack#testing) the way @DarthHater suggests, but reading [the page itself](http://guides.rubyonrails.org/caching_with_rails.html) it seems like that's not needed. Conferred and agreed in the Slack, so this is ready to merge!

PR #821, opened 2017-01-27
made identifier searchable?

Kevin-Wei:

added identifier field
made it searchable?

I even added tests. after fumbling around with regex and aggregation for about an hour


colinxfleming:
A couple BS test failures in `audit_trail_test` but nothing significant. Thanks a ton @Kevin-Wei , this looks good -- going to take it for a test drive shortly and probably push to prod midmonth.

Thanks a ton for hooking this into search, too.

Note that when we roll this out we'll want to re-run indexes.

PR #827, opened 2017-01-31
Removing table stylings from patient notes headers; implementing MM/D…

### LINKED ISSUE #795: Improvements to notes section in patient edit view
There are a few user feedback items pertaining to the notes partial as follows:

* Would be nice to be able to email a user from their notes -- e.g. if not system, link to a mailto of their email
* Notes metadata should not be styled as a table -- we can just have them next to each other in a header. Something like `User.name on YYYY-MM-DD TIMESTAMP` should work nicely.

Stems from #593

AlexandraCasser:
-Removed table styling from patient notes headers
-Added MM/DD/YYYY format for date

-Added padding to call list area
-Replaced "Remove" with glyph icon
-Improved widths

The notes page should look like this:
localhost-3000-patients-589095435e0dd1cbbbdd33a3-edit

And the call list area should look like this:
screen shot 2017-01-31 at 3 18 09 pm

Thanks @colinxfleming and let me know if anything needs a touch up!

Closes #795

### DIFF
--- a/app/assets/stylesheets/dashboards.scss
+++ b/app/assets/stylesheets/dashboards.scss
@@ -24,3 +24,28 @@
 .bootstrap_toggle_spacing .checkbox {
   margin-top: 0px;
 }
+
+.pt-phone,.pt-LMP {
+  width: 100px;
+  white-space: nowrap;
+  padding-right: 5px;
+}
+
+.pt-recent-notes {
+  width: 150px;
+}
+
+.glyphicon-remove {
+  color: red;
+  padding: 0px 10px 0px 10px;
+}
+
+.glyphicon-plus-sign {
+  float: right;
+}
+
+.pt-name {
+  a {
+    padding-left: 10px;
+  }
+}
--- a/app/assets/stylesheets/notes.scss
+++ b/app/assets/stylesheets/notes.scss
@@ -1,0 +1,19 @@
+.name-date-time {
+	padding: 8px;
+  line-height: 1.42857143;
+  vertical-align: top;
+  border-top: 1px solid #ddd;
+  font-size: 14px;
+  color: #323a45;
+  text-transform: uppercase;
+  letter-spacing: 1.5px;
+  background: #f1f1f1;
+  display: inline-block;
+  float: left;
+  width: 100%;
+  font-weight: bold;
+}
+
+.notes-body {
+	display: -webkit-inline-box;
+}
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -12,7 +12,7 @@
         <th>Name</th>
         <th>LMP</th>
         <th>Status</th>
-        <th>Appointment date</th>
+        <th>Appt date</th>
         <th>Notes</th>
         <th></th>
         <th></th>
@@ -24,18 +24,18 @@
           <% if local_assigns[:sortable] %>
             <td class="handle"><span class="glyphicon glyphicon-move"></span></td>
           <% end %>
-          <td><%= patient.primary_phone_display %></td>
-          <td><%= link_to patient.name, edit_patient_path(patient) %></td>
-          <td><%= patient.pregnancy.last_menstrual_period_display_short %></td>
+          <td class="pt-phone"><%= patient.primary_phone_display %></td>
+          <td class="pt-name"><%= link_to patient.name, edit_patient_path(patient) %></td>
+          <td class="pt-LMP"><%= patient.pregnancy.last_menstrual_period_display_short %></td>
           <td><%= patient.status %></td>
           <td><%= patient.appointment_date %></td>
-          <td><%= patient.most_recent_note_display_text %> <%= plus_sign_glyphicon(patient.most_recent_note) %></td>
+          <td class="pt-recent-notes"><%= patient.most_recent_note_display_text %> <%= plus_sign_glyphicon(patient.most_recent_note) %></td>
 
           <% if table_type == 'completed_calls' || table_type == 'urgent_patients' %>
             <td class="blank-td"></td>
           <% else %>
             <% if current_user.patients.include?(patient) %>
-              <td><%= link_to "Remove", remove_patient_path(current_user, patient), method: :patch, data: { confirm: "Are you sure you want to remove #{patient.name} from your call list?" },  remote: true %></td>
+              <td><%= link_to "<span class='glyphicon glyphicon-remove' aria-hidden='true'></span><span class='sr-only'>Remove call</span>".html_safe, remove_patient_path(current_user, patient), method: :patch, data: { confirm: "Are you sure you want to remove #{patient.name} from your call list?" },  remote: true %></td>  
             <% else %>
               <td><%= link_to "Add", add_patient_path(current_user, patient), method: :patch, remote: true %></td>
             <% end %>
--- a/config/initializers/time.rb
+++ b/config/initializers/time.rb
@@ -1,7 +1,7 @@
 # Extends time class with convenience methods
 class Time
   def display_date
-    getlocal.strftime("%Y-%m-%d")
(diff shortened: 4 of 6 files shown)


colinxfleming:
Whoa, CodeCov changed their report format!

Thanks a ton for getting this in @AlexandraCasser ! I'm running on fumes tonight but I promise I'll get to this in the next day or two. I think there's a way we can generalize some of the CSS that'll make this a little more maintainable, and tidy up the notes part of table content a little bit.

colinxfleming:
Also I should say I'm super stoked on this! Having a more pleasant call list has been on the todo list for awhile, and I really appreciate your taking it on -- going to result in a lot of happy case managers, I think!

AlexandraCasser:
@colinxfleming Take your time getting around to it and thanks so much for having a look.  I'm open to any and all suggestions.

app/assets/stylesheets/notes.scss
```diff
@@ -0,0 +1,19 @@
...
+  text-transform: uppercase;
+  letter-spacing: 1.5px;
+  background: #f1f1f1;
+  display: inline-block;
+  float: left;
+  width: 100%;
+  font-weight: bold;
+}
+
+.notes-body {
```
colinxfleming: Not sure this is doing anything (for me, on chrome, anyway) -- what's the intent here?

app/assets/stylesheets/notes.scss
```diff
@@ -0,0 +1,19 @@
+.name-date-time {
```
colinxfleming: It looks like we're basically recreating the existing table styling on a div class rather than a table. What do you think about kind of splitting the difference here and just using existing th and td elements in the `notes_table`? Something like:

```
...
      <% unless note.new_record? %>
        <tr>
          <th><%= note.created_by.name %> <%= note.created_at.display_date %> <%= note.created_at.display_time %></th>
        </tr>
        <tr>
          <td><%= note.full_text %></td>
        </tr>
      <% end %>
...
```

As you can probably guess by the scss patterns, we aren't in that part of the stack too often! So I'm inclined to keep as many changes in the html.erb templates as possible. Your thoughts?

app/views/dashboards/_table_content.html.erb
```diff
@@ -24,18 +24,18 @@
...
           <td><%= patient.appointment_date %></td>
-          <td><%= patient.most_recent_note_display_text %> <%= plus_sign_glyphicon(patient.most_recent_note) %></td>
+          <td class="pt-recent-notes"><%= patient.most_recent_note_display_text %> <%= plus_sign_glyphicon(patient.most_recent_note) %></td>
 
           <% if table_type == 'completed_calls' || table_type == 'urgent_patients' %>
             <td class="blank-td"></td>
           <% else %>
             <% if current_user.patients.include?(patient) %>
-              <td><%= link_to "Remove", remove_patient_path(current_user, patient), method: :patch, data: { confirm: "Are you sure you want to remove #{patient.name} from your call list?" },  remote: true %></td>
+              <td><%= link_to "<span class='glyphicon glyphicon-remove' aria-hidden='true'></span><span class='sr-only'>Remove call</span>".html_safe, remove_patient_path(current_user, patient), method: :patch, data: { confirm: "Are you sure you want to remove #{patient.name} from your call list?" },  remote: true %></td>
```
colinxfleming: This looks awesome. Love how much tidier it feels compared to the word remove!

app/assets/stylesheets/dashboards.scss
```diff
@@ -24,3 +24,28 @@
 .bootstrap_toggle_spacing .checkbox {
   margin-top: 0px;
 }
+
+.pt-phone,.pt-LMP {
```
colinxfleming: Can we collapse these into one class since they're doing the same thing?

Should we also apply a css class to the appointment date and status tds? Interested in your opinion here.

app/assets/stylesheets/dashboards.scss
```diff
@@ -24,3 +24,28 @@
...
+  width: 100px;
+  white-space: nowrap;
+  padding-right: 5px;
+}
+
+.pt-recent-notes {
```
colinxfleming: The note cell feels really cramped to me. Do you think it's a good idea to just reduce the size of the partial note we're displaying instead of splitting it onto two lines? Right now we're displaying the first 40 characters of a note and then the plus sign glyphicon, but I wonder if we want to decrease that so this fits better. (It would also let us ditch the glyphicon-plus-sign float below.) Thoughts?

config/initializers/time.rb
```diff
@@ -1,7 +1,7 @@
 # Extends time class with convenience methods
 class Time
   def display_date
-    getlocal.strftime("%Y-%m-%d")
+    getlocal.strftime("%m-%d-%Y")
```
colinxfleming: Please revert this if you get a chance -- I'd like the standard date format to stay YYYY-MM-DD per iso standard. Kinda picky, I know, but I sometimes feel like my day job is 80% getting bitten in the behind by timezone and date format problems and 20% programming, haha.

app/assets/stylesheets/dashboards.scss
colinxfleming: (actually, I think that float-right solves a diff issue, so let's keep that either way. my mistake!)

config/initializers/time.rb
AlexandraCasser: No worries.  I changed it because @mebates wanted different formatting, but either way is fine by me.

app/assets/stylesheets/dashboards.scss
AlexandraCasser: I think reducing the size of the note is a very good idea.  What's that controlled by?  Have users expressed that it's useful to them?  Because if not I would consider getting rid of it if we're displaying so little text it's not helpful to them.

app/assets/stylesheets/notes.scss
AlexandraCasser: Sounds good!

app/assets/stylesheets/dashboards.scss
AlexandraCasser: I'm guessing we don't need this anymore if we're using the existing table styles, right?

colinxfleming: Let's say this -- until we have some design feedback, let's change this to a single class and keep only the padding. I think it does make it look less cramped on the left, which I think would make me feel a little less stressed out if I were CMing!

app/assets/stylesheets/dashboards.scss
colinxfleming: It's controlled by two things:

* in `app/models/Patient.rb`, around line 135, is how we control how many characters display in the note text:

```
  def most_recent_note_display_text
    note_text = most_recent_note.try(:full_text).to_s
    display_note = note_text[0..40]  #
    display_note << '...' if note_text.length > 41 # 
    display_note
  end
```

* The plus sign expander glyphicon comes from `app/helpers/notes_helper.rb` in the `plus_sign_glyphicon` method - the idea is that if a note's text is longer than `n`, it'll display the plus sign glyphicon. 

I think we could trim that back as far as we need to, since the plus sign glyphicon functionally works as an expander. How about trying to set `note_text[0..x]` to 15, `note_text.length > x` to 16, and the number in `plus_sign_glyphicon` to 16 and see how that looks?

colinxfleming:
@AlexandraCasser so sorry for the delay reviewing this again! I had a few points of feedback (which I'm happy to implement if you prefer), but after that I think we're ready to Smash That Merge Button if you're also feeling good about it! Let me know!

AlexandraCasser:
@colinxfleming Will implement this tomorrow.  Thanks!

AlexandraCasser:
@colinxfleming I think it would actually be easier if you implemented these, much as I hate to load more work on you.  Is that ok?

colinxfleming:
@AlexandraCasser copy that, will adjust to taste and implement!

PR #829, opened 2017-02-02
Initial prop up of lets-encrypt changes

### LINKED ISSUE #748: Investigate cheaper SSL solutions

DarthHater:

This is the start of changes to hopefully use lets-encrypt for FrEe SsL!!111!

* Adds gem (github reference, YIKES) for heroku API
* Adds lets-encrypt heroku gem
* Middleware mods for production

What's left to be done?
* Lots of environment variables, I think based on the documentation. Best to try this all out on staging with a dry run of the code. https://github.com/pixielabs/letsencrypt-rails-heroku/blob/master/README.md#configuring
* Heroku configuration: https://github.com/pixielabs/letsencrypt-rails-heroku/blob/master/README.md#creating-a-heroku-token
* First time use: https://github.com/pixielabs/letsencrypt-rails-heroku/blob/master/README.md#using-for-the-first-time
* Add scheduled task cron whatever to update cert: https://github.com/pixielabs/letsencrypt-rails-heroku/blob/master/README.md#adding-a-scheduled-task
* ???
* PROFIT!!!! (in small amounts, actually)

Prod steps based on Staging Dry Run:
- Tried to run heroku plugins:install heroku-cli-oauth
    - Didn’t run because already installed on staging
- Setup ENV variables
- Change ACME_EMAIL to DCAF email (currently using @colinxfleming )
- Deploy branch
- Won’t work on herokuapp url’s since these are SSL enabled by default
    - Need to setup REAL url’s to make sure it works (not an issue in Prod)
- Need to setup cron job (use Heroku scheduler, copy pasta) 

Production
* May need to do some DNS crap: https://devcenter.heroku.com/articles/ssl#change-your-dns-for-all-domains-on-your-app
* Make sure to shut off current SSL endpoint and Expedited SSL

PROFIT!!!!

* Does code related setup of #748

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem 'mongo_session_store-rails4'
 gem 'mongoid-enum'
 gem 'tzinfo-data', require: false
 gem 'js-routes'
-gem "omniauth-google-oauth2", "~> 0.2.1"
+gem 'omniauth-google-oauth2', '~> 0.2.1'
 gem 'platform-api', git: 'https://github.com/jalada/platform-api', branch: 'master'
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,7 @@ gem 'mongoid-enum'
 gem 'tzinfo-data', require: false
 gem 'js-routes'
 gem "omniauth-google-oauth2", "~> 0.2.1"
-gem 'platform-api', github: 'jalada/platform-api', branch: 'master'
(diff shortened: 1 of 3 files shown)


colinxfleming:
This rules. @DarthHater I'll take a deeper look this weekend and we can figure out how to try this out on staging. Thanks a ton for doing all this digging and for coming up with docs.

colinxfleming:
Keeping the issue open until we make the actual changes in prod, but merging now. Thanks @DarthHater !

PR #851, opened 2017-02-16
Allow for automated tests to be run in specific suites

### LINKED ISSUE #849: Create Distinct Dev and CI Environment Test Settings
Since introducing Rack-Attack in #699, we have tests that are not very useful to run every time a developer wants to test but add value in our CI build. There are probably other tests that meet these criteria today?

 - important to test
 - not likely to fail unless their corresponding code is directly messed with
 - take a relatively long time to run

Downsides: 
 - a user may break something and not know until they do a PR
    - mitigated by only moving very narrow tests
 - :question: 

Resources: 
 - [Customizing Rake Tasks In Rails 4.1 And Higher](http://metaskills.net/2015/02/08/customizing-rake-tasks-in-rails-41-and-higher/)

We may decide this isn't worth doing due to minimal time saved (Rack-Attack's test suite is only 5 seconds) or just not worth doing yet.

DarthHater:

This allows for rake tests to be run in a variety of ways, and switches default behavior:
* `rake test` will still run everything, but it can be modified to drop a specific suite if need be
* `rake test:everything` is now an implicit command to run ALL tests
* `rake test:limited` excludes integration tests
* `rake test:middleware` would run only the middleware tests (currently commented out) 

The end result of this is some narrow tests that could be laborious in nature can be moved to only running under CI.

* Modifications to Rakefile
* Small tweak to circle.yml to use `rake test:everything`

* Fixes #849 
* Mildly related to #820

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,3 +6,35 @@ Rails.application.load_tasks
 task clear_call_lists: :environment do
   User.all.each(&:clear_call_list)
 end
+
+Rake::Task['test:run'].clear
+Rake::Task['test:integration'].clear
+# Uncomment when throttling tests go live
+# Rake::Task['test:middleware'].clear
+
+namespace :test do
+  Rails::TestTask.new(:limited) do |t|
+    t.libs << 'test'
+    t.test_files = FileList['test/**/*_test.rb'].exclude(
+      'test/integration/**/*_test.rb'
+      )
+  end
+
+  Rails::TestTask.new(:everything) do |t|
+    t.libs << 'test'
+    t.test_files = FileList['test/**/*_test.rb']
+  end
+
+  Rails::TestTask.new('integration' => 'test:prepare') do |t|
+    t.libs << 'test'
+    t.pattern = 'test/integration/**/*_test.rb'
+  end
+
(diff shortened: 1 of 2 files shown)


colinxfleming:
@DarthHater @lomky  I'm thinking this thru and I think I'm inclined to close that associated issue as a wontfix for the following reasons: 

* the middleware tests don't take up a severe amount of time. Not a huge deal if they take a little while
* the testing pain point is the integration tests, but it's kind of important to run all of those
* I think the odds of these rake tasks getting much mileage is kind of slim

As an alternative, I think the right way to go might be to just make it clearer in the docs that you can run portions of the rails test suite with `rake test:whatever`, since that's like 90% of what we're trying to do, I think?

I think if we really expand our non model/controller/integration test suite we can and should revisit this question and implement it like this, but I think for now we can get away without this.

Interested in your thoughts @lomky @DarthHater .

DarthHater:
Fine by me, this was just a POC.

colinxfleming:
@DarthHater sounds good. Let's keep this in the back of our minds moving forward if parts of our test suite gets too bulky.

PR #856, opened 2017-02-18
Initial changes to support data volunteers

### LINKED ISSUE #855: Add data entry user so that we can have people enter fulfillment data
Add a non-admin user who enter pledge data.

Entering fulfillment data, so our admins don't have to do it.

By doing the following:
* Adding `data_volunteer` to the `Patient#role` enum, so we can start creating data volunteers
* Going into `views/patients/menu` and `views/patients/edit` and making so that the fulfillment partial renders if a user is an admin OR a data_volunteer
* Tweak the integration test in `tests/integration/pledge_fulfillment_test.rb` to make sure that data_volunteers can see the fulfillment piece
* Tweak the integration test in `tests/integration/create_user_test.rb` to make sure that a data volunteer CANNOT create a new user (basically, another set of tests for the stuff starting L63)

DarthHater:

This allows for data volunteer users to be created. 

* Thus far all implementation details are in #855 minus a typo (added enum to users not patients)

* Fixes #855

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -30,7 +30,7 @@ class User
   # Non-devise generated
   field :name, type: String
   field :line, type: String
-  enum :role, [:cm, :admin]
+  enum :role, [:cm, :admin, :data_volunteer]
   field :call_order, type: Array
 
   ## Database authenticatable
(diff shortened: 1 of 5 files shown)


colinxfleming:
Thank you very much @DarthHater ! :shipit:

PR #862, opened 2017-02-19
Fix db/seeds.rb to fail loudly on DB errors

lomky:

When poking around at Date formats, I noticed my `rake db:seed` was failing to find a patient after their name was updated, implying the update failed silently. Upon examining the `db/seeds.rb` file, I found the `create` and `update` methods are often called without a following `!`. When running in `rake db:seed`, you need to call `create!` or `update!` for any failure to happen loudly,

* `db/seeds.rb` all `create` calls are now `create!`
* `db/seeds.rb` all `update` calls are now `update!`
* `db/seeds.rb` test `Add example of patient with a pledge submitted` now functional
   * `Patient`s are required to have a `clinic_name` and `appointment_date` prior to pledge submission, so that update was moved to happen _before_ the pledge submit line (see lines 86-98) 

Not related to any issue, found while poking around in date format changes and how it affects the seeds file.

Submitted as two commits to illuminate the issue:
* first commit adds the `create!` and `update!` calls, making the `rake db:seed` call fail loudly at the bad update call.
* second commit fixes the failing seed

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -16,18 +16,18 @@ User.destroy_all
 #                         password: 'P4ssword', password_confirmation: 'P4ssword'
 
 # Create two test users
-user = User.create name: 'testuser (admin)', email: 'test@test.com',
+user = User.create! name: 'testuser (admin)', email: 'test@test.com',
                    password: 'P4ssword', password_confirmation: 'P4ssword',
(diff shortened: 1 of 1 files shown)


colinxfleming:
<3 <3 <3 <3 :shipit:

PR #870, opened 2017-02-23
Test fluke after 4:00 Pacific time - fixed date display

cschlessinger:

When running the test suite, I noticed that it started flaking after 4:00 p.m. PST. It appears to be how the display_date method was called in the test suite - it converted it out of UTC time. I changed it to use .strftime, so it should work correctly at any time now.

I didn't see an issue for this, just noticed it while running the test suite.


colinxfleming:
@cschlessinger thanks for fixing the test as you go, you're a gem! I'll review and merge this weekend, but again, I don't expect we'll need to change this any.

PR #865, opened 2017-02-22
Rails five dot oh dot two

DarthHater:
This is actually getting pretty close to "good". 

Currently you can:
* run rails s (app boots, some deprecation warnings)
* run rails db:seed (fails loudly but creates users)
* run rails test (down to like two failures)
* login as dcaf.testing@gmail.com user
* browse around the app
* do pretty much anything, I think the app is fully functional now

This pull request makes the following changes:
* Tons, it's wild

Known issues:
- [x] `rails db:seed` fails loudly (might be fine)
- [x] rack complains about `delete_session` not being implemented (it wasn't, now it is)
- [x] Tests look to be clobbered, probably need to investigate this on CircleCI
  - Most of this was legitimate, we had true failing tests, most are resolved now
- [x] Fix remaining test failures
- [x] Investigate and potentially resolve test deprecation messages (all look to be issues for Rails 5.1)
- [x] Figure out if code coverage results are really that bad?
- [x] 500 bad request on adding new data, can be replicated by opening any patient and then updating some data (anything as far as I can tell)
- [x] clean up as many deprecation warnings as we can (this may be impossible for certain libraries)

Some caveats:
- [x] We had to fork `mongoid-enum` and it's under `https://github.com/DarthHater/mongoid-enum` which seems to be the most logical plan of action. The original library hasn't seen a merge in over a year, the maintainer might have disappeared. Suggest we stick on this and then maybe move off of it in the future (or keep our fork updated). More at #866 
- [x] We had to use a fork of `mongo_session_store` at `https://github.com/drgcms/mongo_session_store.git`, likely stick on this until Rails 5 support is added to mongo-session-store (there is an issue for this). More at #867 
- [x] We had to use a fork of `mongoid_userstamp` and it's under `https://github.com/DarthHater/mongoid_userstamp.git` . For now this seems like the best course of action because when I tried to submit my changes to the upstream lib, I was asked if I wanted to take it over: https://github.com/tbpro/mongoid_userstamp/pull/35 . In the future we should look at moving off of this to something else, or maintaining it ourselves. More at #987 

It relates to the following issue #s: 
* Fixes #791

PR #872, opened 2017-02-23
Patients contacted cleanup

### LINKED ISSUE #712: REPORTING: Create Patients contacted class method

colinxfleming:

Picks up the work of @DarthHater and @agkramer in #865 and #715!

* adds a stub report for patients contacted and first contacted in a week

* Fixes #712

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,6 @@
 /vendor/bundle/
 /vendor/ruby/
 
-# npm stuff
-npm-debug.log
-
 # minimal Rails specific artifacts
 db/*.sqlite3
 /db/*.sqlite3-journal
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -125,29 +125,19 @@ class Patient
   end
 
   def self.contacted_since(datetime)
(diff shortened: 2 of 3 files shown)


colinxfleming:
CI failing for bs reasons. Merging. Thanks all!

PR #874, opened 2017-02-23
Make identifier searchable pt 2

### LINKED ISSUE #727: ACCOUNTING: Make patient identifier searchable

colinxfleming:

Finishes off @Kevin-Wei 's work in #821 

* Fixes #727

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -34,6 +34,7 @@ class Patient
   field :other_phone, type: String
   field :other_contact_relationship, type: String
   field :clinic_name, type: String
+  field :identifier, type: String
 
   enum :voicemail_preference, [:not_specified, :no, :yes]
   enum :line, [:DC, :MD, :VA] # See config/initializers/lines.rb. TODO: Env var.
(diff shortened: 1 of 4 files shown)


colinxfleming:
Making a note here that I ran the following in prod to completely roll this out:
* from terminal: `rake db:mongoid:create_indexes`
* in rails c: `Patient.all.each { |pt| pt.save }`

Smoke tested and looking good!

Issue #800, opened 2017-01-05
Add rake task to generate training accounts


colinxfleming:

colinxfleming:
Let's do this: `rake db:seed:training` should create 20 accounts as follows:

* email: `training-account-#{i}@test.com`
* password: `P4ssword`
* password_confirmation: `P4ssword`
* name: `Training Account #{i}`

colinxfleming:
Bumping to active and marking beginner friendly, since it's a nice to have.

PR #876, opened 2017-02-24
wip changes for db:seed:training

cvaldivia19:
Not yet ready

Relates to issue #800

### DIFF
--- a/lib/tasks/seed_training.rake
+++ b/lib/tasks/seed_training.rake
@@ -1,0 +1,20 @@
+#Let's do this: rake db:seed:training should create 20 accounts as follows:
+
+#email: training-account-#{i}@test.com
+#password: P4ssword
+#password_confirmation: P4ssword
+#name: Training Account #{i}
+
+namespace :db do
+  namespace :seed do
+    task :training => :environment do
+      # Create two test users
+      20.times do |i|
+        User.create! name: "Training Account #{i}",
+                     email: "training-account-#{i}@test.com",
+                     password: "P4ssword",
+                     password_confirmation: "P4ssword"
+      end 
+    end
+  end
+end


colinxfleming:
This looks pretty good to my eyes @cvaldivia19 ! What's left to handle in here? 

Thanks again for tackling this! Glad to have you on the team!

cvaldivia19:
Hi @colinxfleming! @lomky was helping me with this and we didn't finish checking it but I'm not quite sure what still needs to be done...

lomky:
Indeed, we hit a the end time barrier and were hurrying to get through committing the work we'd done. It wasn't working when we committed it, but we did no debugging yet. I think this is probably just a few tweaks from working.

lib/tasks/seed_training.rake
```diff
@@ -0,0 +1,20 @@
+#Let's do this: rake db:seed:training should create 20 accounts as follows:
+
+#email: training-account-#{i}@test.com
+#password: P4ssword
+#password_confirmation: P4ssword
+#name: Training Account #{i}
+
+namespace :db do
+  namespace :seed do
+    task :training => :environment do
```
colinxfleming: So that this shows up when we run `rake -T` (the rails command for seeing available rake tasks), we'll want to add a `desc`. For an example of this, see https://github.com/colinxfleming/dcaf_case_management/blob/master/lib/tasks/user.rake#L2 , and see how the string after `desc` shows up next to the task when you run `rake -T`.

lib/tasks/seed_training.rake
```diff
@@ -0,0 +1,20 @@
+#Let's do this: rake db:seed:training should create 20 accounts as follows:
```
colinxfleming: To protect our prod environment, can you add the following safety check here?

```
fail 'No running seeds in prod' unless [nil, 'Test Sandbox'].include? ENV['FUND']
```

colinxfleming:
Thanks again for getting this going and taking care of it @cvaldivia19 , this is going to be super useful toward our training efforts. Thanks for your first pull request on this project, so glad to have you! 💯

cvaldivia19:
Thanks, Colin, and sorry for my late reply! Is there anything that remains
to be done here?

colinxfleming:
Nope you're all good, thanks! Looking forward to meeting you at code for dc
soon

PR #879, opened 2017-02-26
WIP Fix some persistent integration test failures

colinxfleming:

WIP PR to fix pull request failures as they choke up our builds. I think most of these can be fixed with some good old waiting.


colinxfleming:
Good for now, but to be continued I think!

PR #881, opened 2017-02-26
Complete restyling call list

### LINKED ISSUE #795: Improvements to notes section in patient edit view

colinxfleming:

Jumps off @AlexandraCasser 's clutch work in #827 to make the call list much, much less cramped, and adds a little extra:

This pull request makes the following changes in addition to what's in #827 :
* turns glyphicons into dashboards helper methods for cleaner code and no html safe-ing
* reduces note text in the call list table to first 30 characters
* (etc)

Look at this beautiful call list @AlexandraCasser remixed into existence:

screen shot 2017-02-26 at 3 11 30 am

* Fixes #795


lwaldsc:
Chiming in to say this looks gorgeous. Great job @AlexandraCasser
<https://github.com/AlexandraCasser>!

PR #882, opened 2017-02-26
REPORTING: Build basic report view #659

### LINKED ISSUE #659: REPORTING: Build basic report view
Team UX has whipped up a draft design, and @itsWill set up a reports controller, so it's time to put it to work! 

dcafwireframe091316_data center

Build this out in `app/views/reports/index.html.erb`. Don't worry about populating with actual data for now -- just write down 'TK' or hardcore a fake number.

Sourced from #634

ashlynnpai:
* Adds reports UX with hardcoded data and temporary budget bar

* fixes #659 

reports_1

### DIFF
--- a/app/assets/stylesheets/reports.scss
+++ b/app/assets/stylesheets/reports.scss
@@ -1,3 +1,21 @@
 // Place all the styles related to the Reports controller here.
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: http://sass-lang.com/
+
+$dcaf-color: #5D3E8D;
+
+.report-section{
+  margin-bottom: 50px;
+}
+
+.line-header, .line-header:hover{
+  color: $dcaf-color;
+}
+
+#budget_bar{
+  background-color: #f5f5f5;
+  margin-left: 10px;
+  margin-bottom: 20px;
+  padding: 10px;
+}
+    
--- a/app/assets/stylesheets/reports.scss
+++ b/app/assets/stylesheets/reports.scss
@@ -12,10 +12,3 @@ $dcaf-color: #5D3E8D;
   color: $dcaf-color;
 }
 
-#budget_bar{
-  background-color: #f5f5f5;
-  margin-left: 10px;
-  margin-bottom: 20px;
-  padding: 10px;
-}
-    
--- a/app/views/reports/_call_line.html.erb
+++ b/app/views/reports/_call_line.html.erb
@@ -1,0 +1,45 @@
+<div class="report-section">
+  <h2 class="border-bottom">Week of <%= week_range %></h2>
+
+  <h3><%= link_to "DC Call Line", '#', class: 'line-header' %></h3>
+  <div class="row">
+    <h4 class="col-sm-8" id="budget_bar"><strong>Budget bar, design TK</strong></h4>
+  </div>
+  <!-- UNCOMMENT WHEN BUDGET BAR IS READY
+  <div id="budget_bar">
+    <div class="progress"></div>
+    <div class="progress-bar progress-bar-striped"</div>
+  </div>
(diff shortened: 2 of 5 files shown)


app/views/reports/_call_line.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="report-section">
+  <h2 class="border-bottom">Week of <%= week_range %></h2>
+
+  <h3><%= link_to "DC Call Line", '#', class: 'line-header' %></h3>
```
colinxfleming: Looks like this code repeats a few times, and I bet we can clean it up. I have a few ideas, what do you think is the best one?

* We can either send data to the report in an array as follows and loop thru it:

```
[ 'DC' => { patients_contacted: 46, new_patients_contacted: 23, pledges_sent: 34 }, 'MD' => { ... }, 'VA' => { ... } ]
``` 
 and then do something like:

```rb
<% lines.each do |line| %>
Patients contacted: <%= @data[line][:patients_contacted] %>
...
<% end %>
```

* Or, we could move these to their own partials, and do something like this:

```rb 
# in this partial
<%= render 'reports/contact_report, data: @data['DC'] %>
<%= render 'reports/contact_report, data: @data['MD'] %>
... etc

# in its own subpartial, L4-16 would move
<h3><%= link_to "#{data[:line]} Call Line", '#', class: 'line-header' %></h3>
... etc
```

@ashlynnpai curious as to your thoughts! I think I favor the loop, given what the index looks like. Do you think one sounds more flexible and easier to maintain / expand than the other?

app/views/reports/_past_thirty_days.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="report-section">
+  <h2 class="border-bottom">Past 30 days</h2>
+  <div class="row">
+    <div class="col-sm-4">
+      <h3><%= link_to "DC", '#', class: 'line-header' %></h3>
```
colinxfleming: Same deal, let's see if we can either loop this or partialize it.

app/views/reports/index.html.erb
```diff
@@ -1 +1,5 @@
-<h1>Index for Reports</h1>
+<h1><strong>DCAF Data Center</strong></h1>
+
+<%= render partial: 'reports/call_line' %>
```
colinxfleming: Love this -- this is going to be super straightforward to extend.

app/assets/stylesheets/reports.scss
```diff
@@ -1,3 +1,21 @@
...
+
+.report-section{
+  margin-bottom: 50px;
+}
+
+.line-header, .line-header:hover{
+  color: $dcaf-color;
+}
+
+#budget_bar{
```
colinxfleming: If we can, let's drop this CSS and just make sure we apply a global style to the budget bar when it's done, since it's a separate feature. I think just the text placeholder is fine for now. Thoughts?

app/views/reports/_call_line.html.erb
ashlynnpai: I think I'm missing why we want to put the data into an array first instead of just pulling the data straight from the database. Where does @data get defined?

ashlynnpai:
How about this?

colinxfleming:
@ashlynnpai I love this, it is perfffffff. Excellent work!

colinxfleming:
(I'll give it a merge tomorrow when I am settled in after work.)

colinxfleming:
@ashlynnpai wanted to say thank you again -- this is a really exciting pull request because it starts to move us into up to the minute reporting, which is something that literally no other abortion fund has. I cannot understate how stoked I am to merge this PR. Thank you so much! <3

PR #883, opened 2017-02-26
[#845] limit patient results returned

### LINKED ISSUE #845: Limit patient results returned

Limit the results on patient search to the 15 most recently touched patients.

Making it so that you can't enter a phone number of '202' and get back ~300 patients.

Add a limit of 15 to the `Patient.search` method, sorted by either date of last touch or created, whichever's more straightforward. Best to do this on the query level if at all possible, to optimize.

FeminismIsAwesome:

This adds a limit to the number of patient search results being returned.

-> adds a limit and ability to sort the search results coming back.

* Fixes #845 

Notes:
I didn't know how to deal with the fact updated_at can't be accessed directly in mongo, so if you've got a better way to test the sorting, please let me know!

Staring at the current search functionality, I opted to not try and figure out how to make it AREL-y, since the two different main functions at play find_name_matches and find_phone_matches, aren't either.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -7,6 +7,7 @@ class Patient
   include Mongoid::Userstamp
   include StatusHelper
 
+  SEARCH_LIMIT = 15
   LINES.each do |line|
     scope line.downcase.to_sym, -> { where(:_line.in => [line]) }
   end
@@ -215,10 +216,18 @@ class Patient
       all_matching_names = find_name_matches name_regexp, lines
       all_matching_phones = find_phone_matches phone_regexp, lines
 
-      (all_matching_names | all_matching_phones)
+      sort_and_limit_patient_matches(all_matching_names, all_matching_phones)
     end
 
     private
+
+
+    def sort_and_limit_patient_matches(all_matching_names, all_matching_phones)
+      matches = (all_matching_names | all_matching_phones)
+      matches.sort{|a,b|
+        b.updated_at <=> a.updated_at
+      }.first(SEARCH_LIMIT)
+    end
 
     def find_name_matches(name_regexp, lines = LINES)
       if nonempty_regexp? name_regexp
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -215,16 +215,19 @@ class Patient
 
       all_matching_names = find_name_matches name_regexp, lines
       all_matching_phones = find_phone_matches phone_regexp, lines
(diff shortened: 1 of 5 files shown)


app/models/patient.rb
```diff
@@ -215,11 +216,19 @@ def search(name_or_phone_str, lines = LINES)
       all_matching_names = find_name_matches name_regexp, lines
       all_matching_phones = find_phone_matches phone_regexp, lines
 
-      (all_matching_names | all_matching_phones)
+      sort_and_limit_patient_matches(all_matching_names, all_matching_phones)
```
colinxfleming: You may want to pull from master -- @Kevin-Wei added a `search by identifier` to this concatenation.

app/models/patient.rb
```diff
@@ -215,11 +216,19 @@ def search(name_or_phone_str, lines = LINES)
...
     end
 
     private
 
+
+    def sort_and_limit_patient_matches(all_matching_names, all_matching_phones)
+      matches = (all_matching_names | all_matching_phones)
+      matches.sort{|a,b|
```
colinxfleming: This'll work very nicely, I think it's great!

The world's most minor style suggestion: The ruby style for blocks that span one line is  `{ |x| ... }`, and for blocks with line breaks it's `do |x|`. Just so it's a little more 'ruby' stylistically,  you might want to toss the block on one line. NBD if not though, up to you!

app/models/patient.rb
```diff
@@ -215,11 +216,19 @@ def search(name_or_phone_str, lines = LINES)
...
+        b.updated_at <=> a.updated_at
```
colinxfleming: I love that you used the spaceship operator here, because it reminds me that the spaceship operator is a thing

FeminismIsAwesome:
@colinxfleming , i think i've got the formatting right on the blocks this time? Let me know if there's something still there to fix. Also because of the number of arrays we were shoveling into method, I felt compelled to splat it, but I could also just combine the matches before sorting/limiting them. Also added tests for the identifier, because the one that was there wasn't really specific enough.

FeminismIsAwesome:
@colinxfleming , the test failing is an integration spec i've verified works locally---the call list test that shoots off an email and just sleeps for a second

colinxfleming:
Works like a charm and looks great! Thanks a ton! This closes a bit of a security issue for us, and I think the sort by last mod is a really nice CM experience benefit as well. Thanks so much @FeminismIsAwesome !

PR #890, opened 2017-02-28
add change line functionality for admins

### LINKED ISSUE #868: Allow admin users to change a patient's line

Allow admin users to change a patient's line.

The most common reason I fire up `rails c` these days is to change a patient's line. I want to empower the case management directors to fix this so the team doesn't have to go thru me.

We should: 
* add a select option to the `patient_information` partial for `Patient#line`, and have it display only if a user is an admin.
* add a test to confirm that it doesn't show up if you aren't an admin

tingaloo:

Adds a line dropdown box for admins

* Adds line dropdown on patient information tab.
* Checks if line option is visible for non-admins and for admins

Uploading images not working on github.

* Fixes #868

### DIFF
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -25,6 +25,10 @@ module PatientsHelper
      'No insurance', 'Don\'t know', 'Other (add to notes)']
   end
 
+  def line_options
+    ['DC', 'MD', 'VA']
+  end
+
   def income_options
     [nil,
      ['Under $9,999 ($192/wk - $833/mo)', 'Under $9,999'],
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -9,6 +9,12 @@
 
       <div class="row">
         <div class="info-form-left col-sm-6">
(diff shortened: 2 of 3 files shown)


app/helpers/patients_helper.rb
```diff
@@ -25,6 +25,10 @@ def insurance_options
      'No insurance', 'Don\'t know', 'Other (add to notes)']
   end
 
+  def line_options
```
colinxfleming: I think we can probably consume an existing method for this in `lines_helper`, `lines`. (which just invokes the `LINES` constant, which comes from an env var). The instinct is totally right on though!

app/views/patients/_patient_information.html.erb
```diff
@@ -9,6 +9,12 @@
 
       <div class="row">
         <div class="info-form-left col-sm-6">
+          <% if current_user.admin? %>
+            <%= f.select :line,
+                       options_for_select(line_options,
```
colinxfleming: Go ahead and swap out `line_options` for `lines` here, if you don't mind

colinxfleming:
Perfect! Thanks a ton @tingaloo !

PR #902, opened 2017-03-04
Update setup

zorkian:
This fleshes out slightly the Docker steps.

This is a documentation update for Docker setup. I can't vouch this is 100% correct but it worked for me. Feedback/tweaks/whatever is always welcome.

* Update setup documentation

### DIFF
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -11,10 +11,27 @@ For the rest of the setup, you have three options: Docker, installing everything
 ## Docker
 
 We've dockerized this app, to manage the dependencies and save us some headache. If you've got [Docker installed already](https://docs.docker.com/engine/installation/), you can be up and running with three commands:
-* `docker-compose build`
+
+* `docker-compose pull`
+* `docker-compose build` (this may say 'uses an image, skipping' a few times, that's OK)
 * `docker-compose run web rake db:seed # to populate the database`
 * `docker-compose up`
 
+The last command will take a moment and should print a number of things. When it's ready
+to go, it should say something like:
+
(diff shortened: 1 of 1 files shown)


docs/SETUP.md
```diff
@@ -11,10 +11,27 @@ For the rest of the setup, you have three options: Docker, installing everything
...
+to go, it should say something like:
+
+    web_1  | [1] * Listening on tcp://0.0.0.0:3000
+
+In order to connect to the application you will need the IP address of your Docker instance.
+This can be found with the `docker-machine ip default` command. With the IP, and the port
+from the output of the original `up` command, you can now go to your browser and navigate
+to the application.
+
+For example, if your Docker machine's IP is `192.168.99.100`, you would go to your browser
```
colinxfleming: I usually do `localhost:3000` but that's probably a YMMV thing -- can you try hitting a localhost and confirm it doesn't work? If that's right let's def keep this in

rfroetscher: this depends on what version of docker you are running. if you are using docker for mac there is no need to do the docker machine stuff. sounds like you are using the older docker toolbox.

colinxfleming: copy that, thanks!

colinxfleming:
image

docs/SETUP.md
zorkian: localhost didn't work on my side, neither did 127.0.0.1. I may indeed be on an 'old docker' but I just installed it via `brew update && brew install docker` so I'd be surprised -- but very likely they way that Homebrew installs it is... special?

I can undo it and use the official distribution, I just default to using Homebrew because it's usually easy and (mostly) correct.

PR #908, opened 2017-03-04
Adjust status helper for rare scenario

### LINKED ISSUE #676: Patient status in cases of appt date left in voicemails
CM Joy and @lwaldsc flagged a scenario where a PT will give an apt date in the voicemail, but we may not actually contact the PT. If the CM fills in the appointment date, the PT's status will be fundraising, regardless if we have ever talked to them. It is rare of enough an occasion, we may not have to do anything, but just want to flag that.

Should we adjust the behavior of status to account for this? 

cc @NerdyGirl537

zorkian:
There's an edge case where a PT calls in and leaves a voicemail
indicating they have an appointment date. If we then set the appointment
date in the system, the status would have been 'Fundraising' but in
reality we would prefer it to say 'No Contact Made'.

This fix adjusts the priority of the statuses slightly that No Contact
Made will come up if we haven't talked to a user yet, regardless of
their appointment status.

Fixes issue #676.

* Adjust status helper as above
* Adjust tests to test new behavior

* Fixes #676

### DIFF
--- a/app/helpers/status_helper.rb
+++ b/app/helpers/status_helper.rb
@@ -12,9 +12,9 @@ module StatusHelper
   def status
     return STATUSES[:resolved] if pregnancy.resolved_without_dcaf?
     return STATUSES[:pledge_sent] if pregnancy.pledge_sent?
+    return STATUSES[:no_contact] if not contact_made?
     return STATUSES[:fundraising] if appointment_date
-    return STATUSES[:needs_appt] if contact_made?
-    STATUSES[:no_contact]
+    STATUSES[:needs_appt]
   end
 
   private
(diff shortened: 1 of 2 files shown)


colinxfleming:
this looks good @zorkian ! I'm going to spend a little bit of time charting out the scenarios to make absolutely sure we've got the statuses right, so I'll flag if anything needs a shift, but this is excellent. 

Thank you so much for the test fixer upper as well!

zorkian:
This test works locally ... can you trigger a retry on circleci?

colinxfleming:
Thanks again for this fix, @zorkian ! Now that I'm much more rested than I was a month ago it's a lot easier to follow the logic, haha.

PR #911, opened 2017-03-04
Attempt 1

phoenixeliot:
Trying some stuff.


colinxfleming:
@phoenixeliot I think this is on the right track for sure!

When I try it out as is now, here's what I'm seeing: 

image

so it looks like something somewhere isn't firing off quite right. But I think this is definitely on the right track!

Issue #524, opened 2016-07-18
Add ability for a user to remove their own calls


colinxfleming:
A time or two during the pilot, @lwaldsc accidentally clicked 'left voicemail' on an 'unable to reach patient' call, and right now we aren't able to remove calls from the call log.

Potential solution here that I would like to run by UX: Maybe do it kinda like facebook does, where you can remove a call within a certain time period (5m? 10m?) of creating it, but then it's in there for keeps? Or should we just have it so that users can edit their own calls? Or should we just open it up and have everything editable?

eheintzelman:
Discussed with @lwaldsc - Would it be possible to make calls editable during a CM's current session?  Once the sign out or their session times out, they no longer have the ability to edit.

colinxfleming:
@eheintzelman I think we can swing that, or do something that's functionally identical. I'll change the issue accordingly.

colinxfleming:
Technical requirements here are as follows:

Minimal:
* A user should be able to delete their own calls from the call log
* A user should not be able to delete other users' calls from the call log
* We should adjust the `call_log` partial to add the delete option, with a javascript alert confirmation

Gold medal version:
* a user should only be able to delete their own calls within their own session or within 8 hours, whichever is easier
* Both the views and model logic should reflect that behavior

colinxfleming:
TBH I think this is a nice to have rather than a must have. I'm marking backlog, please let me know if you disagree @lwaldsc @NerdyGirl537 .

colinxfleming:
Promoting this to active!

mebates:
For the UI, please add a column in the Call Log table with the word "Remove" in it, as a red underlined link. This resembles the 'Remove' behavior on the call list. 

+1 to the javascript alert confirmation

charleshuang80:
I'll work this issue today

colinxfleming:
@charleshuang80 I'd put good money on you being in good shape on this, but let me know if you need an un-stick!

colinxfleming:
Promoting to hackathon!

yf6:
I'd like to give this a try during the hackathon, just to confirm, we are talking about this table here?
<img width="1191" alt="screen shot 2017-03-03 at 11 27 43 pm" src="https://cloud.githubusercontent.com/assets/1358856/23577014/0f685806-0069-11e7-81b3-71a11a043c83.png">

colinxfleming:
@yf6 yep, that's the one!

PR #913, opened 2017-03-05
Allow a user to remove their own calls

yf6:
Add a destroy endpoint to calls controller. 
Add delete button to calls in call log that's created recently by the current user.

* Fixes https://github.com/colinxfleming/dcaf_case_management/issues/524

@colinxfleming sending this your way while I write some tests and hopefully find a way to get the tests running locally.

### DIFF
--- a/app/controllers/calls_controller.rb
+++ b/app/controllers/calls_controller.rb
@@ -1,6 +1,6 @@
 # Create method for calls, plus triggers for modal behavior
 class CallsController < ApplicationController
-  before_action :find_patient, only: [:create]
+  before_action :find_patient, only: [:create, :destroy]
 
   def create
     @call = @patient.calls.new call_params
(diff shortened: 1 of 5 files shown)


yf6:
Tests added! Ready for reviews!

colinxfleming:
ahhhh the gif

mebates:
Kind of a BOLD action for something that may not be used much. Could we convert it from a button to just an underlined red link?

yf6:
my 2c: buttons and links are semantically different, a button is an action in the current context and a link navigates the user to a different resource. If the concern is the big red button is too bold, we can style it differently, but a button makes more sense here.

colinxfleming:
Clutch part of this was getting the functionality in place -- if we need to tweak the styling let's record that in a separate issue and fix with a separate PR. 

Thanks so much @yf6 , hope you're caught up on sleep by now!

PR #924, opened 2017-03-05
Make the call tables' sortable by clicking columns

### LINKED ISSUE #896: Add sortability to call list dashboard table

Our case managers have asked for the ability to sort the patient dashboard by the column headers.

image

Case managers often want to sort their call list or urgents list by appointment date, so they can prioritize their calls easier.

Probably a JS library, but this is something I don't know much about, so chef's choice!

* **Anything else?**
If there's an easy path to this then that's awesome, but we have this implemented in a limited way already with drag and drop. 

The file here that generates tables is located in `app/views/dashboards/_table_content.html.erb`.

janinedog:

I made it so the "Status" and "Appt Date" columns in the three call tables can be clicked to sort the rows in the tables.  I used a lightweight jQuery plugin called [StupidTable](http://joequery.github.io/Stupid-Table-Plugin/).

Note that I added the `autosortable` parameter to the `calls/create.js.erb` and `users/refresh_patients.js.erb` files because the `sortable` parameter (which does the drag-and-drop behavior) was in those templates, but I don't know if those files are used.

screen shot 2017-03-04 at 7 40 33 pm

* Fixes #896


app/views/dashboards/_table_content.html.erb
```diff
@@ -11,8 +21,8 @@
         <th>Phone</th>
         <th>Name</th>
         <th>LMP</th>
-        <th>Status</th>
-        <th>Appt date</th>
+        <th<%= string_sort.html_safe %>>Status<%= autosort_span.html_safe %></th>
```
colinxfleming: Extra `>` here by accident?

app/views/dashboards/_table_content.html.erb
```diff
@@ -1,4 +1,14 @@
-<!-- locals: title, table_type (for div id and tbody div id), pregnancies, sortable -->
+<!-- locals: title, table_type (for div id and tbody div id), pregnancies, sortable, autosortable -->
+
+<%
```
colinxfleming: Can you turn these into methods in `dashboards_helper`? Something like: 

```
def string_sort(local_assigns)
  local_assigns[:autosortable] ? ' data-sort="string"' : ''
end
```
etc

app/assets/javascripts/calls.coffee
```diff
@@ -38,5 +38,16 @@ jQuery ->
...
     )
 
-
-
+    # make the columns of all of the tables sortable
+    autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table' ]
+    autosortable_tables.forEach ( table ) ->
+        stupidtable = $( table ).stupidtable()
+
+        # when a column is clicked, make sure to reset all of the other columns back to [-]
```
colinxfleming: <3 <3 <3 for explanatory comments!

app/views/dashboards/_table_content.html.erb
janinedog: Nope!  One is for the template closing tag, one is for the HTML closing tag.

colinxfleming: oh, I see! copy that!

janinedog:
@colinxfleming I made the requested changes.  Let me know how it looks!

app/views/dashboards/_table_content.html.erb
```diff
@@ -11,8 +22,8 @@
         <th>Phone</th>
         <th>Name</th>
         <th>LMP</th>
-        <th>Status</th>
-        <th>Appt date</th>
+        <th<%= string_sort.html_safe %>>Status<%= autosort_span.html_safe %></th>
+        <th<%= string_sort.html_safe %>>Appt date<%= autosort_span.html_safe %></th>
```
colinxfleming: Can we change this to something like `<th<%= autosort_string_html_attr(local_assigns) %>>`?

colinxfleming:
@janinedog this looks good to me! I have a super slight style edit (I tend to follow the practice of not assigning variables in views where possible) but I'm going to make it in a separate slightly adjusted PR tomorrow morning, since that's really just my own weird junk and there's no reason to make you jump thru that hoop on my behalf.

In other words, 

image

janinedog:
Got it!  I actually meant to pull out those variables completely but apparently forgot.  Thanks for the modification--you're right.  :)

mebates:
So is the sort action this braketed minus sign? [-]
Could it be an arrow? Up or down based on which ones are sorted or not?

colinxfleming:
If you mash the columns they sort, and then the dash turns into an arrow up
or down. If there is a symbol to better represent 'mash this here button to
sort' then we can see if it's possible to sub in something else. Maybe
let's  open a new issue for that?

janinedog:
@mebates Yeah, as @colinxfleming said, the [-] symbol is what it shows when there's no active sort (i.e. it's just in the default order that it's in on page load).  If you click that symbol, it becomes [&uarr;] or [&darr;], to show the direction of the sort.  Sorry I forgot to take a screenshot of that part!

mebates:
Ok good with me! Thanks!

lwaldsc:
Just want to say thanks on this. It seems like a small feature but it is
*super* helpful when case managing!

PR #927, opened 2017-03-05
Slight edit to sortable table content headers

colinxfleming:

Very slight adjustments to @janinedog 's work in #924  - relies on the helper methods rather than setting vars in view

* adjusts her great work per my own weird junk

* Fixes #924


janinedog:
👍  Thanks!

PR #929, opened 2017-03-05
Add clinic editing functionality

### LINKED ISSUE #146: Clinic modeling
We have a limited number of clinics that doesn't change much, that should be admin-editable. It's not optimal behavior to edit this from 

IDK what the UX component of this looks like, but the proper data model is probably a `has_one :through` between pregnancy and clinic, so we can assign an identifier from a list of clinics.

jingpei:
* Edit views for clinics
* Restructured clinic form to be used by both edit and add views

Added Edit link to Clinics index page
screen shot 2017-03-05 at 11 13 46 am

Added Edit View
screen shot 2017-03-05 at 11 14 49 am

* Fixes #146

### DIFF
--- a/app/controllers/clinics_controller.rb
+++ b/app/controllers/clinics_controller.rb
@@ -1,6 +1,6 @@
 class ClinicsController < ApplicationController
   before_action :confirm_admin_user
-  before_action :find_clinic, only: [:update, :destroy]
+  before_action :find_clinic, only: [:update, :edit, :destroy]
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
 
   def index
@@ -23,14 +23,22 @@ class ClinicsController < ApplicationController
     @clinic = Clinic.new
   end
 
-  # def update
-  #   clinic = Clinic.find params[:id]
-  #   if clinic.save
-  #     head :ok
-  #   else
-  #     head :bad_request
-  #   end
-  # end
+  def edit
+  end
+
+  def update
+    if @clinic.update_attributes clinic_params
+      flash[:notice] = 'Successfully updated clinic details'
+      redirect_to clinics_path
+    else
+      flash[:notice] = 'Error saving clinic details'
+      redirect_to clinics_path
+    end
+  end
+
+  def find_clinic
+    @clinic = Clinic.find params[:id]
+  end
 
   # def destroy
   # end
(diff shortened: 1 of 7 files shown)


app/controllers/clinics_controller.rb
```diff
@@ -23,14 +23,22 @@ def new
...
-  # end
+  def edit
+  end
+
+  def update
+    if @clinic.update_attributes clinic_params
+      flash[:notice] = 'Successfully updated clinic details'
+      redirect_to clinics_path
+    else
+      flash[:notice] = 'Error saving clinic details'
```
lomky: Could this please `flash[:alert]` instead of `flash[:notice]`? That will make it appear styled red to grab attention on failures.

lomky:
Looks great! There's one small tweak to improve user experience, but otherwise I think this is good to go @colinxfleming

lomky:
Leaving this screenshot for @mebates design feedback

screenshot from 2017-03-06 20-14-55

mebates:
Awesome work @jingpei !!! Here's some design thoughts: 
How many clinics do we anticipate will be on this page? If there are a lot, a sortable table might be helpful, or maybe two columns. Will we order the clinics alphabetically, or by state, or something else?

I would like to see more differentiation between the three actions here. "Add a new clinic" could be a purple button. "More details" could look more like an accordion - perhaps each clinic has a field behind it, creating a box that expands? Or at the least, it can be "+ More details" with a plus sign icon in a circle.

colinxfleming:
We are probably looking at like 30+ clinics.

Jing and I talked about doing this via accordion so it's cool to have
guessed that right. You're rubbing off on me mollie!

Mollie - Since this is still a page nobody is using in production yet,
what's the feeling around merging as is and doing the styling in a separate
pr?

lwaldsc:
@colinxfleming she'll have you mastering indesign next

jingpei:
@mebates I also thought a sortable (or at least searchable) table would be cool, one where the initial line is clinic details (name, address, phone), and the row right below an expandable row that takes up the entire colspan. That'll probably take some reworking though I'd love to see that

I worked on this page last year and thought it was a pretty cool interaction -- mainly how the first row of each card is the details but the "more details" section is kind of it's own collapsible row. Obviously it won't look as fancy as this.
UX I was thinking of:
https://www.nerdwallet.com/the-best-credit-cards

mebates:
Fine with this merging without the styling! @jingpei I like your idea of a table with an expandable row below it. Thanks!

colinxfleming:
Perf. @mebates I opened a UX ticket for styling this. Merging as is for the time being.

PR #939, opened 2017-03-05
just have the js on the specific page plus fix that array

rfroetscher:
Quick bugfix here. This was causing test problems
@colinxfleming


colinxfleming:
$$$

PR #941, opened 2017-03-05
Add tooltip support next to form field labels

### LINKED ISSUE #793: Include pledge recommendations and guidelines based on LMP
* **What are we trying to do?** DCAF has some guidelines on how much money to grant based on LMP; CMs should have pledge caps on hand when they're talking to patients, especially since they sometimes change.

* **What feature or behavior is this required for?** As a user, I want to not accidentally blow my budget and make the CM directors mad at me.

* **How should we solve this issue?** TBD

* **Anything else?** This probably requires some UX work, but in no way is this a priority.

Stems from #593

janinedog:

* Added support for tooltip help thingys next to form field labels
* Added a tooltip next to the DCAF Pledge field that shows recommended pledge limits

screen shot 2017-03-05 at 2 30 04 pm

screen shot 2017-03-05 at 2 29 12 pm

* Fixes #793

### DIFF
--- a/app/assets/javascripts/patients.coffee
+++ b/app/assets/javascripts/patients.coffee
@@ -44,4 +44,14 @@ ready = ->
   if $("#patient_pregnancy_procedure_cost").val()
     updateBalance()
 
+  # put a help icon next to form field labels that have the "tooltip-header" class
+  # the text that shows up in the tooltip is from the label's corresponding form field's "data-tooltip-text" attribute
+  $( '.tooltip-header' ).append( ' <span class="tooltip-header-help">(?)</span>' )
+  $( '.tooltip-header' ).parent( '.form-group' ).find( '.form-control' ).each ->
+    $(@).parents( '.form-group' ).find( '.tooltip-header-help' ).tooltip( {
+      html: true,
+      placement: 'top',
+      title: $(@).data( 'tooltip-text' )
+    } )
+
 $(document).on 'ready page:load', ready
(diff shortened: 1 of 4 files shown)


mebates:
Looking good! @colinxfleming - this is the first tooltip I can think of in the app. Can you think of others? I don't love tooltips, but I also don't hate this version of a tooltip! If we're thinking of using more tooltips throughout the app we may want to create a standard. 

@janinedog did you use a code framework? How might this be accessed for someone using a screen reader? If you're up for making some front end tweaks, I would add some more padding around the type so it's not so squished.

mebates:
Just told @lomky that this can be pushed into production w/o any design things fixed cause it already looks pretty darn good!!

colinxfleming:
I think there's a lot of potential value add for explaining exactly what
the fields mean and represent via these. E.g. what constitutes income, how
household ing works, etc. If the code looks good @lomky let's mash this in
and then figure out a way to abstract it out into a standard thing later,
once we get copy from team cmd.

lwaldsc:
Popping in to say this tool tip is super helpful! I could never remember
limits when I first started CMing!

janinedog:
@mebates I used the built-in bootstrap tooltip to do this.  With this change, it's easy to add a tooltip to any form field going forward; just follow the directions in this comment:

```
# put a help icon next to form field labels that have the "tooltip-header" class
# the text that shows up in the tooltip is from the label's corresponding form field's "data-tooltip-text" attribute
```

colinxfleming:
Test failure is lying. Merging!

PR #953, opened 2017-03-18
Upgrade CircleCI to 2.0

colinxfleming:
Work-in-progress PR to convert our CI builds to CircleCI 2.0, for more docker friendliness and hopefully getting our flaky tests to chill.


colinxfleming:
Closing without merging for the following reasons:

a) this isn't fixing the flaky tests
b) it's way more complicated and precise than our current build. Like, we just need to run rake test and a few cli gem commands, not set up a complex multistep docker build
c) I would be amped for there being an easy way to just do 'docker-compose run web rake test` and there doesn't seem to be one
d) I have zero interest in figuring out how to do the 

Since we're a humble volunteer project rather than an enterprise level company that needs a very granular CI system, I don't think this really suits our needs at this time. Cool that we got into the beta though.

PR #958, opened 2017-03-23
Back out Lets Encrypt stuffz

### LINKED ISSUE #957: Unhack our SSL cert

Investigate whether or not Heroku's free SSL certs for custom domains is going to work for us

Moving off of our hacky way of doing this to a service-level handling.

https://blog.heroku.com/announcing-automated-certificate-management

* **Anything else?**
No good deed goes unpunished, eh @DarthHater ?

DarthHater:

This allows for moving to just using Heroku's SSL stuff now that it is released

* Removes platform api dependency
* Removes letsencrypt middleware (not needed)

* Fixes #957

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,6 @@ gem 'js-routes'
 gem "omniauth-google-oauth2", "~> 0.2.1"
 gem 'rack-attack', '~> 5.0.1'
 gem 'rack-test', '~> 0.6.3', require: 'rack/test'
-gem 'platform-api', git: 'https://github.com/jalada/platform-api', branch: 'master'
 gem 'clinic_finder', '~> 0.0.1'
 gem 'geokit'
 
@@ -74,5 +73,4 @@ end
 group :production do
   gem 'puma'
   gem 'rails_12factor'
-  gem 'letsencrypt-rails-heroku'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,16 +1,6 @@
-GIT
-  remote: https://github.com/jalada/platform-api
-  revision: 45ddb3c1a7e2c7f85d979c0791db18e99affb237
-  branch: master
-  specs:
-    platform-api (0.8.0)
-      heroics (~> 0.0.17)
-
 GEM
   remote: https://rubygems.org/
   specs:
-    acme-client (0.4.1)
-      faraday (~> 0.9, >= 0.9.1)
     actionmailer (4.2.7.1)
       actionpack (= 4.2.7.1)
       actionview (= 4.2.7.1)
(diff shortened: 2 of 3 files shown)


colinxfleming:
Starting this process now in sandbox. DOES IT BLEND?

colinxfleming:
Heroku docs: https://blog.heroku.com/announcing-automated-certificate-management
Original implementation: #829 

What I did: 

* pushed this branch to sandbox
* ran `heroku authorizations` and removed ExpeditedSSL, and multiple LetsEncrypt entries (nb which may have wrecked things)
* removed the `ACME_*`, `HEROKU_TOKEN`, `HEROKU_APP`, env vars
* removed the scheduled renew task
* confirmed that this did not change the estimated bill for staging
* ran `heroku certs:auto:enable`
* ran `heroku certs:auto`, which started with a status of 'Waiting' and stayed that way for awhile, then went to (Heroku says the process takes ~15m).
* confirmed that we still had a valid SSL cert

(which it turns out we didn't). 

I'll try and resolve tomorrow. In the meantime, I alerted the CMDs to the sandbox outage.

colinxfleming:
Steps I took to restoring ssl in sandbox:

* `heroku rollback vXXX`
* checked that the env vars were still there
* `heroku certs:auto:disable` (rollback doesn't revert these configs)
* `heroku authorizations:create -d "LetsEncrypt"` to create a new token; pasted that into the env var
* `heroku run rake letsencrypt:renew` to make a new ssl cert
* set the autorenew job back up (`if [ "$(date +%d)" = 01 ]; then rake letsencrypt:renew; fi`) in the scheduler
* check that SSL cert is back (it is)
* check that `heroku authorizations` lists the letsencrypt
* check that `heroku certs:auto` shows up disabled
* made sure that git master did not have those changes we were experimenting with (this involved a reset to ab6eab3 for us)
* `git push heroku master -f` (yolo)
* redeployed master from circleci, just in case

We're back up to speed now: 
image

colinxfleming:
A few interesting quirks here:

* turning on `certs:auto:enable` doesn't seem to do anything meaningful. I wonder if there's a separate box upgrade we have to do
* there doesn't seem to be a terribly clear way to point things at a specific SSL certificate using this paradigm
* before we start this process in the future, we should make sure the SSL cert expiration date changes. That was how I noticed that the old one was still in place after `certs:auto:enable`, and we'd be in a world of hurt when the certs expired had we not done that

DarthHater:
Frustrating, were we able to open a support ticket or anything akin to see why it failed? As far as I can tell from their docs, enabling the heroku task takes care of everything for you (probably running something akin to what we are currently running ourselves.

colinxfleming:
Going to take another stab at this when I get a chance soon -- there was a decent amount of consternation caused by the `platform-api` gem this wknd, would love to shoot this requirement into space as soon as possible!

Issue #960, opened 2017-03-23
Account for clinic costs in clinic model


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Account for clinic costs by week, as individual clinics. 

* **What feature or behavior is this required for?**
Clinic-patient pairing / clinic_finder functionality.

* **How could we solve this issue? (Not knowing is okay!)**
* Individual fields on clinic model for every week LMP from 5 to 30. 
* Representing it in `/clinics`, dirty.
* Representing it in the `clinic` form.

* **Anything else?**
We're punting on anesthesia or anything even slightly complicated for now. Just pretend that everything has one type of abortion and it's a single cost.

colinxfleming:
Resolved by #961 and #977, so we'll close when #977 gets in

colinxfleming:
#977 is in! Closing

PR #961, opened 2017-03-24
Clinic Costs now By Week

lomky:

First hash at the Clinics having per-week costs

  - The model now has fields for cost from week 5 through week 30
  - The view has a basic display of all 26 costs
  - The edit now has a very dirty set of fields for entering these

Bumps #960


app/models/clinic.rb
```diff
@@ -17,13 +17,33 @@ class Clinic
   field :active, type: Boolean, default: true
   field :accepts_naf, type: Boolean, default: false
   field :gestational_limit, type: Integer
-  field :costs_9wks, type: Integer
+  field :costs_05wks, type: Integer
```
colinxfleming: I think for ease of translation, can we strip the leading zeroes here? (Unless there's a reason for it I'm forgetting, in which case let's document it here!)

app/models/clinic.rb
```diff
@@ -17,13 +17,33 @@ class Clinic
...
+  field :costs_06wks, type: Integer
+  field :costs_07wks, type: Integer
```
colinxfleming: I wonder if we can do this with a range of some sort -- I THINK, though I haven't tried it, that mongoid will let you dynamically define model fields, since it's all ruby code anyway. Maybe something like `(5..30).each { |i| field "costs_#{i}wks".to_sym, type: Integer }`?

app/views/clinics/_clinic_form.html
```diff
@@ -12,13 +12,34 @@
       </div>
       <div class="info-form-right col-sm-6">
         <%= f.number_field :gestational_limit, autocomplete: 'off' %>
-        <%= f.number_field :costs_9wks, autocomplete: 'off' %>
+        <%= f.number_field :costs_05wks, autocomplete: 'off' %>
```
colinxfleming: This we can almost definitely do with a range. Should make it a little more maintainable and do future us a favor! Thoughts?

app/views/clinics/index.html.erb
```diff
@@ -29,21 +29,63 @@
...
-            <label>Clinic Costs</label>
+            <label>Clinic Costs by Week</label>
             <table class="clinic-costs-table table">
               <thead>
-                <th>9 Weeks</th>
-                <th>12 Weeks</th>
-                <th>18 Weeks</th>
-                <th>24 Weeks</th>
-                <th>30 Weeks</th>
+                <th>05 </th>
```
colinxfleming: No surprises about what I'd suggest here, haha

app/models/clinic.rb
lomky: The only reason was making it look nice and lined up. I have no preference, @Kevin-Wei liked it better with the leading zero.

colinxfleming: Let's go without leading zeroes, if that's okay with you. Happy to handle since this is my own pedantic junk, if you prefer.

lomky: Naw, I got it! I like putting polish on code. :1st_place_medal:

Issue #841, opened 2017-02-09
Resource guide


lwaldsc:
***Currently our resource folder also includes the budget tracker and pledge form, so those features should be completed before we can migrate everything.

**Essential information we need**
Clinic contact information/price guide 
Other local funds' contact information and eligibility
Other social resources contact information (could be supported in a follow up program) 
LMP calculator (could be supported in another feature in DARIA) 
Suggestions for fundraising
Call script
How to fax document (could be included in the pledge box?)
Policies and procedures
Meghan's weekly email content (separated into "Basics" and "New Information")

**Where does it live? How do I get to it?** 
Pop-out new window/tab that houses a mini-wiki
Links in ? helper buttons near certain features to related page (Clinic details would link to the list of clinics; additional funding would link to the list of funds, etc)
Ideally, the New Information would be housed on the home page of the wiki as a reminder to CMs who didn't read the email.

**Wiki tabs**
Home
Basics
Clinics
Funds
Resources
LMP Calculator
Fundraising
Script
Hard Cases
Policies
History
(Using anchor links to go to specific information under each tab.) 

**How to keep it updated? What changes a lot, what doesn't? (Jekyll in github/CRM/links to PDF?)**
Girl I have no idea. Information does not change that often. Parts of Meghan's weekly email does, alerting CMs to any new information or changes and who is on the lines this week.

**Relationship to other DCAF resources**
Currently in googledrive/crypto but hoping to get rid of both

mebates:
I see in #962 that we added this to the top nav. Since this opens a new tab outside of the app, I might consider adding it to the footer instead. Thoughts @lwaldsc ?

colinxfleming:
ping @lwaldsc  -- if you think it's good at the top go ahead and hit close, otherwise let's figure out where it should live.

colinxfleming:
Area Man Mashes Wrong Button

lwaldsc:
Honestly I would leave it on the top so it's easier for CMs to find. I know you're the expert Mollie but I know there will be a bunch of "where is the link" kind of questions soooooooooo

PR #962, opened 2017-03-25
Add link to CM resources

### LINKED ISSUE #658: What cheatsheets should we port over from Crypto?
Brought up in #614 -- There are a few valuable reference sheets that still live in crypto, so as part of our effort to completely get out of crypto by the end of the year, it would be awesome to port these over. 

@NerdyGirl537 and @lwaldsc can you log the types of information in there, if there's anything besides clinic info and contact for other funds we work with?

colinxfleming:

The CM team has a locked down google drive directory with some fund resources that CMs occasionally reference. As the most straightforward and maintainable path that doesn't involve checking semisensitive information into version control, this adds a link to that resource.

In the abstract, this lets funds utilize their own document storage systems for resources, whatever those are (as long as they're web browser accessible).

Note that this expects the funds to handle their own document security -- it's fund's problems to make sure that, for example, the google drive account has the proper permissions set, etc.

* Adds an env var, `CM_RESOURCES_URL`
* Links to that env var url in the top navbar

screen shot 2017-03-25 at 12 51 49 pm

* Fixes #658 
* Bumps #841


DarthHater:
+1 story checks out. No comments, looks like a standard approach.

colinxfleming:
Added the resource to prod just now. Didn't have it in time for the deployment today so this will go out in two weeks (though I'll merge tomorrow).

colinxfleming:
Failure here is for a BS reason (circle using an old build despite the code being up to speed, geez). I'm merging.

Issue #903, opened 2017-03-04
Failing integration test


zorkian:
* **What are we trying to do?**

Submit documentation update, and get the integration tests to pass. Looks like `test/integration/pledge_fulfillment_test.rb` has a failing test: `test_0002_should show a link to the pledge fulfillment tab after pledge sent - PledgeFulfillmentTest::visiting the edit patient view as a data volunteer`.

* **What feature or behavior is this required for?**

Onboarding and submitting PRs etc.

* **How could we solve this issue? (Not knowing is okay!)**

Make the test pass.

* **Anything else?**

Not this second!

zorkian:
Link to a failure: https://circleci.com/gh/colinxfleming/dcaf_case_management/1731?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

zorkian:
Failing output:

```
Expected false to be truthy.
test/integration/pledge_fulfillment_test.rb:90:in `block (2 levels) in <class:PledgeFulfillmentTest>'
```

lilliealbert:
I got one locally, but when I rerun with the same seed, I get a different failure 👻  :sob:


### failure No 1:
```
lillie@lemur-8 dcaf_case_management (master) $ rake
Started with run options --seed 23038

    HTML screenshot: /Users/lillie/projects/dcaf_case_management/tmp/capybara/screenshot_2017-03-04-11-05-34.178.htmlETA: 00:02:07
    Image screenshot: /Users/lillie/projects/dcaf_case_management/tmp/capybara/screenshot_2017-03-04-11-05-34.178.png
 FAIL["test_0001_should let you mark a pledge submitted", #<Class:0x007feb0d5b0888>, 102.41897805500776]
 test_0001_should let you mark a pledge submitted#SubmitPledgeTest::submitting a pledge (102.42s)
        Expected false to be truthy.
        test/integration/submit_pledge_test.rb:37:in `block (2 levels) in <class:SubmitPledgeTest>'

58bb105df47575191fbc7f51,58bb105df47575191fbc7f56,58bb105df47575191fbc7f4c,58bb105df47575191fbc7f47 Time: 00:03:16,  ETA: 00:00:08
  328/328: [=================================================================================] 100% Time: 00:03:23, Time: 00:03:23

Finished in 203.84374s
328 tests, 692 assertions, 1 failures, 0 errors, 0 skips
```

### failure No 2:
```
[...]

PR #963, opened 2017-03-25
Upgrade phantomJS in build

colinxfleming:

Version-locks phantomjs in CI and in docker to 2.1.x.

No significant issues or changes, though I guess it's a nudge at #903 in a way.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Colin Fleming <c3flemin@gmail.com>
 # configure environment variable
 # note: move this to three ARG commands when CircleCI updates their docker
 ENV DCAF_DIR=/usr/src/app \
-    BUILD_DEPENDENCIES="build-base libxml2-dev libxslt-dev linux-headers bash git openssh" \
+    BUILD_DEPENDENCIES="build-base libxml2-dev libxslt-dev linux-headers bash git openssh fontconfig" \
     APP_DEPENDENCIES="nodejs"
 
 # get our gem house in order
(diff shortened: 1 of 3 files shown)


colinxfleming:
This winds up being pretty surgical (all CI or dockerfile changes) but I think this gets us there. @DarthHater can you review?

Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN apk update && apk upgrade && \
 RUN ln -s `which nodejs` /usr/bin/node
 
 # install phantomjs
-RUN npm install -g phantomjs-prebuilt
+RUN npm install -g phantomjs-prebuilt@2.1.11
```
DarthHater: This is so odd to me. I think in theory you should use 2.1.14 instead of 2.1.11 (may as well use the newest, plus uh, 2.1.11 isn't even a tag on their repo?)

colinxfleming: I think you're right about this. I'll bump to 2.1.14.

circle.yml
```diff
@@ -1,6 +1,9 @@
 machine:
   ruby:
     version: 2.4.1
+  pre:
+    - sudo apt-get install libfontconfig
+    - sudo curl --output /usr/local/bin/phantomjs https://s3.amazonaws.com/circle-downloads/phantomjs-2.1.1
```
DarthHater: TECHNICALLY this is a bit odd since you use prebuilt in Docker but then use the standard phantomjs here. Practically it probably doesn't matter, but worth pointing out. 

Also curious if https://github.com/Medium/phantomjs#continuous-integration instructions on caching can be applied for CircleCI.

colinxfleming: Yeah, this is an example of copy and paste from stackoverflow (or in this case, a CircleCI help thread). Worth keeping an eye on but probably not worth stressing.

PR #969, opened 2017-03-29
Fix modals with Turbolinks UPGRAYEDD

### LINKED ISSUE #968: Pledge modal popup not loading properly

Hitting submit pledge on a patient pops up a modal but doesn't load the modal content, including the buttons which we need to ship a pledge. If you reload and then click submit pledge again, the modal seems to render fine.

Pretty core feature, submitting a pledge!

Not sure. I suspect that this is an asset delivery issue and probably involves something we did in #955 , since javascript console logs seem to be coming back cleanly. 

* **Anything else?**
This is reproducible in the sandbox env as follows:
- log in
- go to the patient 'Clinic and Appt - 2'
- hit submit pledge'
- observe the following, no working buttons

screen shot 2017-03-29 at 10 23 19 am

- reload the page and click 'submit pledge' again, and then see the working buttons

DarthHater:

Fixes issue with modals not loading on initial page load, follows top answer from: https://stackoverflow.com/questions/17317816/rails-javascript-only-works-after-reload

Of note, there are other spots in the code base where it seems this could be applied, this PR for now only updates the pledges modal

* Change to the event that the modals use from page:load to the new turbolinks:load

* Fixes #968

### DIFF
--- a/app/assets/javascripts/external_pledges.coffee
+++ b/app/assets/javascripts/external_pledges.coffee
@@ -4,6 +4,6 @@
 
 jQuery ->
   # use event that works with rails turbolinks
-  $(document).on 'ready page:load', ->
+  $(document).on 'ready turbolinks:load', ->
     # select all elements with id ending with "modal"
     $('[id$=modal]').modalSteps()


colinxfleming:
:shipit:

colinxfleming:
Gonna close this out so we can address it more holistically down the line as part of a more full resolution of #970  . Thanks again for hacking on this with me @DarthHater !

PR #971, opened 2017-03-30
Enforce uniqueness on Clinic#name

colinxfleming:

Hotfixes an issue that was allowing duplicate clinics. Also adds fax number.

* Require uniqueness on name for clinics
* Add fax

No issue #

### DIFF
--- a/app/models/clinic.rb
+++ b/app/models/clinic.rb
@@ -14,6 +14,7 @@ class Clinic
   field :state, type: String
   field :zip, type: String
   field :phone, type: String
+  field :fax, type: String
   field :active, type: Boolean, default: true
   field :accepts_naf, type: Boolean, default: false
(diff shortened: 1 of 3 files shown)


app/models/clinic.rb
```diff
@@ -22,6 +23,7 @@ class Clinic
 
   # Validations
   validates :name, :street_address, :city, :state, :zip, presence: true
+  validates :name, uniqueness: true
```
DarthHater: Minor, and you can roll with the way you have it but you can also use: `validates_uniqueness_of :name` which is more descriptive and you don't have to pass a param too.

colinxfleming: Believe it or not the rubocop styleguide prefers it this way! Going to keep it this way, at least until they change the style guide (which, most likely, they will :( )

PR #974, opened 2017-04-01
fix tests to allow for new interface plus type changes

FeminismIsAwesome:

* (your change here)
* (another change here)
* (etc)

* Fixes #X
* Bumps #Y

### DIFF
--- a/app/assets/javascripts/patients.coffee
+++ b/app/assets/javascripts/patients.coffee
@@ -37,6 +37,14 @@ ready = ->
   $(document).on "change", "#abortion-information-form input, #external_pledges input", ->
     updateBalance()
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/patients.coffee
```diff
@@ -37,6 +37,14 @@ ready = ->
   $(document).on "change", "#abortion-information-form input, #external_pledges input", ->
     updateBalance()
 
+  $(document).on "submit", "#generate-pledge-form form", ->
+    if($("#case_manager_name").val())
+      true
+    else
+      $("#generate-pledge-form .alert").show()
```
colinxfleming: Really, really love this -- very sharp of you to flag this and punch it into the workflow!

PR #975, opened 2017-04-01
Move clinics away from env vars to model

### LINKED ISSUE #973: Move clinics to model from env vars

Rather than referring to a comma separated list of clinics, we're trying to move everything to the clinic model.

Having the CM team able to update and maintain the list of clinics, instead of programmers doing it; also opens up a bunch of other clinic-related cool stuff we can do.

There are a couple of things here:
* make a patient `has_one` clinic, which refers to the clinic model instead of a string
* in the patient edit form, make the list of clinics populate from the model rather the env var
* change all references to `Patient#clinic_name` to `Patient.clinic`
* adjust seedfile
* get rid of the env var
* run a script that changes `Patient#clinic_name` to the appropriate clinic

* **Anything else?**
This is in place from the CMD perspective -- the clinic objects are all set up -- so this is ready to go. Would love to have this out as part of the deployment in 

This is the last step in #818 and #642

colinxfleming:

This takes remaining steps to move us away from clinics as defined by environment variables into clinics as defined by a model.

Turns out there are a lot of moving parts in those env vars.

<strike>TODO: null out clinic, fix tests</strike> (donezo!)

* Depreciates patient#clinic_name
* bumps seedfile
* establishes `clinic has_many patients` / `patient belongs_to clinic` relations
* adjusts a ton of stuff in support of that

No view changes, I hope!

* Fixes #973 

(s/o to past us for writing good tests)

### DIFF
--- a/app/models/pregnancy.rb
+++ b/app/models/pregnancy.rb
@@ -71,7 +71,7 @@ class Pregnancy
   def pledge_info_presence
     errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank') if dcaf_soft_pledge.blank?
     errors.add(:pledge_sent, 'Patient name cannot be blank') if patient.name.blank?
-    errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic_name.blank?
+    errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic.blank?
     errors.add(:pledge_sent, 'Appointment date cannot be blank') if patient.appointment_date.blank?
   end
 end
--- a/app/views/patients/data_entry.html.erb
+++ b/app/views/patients/data_entry.html.erb
@@ -40,7 +40,7 @@
 
   <%= f.text_field :age, autocomplete: 'off' %>
   <%= f.select :race_ethnicity, options_for_select(race_ethnicity_options), label: 'Race / Ethnicity', autocomplete: 'off' %>
-  <%= f.select :clinic_name, options_for_select(clinic_options) %>
+  <%= f.select :clinic_id, options_for_select(clinic_options) %>
 
   <%= f.text_field :appointment_date, label: 'Appointment date', autocomplete: 'off', format: '%Y-%m-%d', placeholder: 'YYYY-MM-DD' %>
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -29,7 +29,7 @@ user3 = User.create! name: 'testuser3', email: 'dcaf.testing@gmail.com',
 Clinic.create! name: 'Sample Clinic 1 - DC', street_address: '123 Fake Street',
                city: 'Washington', state: 'DC', zip: '20011'
 Clinic.create! name: 'Sample Clinic 2 - VA', street_address: '123 Fake Street',
-               city: 'Arlington', state: 'DC', zip: '22204'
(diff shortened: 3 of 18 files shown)


colinxfleming:
Tagging in @lomky for eyes!

app/views/patients/data_entry.html.erb
```diff
@@ -40,7 +40,7 @@
 
   <%= f.text_field :age, autocomplete: 'off' %>
   <%= f.select :race_ethnicity, options_for_select(race_ethnicity_options), label: 'Race / Ethnicity', autocomplete: 'off' %>
-  <%= f.select :clinic_name, options_for_select(clinic_options) %>
+  <%= f.select :clinic_id, options_for_select(clinic_options) %>
```
lomky: Why is this call different than the one above?

app/models/pregnancy.rb
```diff
@@ -71,7 +71,7 @@ def updating_pledge_sent?
   def pledge_info_presence
     errors.add(:pledge_sent, 'DCAF soft pledge field cannot be blank') if dcaf_soft_pledge.blank?
     errors.add(:pledge_sent, 'Patient name cannot be blank') if patient.name.blank?
-    errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic_name.blank?
+    errors.add(:pledge_sent, 'Clinic name cannot be blank') if patient.clinic.blank?
```
colinxfleming: Just checked to be absolutely sure and it's looking like yes: 

```
> a  = Patient.first
=> #<Patient _id: 58e00ae2311c884c34bedf98, ...>

> a.clinic
=> nil

> a.clinic.blank?
=> true
```

Thanks for flagging this!

db/seeds.rb
```diff
@@ -25,6 +25,12 @@
 user3 = User.create! name: 'testuser3', email: 'dcaf.testing@gmail.com',
                     password: 'P4ssword', password_confirmation: 'P4ssword'
 
+# Seed a pair of clinics, Sample 1 and Sample 2
+Clinic.create! name: 'Sample Clinic 1 - DC', street_address: '123 Fake Street',
+               city: 'Washington', state: 'DC', zip: '20011'
+Clinic.create! name: 'Sample Clinic 2 - VA', street_address: '123 Fake Street',
+               city: 'Arlington', state: 'DC', zip: '22204'
```
colinxfleming: world geography champion @colinxfleming . Fixed!

app/views/patients/data_entry.html.erb
colinxfleming: The select field `clinic_name` is looking for `patient#clinic_name`, while `clinic_id` is looking for `patient#clinic_id` (which, it turns out, is the field that gets quietly generated in the json when you set up a `belongs_to`, and activerecord is smart enough to translate `patient.clinic` into `Clinic.find(patient.clinic_id)`. 

I wound up having to use `clinic_id` here and not `clinic` because the latter does a lookup (as in, runs `Clinic.find(id)`) and chokes on nil values.

This post brought to you by learning this the hard way this afternoon 😑

colinxfleming:
kaboom! Last step is to write the transition script.

colinxfleming:
I wrote a transition script so we'll be rolling this out tomorrow.

PR #977, opened 2017-04-02
Clinic params

### LINKED ISSUE #967: Develop automated backup system

Twice a day, run a task that takes a db snapshot and securely stories it somewhere, encrypted.

Me sleeping at night; data recovery in case of catastrophic failure. (I'm taking backups manually now.)

No idea but I am highly suggestible!

* **Anything else?**

colinxfleming:

A CMD pointed out that some of the attributes on clinic weren't saving properly, which is because they weren't whitelisted. This jams everything into strong params before I forget, and makes a few edits to the form to make sure some additional attributes (namely, `fax` and `active`) that I forgot to put in the first time make it in.

TK need to check and make sure there aren't any tests I need to write. 

@lomky can you review at your leisure?

* Adds forgotten fields to strong params
* Adds forgotten fields to form

Fixes #967


colinxfleming:
`clinics.yml` is a piece of machinery that the `clinic_finder` gem currently depends on for demo purposes. No need to bump it, I don't think.

PR #986, opened 2017-04-08
add NAF filter for clinics

### LINKED ISSUE #979: Limit Clinic Selection to clinics that accept NAF for some PTs

We now have great data stored on clinics, and we want to reveal it to the CMs. Add a selector below 'Clinic details > Clinic Name' that says "Only show clinics that accept NAF funding". If the CM checks this box, limit the list to clinics that accept NAF.

tingaloo:

Adds a checkbox that DISABLES clinics that DO NOT accept NAF funding.

* Adds checkbox to filter NAF

screen shot 2017-04-08 at 7 38 26 pm

Clinics that do not define accepts_naf will be sorted as false automatically, in this case 'Sample Clinic 1' and 'Sample Clinic 2'. 'SAMPLE CLINIC without NAF' is disabled but still selected.
<br>

I made this more Ruby heavy than JS, doing all lookups and array manipulating in the patient helper. For now I disabled clinics because it is easier and I have no idea how many clinics are in production. If we want to remove them, I think we'd have to override some bootstrap styles because option tags.

Currently works in development, and should work with any array passed in.

* Fixes #979


colinxfleming:
MY DUDE. Thank you! I'm running around this evening and will review
tomorrow or monday.

colinxfleming:
This PR is a beaut. I think we can merge as is, I'll plan on getting it in this week. Thanks a ton for knocking this out so quickly.

colinxfleming:
whoops, I can't quite get CI to pass. I suspect it has something to do with the JS causing some kind of redirect -- I would **guess** that moving it away from `document` to something else will resolve. Not sure what that something else is off the top of my head but I'm sure there's an example you can crib off.

This test is probably the canary and I suspect when you get this to pass everything else will fall into place:

```
Expected response to be a <success>, but was <302>
test/controllers/patients_controller_test.rb:63:in `block (2 levels) in <class:PatientsControllerTest>'
```

Sorry for the false flag! (It does look good!)

tingaloo:
Actually not a false flag, and not a JS problem either.

This line fails with good reason (not sure why it didn't fail before).
```clinic_with_naf = array.map { |clinics| [clinics, :"data-naf" => Clinic.find_by(name: clinics).accepts_naf] }```

```Clinic.find_by``` currently doesn't have enough time to fetch more than 2 clinics. So it kind of looks like this.

```
MONGODB | localhost:27017 | dcaf_case_management_development.find | STARTED | {"find"=>"clinics", "filter"=>{"name"=>"Sample Clinic 1 - DC"}}
MONGODB | localhost:27017 | dcaf_case_management_development.find | SUCCEEDED | 0.001824s
#<Clinic:0x007fc02bf702b8>
MONGODB | localhost:27017 | dcaf_case_management_development.find | STARTED | {"find"=>"clinics", "filter"=>{"name"=>"Sample Clinic 2 - VA"}}
MONGODB | localhost:27017 | dcaf_case_management_development.find | SUCCEEDED | 0.001461s
#<Clinic:0x007fc0320d3870>
MONGODB | localhost:27017 | dcaf_case_management_development.find | STARTED | {"find"=>"clinics", "filter"=>{"name"=>"Sample Clinic with NAF"}}
MONGODB | localhost:27017 | dcaf_case_management_development.find | SUCCEEDED | 0.00132s
  Rendered patients/_abortion_information.html.erb (15.2ms)
  Rendered patients/edit.html.erb within layouts/application (111.6ms)
Redirected to http://localhost:3000/
Completed 302 Found in 178ms
```
Basically if there's more than 2 clinics, the page will load before Clinic.find_by gets a chance to fetch. I'll rewrite this in JS since it's frontend and async.

**Comment by colinxfleming** [comment]
Ah, i see. We are currently in the process of moving away from a list of
clinics as an env var to a proper clinic object system. There's an open pr
moving over the rest ofthe way that might be helpful to refer to.

**Comment by colinxfleming** [comment]
( #975 is that PR)

**Comment by tingaloo** [comment]
Not really fond of jquery ajax and coffeescript, but I got it to work.

Patients javascript is quite bloated now. I added some routes for my JS calls because I'm not sure how to call helper methods using AJAX, so the code can be a bit janky.
 
I wrote around this line from #975  
```Clinic.where(active: true).map { |clinic| [clinic.name, clinic.id] }```
Overall it should be good to go.

File: app/controllers/patients_controller.rb [hunk 1]
```diff
@@ -56,6 +56,17 @@ def data_entry_create # temporary
       render 'data_entry'
     end
   end
+
+
+  def fetch_clinic_naf
```
colinxfleming: What are your thoughts about moving this to the `clinics` controller? Can rig it up something like `ClinicsController#naf_status`?

app/controllers/patients_controller.rb
```diff
@@ -56,6 +56,17 @@ def data_entry_create # temporary
...
+    accepts_naf =  Clinic.find_by(id: params['id']).accepts_naf
```
colinxfleming: You can just use `Clinic.find(params['id'])` here, no need to declare the field. A nice little rails-ism!

app/controllers/patients_controller.rb
```diff
@@ -56,6 +56,17 @@ def data_entry_create # temporary
...
+    render json: [params['id'], accepts_naf]
+  end
+
+  def get_clinics
```
colinxfleming: Same deal -- lets move this to the clinics controller. Might make a good thing to return from the index route if you're requesting JSON?

config/routes.rb
```diff
@@ -10,6 +10,8 @@
       resources :notes, only: [ :create, :update ]
       resources :external_pledges, only: [ :create, :update, :destroy ]
     end
+    get 'get_clinics', to: 'patients#get_clinics', as: 'get_clinics'
```
colinxfleming: Per above, I think these might be better suited for the clinics controller. What are your thoughts?

app/assets/javascripts/patients.coffee
```diff
@@ -22,6 +21,39 @@ calculateRemainder = ->
 valueToNumber = (val) ->
   +val || 0
 
+getClinics = ->
```
colinxfleming: We have a clinics coffeescript file, no? Let's move stuff pertaining to this feature over there, since as you very correctly pointed out, the patients JS is getting a bit massive.

tingaloo:
All tidied up! Moving the logic to `Clinics` cleaned A LOT of the logic in `Patients`. Also fetching all the clinics at once via `index` made things much more simple. I got tunnel vision working on this the first time, now it looks way better.

Only thing that's left is an ajax integration test, I tried writing [a test](https://github.com/colinxfleming/dcaf_case_management/pull/986/commits/bcced15a14e9072854241f41fb1cb4fdea8487a4) but it kept failing. I'll give it another shot at the next hacknight.

Happy Easter!

test/controllers/clinics_controller_test.rb
```diff
@@ -21,6 +21,16 @@ class ClinicsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  describe 'naf_status' do
+    before {
```
colinxfleming: You can simplify this a bit -- `create` lets you pass in a hash of stuff as an argument, and it'll set the object attributes accordingly. I think something like `before { @clinic = create :clinic, accepts_naf: true }` should fly.

config/routes.rb
```diff
@@ -8,19 +8,22 @@
...
-      resources :notes, 
+      resources :notes,
                 only: [ :create, :update ]
-      resources :external_pledges, 
+      resources :external_pledges,
                 only: [ :create, :update, :destroy ]
     end
 
+    get 'naf_status', to: 'clinics#naf_status', as: 'naf_status'
+    get 'clinics', to: 'clinics#index', as: 'clinics'
```
colinxfleming: I think we already have this route set as part of `resources :clinics` -- I think we can just use that. (We should keep `naf_status` though for sure!)

app/controllers/clinics_controller.rb
```diff
@@ -5,6 +5,7 @@ class ClinicsController < ApplicationController
 
   def index
     @clinics = Clinic.all
+    render json: @clinics.to_json
```
colinxfleming: To make it a bit clearer that this endpoint can be either json or HTML, let's throw this in a `respond_to` block. Something like:

```ruby
respond_to do |format|
  format.html
  format.json { render @clinics.to_json }
end
```

config/routes.rb
```diff
@@ -8,19 +8,22 @@
...
                 only: [ :create, :destroy ]
-      resources :notes, 
+      resources :notes,
                 only: [ :create, :update ]
-      resources :external_pledges, 
+      resources :external_pledges,
                 only: [ :create, :update, :destroy ]
     end
 
+    get 'naf_status', to: 'clinics#naf_status', as: 'naf_status'
```
colinxfleming: Do we need this endpoint? Javascript is admittedly not my strong suit but it doesn't look like it's getting called anymore. Thoughts?

test/controllers/clinics_controller_test.rb
```diff
@@ -21,6 +21,16 @@ class ClinicsControllerTest < ActionDispatch::IntegrationTest
...
+      @clinic = create :clinic
+      @clinic[:accepts_naf] = true
+     }
+    it 'should render true' do
```
colinxfleming: HMU if you have a tough time with this -- rails 5 changed how they did controller tests. http://guides.rubyonrails.org/testing.html#functional-tests-for-your-controllers is the documentation but I spent a lot of quality time with this recently and it's fresh in my mind.

test/controllers/clinics_controller_test.rb
```diff
@@ -21,6 +21,16 @@ class ClinicsControllerTest < ActionDispatch::IntegrationTest
     end
```
colinxfleming: An additional test in this controller to confirm that it responds to `index` as json would be great! Just a ping to confirm that the endpoint returns JSON data roughly in the format you're consuming.

tingaloo:
I'd say it's done! Made changes and removed the naf_status function.

Removed the capybara test because the CSS attribute never returned anything useful. Maybe somebody with more capy exp can give it a go.

PR #995, opened 2017-04-14
Removed 12factor, no longer needed

### LINKED ISSUE #988: Can we remove the rails_12factor gem?

Rails 5 can do all the 12 factor app stuff by default, rather than requiring an external library. With some configuration.

Keeping the gemfile clean!

* Remove the `rails_12factor` gem
* take the steps in https://github.com/heroku/rails_12factor#migrating-to-rails-5

* **Anything else?**
Stems from #865

DarthHater:

Removes 12factor, no longer needed in Rails 5.

* Removes 12 factor gem

Testing for this needs to mainly be on Heroku, since it relies on ENV variables that Heroku provides. Code changes are minimal. 12 Monkeys is a good Terry Gilliam movie.

* Fixes #988

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,6 +73,5 @@ end
 
 group :production do
   gem 'puma'
-  gem 'rails_12factor'
   gem 'letsencrypt-rails-heroku'
 end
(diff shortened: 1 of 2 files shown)


colinxfleming:
Made an associate env var change in staging and prod, I think we can declare victory here. Thanks again for taking care of this @DarthHater !

Issue #860, opened 2017-02-19
Standardize dates as MM-DD-YYYY


colinxfleming:
* **What are we trying to do?**
As much as it pains me to admit this, I am the only person, possibly in the universe, who wants to set all dates as YYYY-MM-DD. Let's please set our dates as MM-DD-YYYY from here out.

* **What feature or behavior is this required for?**
Not driving our CMs up the wall by having different date formats in different places.

* **How should we solve this issue?**
A few things:
* crack open `config/initializers/time.rb` and change our global convenience method for `display_time` to be MM-DD-YYYY
* on the dashboards page, edit the `appointment date` field so it shows up properly formatted
* in the new patient partial, edit `initial call date` so it shows up properly formatted and with no datepicker
* on the patient dashboard partial, change the appointment date and `called on` dates to proper format
* on the pledge fulfillment partial, take away the calendar icon for procedure date and standardize date of check and procedure date to MM-DD-YYYY
* fix up any tests this breaks

If this turns out to be a huge pain in the behind (since mongoid may be very strict about that date format), just settle everything on YYYY-MM-DD for now and we'll figure out a longer term strategy later. i want to emphasize that this is not worth sinking a ton of time into, so as soon as you encounter significant resistance, scope it down to just getting everything formatted the same. 

* **Anything else?**
Stems from discussion in #660

lomky:
I started work on this, but ran into issues with seeding the database that revealed a quiet existing bug. That bug is fixed in PR #862, not sure I can finish fixing this issue without the fix in that PR.

lomky:
I took an initial stab at this, not even worth PRing yet. Something is going wonky when you format dates as MM-DD-YYYY and then try to generate them via Date methods (validating fails). Might be simple to fix for someone more familiar with Dates in ruby & rails. 

 - :exclamation:  This is open for others to grab, unless I get back to it first :slightly_smiling_face: 

The [branch here](https://github.com/lomky/dcaf_case_management/tree/860_standardize_dates) has:

* instances of `YYYY-MM-DD` changed to `MM-DD-YYYY`
* instances of `%Y-%m-%d` changed to `%m-%d-%Y`
* tests with explicit dates updated (i.e. 2016-03-29 is now 03-29-2016)

Currently failing to populate the DB with validation errors (Post PR #862 seed `!` fixes - otherwise fails silently) and throwing errors on display due to Date manipulation issues in `last_menstrual_period_display_short` - and maybe other issues.

[...]

PR #1002, opened 2017-04-15
Unbreak tests

colinxfleming:

The clinics PR we just merged has some rails 5 issue that's breaking the data entry form. This is a WIP placeholder to fix master. 

Exact test is something like 'undefined method '>' for nilclass' on L274 of Patient model.

### DIFF
--- a/test/integration/data_entry_test.rb
+++ b/test/integration/data_entry_test.rb
@@ -16,7 +16,7 @@ class DataEntryTest < ActionDispatch::IntegrationTest
     before do
       # fill out the form
       select 'DC', from: 'patient_line'
-      fill_in 'Initial call date', with: 2.days.ago.strftime('%m/%d/%y')
+      fill_in 'Initial call date', with: 2.days.ago.strftime('%Y-%m-%d')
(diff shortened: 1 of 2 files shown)


colinxfleming:
Rails made a very microscopic change to how they handle date inputs, which made the data entry form choke. This fixes that now. Sorry everyone. 

@DarthHater for review

test/integration/data_entry_test.rb
```diff
@@ -16,7 +16,7 @@ class DataEntryTest < ActionDispatch::IntegrationTest
     before do
       # fill out the form
       select 'DC', from: 'patient_line'
-      fill_in 'Initial call date', with: 2.days.ago.strftime('%m/%d/%y')
+      fill_in 'Initial call date', with: 2.days.ago.strftime('%Y-%m-%d')
```
DarthHater: Is there a reason you switched the format here?

colinxfleming: Yeah -- if we have this as a text field instead of a date field, rails is looking for something in this format and it'll swallow it otherwise (or set it to 1900-01-01, I forget which). See #860 for the longer running issue on this.

Issue #965, opened 2017-03-28
Add 'download anonymized CSV of patients' functionality


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Create a 'download patient data to CSV' endpoint.

* **What feature or behavior is this required for?**
Letting non-programmer data analysts, of which DCAF has a decent amount, poke at our data in excel. I was against this for a long time for security reasons but I think that the benefit of having this in place will become too big to not implement something like this, especially as other funds start to explore it.

* **How could we solve this issue? (Not knowing is okay!)**
The MVP of this is setting up an endpoint restricted to admins/data volunteers that dumps our entire patient list after obscuring some information. Here's a technical list of stuff:
* `patients` index route which responds to csv
* csv should be one line per patient
* should have object bson id and `identifier` field, but not name/phone
* should have a boolean for 'has alt contact' instead of alt name/alt phone/alt relationship (e.g. 'has alternate contact')
* no notes

* **Anything else?**
nope!

lomky:
A thought: Let's make sure that all fields being exported in the CSV have to be specified, rather than filtering out sensitive ones. A whitelist of fields that are okay for the CSV export.

That way, if we add any new, sensitive fields to patients down the road we won't have to remember _then_ to exclude it, possibly putting privacy at risk if we forget.

lwaldsc:
Agreed 1000%

colinxfleming:
Marking this active. This is something that would open up a lot of analysis potential for DCAF and I think we should make it a priority after we sort out merging the patient and pregnancy models. If someone wants to work on a proof of concept with just `patient` fields, that'd be a great starting place.

cfinucane:
This seems as good a place as any to dive in, so I'll try doing a PoC this weekend.

cfinucane:
OK, I've attached an initial PR with a proof of concept like Colin described.  Aside from any technical issues in my code, it seems like some next steps would be:

- Deciding on what exactly goes in the whitelist
- Thinking about what exactly we mean by "anonymous."  I'm not sure if I'm just overthinking this, but making records truly un-linkable to individual people can be trickier than it might seem (see, e.g. https://crypto.stanford.edu/~pgolle/papers/census.pdf)

Anyways, hope everyone's having a good weekend :)

lwaldsc:
[...]

PR #1004, opened 2017-04-16
Patient CSV export

cfinucane:

Here's my bare-bones first take at the patient data CSV export functionality described in #965.
The things I'm most unsure about are whether I put the export configuration in the best place, and the tests.

* add the notion of a `has_data_access` composite pseudo-role (I just made that word up)
* add the `has_alt_contact?` calculated boolean value to Patient
* add a CSV-only index for Patient

* Begins work on #965


app/views/patients/index.csv.erb
```diff
@@ -0,0 +1,8 @@
+<%=
```
colinxfleming: I think rendering as CSV is generally handled by the `respond_to`, and we just want to have this look like a regular-rails template -- no need for the CSV.generate in here. Instead, we can basically pretend like we're writing a regular HTML form by ditching the CSV generate and doing something like this:

```
<%= Patient::CSV_EXPORT_FIELDS.keys.join(',') %> # will render as `BSON ID, Identifier, Has alt contact?`
<% @patients.each do |patient| %>
  ...
<% end %>
```

app/views/patients/index.csv.erb
```diff
@@ -0,0 +1,8 @@
...
+CSV.generate do |csv|
+  csv << Patient::CSV_EXPORT_FIELDS.keys # Header line
+  @patients.each do |patient|
+    csv << Patient::CSV_EXPORT_FIELDS.values.map { |field| patient.send(field) }
```
colinxfleming: I deeply love `send` but it's a little risky, since it'll run whatever you feed it.

I think instead we should use hash attributes here -- a nice gift from our friends at rails is that you can get at object attributes by passing it a symbol, so we could do something like this -- 

```rb
Patient::CSV_EXPORT_FIELDS.values.map { |field| patient[field.to_sym] }
```

That might not help for `has_alt_contact?` though, I'm not sure if this works for methods. If that winds up being the case and we totally have to go with send (probably don't spend more than like 10m fighting with it), let's use the slightly safer version, `public_send` which will choke if we try a private method.

config/initializers/dependencies.rb
```diff
@@ -0,0 +1 @@
+require 'csv'
```
colinxfleming: Can probably kill this, since we're getting rid of `CSV.generate`? Thoughts?

test/controllers/patients_controller_test.rb
```diff
@@ -12,6 +15,65 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
     @clinic = create :clinic
   end
 
+  describe 'index method' do
+    before do
+      # Not sure if there is a better way to do this, but just calling
+      # `sign_in` a second time doesn't work as expected
+      delete destroy_user_session_path
```
colinxfleming: Oh word! Nah, this is exactly what you should do here.

test/controllers/patients_controller_test.rb
```diff
@@ -12,6 +15,65 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
...
+    end
+
+    it 'should reject users without access' do
+      sign_in @user
+
+      get patients_path
+      assert_redirected_to root_path
+
+      get patients_path + '.csv'
```
colinxfleming: I haven't had to do this yet, but I THINK the new way to do this is by strapping `as: :csv` at the end, instead of the string concat. So something like `get patients_path, as: :csv`. See http://api.rubyonrails.org/classes/ActionDispatch/IntegrationTest.html (control-f 'Changing the request encoding).

app/controllers/patients_controller.rb
```diff
@@ -1,9 +1,15 @@
 # Create and update patients, plus the main patient view in edit
 class PatientsController < ApplicationController
+  before_action :confirm_user_has_data_access, only: [:index]
   before_action :find_patient, only: [:edit, :update]
   rescue_from Mongoid::Errors::DocumentNotFound,
               with: -> { redirect_to root_path }
 
+  def index
+    @patients = Patient.all
+    respond_to :csv
```
colinxfleming: It's been a few rails versions since I've had to write this, but I think we can rewrite this line as follows to just have it download, using the `send_data` method (which is basically saying to rails 'download this instead of showing it in browser'):

```rb
respond_to do |format|
  format.csv { send_data @patients }
end
```

app/models/patient.rb
```diff
@@ -186,6 +190,10 @@ def save_identifier
     self.identifier = "#{line[0]}#{primary_phone[-5]}-#{primary_phone[-4..-1]}"
   end
 
+  def has_alt_contact?
+    !!(other_contact || other_phone || other_contact_relationship)
```
cfinucane: Yeah, I had a similar hesitation because the `!!` really confused me when I first encountered it myself, and it's a tough thing to search for, to boot :)  From my perspective using `blank?` had readability issues of its own, but I just found `present?` which I think will do the trick!

app/views/patients/index.csv.erb
cfinucane: The one upside to using the `CSV.generate` is we don't have to handle the special case of escaping commas by quoting any affected values.  This is the only special case I can think of, but I like using purpose-built libraries because we don't have to think about the edge cases.

cfinucane: I do agree that it's pretty silly to have a template that's just a single big code block, and wish there were a best-of-both-worlds solution...

app/views/patients/index.csv.erb
cfinucane: Oh, good call!  I'll try it out.

test/controllers/patients_controller_test.rb
cfinucane: Woo!  I tried making a `sign_out` method to parallel `sign_in` in `test_helper.rb`, but there's an existing one that is Capybara-specific and is paired with `log_in` :)

test/controllers/patients_controller_test.rb
cfinucane: Thanks! I was trying to find something like this but couldn't and gave up.  The pointer is much appreciated.

app/controllers/patients_controller.rb
cfinucane: Hm, I think `send_data` may only be for sending a binary stream (https://apidock.com/rails/ActionController/DataStreaming/send_data).  Is your main concern making sure we get the `Content-Disposition` header?

cfinucane: Actually, there's a CSV example on that page.   So basically we'd be bypassing the view entirely, which might make sense given how silly the view code is.  I think there's probably a subjective MVC separation call to make here.

cfinucane:
Awesome, thanks for your thoughtful comments!  When you say "add the fields on the patient object as a starter," do you really mean all of them?  If not, which ones?

config/initializers/dependencies.rb
cfinucane: See discussion above :)

app/views/patients/index.csv.erb
cfinucane: One alternative I had also considered is adding something like a `to_csv_entry` method on `Patient` itself that just returned a nice hash representing the given instance, thereby bypassing this entirely.  The downside though was that to print the header I'd have to do something kinda gross like get `@patients.first.to_csv_entry.keys`.

test/controllers/patients_controller_test.rb
colinxfleming: haha yeah, that's something I should have sorted out when we moved onto rails 5 and the controller test class changed. Sorry to make you pay for my sins here, I'll file a backlog ticket to fix this up.

app/controllers/patients_controller.rb
colinxfleming: I think given the patterns we're going to have elsewhere in the app (thinking of the PDF generation PR in particular, here), it kinda might make sense to cut out the view and just do the `CSV.generate` stuff in the patient model, and then `send_data` that? That would be consistent, at least. (Sorry for the back and forth on this!)

colinxfleming:
For the patient attributes list, I'm inclined to do `voicemail_preference` thru `special_circumstances`, except for `zip` and `city`, and work it down from there. I imagine that's stripped of easy to match stuff. Thoughts?

cfinucane:
OK, @colinxfleming, I've updated this PR to incorporate your comments.

I've also added a default filename for generated CSVs, to align with the pledge PDF generation, but we can leave that out if it seems like an unnecessary complication.  

I'd love to get any feedback people have on optimal formatting for columns, like `special_circumstances`.

Lastly, I think the CI checks are always going to show failures until this branch gets rebased :)

app/models/patient.rb
```diff
@@ -142,6 +166,24 @@ def self.contacted_since(datetime)
       first_contacts: first_contact.length, pledges_sent: 20 }
   end
 
+  def self.to_csv
+    CSV.generate(encoding: 'utf-8') do |csv|
+      csv << CSV_EXPORT_FIELDS.keys # Header line
+      all.each do |patient|
+        csv << CSV_EXPORT_FIELDS.values.map do |field|
+          value = patient.public_send(field)
```
colinxfleming: If I understand this correctly, we're checking the class of the field (e.g. string? array? int? decimal?) and if it's an array, stripping out blanks and then joining with commas and a space. That is pretty perfect!

I wonder if it's possible to write this just a tiny bit clearer -- would `value = patient[field.to_sym].class` fly here? I'll try that out and if it doesn't work out of the box we can roll with the `public_send`.

cfinucane: Oh, I should have mentioned that I tried that based on your earlier suggestion but it didn't seem to work for 'has_alt_contact?'

cfinucane: Otherwise, yes, your understanding is correct!  But maybe it would be more readable as something like:
```
if value.is_a?(Array)
  value.reject(&:blank?).join(', ')
else
  value
end
```
What do you think?

colinxfleming: Oh durf, the public send thing choking on has_alt_contact totally makes sense! My mistake, let's definitely go with `public_send` in that case.

re: `special_circumstances`, an unoptimized facet of the way we're saving them now is that we save a couple blank strings. It's an inconvenience but it's not causing any particular harm and we don't generate data that quickly, so it's on the backlog down a ways. But great eye noting that for sure, and great call working around it.

I really love the `is_a?` option you present here. It's really solid and crisp, and super clear what it's checking for! Would love to go with that, if it's cool with you.

cfinucane: Pushed :)

colinxfleming:
I think this is shaping up beautifully. 

@cfinucane I'd flag that we did a little restructuring of the patient object this week in #1013 -- this groups the many convenience methods we have on this object into relevant `concerns`. I think we can chuck `to_csv` and other new methods on the model into a concern called something like `Exportable` and then declare victory on this pull request. The `Searchable` concern in `app/models/concerns/searchable.rb` is probably the closest analog to your work here, so I'd recommend using that as a base.

After that I think we merge this and start building off it! Your thoughts @cfinucane ?

colinxfleming:
(Also just want to say that I really appreciate your being so collaborative in working thru these problems -- very happy to have you!)

colinxfleming:
#1014 is in! When you get a chance, go ahead and resolve the merge conflicts with a pull, and from there we can either merge as is or take the additional steps in #965 and then merge. Let me know if you run into any snags resolving conflicts, especially in the patient model.

app/models/patient.rb
```diff
@@ -22,6 +25,27 @@ class Patient
     scope line.downcase.to_sym, -> { where(:_line.in => [line]) }
   end
 
+  CSV_EXPORT_FIELDS = { "BSON ID" => :id,
```
colinxfleming: Let's chuck this in a concern, in `app/models/concerns/exportable.rb`, to match our new-ish patterns elsewhere.

app/models/patient.rb
```diff
@@ -1,4 +1,7 @@
 # Object representing core patient information and demographic data.
+
+require 'csv'
```
colinxfleming: This can go in the concern too.

app/models/patient.rb
```diff
@@ -134,10 +158,31 @@ def self.pledged_status_summary(num_days = 7)
     { pledged: outstanding_pledges, sent: sent_total }
   end
 
+  def self.to_csv
```
colinxfleming: This can go in the concern too (in a `class_methods` block, like in `app/models/concerns/callable.rb`)

app/models/patient.rb
```diff
@@ -134,10 +158,31 @@ def self.pledged_status_summary(num_days = 7)
...
+        end
+      end
+    end
+  end
+
   def save_identifier
     self.identifier = "#{line[0]}#{primary_phone[-5]}-#{primary_phone[-4..-1]}"
   end
 
+  def has_alt_contact?
```
colinxfleming: This can go in the concern too.

cfinucane:
OK, I've moved the code from the model to a new concern.   Let me know if this in line with what you were thinking.

app/models/concerns/exportable.rb
```diff
@@ -0,0 +1,56 @@
...
+    "Initial Call Date" => :initial_call_date,
+    "Urgent?" => :urgent_flag,
+    "Special Circumstances" => :special_circumstances 
+  }
+
+  def has_alt_contact?
+    other_contact.present? || other_phone.present? || other_contact_relationship.present?
+  end
+
+  def get_field_value_for_serialization(field)
```
colinxfleming: 👍 , beautifully clean breakout

Issue #997, opened 2017-04-14
Remove google analytics


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Get google analytics out of our app!

* **What feature or behavior is this required for?**
It should make things a little snappier, and we aren't really doing anything with it currently. So it's just unnecessary bulk.

* **How could we solve this issue? (Not knowing is okay!)**
Remove it from the assets; remove the attribute from our CSRF in `ApplicationController`; look around for other vestiges and get rid of them.

* **Anything else?**

colinxfleming:
Closed via #1010

PR #1010, opened 2017-04-18
Remove Google Analytics from vendor, CRSF and assets

prostko:
I rule and have completed some work on Google Analytics that's ready for review!

Removed Google analytics from the the Assets, vendor assets and the CSRF

* /app/assets/javascripts/application.js (removes require google_analytics)
* removes the 'script src' line from the CSRF
* removes vendor/assets/javascripts/google_analytics

* Fixes Remove google analytics #997

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,6 +28,7 @@ class ApplicationController < ActionController::Base
   def csp_headers
     response.headers['Content-Security-Policy-Report-Only'] =
       "default-src 'self'; " \
+      "script-src 'self' 'sha256-1kYydMhZjhS1eCkHYjBthAOfULylJjbss3YE6S2CGLc=' 'unsafe-eval'; " \
       "font-src 'self' fonts.gstatic.com; " \
       "style-src 'self' 'unsafe-inline'; " \
       'object-src; ' \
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -28,7 +28,6 @@ def prevent_caching_via_headers
   def csp_headers
     response.headers['Content-Security-Policy-Report-Only'] =
       "default-src 'self'; " \
-      "script-src 'self' www.google-analytics.com 'sha256-1kYydMhZjhS1eCkHYjBthAOfULylJjbss3YE6S2CGLc=' 'unsafe-eval'; " \
```
colinxfleming: I think we can just ditch the `www.google-analytics.com` part of this and keep the `script-src` stuff around it. Basically what this is doing is it's throwing a flag if any javascript loads that doesn't have the `sha` hash of `1kYydMhZjhS1eCkHYjBthAOfULylJjbss3YE6S2CGLc`. 

You can check out some more info on CSPs here: https://content-security-policy.com/

colinxfleming:
Looking great, thanks a ton @prostko ! Welcome to the team, we're glad to have you! 🍾

PR #1012, opened 2017-04-18
added dropoff functionality

### LINKED ISSUE #736: New status for non-response
Another new status -
"Patient dropoff"
120 days since last contact, PT automatically goes into that status. 
Should scramble PII in the same way "Resolved without DCAF" will.

maikelnabil:
helpers/status_helper.rb
I've added a new status in the STATUSES hash, then added a condition in the status method to show status dropoff if more_than_120 evaluates to true.

helpers/patients_helper.rb
I've added a method to the patients helper "duration_since_last_call" which returns more_than_120 if the last call was made 120 days or more.

* Adds a new status "dropoff" for patients who hasn't made contact for more than 120 days.

No changes to the views

I haven't tested this functionality. It should be tested before merging!

Fixes #736


app/helpers/patients_helper.rb
```diff
@@ -73,4 +73,11 @@ def dcaf_pledge_limit_help_text
 
     "Pledge Limit Guidelines:<br />1st trimester (7-12 weeks): $#{first_tri}<br />2nd trimester (12-24 weeks): $#{second_tri}<br />Later care (25+ weeks): $#{later_care}"
   end
+
+  def duration_since_last_call
+    @patient = Patient.find(params[:id])
```
colinxfleming: Rails is smart/magic enough to infer the `patient` here, so you can ditch this DB call and just have the relation below be `calls.last.created_at`.

app/helpers/patients_helper.rb
```diff
@@ -73,4 +73,11 @@ def dcaf_pledge_limit_help_text
...
+    days_since_last_call = (Time.now - @patient.calls.last.created_at).to_i / 86400
```
colinxfleming: Saying it again -- I really love how elegant this is. I would have written something way more complicated, but this nails it and is way nicer.

I think there are a few ways we can tighten this up a little bit -- interested in your thoughts:
* mongo tends to be spotty with maintaining orders of stuff. Can we do a `sort_by` on created at so we can be sure to pick the right last one?
* can we assign the 86400 to a variable (eg `DAY = 86400`), so we can be absolutely sure what the number refers to? This is burned into my brain too, but we can't assume that for everyone necessarily.

Have you written a test before? If so (or if you're interested in learning!) then this would be a great thing to write a test for in `test/helpers/status_helper_test.rb`, but if not, that's cool too!

app/helpers/patients_helper.rb
```diff
@@ -73,4 +73,11 @@ def dcaf_pledge_limit_help_text
...
+    if days_since_last_call => 120, return more_than_120
```
colinxfleming: We can just return `days_since_last_call` here, I don't think we need to overthink it and we can use this in a bunch of places.

app/helpers/status_helper.rb
```diff
@@ -6,14 +6,16 @@ module StatusHelper
...
+    resolved: 'Resolved Without DCAF',
+    dropoff: 'Patient Has dropped Off'
   }.freeze
 
   def status
     return STATUSES[:resolved] if pregnancy.resolved_without_dcaf?
     return STATUSES[:pledge_sent] if pregnancy.pledge_sent?
     return STATUSES[:no_contact] if not contact_made?
     return STATUSES[:fundraising] if appointment_date
+    return STATUSES[:dropoff] if more_than_120
```
colinxfleming: Can you swap `more_than_120` out here for `duration_since_last_call > 120`, and move this to the top of the `status` method? We want to check for this first in the sequence.

app/helpers/patients_helper.rb
```diff
@@ -73,4 +73,11 @@ def dcaf_pledge_limit_help_text
 
     "Pledge Limit Guidelines:<br />1st trimester (7-12 weeks): $#{first_tri}<br />2nd trimester (12-24 weeks): $#{second_tri}<br />Later care (25+ weeks): $#{later_care}"
   end
+
+  def duration_since_last_call
```
colinxfleming: Since we're going to ask for this to return days since last call as an int, should we rename it a little to reflect that? Maybe we straight up rename it `days_since_last_call`? (Note that we might have to rename the days variable within it, if that's the case.)

maikelnabil:
Hey @colinxfleming I made the changes you requested. I don't know how to write tests. Let me know how to be more helpful!

Issue #948, opened 2017-03-12
Should we collapse patient and pregnancy objects?


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Simplify our object model, potentially collapsing patient and pregnancy.

* **What feature or behavior is this required for?**
Writing simpler code!

* **How could we solve this issue? (Not knowing is okay!)**
Quick history lesson: When we originally worked this out we were planning on patients connecting to pregnancies with an embeds_many instead of the current embeds one, but I ratcheted that back before the initial deployment in September. The original reasoning was that DCAF is pretty interested in what people who seek our services multiple times look like, but we've since worked our way around that by deciding to just archive and scramble patients when we're done with them and create entirely new patient objects to represent new pregnancies.

I figured at the time it was better to separate things out rather than have a monolithic patient object, so we deployed with that and I think we've all been kind of paying for that mistake ever since. I think fulfillments are a bit different since not everybody has them, but every patient we log has a pregnancy. 

The potential gains here are: 

* one object to keep track of and report on instead of two
* way easier to filter and report on, and write code against
* simplified object model (no more 'so what's under pregnancy and what's under patient?')

The potential drawbacks are:
* bulkier core object might be doing too much

* **Anything else?**
I'm interested in thoughts from the team on this idea. In particular, would this make our object model easier to understand, explain, or write code against?

colinxfleming:
I poked at this just for funsies tonight and am now convinced it's a good idea -- I think we haven't really touched `Patient` or `Pregnancy` except to bolt on new stuff, and this gives us a chance to do a clean refactor.

PR #1013, opened 2017-04-21
Move patient methods to concerns

colinxfleming:

The patient object was getting a little massive. 300 line models is no way to live.

This begins the process in #948 by moving relevant sets of methods into their own concerns.

* Breaks patient object into a core, plus Urgency, Callable, Notetakeable, Searchable, and AttributeDisplayable concerns

@lomky seeking your thoughts on the patterns here -- does dividing things up this way and the names I picked make sense to you? Anything else I should extract?

### DIFF
--- a/app/models/concerns/attribute_displayable.rb
+++ b/app/models/concerns/attribute_displayable.rb
@@ -1,0 +1,13 @@
+module AttributeDisplayable
+  extend ActiveSupport::Concern
+
+  def primary_phone_display
+    return nil unless primary_phone.present?
+    "#{primary_phone[0..2]}-#{primary_phone[3..5]}-#{primary_phone[6..9]}"
+  end
+
+  def other_phone_display
+    return nil unless other_phone.present?
+    "#{other_phone[0..2]}-#{other_phone[3..5]}-#{other_phone[6..9]}"
+  end  
+end
--- a/app/models/concerns/attribute_displayable.rb
+++ b/app/models/concerns/attribute_displayable.rb
@@ -1,3 +1,4 @@
+# Methods related to displaying attributes on the patient model
 module AttributeDisplayable
   extend ActiveSupport::Concern
 
@@ -9,5 +10,5 @@ module AttributeDisplayable
   def other_phone_display
     return nil unless other_phone.present?
     "#{other_phone[0..2]}-#{other_phone[3..5]}-#{other_phone[6..9]}"
-  end  
+  end
 end
--- a/app/models/concerns/callable.rb
+++ b/app/models/concerns/callable.rb
@@ -1,3 +1,4 @@
+# Methods relating to calls or contacts
 module Callable
   extend ActiveSupport::Concern
 
--- a/app/models/concerns/callable.rb
+++ b/app/models/concerns/callable.rb
@@ -1,0 +1,29 @@
+module Callable
+  extend ActiveSupport::Concern
+
+  def recent_calls
+    calls.includes(:created_by).order('created_at DESC').limit(10)
+  end
+
+  def old_calls
+    calls.includes(:created_by).order('created_at DESC').offset(10)
+  end
+
+  class_methods do
+    def contacted_since(datetime)
+      patients_reached = []
+      all.each do |patient|
+        calls = patient.calls.select { |call| call.status == 'Reached patient' &&
+                                              call.created_at >= datetime }
+        patients_reached << patient if calls.present?
+      end
+
+      # Should we use this or first call?
+      first_contact = patients_reached.select { |patient| patient.initial_call_date >= datetime }
+
+      # hard coding in first_contacts and pledges_sent for now
+      { since: datetime, contacts: patients_reached.length,
+        first_contacts: first_contact.length, pledges_sent: 20 }
+    end
+  end
+end
--- a/app/models/concerns/notetakeable.rb
+++ b/app/models/concerns/notetakeable.rb
@@ -1,3 +1,4 @@
+# Methods relating to notes or parsing notes
 module Notetakeable
   extend ActiveSupport::Concern
 
--- a/app/models/concerns/notetakeable.rb
+++ b/app/models/concerns/notetakeable.rb
@@ -1,0 +1,14 @@
+module Notetakeable
+  extend ActiveSupport::Concern
+
+  def most_recent_note_display_text
+    note_text = most_recent_note.try(:full_text).to_s
+    display_note = note_text[0..30]
+    display_note << '...' if note_text.length > 31
+    display_note
+  end
+
+  def most_recent_note
+    notes.order('created_at DESC').limit(1).first
+  end
+end
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -1,0 +1,67 @@
+module Searchable
+  extend ActiveSupport::Concern
+
+  SEARCH_LIMIT = 15
+
+  class_methods do
+    # Case insensitive and phone number format agnostic!
+    def search(name_or_phone_str, lines = LINES)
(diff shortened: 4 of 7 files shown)


app/models/concerns/attribute_displayable.rb
```diff
@@ -0,0 +1,13 @@
+module AttributeDisplayable
```
lomky: This is the only one whose name is a bit confusing to me. What else would go in the concern?

lomky: Actually, Having some guidance for what goes in any given concern vs in the main patient might be good for all of them.

app/models/concerns/callable.rb
```diff
@@ -0,0 +1,29 @@
+module Callable
+  extend ActiveSupport::Concern
```
lomky: So this Concern is where "Functions related to calling patients" go?

app/models/concerns/notetakeable.rb
```diff
@@ -0,0 +1,14 @@
+module Notetakeable
```
lomky: This concern is for "Functionality related to Notes on Patients"?

app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,67 @@
+module Searchable
+  extend ActiveSupport::Concern
```
lomky: This concern is for "Search functions around Patients"?

Is something is both search-ish and, for example, notetaking-ish, which Concern does it belong in? Looks like Searchable?

app/models/concerns/urgency.rb
```diff
@@ -0,0 +1,35 @@
...
+        end
+      end
+    end
+  end
+
+  private
+
+  def recent_history_tracks
+    history_tracks.select { |ht| ht.updated_at > 6.days.ago }
+  end
```
lomky: Why does this function go here in Urgency?

app/models/patient.rb
```diff
@@ -1,13 +1,18 @@
...
   include Mongoid::Timestamps
   include Mongoid::Enum
   include Mongoid::History::Trackable
   include Mongoid::Userstamp
   include StatusHelper
+  include Urgency
+  include Callable
+  include Notetakeable
+  include Searchable
+  include AttributeDisplayable
```
lomky: Maybe a note at the top here stating that these are Concerns (with a link to wtf Concerns are?) and any related functionality should go in them instead of directly in Patient?

app/models/patient.rb
```diff
@@ -126,51 +118,16 @@ def self.pledged_status_summary(num_days = 7)
...
-  def most_recent_note
-    notes.order('created_at DESC').limit(1).first
-  end
-
   # TODO: reimplement once pledge is available
-  #def most_recent_pledge_display_date
-  #  display_date = most_recent_pledge.try(:sent).to_s
-  #  display_date
-  #end
+  def most_recent_pledge_display_date
```
lomky: Did you mean to commit this in this PR? It's not mentioned in the description.

app/models/patient.rb
```diff
@@ -126,51 +118,16 @@ def self.pledged_status_summary(num_days = 7)
...
+   display_date = most_recent_pledge.try(:sent).to_s
+   display_date
+  end
 
   # TODO: reimplement once pledge is available
-  #def most_recent_pledge
-  #  pledges.order('created_at DESC').limit(1).first
-  #end
+  def most_recent_pledge
```
lomky: Did you mean to commit this in this PR? It's not mentioned in the description.

app/models/patient.rb
lomky: I found the [default docs](http://api.rubyonrails.org/classes/ActiveSupport/Concern.html) a bit lacking in use depth, but [this nice writeup was helpful](http://vaidehijoshi.github.io/blog/2015/10/13/stop-worrying-and-start-being-concerned-activesupport-concerns/)

app/models/concerns/callable.rb
colinxfleming: Yeah, that's right -- stuff pertaining to parsing calls or contacting patients goes here. I'll add a comment to reflect.

app/models/concerns/notetakeable.rb
colinxfleming: Yep! Methods relating to parsing notes for a given patient. Comment added.

app/models/concerns/attribute_displayable.rb
colinxfleming: AttributeDisplayable was kind of a shot in the dark, but this one pertains to convenience methods for displaying attributes on the patient model in particular. It was kind of a catchall nothing else really fit into here, at least prior to merging pregnancy in.

I added a comment to indicate that it displays attributes on the patient model, but this really could just as easily be `PhoneDisplayable`. Thoughts?

app/models/concerns/searchable.rb
colinxfleming: Yeah, I think something that involved both would wind up in Searchable, but I'm admittedly not sure how to resolve ties here. Thoughts?

app/models/concerns/urgency.rb
colinxfleming: It's consumed by `still_urgent?` and not really used anywhere else, but it probably does have wider application. We could move it out into `Trackable` and try to move some of the `track_history on: ...` boilerplate into that. Which sounds like a really good idea now. Should try it out.

app/models/patient.rb
colinxfleming: whoops. nope. We can remove these since we went a different direction with our pledge. I'm gonna strip 'em.

app/models/patient.rb
colinxfleming: I think this is dead on. Added a comment and the second link.

colinxfleming:
There are a few additional things here:

* move StatusHelper into a concern where it belongs
* Break up `patient_test` into individual concern tests

colinxfleming:
A couple changes in that round of commits:

* History track related stuff moved to its own concern
* Comments added to the top of concerns
* Removed methods we don't use which I accidentally uncommented. whoops

@lomky getting closer?

app/models/concerns/searchable.rb
lomky: I think search functions benefit from being clustered together. Their form is more similar to each other than to the different objects they are querying, so when you're going in to add any new search capability having the other examples at hand seems helpful.

app/models/concerns/attribute_displayable.rb
lomky: I think I like the existing name after you've explained it.

app/models/patient.rb
```diff
@@ -126,150 +115,17 @@ def self.pledged_status_summary(num_days = 7)
...
-    # Verify that a pregnancy has not been marked urgent in the past six days
-    return false if recent_history_tracks.count == 0
-    return false if pregnancy.pledge_sent || pregnancy.resolved_without_dcaf
-    recent_history_tracks.sort.reverse.each do |history|
-      return true if history.marked_urgent?
-    end
-    false
-  end
-
   def assemble_audit_trails
```
lomky: would this be a good candidate for moving into HistoryTrackable?

colinxfleming: Yes it would! Moving it over.

colinxfleming:
I had to move the `track_history on: ...` stuff back to the patient object due to some demonry w/r/t `still_urgent?` that I don't quite understand. Not a significant change. 

Moved `assemble_audit_trails` to the `HistoryTrackable` concern as well per a(nother) great suggestion from @lomky .

Going to make sure tests pass and then merge if it comes back clean.

colinxfleming:
Here's to a reduced terror factor!

PR #1014, opened 2017-04-21
WIP Merge patient and pregnancy models

### LINKED ISSUE #948: Should we collapse patient and pregnancy objects?

colinxfleming:

I wanted to see how much of a pain this would be and it doesn't seem like it's going to be that brutal. Here's a draft PR that merges the pregnancy fields we're using into the ones that patient uses, merges

Everything else we'll have to do: 

* get `fields_for patient.pregnancy` out of our views
* fix the controllers since we no longer need to create a pregnancy object at point of create
* probably fix some reporting stuff
* break up the patient test, which is now approaching 400 lines or so
* write a huge migration script to move over the pregnancy fields to the patient fields

See aso #1013 

@lomky @Kevin-Wei interested in your thoughts about this as people who probably often have to manage large object oriented code. 

* Merges patient and pregnancy into one mega object
* other stuff tk

* Fixes #948 , sooner or later


lomky:
My first instinct here is that this feels completely right. We do not, so far as I'm aware, have any reason to consider 'Patient' sans 'Pregnancy', nor a 'Pregnancy' without a 'Patient'. Reducing it to the 'Patient' object seems like the absolutely correct choice.

My first thoughts on this:

  - Before this change, we could have a Patient object which had not yet had the Pregnancy object created. 
    - was this a meaningful state for the object that we want to preserve somehow?
    - My instinct is no, but I want to run that by others.

app/models/patient.rb
```diff
@@ -55,12 +66,16 @@ class Patient
   field :household_size_adults, type: Integer
   field :insurance, type: String
   field :income, type: String
+  field :special_circumstances, type: Array, default: []
```
lomky: Looking at the relocated field names, this is the only one that strikes me as becoming unintuitive when moved from belonging to a "pregnancy" to belonging to a "patient". Are we talking about special circumstances wrt the pregnancy, the patient, or either? If either, then I think we've actually improved the semantic meaning here. :100:

colinxfleming: I think in practice it winds up being either. Some of 'em are special circumstances for the pregnancy (a health diagnosis of some kind) and some of 'em are a patient's circumstances that give us more context into additional ways we can help them (per my understanding, stuff on the order of 'by the way, here's our person at a shelter', etc). But since a goal of DCAF is to center patients in abortion care I think this a step toward reflecting that in our codebase.

colinxfleming:
Def not merge-ready -- plenty of work to do in the views in particular!

I think 'artificially prying apart' is exactly right. At the beginning of this project, a story which we probably featured too heavily was cases where patients use our services a second time, so we initially had a `patient has_many pregnancies` going on, until it became clear before deployment that was going to dig us into a hole. With `identifier` and patient archiving we have a way to capture that story in such a way that doesn't result in a confusing object model.

A patient without a pregnancy is not a state we need to preserve, as we create them both at the same time (along with fulfillment, which tbh we should probably stop doing). See the very hacky code around the create actions in `PatientsController` for an example here.

Having received this blessing, I'll continue working on this after getting the pdf stuff in. Thanks a ton @lomky !

colinxfleming:
@lomky the integration and controller tests aren't all passing yet, but this is getting closer to the final version (views code and controller code mostly sorted out at this point). It's probably closer to done than not at this point. Would love your eyes and any feedback if you get a chance!

app/controllers/patients_controller.rb
```diff
@@ -33,30 +31,22 @@ def update
...
+  def data_entry_create
     @patient = Patient.new patient_params
     @patient.created_by = current_user
-    @pregnancy = @patient.pregnancy || @patient.build_pregnancy
-    @pregnancy.created_by = current_user
-    (@patient.fulfillment || @patient.build_fulfillment).created_by = current_user
 
     if @patient.save
       flash[:notice] = "#{@patient.name} has been successfully saved! Add notes and external pledges, confirm the hard pledge and the soft pledge amounts are the same, and you're set."
-      current_user.add_patient @patient
```
lomky: I don't understand this line's deletion. It doesn't seem related to pregnancy or fulfillment objects?

app/controllers/patients_controller.rb
```diff
@@ -75,6 +65,9 @@ def patient_params
       :household_size_adults, :household_size_children, :insurance,
       :referred_by, :initial_call_date, :urgent_flag,
       :clinic_id,
+
+      :last_menstrual_period_days, :last_menstrual_period_weeks,
```
lomky: Are the redundant with the same-name ones under `pregnancy` two lines down? Or should those be deleted? Or do I not understand what's happening here?

app/controllers/patients_controller.rb
colinxfleming: Right you are. I'll restore that.

app/controllers/patients_controller.rb
colinxfleming: Yeah, they're redundant, and straightening out strong params definitely slipped my mind, so thanks for flagging this! Since there's no longer a `pregnancy` object here, we can move them all over and totally eliminate the nesting. I might do a little bit of additional cleanup refactoring here by moving sets of attributes into constants, reflecting where they are in the form (e.g. `PATIENT_DASHBOARD_PARAMS = [:name, :primary_phone, :appointment_date, ...]`).

(In the unlikely event that you haven't seen these before -- Strong parameters are one of rails' most 'Why the isn't this working' pieces of the framework, but they're great. They make it so you can't mass assign attributes thru a form unless all the attributes are cleared. The API writeup isn't bad: http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters )

app/controllers/patients_controller.rb
colinxfleming: Fixed in 16faa09

colinxfleming:
Straightened out everything except one lonely reporting test I'm going to have to dig into further. Also changed base branch to master. This SHOULD be good to 

@lomky would welcome another eyes. sorry it's gigantic, turns out when you depreciate an object it involves a lot of code changes, weird.

lomky:
@colinxfleming on it! I'll also do a grep pass in the code as a whole, like I should have the first time :sweat_smile:

app/models/concerns/last_menstrual_period_measureable.rb
```diff
@@ -1,41 +1,39 @@
...
   def last_menstrual_period_display
-    return nil unless last_menstrual_period_now
-    display_as_weeks last_menstrual_period_now
+    return nil unless _last_menstrual_period_now
+    _display_as_weeks _last_menstrual_period_now
   end
 
   def last_menstrual_period_display_short
-    return nil unless last_menstrual_period_now
+    return nil unless _last_menstrual_period_now
```
colinxfleming: Convention only! It more or less translates to 'treat this as private, but invoke it if absolutely necessary'. There was some fight I was having with the object that made this a good decision but I forget the finer details, can dig it up if you like.

colinxfleming:
Removed @lomky 's flagged refs (thank you!), except for the reporting test which is a little more involved. Working on the reporting test now.

app/controllers/patients_controller.rb
```diff
@@ -54,26 +55,36 @@ def find_patient
     @patient = Patient.find params[:id]
   end
 
+  # Strong params divided up by partial, in order of appearance
```
colinxfleming: yeah, in order of appearance where. I meant to say 'appearance in their respective partials', but I think just partial is fine. Removing everything after the comma here for simplicity.

colinxfleming:
Fixed the reporting test. Fingers crossed but I think we SHOULD be good?

colinxfleming:
AWWWW YISSSSS. Build is passing!

colinxfleming:
I'm not super concerned about the coverage drop since they seem to be due to the reduction of loc in general. Which is good!

@lomky requesting another review from you. I'd love to roll this out as part of the deployment on Sunday.

colinxfleming:
awww yissssss

I'll fix the merge conflicts during my lunch break today. This is also going to cause a whole host of them so let's be ready for that. 

The final ticket item here is writing a migration script that will move over individual pregnancy fields manually (something like `Patient.each { |pt| pt.update field: pt.pregnancy.field }` or a more efficient equivalent) and then finally unsetting the pregnancy object. Steps there should be:

* If prod: take a db backup
* Create a test fixture with every field filled out
* Run the migration script
* Check that the test fixture shows up the same
* Unset pregnancy object
* Check that the test fixture shows up the same
* If prod: take a db backup

lomky:
Did an additional review via command line and prodded around my local instance. I'm solidly on board! :100:

colinxfleming:
Here's a short block of code we can run to nuke everyone. Not terribly efficient, but who cares, it's a perk of working with small data. 

```
# migrate over pregnancy fields onto patient object
Patient.each do |patient|
  patient.update  last_menstrual_period_days: patient[:pregnancy][:last_menstrual_period_days]
  patient.update  last_menstrual_period_weeks: patient[:pregnancy][:last_menstrual_period_weeks]
  patient.update  referred_to_clinic: patient[:pregnancy][:referred_to_clinic]
  patient.update  resolved_without_dcaf: patient[:pregnancy][:resolved_without_dcaf]
  patient.update  procedure_cost: patient[:pregnancy][:procedure_cost]
  patient.update  pledge_sent: patient[:pregnancy][:pledge_sent]
  patient.update  patient_contribution: patient[:pregnancy][:patient_contribution]
  patient.update  naf_pledge: patient[:pregnancy][:naf_pledge]
  patient.update  dcaf_soft_pledge: patient[:pregnancy][:dcaf_soft_pledge]
end

# destroy info in `patient.pregnancy`
Patient.each { |patient| patient.unset :pregnancy }
```

colinxfleming:
That seems to work in a dev environment based on running `rake db:seed` on master and then switching to this branch. I'll test it out on the sandbox before running in production.

colinxfleming:
Let's party!

colinxfleming:
I've confirmed that the migration script runs clean in sandbox just now. 

Bummer: Does not fix the report endpoint issue.

Issue #490, opened 2016-07-07
Patient Info in pledge and status isn't reflected until page refresh


MiroFurtado:
Pretty self explanatory. In order for the patient info to be saved (for example if I want to show the state of the patient in the pledge popup box), I have to click on the dashboard and then back on the patient.

colinxfleming:
So if I'm understanding correctly, the steps to recreate are: 

* I am looking at a pregnancy edit page
* I edit an attribute such as phone number
* I click on submit pledge; submit pledge has the wrong phone number
* I reload the page and click submit pledge again; submit pledge now has the right phone number

colinxfleming:
Potential solution here is to have the submit pledge modal populate by js injection rather than on page load. I'll investigate tomorrow night.

colinxfleming:
Thanks a ton for flagging this @MiroFurtado . Good call.

colinxfleming:
This probably isn't a Monday thing. Sorry @lwaldsc .

colinxfleming:
This is currently a major pain point so I'm going to slate it for the hackathon.

phoenixeliot:
I'm taking a look at this one.

colinxfleming:
I'm popping this into active. I think there's a useful feedback win here. 

The technical step is to probably update `app/views/patients/update.js.erb` to reload the patient's status on the pregnancy edit page with new information.

colinxfleming:
#1015 resolves for the submit pledge flow (which, A++++++++++). Still need to include a couple other dynamic fields (some fields in patient information, fields in fulfillment partial, probably the changelog?).

PR #1015, opened 2017-04-22
Single modal for call lists

### LINKED ISSUE #257: Call List: change use of new call modal to only use one modal in the DOM
Currently every entry (row) in a call list, using partial _table_content, is inserting its own modal in the row, based on the partial in views/calls/_new_call. We should change the modal, where it is inserted, and how it is referenced in _table_content so that we can have a single modal that can be called by every entry (row) in the table, or even possibly by multiple tables because the dashboard index has multiple call tables.

tingaloo:

Global modal on load, fetches call data on click so it doesn't render a modal per call.
It may break some tests, but we can lay this issue to rest.

* Single modal

* Fixes #257


tingaloo:
Actually I got something off, I didn't realize data was still being called a lot.

app/views/dashboards/_table_content.html.erb
```diff
@@ -42,9 +42,20 @@
             <% end %>
           <% end %>
 
-          <td><%= link_to call_glyphicon, "#call-#{patient.primary_phone_display}", data: {toggle: "modal"} %></td>
+      <td>
+        <%= link_to call_glyphicon, "#call-#{patient.primary_phone_display}", class:"call_button",
+        data: {
+          patientname: patient.name,
```
colinxfleming: This is kind of a cool implementation of this! I am going to keep this in the back of my mind when I run into this problem in the future. That said, I worry that we're basically moving the 100 different modals into 100 different data fields and not actually generating any less code with this solution.

It looks like this is basically overruling what's in `calls/_new_call` with a bit of additional javascript to pop in the modal, and I do think there's some merit to keeping that as the flow -- slightly more restful.

I think a way around this might be using rails UJS -- A nice feature of rails views is that they can fire off javascript/vanilla jquery as an action -- e.g. we could rig up `calls/new` to `respond_to :js`, and have a `calls/new.js.erb` that fired off a jquery event like: `$('.call-modal').html(<%= escape_javascript(render 'calls/_new_call', patient: patient %>);`. See `app/views/notes/create.js.erb` for something similar that we're doing. Then we could rig this link to just fire an ajax `get` at a patient new call path. Some reading on how rails does this: http://guides.rubyonrails.org/working_with_javascript_in_rails.html#server-side-concerns https://coderwall.com/p/kqb3xq/rails-4-how-to-partials-ajax-dead-easy

Your thoughts?

app/assets/javascripts/dashboards.coffee
```diff
@@ -1,3 +1,22 @@
...
+
+# File dynamically loads in all patient data on click.
+
+jQuery ->
+  $(document).on "click", ".call_button", ->
+
+    $('.new-call').attr('id', $('.call_button').data('primaryphone'))
+    el = $('.modal-title')
+
+    el.html("")
```
colinxfleming: I'm sure whatever we do here will involve some coffeescript like this -- I'd love to wrap whatever we do in a named function so it's a little easier to read on the fly, if that's cool with you.

app/views/dashboards/_modal.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="modal fade new-call" tabindex="-1" role="dialog"  %>">
```
colinxfleming: 👍

app/views/dashboards/_table_content.html.erb
tingaloo: > That said, I worry that we're basically moving the 100 different modals into 100 different data fields and not actually generating any less code with this solution.

This was the big whoops, I realized this when all the ```data``` variables were called regardless. I wasn't sure how the data attribute worked at first.

> It looks like this is basically overruling what's in calls/_new_call with a bit of additional javascript to pop in the modal, and I do think there's some merit to keeping that as the flow -- slightly more restful.

Yup. I'm brute forcing JS in here ha.

My original idea was to generate a Rails method that calls all the helper methods, but controllers cannot call helper methods. So I went a little off rails..

Based on what you said and reading those guides, I am to generate a Rails method that calls ```js.erb```, render a partial which calls all the helper methods.

Can you show me a snippet where notes#create is being called?

colinxfleming: Yeah, sounds like we're in agreement here. I think here's the path to take:

* Add `:new` to the `resources :calls` array in `routes.rb`, and a `CallsController#new` method to match
* Fire up a new file `app/views/calls/new.js.erb` which renders the `_new_call` partial (pretty much what the coffeescript is doing, just with rails `escape_javascript(render ...)` helper deal)
* Have the phone icon link trigger that controller action. (We do something similar with `display_reached_patient_link` in `CallsHelper`.

I think from there it's going to be stuff on the order of making sure the js still works (modal clickaway, being able to create new calls via the link, etc).

For ref, we're calling `notes#create` here -- https://github.com/colinxfleming/dcaf_case_management/blob/master/app/views/patients/_notes.html.erb#L15 -- and this is the block that's telling rails to look for `create.js.erb`.

Thanks again, @tingaloo ! Feel like I'm picking up a decent amount of js just by reading your PRs.

colinxfleming:
Yeah, sounds like we're in agreement here. I think here's the path to take:

* Add `:new` to the `resources :calls` array in `routes.rb`, and a `CallsController#new` method to match
* Fire up a new file `app/views/calls/new.js.erb` which renders the `_new_call` partial (pretty much what the coffeescript is doing, just with rails `escape_javascript(render ...)` helper deal)
* Have the phone icon link trigger that controller action. (We do something similar with `display_reached_patient_link` in `CallsHelper`.

I think from there it's going to be stuff on the order of making sure the js still works (modal clickaway, being able to create new calls via the link, etc).

For ref, we're calling `notes#create` here -- https://github.com/colinxfleming/dcaf_case_management/blob/master/app/views/patients/_notes.html.erb#L15 -- and this is the block that's telling rails to look for `create.js.erb`.

Thanks again, @tingaloo ! Feel like I'm picking up a decent amount of js just by reading your PRs.

colinxfleming:
A beaut! Good to merge once CI passes 💯

app/views/calls/_new_call.html.erb
```diff
@@ -1,25 +1,14 @@
...
-      <div class="modal-body">
-        <div class="calls-layout">
-          <button type="button" class="close close-terms" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-          <h4 class="modal-title calls-request">Call <b><%= patient.name %></b> now:</h4>
-          <br>
-            <!-- wrapper for ajaxing when recording new call  -->
-            <%= display_other_contact_and_phone_if_exists(patient) %>
 
-            <%= content_tag :h4, patient.primary_phone_display, class: 'calls-phone' %>
+Call <b><%= patient.name %></b> now:</h4>
```
colinxfleming: Looks like the opening of the h4 tag here is in `_modal`. Can we move that in here?

app/views/dashboards/_modal.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="modal fade new-call" tabindex="-1" role="dialog">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-body">
+        <div class="calls-layout">
+          <button type="button" class="close close-terms" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+          <h4 class="modal-title calls-request">
```
colinxfleming: As noted above, would love to move this h4 into the other partial. Can replace it with a `div` so the javascript still works, I think?

tingaloo:
Haha, I know what I have to do. EVERY modal needs to be dynamically populated since everyone is sharing the same one and doing different mutations on it. There's only two though, i'll get to it sometime tomorrow.

app/views/dashboards/_modal.html.erb
tingaloo: Since everyone (patient submit/cancel pledge) is trying to populate the modal, I am going to have it clear at the ```<div class="modal-content">``` for consistency. In the future, every new modal popup needs to follow the same rules as calls#new

app/views/patients/pledge.js.erb
```diff
@@ -0,0 +1,8 @@
+$('.modal-content').html('')
+$('.modal-content').append(
+  '<%= escape_javascript(render partial: "patients/pledge", locals: { patient: @patient, disable_next: disable_continue?(@patient) })%>');
+
+$('.modal').modal('toggle');
+$('.js-title-step span').slice(1).remove();
+
```
tingaloo: Here is the issue, new text gets appended to ```.js-title-step```. I don't want to do a direct string check if I can help it. The slice method gets rid of the 1 label number that keeps being appended.

tingaloo:
@colinxfleming There's one slight bug left. If a CM repeatedly clicks submit pledge over and over, the below text will keep appending. This stems from ```$('[id$=modal]').modalSteps()``` being called on each toggle. There's a few solutions to this but I wanted to know what you think.

screen shot 2017-05-14 at 11 54 56 am

tingaloo:
Nevermind, i finally got it. I'm done with jquery for awhile. Please look at ```pledge.js.erb``` and tell me if that flies.

tingaloo:
@colinxfleming bumping this, also should fix #490. I have some other changes in mind but would like this to be merged first.

test/controllers/calls_controller_test.rb
```diff
@@ -7,6 +7,14 @@ class CallsControllerTest < ActionDispatch::IntegrationTest
     @patient = create :patient
   end
 
+  describe 'new method' do
+    it 'should respond successfully' do
+      @call = attributes_for :call
```
colinxfleming: Might be able to ditch this line, give it a shot and let me know?

app/views/patients/pledge.js.erb
```diff
@@ -0,0 +1,16 @@
+$('.modal-content').html('')
+$('.modal').attr('id', 'pledge-modal')
+$('.modal-content').append(
+  '<%= escape_javascript(render partial: "patients/pledge", locals: { patient: @patient, disable_next: disable_continue?(@patient) })%>'
+);
+
+$('.modal').modal('toggle');
```
colinxfleming: quick clarification on what this set of javascript is doing? Worth rolling into a function?

app/views/patients/_call_log.html.erb
```diff
@@ -2,11 +2,11 @@
   <div class="margin-bottom col-sm-12">
     <h2>
       Call Log
-      <span class="pull-right new-call-entry"><small><%= link_to "Record new call", "#call-#{@patient.primary_phone_display}", data: {toggle: "modal"} %></small></span>
+      <span class="pull-right new-call-entry"><small>
+        <%= link_to "Record new call", new_patient_call_path(@patient), class:"call_button", remote: true %>
```
colinxfleming: can you slap an additional class on this element? `call-{phone number}` please!

app/views/dashboards/_table_content.html.erb
```diff
@@ -41,10 +41,8 @@
               <td><%= link_to "Add", add_patient_path(current_user, patient), method: :patch, remote: true %></td>
             <% end %>
           <% end %>
+          <td><%= link_to call_glyphicon, new_patient_call_path(patient), class:"call_button", remote: true %> </td>
```
colinxfleming: Same thing here, would you mind slapping a class `call-{phone-number}` on this link? e.g. `call-123-123-1234`

test/controllers/calls_controller_test.rb
tingaloo: good catch!

app/views/patients/pledge.js.erb
tingaloo: I added a new route for submitting pledge via ujs. The disabled next stuff is for the jquery multi pledge modal. Maybe we can roll it in a function, but this is pretty specific and i'm comfortable leaving as is until another modal comes around.

colinxfleming: Got it. I tend to go 50-50 on clearly named functions for stuff like this, just because JS is so blasted hard to read, but I hear you on the specificity of this. Let's merge as is.

colinxfleming:
Can confirm that this addresses part of #490 , though we'll want to flesh it out a little further so a couple other live-changing fields are altered. (Namely, a bunch of stuff in the top patient dashboard, a few fields in fulfillment, and maybe the changelog.). I'll identify them all specifically in a bit.

PR #1021, opened 2017-04-29
Add brakeman pro to ci

### LINKED ISSUE #1019: Switch out brakeman for brakeman pro gem

THe company has very generously given us a license to use the gem. We should take steps to set that up in CI as part of runs.

~ security upgrade ~ 

https://brakemanpro.com/docs/engine/commandline

* **Anything else?**
Requires keys -- assigning to myself

colinxfleming:

Swaps out brakeman for brakeman pro in CI (thank you Justin and team brakeman)

* Swaps out brakeman pro in CI
* Fixes a warning (in progress, not a bad one)

* Fixes #1019

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -39,7 +39,7 @@ module DashboardsHelper
     # Throw a flag unless sorting by string, int, or float
     raise 'Bad datatype' unless %w(string string-ins int float).include? type
     content_tag :th, data: { sort: type } do
-      safe_join [column_name, autosort_arrow_span(local_assigns)], ' '
+      safe_join([column_name, autosort_arrow_span(local_assigns)], ' ').strip
     end
   end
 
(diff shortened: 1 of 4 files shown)


colinxfleming:
This is good to go from my end (assuming I can get the flaky tests to pass). The changes to the view fix a warning, which is almost def a false flag but which I'm fixing in the name of a clean report. @DarthHater can you review when you get a minute? No rush.

colinxfleming:
May need to adjust this due to @DarthHater 's discovery that env vars are not available in forks; game plan TBD

colinxfleming:
@DarthHater 's work on gemnasium PRs has made clear that env vars aren't available in forks, which means merged as is, this PR breaks every build from a fork. To compensate in the interim, this just builds on the main master, so while we won't catch things before they hit master, it'll serve as a safety fuse before stuff makes it into sandbox.

colinxfleming:
Thank you @lomky !

PR #1025, opened 2017-04-29
autocheck fulfilled

### LINKED ISSUE #983: Autocheck 'Fulfilled' checkbox when entering stuff in any fulfillment field

In this fulfillment form, there's a boolean checkbox for 'Fulfilled' that's easy to miss -- we should make it so it autochecks if you enter any fields in the fulfillment form.

Make sure our data entry vols don't accidentally mark someone as not fulfilled. 

The code to this partial is in `app/views/patients/_fulfillment.html.erb`. Any js should go in `app/assets/javascripts/patients.coffee`. 

Here's a screenshot of the partial in action:
image

I'm agnostic as to the best way to do this, so taker's choice. Probably the most straightforward way to do this is to just check the box if any fields in the form aren't blank or empty. 

* **Anything else?**
There's a patient created from `rake db:seed` that you should use, since the Fulfillment tab doesn't show up until a patient's marked pledge sent. (Name: `Pledge Submitted - 3`)

Stems from #864 .

Reserving for Jon, who needs a js issue to hammer on!

tingaloo:

On pledge fulfillment, any field change autochecks Pledge fulfilled.

* Fixes #983


test/integration/pledge_fulfillment_test.rb
```diff
@@ -82,4 +83,39 @@ class PledgeFulfillmentTest < ActionDispatch::IntegrationTest
       assert has_text? 'Check #'
     end
   end
+
+  describe 'pledge fulfilled autochecking on change' do
```
colinxfleming: My dude 👍

app/assets/javascripts/patients.coffee
```diff
@@ -23,6 +23,23 @@ valueToNumber = (val) ->
   +val || 0
 
 ready = ->
+  fields = [
```
colinxfleming: Can we rename this variable a bit for clarity's sake? Not sure I'd know what this did just by looking at it if I saw it out outside the context of this PR.

app/assets/javascripts/patients.coffee
```diff
@@ -37,6 +54,26 @@ ready = ->
   $(document).on "change", "#abortion-information-form input, #external_pledges input", ->
     updateBalance()
 
+  $(document).on "change", field_selectors, ->
+    i = 0
```
colinxfleming: Could we wrap this in its own named function? Something like `mark_fulfilled_when_fields_checked` or something similar that captures the spirit of what this does?

app/assets/javascripts/patients.coffee
```diff
@@ -37,6 +54,26 @@ ready = ->
...
+          break;
+        else
+          el.prop 'checked', true
+        i++
+      else
+        i++
+    if empty == true
+          el.prop 'checked', false
+
+    $(this).submit()
```
colinxfleming: Not sure we need this, it should be submitting per L31 in `patients.coffee` -- can you confirm?

app/assets/javascripts/patients.coffee
```diff
@@ -23,6 +23,23 @@ valueToNumber = (val) ->
...
+    '#patient_fulfillment_procedure_cost'
+    '#patient_fulfillment_check_number'
+    '#patient_fulfillment_gestation_at_procedure'
+    '#patient_fulfillment_date_of_check'
+    '#patient_fulfillment_procedure_date'
+  ]
+  i = 0
+  field_selectors = ''
+
+  while i<fields.length
```
colinxfleming: As a 'doesn't write that much js person', I'm not quite sure what this is doing and why, could I trouble you to write a quick descriptive comment about what's going on here?

tingaloo: Sure! basically ```field_selectors``` looks something like this

```'#patient_fulfillment_procedure_cost, #patient_fulfillment_check_number, #patient_fulfillment_gestation_at_procedure, ...'```

This just programmatically concats the fields for less mistyping. And I can feed this variable to the onChange jquery handler without too many problems.

File: app/assets/javascripts/patients.coffee (hunk 4 above)
Comment (tingaloo) [inline app/assets/javascripts/patients.coffee]: We didn't need it, thanks for catching this.

File: app/assets/javascripts/patients.coffee (hunk 5 above)
Comment (tingaloo) [inline app/assets/javascripts/patients.coffee]: Then I just realized that I can just use ```  $(document).on "change", ".edit_patient", ->``` so I'll be patching it right up.

File: app/assets/javascripts/patients.coffee [hunk 6]
```diff
@@ -22,13 +22,41 @@ calculateRemainder = ->
...
+
+
 ready = ->
   $(document).on "click", "#toggle-call-log", ->
     $(".old-calls").toggleClass("hidden")
     html = if $(".old-calls").hasClass("hidden") then "View all calls" else "Limit list"
     $("#toggle-call-log").html(html)
 
   $(document).on "change", ".edit_patient", ->
+    markFulfilledWhenFieldsChecked()
```
colinxfleming: For patients where there isn't a pledge yet, this seems to be breaking the ability to save. To replicate: 

* run `rake db:seed`
* go to a patient who hasn't been marked pledge sent
* try and save '2017-10-10' in the appointment date input
* check the console and see the error on no `length` field

Maybe a good way around this is to pop into `_fulfillment`, give this form a specific id, and then only fire this action when that id changes?

tingaloo:
Got em.

PR #1028, opened 2017-04-30
WIP Turn on test parallelization again

### LINKED ISSUE #992: Auto-parallelize in CI with knapsack gem

Test more faster by balancing tests across CI nodes

Test more faster

https://github.com/ArturT/knapsack#step-for-minitest and https://circleci.com/docs/1.0/parallel-manual-setup/

* **Anything else?**
Stems from noticing that #865 breaks test parallelizing in CI

colinxfleming:
A rails 5 regression on CI was not splitting tests between four nodes, which makes the suite take one minute instead of four. This PR turns that back on (eventually). 

Fixes #992


colinxfleming:
That was not pretty but I think we're in business again. @DarthHater I wound up adding an ext dependency that plugs into Circle, `knapsack`, since the old way of just `parallel: true` no longer works with rails 5 for some godforsaken reason. 

@DarthHater can you review?

colinxfleming:
yolo!

PR #1031, opened 2017-05-02
#1008 Stop incrementing LMP after a certain point

### LINKED ISSUE #1008: Stop incrementing LMP after a certain point

One of the most useful features in the app is our `LMP counter`, which given a date of initial call and another date will tell you how many weeks since someone's last menstrual period. However, we don't have it stop ever, so as we reconcile we're starting to see some people who are, like, 60 weeks pregnant, which, yikes. 

We should make it so this stops incrementing at either the date of appointment, or at 35 weeks 0 days. 

Reducing the cringe factor for our data entry team.

The code for this is in `app/helpers/last_menstrual_period_helper.rb`. Probably it involves adding some special case handling to the methods in there. You can see an example of how we're using it in practice at `app/views/patients/_patient_dashboard.html.erb`.

* **Anything else?**
Stems from #864

KatSDC:

I added case handling to prevent the LMP from incrementing past 40 weeks 0 days. Also had to use rails console to update a patient's LMP past 40 weeks in order to see it this change in the view.

* app/helpers/last_menstrual_period_helper.rb 

image

* Fixes #1008


colinxfleming:
Looks good! Let's merge as is and add a test separately. 

Might need to move this over to the concern once the pregnancy PR gets in. I'll be in touch on that if there's additional action to take. 

Thanks a ton @KatSDC !

colinxfleming:
@KatSDC we just merged #1014 so this might need a bump -- I think this code now lives in `app/models/concerns/last_menstrual_period_measureable.rb`, could you do a git pull from master and then move your changes over to that concern? (Sorry about that! #1014 was a lot of code.)

KatSDC:
I have updated the PR to address this!

app/models/concerns/last_menstrual_period_measureable.rb
```diff
@@ -30,7 +30,11 @@ def _last_menstrual_period_on_date(date)
     return nil unless initial_call_date && last_menstrual_period_weeks
     weeks = 7 * (last_menstrual_period_weeks || 0)
     days = (last_menstrual_period_days || 0)
-    (date - initial_call_date) + weeks + days
+    if (date - patient.initial_call_date) + weeks + days > 280
```
colinxfleming: Hey, looks like the test failures are legit, and I think the cause of them is the use of `patient` here. Since these methods in this concern are essentially methods on `patient`, it's trying to hit `patient.patient.initial_call_date` instead of `patient.initial_call_date` and choking.

tldr: I think you can remove the word `patient` from `patient.initial_call_date` here and below and it'll fly.

PR #1034, opened 2017-05-02
added county text_field to _patient_information

### LINKED ISSUE #1016: Add optional county field

Add a `county` field (string) on `Patient` object, and then add it to the form in `app/views/patients/_patient_information` if an env var for `TRACK_COUNTY` is present.

Baltimore Abortion Fund tracks county as a business need. We should

Nice and straightforward -- add a line to the appropriate place in `app/models/patient.rb`

* **Anything else?**
Needed for BAF deployment. 

There's probably an additional ticket as we generalize to allow toggling particular fields but I think we should wait until that crystalizes a little more (e.g. wait for another fund to give us a sense of what their particular business needs are)

electricflannel:

After examining, patient.rb, I found that ```field :county, type: String``` was already there.  So I only needed to add a text field to the _patient_information form.

It adds a text input into the __patient_information form.  Corrects issue "Add optional county field #1016"

Added county text input under City and State
screen_shot

* Fixes #1016


colinxfleming:
looks good @electricflannel ! We'll get this in after the patient/pregnancy PR gets merged. Thanks a ton!

lomky:
Deep dug into the patient params atm - I think we need to add this to the strong params as well?

colinxfleming:
@lomky is right on here -- this'll get swallowed by the controller if we don't bump it. My mistake, sorry all!

@electricflannel could you pop into `patients_controller` and add  `:county` to the `PATIENT_INFORMATION_PARAMS` constant when you get a sec?

electricflannel:
Hey @colinxfleming and @lomky, I just added the :county to strong params, so I think we are good to go!

app/controllers/patients_controller.rb
```diff
@@ -101,6 +101,23 @@ def find_patient
...
+      :household_size_adults, :household_size_children, :insurance,
+      :referred_by, :initial_call_date, :urgent_flag,
+      :clinic_id,
+      pregnancy: [:last_menstrual_period_days, :last_menstrual_period_weeks,
+                  :resolved_without_dcaf, :referred_to_clinic, :procedure_cost,
+                  :pledge_sent, :patient_contribution, :naf_pledge, :dcaf_soft_pledge],
+      special_circumstances: [],
+      fulfillment: [:fulfilled, :procedure_date, :gestation_at_procedure,
+                    :procedure_cost, :check_number, :check_date]
+
```
lomky: Looks like the merge got a bit mucked up! These lines (104-120) should be deleted, as all of them are redundant to lines 77-100 + line 121. 

Not too surprising considering how much we changed the object around on you mid ticket :)

colinxfleming:
This looks great, thanks @electricflannel ! I'm going to spin up a test in a separate PR to give you an example of that, so you know what you're looking for and next time you can learn some unit testing. Hang tight.

PR #1045, opened 2017-05-05
Test: addBalance test

### LINKED ISSUE #1039: Log of tests that need wwriting

Maintain high test coverage for our app by logging corners we cut, so we can add tests for them later.

Not accidentally breaking things as move into production

WIth that, the log: 

* <strike>updateBalance function in patients.coffee (should be an integration test confirming that the math in abortion information fires and is accurate)</strike> ( #1045 , thanks @tingaloo )

tingaloo:

Add integration test for addBalance()

* semi-Fixes #1039


colinxfleming:
Nicely done, thank you so much!

PR #1048, opened 2017-05-06
Whitelist clinics endpt

### LINKED ISSUE #1047: Display only NAF clinics only works as an admin

CMs (with that user role) should be able to filter to just NAF clinics using the button on the frontend. Instead, it seems to be just limited to admins.

Filtering to just NAF clinics.

I think the problem here is a permissioning issue -- the `before_action :confirm_admin_user` is tripping us up. Here's a proposed way to handle: only fire that `before_action` on `except: [ :index ]`. I'd like to have the json accessible as non-admin and the html not, but want the same endpoint more, so I'm not going to stress about it.

* **Anything else?**
(Should add a test for this too! whoops)

colinxfleming:

As noted in #1047 we made a slight oversight and only tested with admin users, and it turns out regular CMs can't make the api call. This whitelists the clinics index endpoint so they can.

@tingaloo could you review, if you have time between now and Saturday night? I feel good about this but want to get it in before the deploy on Sunday morning.

* Whitelists clinic endpoint for everyone to allow the JSON call
* Some code quality/style bumps
* Adds an integration test for the naf clinics filter feature

* Fixes #1047


colinxfleming:
(This'll need to be bumped now that #1014  is in)

test/integration/filter_naf_clinics_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    click_link 'Abortion Information'
+  end
+
+  describe 'filtering to just NAF clinics' do
+    it 'should filter to only NAF clinics when box is checked' do
+      assert has_select? 'patient_clinic_id', options: ['', @naf_clinic.name,
+                                                             @nonnaf_clinic.name]
+
+      check 'naf_filter'
+      sleep 1
```
tingaloo: ```wait_for_ajax ``` or not having any wait method works for some reason.

tingaloo:
This looks great, especially that test.

colinxfleming:
Thanks much @tingaloo ! Really appreciate it.

PR #1049, opened 2017-05-06
Update forks and remove one now that the upstream is updated for Rails 5

### LINKED ISSUE #867: Rails5 Technical Debt: Move Mongoid-Session-Store-rails4 to a Rails5 lib

* Well, right now we rely on a lib that is rails4 specific. It seems this lib is going to be absorbed into an official mongoid-session-store lib, and support for Rails 5 will be added there
* Problem is, they haven't released the Rails 5 support yet
* Bigger problem is, I added a reliance on a branch of a fork of the original lib where there is Rails 5 support
    - See: https://github.com/drgcms/mongo_session_store/tree/rails5
    - drgcms lib didn't work (he didn't implement a method we needed) so we forked to our own: https://github.com/DarthHater/mongo_session_store

* We want to make sure to use libs that are seeing active development, and won't be abandoned
* If we don't do this we will probably be decently mad at ourselves when Rails 6 or something else comes along

* Move to using the mongoid-session-store lib once they release a Rails 5 version
    - This one: https://github.com/mongoid/mongo_session_store

Related to the work at #865

DarthHater:

This updates all forks to be DCAFEngineering org owned, and removes one now that the upstream is good for rails 5. 

* Small changes to Gemfile and Gemfile.lock
* Upstream repos changed to be owned by DCAFEngineering

* Fixes #867


test/controllers/clinics_controller_test.rb
```diff
@@ -21,7 +21,7 @@ class ClinicsControllerTest < ActionDispatch::IntegrationTest
     end
 
     it 'should render json' do
-      get clinics_path, :format => :json
+      get clinics_path, xhr: true
```
colinxfleming: MY DUDE

PR #1052, opened 2017-05-13
Add dropoff status cont'd 

### LINKED ISSUE #736: New status for non-response

colinxfleming:

Jumps off @maikelnabil 's outstanding work in #1012 to implement a dropoff status, plus a unit test.

@maikelnabil your concept was really elegant and sound! I made it a little more 'ruby' in a couple ways and added a test. I'll explain some of my choices in the branch.

A thing I couldn't quite get perfectly was hitting it so that at 121 days exactly it moves over to dropoff. There seems to be an error of an additional day. Not sure it's worth sweating but would welcome thoughts in case I muffed something here.

@lomky and @maikelnabil would appreciate your thoughts!

* Dropoff status
* Test for dropoff status

* Fixes #736


app/helpers/status_helper.rb
```diff
@@ -1,28 +0,0 @@
-# Methods relating to inferring patient status
```
colinxfleming: This got moved into `app/models/concerns/statusable.rb` in the great patient-pregnancy merging.

app/models/concerns/callable.rb
```diff
@@ -10,6 +10,10 @@ def old_calls
     calls.includes(:created_by).order('created_at DESC').offset(10)
   end
 
+  def last_call
```
colinxfleming: This is pretty much just a drop-in replacement for `calls.sort_by(created_at).last` in `days_since_last_call`. Makes it look a little cleaner, I thought.

app/models/concerns/statusable.rb
```diff
@@ -27,4 +29,11 @@ def contact_made?
     end
     false
   end
+
+  def days_since_last_call
+    return 0 if calls.blank?
```
colinxfleming: A couple notes about this:

* If a patient doesn't have any calls this will choke, so we add the safety condition up top
* `calls.sort_by...` is replaced with `last_call`, but the core of your flow here is still much intact

lomky:
@colinxfleming my guess for why the 121 days isn't quite working is that we're treating all days as 24 hours with the `day=86_400`, but at least twice a year we have a non-24 hours day (a 23 hour day each spring and a 25 hour day each fall), so when you are doing your time-math and then testing with Ruby's date-math, you get disagreement.

I suspect if you can do the function-math with the ruby date-logic, the test would stop failing.

colinxfleming:
Per usual @lomky was right about this. Made an adjustment to treat them both like dates and it works much better!

Will merge after CI clears.

test/models/patient_test.rb
```diff
@@ -524,6 +524,13 @@ class PatientTest < ActiveSupport::TestCase
 
       # it 'should update to "Pledge Paid" after a pledge has been paid' do
       # end
+      it 'should update to "No contact in 120 days" after 120ish days of no calls' do
+        create :call, patient: @patient, status: 'Reached patient', created_at: 121.days.ago
+        assert_equal Patient::STATUSES[:dropoff], @patient.status
+
+        create :call, patient: @patient, status: 'Left voicemail', created_at: 119.days.ago
```
lomky: The twitchy bit in my head wants this to be 120 instead of 119 to make the boundary clear, but that isn't critical.

colinxfleming: I think that's pretty legit! I'll alter that now.

PR #1056, opened 2017-05-15
Fixes issue #1053 remove zip code field

### LINKED ISSUE #1053: Remove zip code

We don't really need to track zip code in this app, so let's take it out.

Patient information security, not keeping extra stuff around.

Remove the field from the model in `app/models/patient.rb`!

* **Anything else?**
See discussion in #306

itroman:

Fixes issue #1053 remove zip code field

* Remove the field from the model in `app/models/patient.rb!`

* Fixes #1053

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -61,7 +61,6 @@ class Patient
   field :age, type: Integer
   field :city, type: String
   field :state, type: String
-  field :zip, type: String
   field :county, type: String
   field :race_ethnicity, type: String
   field :employment_status, type: String
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -41,8 +41,6 @@
 
           <%= f.text_field :county, label: 'County', autocomplete: 'off' %>
 
-          <%= f.text_field :zip, label: 'ZIP', autocomplete: 'off' %>
-
           <h3>Other Contact</h3>
             <%= f.text_field :other_contact, autocomplete: 'off', label: 'Other contact name' %>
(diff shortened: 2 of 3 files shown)


colinxfleming:
Looking good @itroman , thanks for the quick turnaround on this!

No good deed goes unpunished, though -- It looks like our test suite is failing, and that we need to remove ZIP from a few other places. I would guess we'll need to do it in the following places:

* `app/views/patients/_patient_information.html.erb` (the form field)
* `app/views/patients/data_entry.html.erb` (form field)
* `test/integration/update_patient_information_test.rb` (two spots, the `fill_in` and the `assert`)
* `test/integration/data_entry_test.rb` (two spots, the `fill_in` and the `assert`)

After those changes, you can run `rake test` in a terminal to run the test suite. If it comes back green, we're good -- if it comes back red, we might have a little more debugging to do, but it should spit out some helpful messages.

Sorry this turned out to be a little more complicated than I thought!

(PS if you get a chance please pop into our slack channel! Link in the readme.)

itroman:
Alrighty!  I'll get goin on this.

itroman:
Made additional changes to:

- `app/views/patients/_patient_information.html.erb`
- `test/integration/update_patient_info_test.rb`

Test results were:  `366 tests, 783 assertions, 0 failures, 0 errors, 0 skips`

so hopefully that should do it.

colinxfleming:
Looking good, my dude! Thanks for digging in and turning these changes around so quick. Merging now, and this will go into prod on Sunday. 

How'd you find us, if you don't mind my asking?

colinxfleming:
(ps welcome aboard!  💯 🔥 🌻 )

PR #1059, opened 2017-05-17
Omnibus PR to abstract DCAF stuff out of the app

### LINKED ISSUE #999: Abstract DCAF specific stuff out of app
There's some DCAF specific stuff in the app that we need to abstract out. Here's our general ticket to log, for squashing in other tickets. 

Stuff to generalize off the top of my head:

- [x] the pledge attribute is called `dcaf_soft_pledge`
- [x] PDF generation is hardcoded to dcaf's needs -- I have an example of the form but this is going to be pretty involved. In the short term we should just shut it off for Baltimore I think and aim to have it in place when they start it up in June.

colinxfleming:

This weekend is the first milestone of rollout, when we spin up an instance for them to start doing data entry. This is an omnibus PR to get that out.

(Please pardon the merge commit scruff!)

Should be one commit per thing in the related issue #999.

* Fixes #999


colinxfleming:
On further reflection, we definitely should not do this as an omnibus because it's going to touch a lot more files than I thought it would. I'm going to do it in component PRs instead, starting with the model changes which we have to roll out, and then we'll do it by grep after that.

PR #1062, opened 2017-05-20
Add optional ability to shut off pledge generator

colinxfleming:

We're unlikely to have PDF generation in place for Baltimore by the time they launch. This adds the ability to shut it off with an env var, and display a message about filling out the form the old way instead.

* Shut off pledge generator via env var

No issue #.


colinxfleming:
Tagging in @DarthHater for review

PR #1063, opened 2017-05-20
Enable setting user role on account creation

colinxfleming:

Long overdue, this adds the ability to set another user as an admin.

* Allows an admin to set another user's role on create

Nice and simple:

screen shot 2017-05-20 at 1 44 52 am

No issue.

@lomky and/or @DarthHater would appreciate a review, since this has a security implication!


colinxfleming:
Thank you @lomky !

Issue #274, opened 2016-05-08
Render inline errors in forms with rails-bootstrap-forms gem


colinxfleming:
@ewiggins and I were working through a validation issue in #271  and in the process, it became clear that we aren't doing an excellent job of rendering form errors.

It's an open question of whether we should do this before MVP in July or not, but I do want to get it on the radar either way. 

@mebates @adinneen want to weigh in here? Should we roadmap this as part of mvp? 

Implementation-wise, the paths of least resistance are a flash notification up top or inline form errors. Will leave it to our UX friends to determine what's best here.

mebates:
If we can build them in by July, that would be awesome. Could be a good think to work on in the next two weeks. I would recommend following the USWDS design, if that's easy enough: https://standards.usa.gov/form-controls/#text-inputs

colinxfleming:
Does the uswds have a standard library or something like that, by chance?

The path of least resistance is probably just using whatever is standard in
the bootstrap forms library we are using, which is probably fairly similar.

Agree that having this in before launching would be fantastic. Let's make
sure we have it on the list.

colinxfleming:
For MVP let's go with the functionality baked into the bootstrap forms gem we're using. Here's a link to this: https://github.com/bootstrap-ruby/rails-bootstrap-forms#alert-messages

Let's start by implementing for the main form in `app/views/pregnancies/edit` and `app/views/pregnancies/_new_pregnancies`. 

(A possible sticking point is that we have fairly minimal model validation at this time; feel free to whip up some fake stuff to make sure errors are rendering properly in testing.)

colinxfleming:
Renaming and retagging to reflect the game plan.

colinxfleming:
Dropping this from MVP

colinxfleming:
Wonder if it would also be nice to have some marker of 'saved!' pop up when a field saves.

colinxfleming:
This is currently our oldest not-fixed issue, and it's kind of a significant one. I would love to prioritize this. Here is a potential strategy that I think might work:

* create a file called something like `update_failed.js.erb`
* change the controller action in the case of error to return `update_failed.js.erb` instead of an internal server error
* if a patient saves cleanly, flash a success message `Patient data saved!` at the top of the form (rendering the `messages` partial should do the trick here)
[...]

PR #1064, opened 2017-05-20
Submit Pledge Dynamic Update

tingaloo:

Add ajax reloading after submitting/cancelling pledge

* Refreshes partials on pledge_sent update
* (etc)

* Fixes #X
* Bumps #Y


tingaloo:
@colinxfleming Here is what I had setup, you can play around with it and see what can be improved. It mucks up the patient controller but I wanted to see what you think.

colinxfleming:
This is really cool! Thanks for this initial stab at it! I'll take a deeper look once Baltimore is live, if that's okay.

Just so you know where my head's at generally on this front: I think a thing that we currently do a bad job at is providing confirmation that changes our CMs are making stuck. See for example #274 in addition to #490 , and probably a few other spots as well. Your changes in making the pledge modal populate dynamically is a major step in the right direction, on that front.

tingaloo:
This is feedback UX dependent so this can wait until you have some workflows down. This is a PR I don't mind keeping open.

Issue #812, opened 2017-01-14
Patient archiving


colinxfleming:
Time to make an issue for this. Here is the issue: 

* There are some patient records who we are functionally done with; either people who have elected to carry their pregnancies, or who have had their procedures with assistance from us, or who have just timed out past the point where they need our help (120 days or so).
* We no longer really need a lot of their personally identifiable information (phone, name, etc) after this point.
* We also may need to create another entry for them if they seek our services again in the future, which conflicts with our validation on unique phone. (Right now we have a hacky workaround of setting a clearly fake phone number.)
* Thus, we need a way to 'archive' a patient to strip their information and get them out of rotation except for reporting. 

I think what this looks like from a technical perspective is wiping the phone numbers, setting an archive flag, and changing the name to a hash. A few caveats:

* People with the same phone number and name should have the same hash, so we can still identify people who are second people;
* Archiving should be irreversible.

nimiro:
The issue is pretty well summarized above, but here are tonight's discovery notes just in case:
https://github.com/colinxfleming/dcaf_case_management/wiki/Anonymization-notes

colinxfleming:
I'm ready for this. This is one of the last bigger technical lifts we need to worry about. @lomky you're up. 

* Let's throw code relating to this in a `patient_archiving_helper` or `patient_archiving_concern` and include that in the patient model
* Probably the sum result of this should be something like `@patient.archive!` that does everything we need it to

lomky:
It may be useful to maintain the data in 'Relationship to other contact', but that feels a little iffy. We probably do want to know whether they had any Other Contact data or not, even if we scramble the data itself

lomky:
We probably need to nuke the whole change log, but how can we save useful info safely?

Create fields for "changed appointment date", "changed clinic location", those but after the pledge is sent? etc?

Maybe follow the same process for generating a CSV export  blacklisting entry by entry bsaed on the field being changed? "name" -> scramble, "notes" -> scramble, appt date -> leave be

colinxfleming:
@lomky you are definitely right that we want to know whether someone had an alt contact or not, and that we don't want to know the PII of that person. Relationship is a little borderline and I think I'd rather err on the side of not saving it. 

re: persisting other contact info, in #1004 we're rolling out a `has_other_contact?` convenience method. Maybe we can have that as a persisted boolean field, and have it as part of a `after_save` hook.

[...]

PR #1061, opened 2017-05-20
Patient model changes: DCAF soft pledge => fund pledge, resolved without dcaf => resolved without fund

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

This changes `patient.resolved_without_dcaf` to `patient.resolved_without_fund`, and `patient.dcaf_soft_pledge` to `patient.fund_pledge`. And makes a lot of associated changes in code. 

It also adds a pair of convenience constants, `FUND` and `FUND_FULL`, corresponding to fund abbreviation and full name, respectively.

This starts abstracting DCAF specific stuff out of the app, so we can ~ fulfill our destiny ~ and have a generalized solution.

This pull request makes the following changes:
* `patient.resolved_without_dcaf` to `patient.resolved_without_fund`
* `patient.dcaf_soft_pledge` to `patient.fund_pledge`
* add convenience constants

It relates to the following issue #s: 
* Bumps #999 

@DarthHater for review please!

PR #1065, opened 2017-05-20
Moar export! Add pregnancy fields, clinic name, age range

colinxfleming:

@cfinucane you built a really remarkably extendable baseboard for the export function, color me impressed!, and also grateful! This adds fields I think we definitely want to export from patient, but please check my work.

It removes special circumstances and switches age to a range as well, and adds clinic info.

* Adds pregnancy fields
* Trims or adjusts some others per conversation in #812 
* Adds a model for getting at related objects

Still TK, handling calls and fulfillments.

* Bumps #965 

@cfinucane @lomky for review


app/models/concerns/exportable.rb
```diff
@@ -21,16 +21,56 @@ module Exportable
...
 
+  def export_clinic_name
+    clinic.try :name
+  end
+
+  def age_range
+    case age
+    when nil
+      nil
+    when 0..18
```
lomky: `0..17`?

app/models/concerns/exportable.rb
```diff
@@ -21,16 +21,56 @@ module Exportable
...
-    "Special Circumstances" => :special_circumstances 
-  }
+    # "Special Circumstances" => :special_circumstances,
+    "LMP at intake (weeks)" => :last_menstrual_period_weeks,
+    "Race or Ethnicity" => :race_ethnicity,
+    "Employment status" => :employment_status,
+    "Abortion cost" => :procedure_cost,
+    "Patient contribution" => :patient_contribution,
+    "NAF pledge" => :naf_pledge,
+    "Fund pledge" => :dcaf_soft_pledge,
```
lomky: assuming this is after #1061, this will need a bump

app/models/concerns/exportable.rb
```diff
@@ -21,16 +21,56 @@ module Exportable
...
+    "Clinic" => :export_clinic_name,
+    "Pledge sent" => :pledge_sent,
+    "Resolved without fund assistance" => :resolved_without_dcaf,
```
lomky: assuming this is after #1061, this will need a bump

app/models/concerns/exportable.rb
```diff
@@ -21,16 +21,56 @@ module Exportable
...
+      '18-24'
+    when 25..34
+      '25-34'
+    when 35..44
+      '35-44'
+    when 45..54
+      '45-54'
+    when 55..100
+      '55+'
+    end
```
DarthHater: While improbable you could likely do a else case to handle a typo. Suggest 'Centarian' :)

app/models/concerns/exportable.rb
DarthHater: Damn! Lomky with the good eyes

app/models/concerns/exportable.rb
colinxfleming: The test is a good flag! So is the default case. I made it 'bad value' which seemed okay.

colinxfleming: Resolved in ded1e93

app/models/concerns/exportable.rb
colinxfleming: oh man, great, great catch.

app/models/concerns/exportable.rb
colinxfleming: Bumped in 5866b74

colinxfleming:
Removed `identifer` export in aea08ab5ca43ca5b85c0a8b43bbd1ad673fd6aa9 . Thanks all for the sharp and thoughtful feedback.

PR #1069, opened 2017-05-21
Correct line handler in patients to take a correct env var

colinxfleming:

It turns out that `LINES` is a reserved env var in heroku, so this moves the line setter to `DARIA_LINES`, which is thankfully NOT a reserved env var.

Requires a heroku env var change after merge.

* Changes `LINES` variable to draw from the env var `DARIA_LINES`
* Alters patient model to use `LINES` instead of hardset option

Bumps #999

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -50,7 +50,7 @@ class Patient
 
   # Contact-related info
   enum :voicemail_preference, [:not_specified, :no, :yes]
-  enum :line, [:DC, :MD, :VA] # See config/initializers/lines.rb. TODO: Env var.
+  enum :line, LINES # See config/initializers/env_vars.rb
   field :spanish, type: Boolean
   field :initial_call_date, type: Date
   field :urgent_flag, type: Boolean
--- a/config/initializers/env_var_constants.rb
+++ b/config/initializers/env_var_constants.rb
@@ -1,6 +1,6 @@
 # Definition of line constants
-LINES = if ENV['LINES'].present?
-          ENV['LINES'].split(',')
+LINES = if ENV['DARIA_LINES'].present?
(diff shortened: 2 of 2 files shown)


app/models/patient.rb
```diff
@@ -50,7 +50,7 @@ class Patient
 
   # Contact-related info
   enum :voicemail_preference, [:not_specified, :no, :yes]
-  enum :line, [:DC, :MD, :VA] # See config/initializers/lines.rb. TODO: Env var.
+  enum :line, LINES # See config/initializers/env_vars.rb
```
lomky: Should this be `DARIA_LINES`?

colinxfleming: Nah, it reads from the constant set in congig/initializers/env_vars

colinxfleming:
Take that, ci

app/models/patient.rb
colinxfleming: No longer in transit, so better and more complete answer (sorry @lomky !)

`LINES` is set as an app constant in `config/initializers/env_var_constants.rb`, reading from `ENV['DARIA_LINES']` and turning that into an array of symbols (generally, `[:DC, :MD, :VA]`. In turn, we can use `LINES` as a constant throughout the app.

If I'm reading you right, it sounds like there's some wisdom to changing the constant `LINES` to `DARIA_LINES` to match the env var names, and I agree with that - if I am in fact reading you right, let's file that as an issue. Thoughts?

colinxfleming:
Going to invoke admin and merge to get baltimore app out.

TODOs from this PR: 

* change `LINES` constant to `DARIA_LINES` per PR above
* document how lines are used by the app better, per conversation, because this needs to be clearer and it's not terribly clear from the code

PR #1073, opened 2017-05-25
Draft user management panel

### LINKED ISSUE #686: Build out draft user admin panel

tingaloo:

I made some headway with the admin panel.  I went fast and threw lots of features with minimal testing (will come later!) Lots of code overlap with ```patients``` (#search, call sorting, helpers), which needs addressing. Most of the features (locking, search, sort) work. Again no tests yet but I wanted to throw this out and get a review before finishing up. Also this may be too much in one PR so let me know if you want me to break it down to smaller PRs!

On the design PDFs, there were checkboxes on index to select MULTIPLE users to lock or reset password. Was this meant as a multiple selection or should it be more so like a radio selection? I left this as WIP.

* Add search, sort by date_created, lock, and password reset features.

screen shot 2017-05-25 at 8 12 05 am

screen shot 2017-05-25 at 8 12 42 am

* Fixes #686 (without tests)


app/assets/javascripts/calls.coffee
```diff
@@ -39,7 +39,7 @@ jQuery ->
     )
 
     # make the columns of all of the tables sortable
-    autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table' ]
+    autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table', '#user-list table']
```
tingaloo: should this be in a shared .js file, users.coffee? It certainly does not belong here.

app/controllers/users_controller.rb
```diff
@@ -50,6 +106,15 @@ def reorder_call_list
 
   private
 
+  # copied from DashboardsController
+  def searched_for_name?(query)
+    /[a-z]/i.match query
```
tingaloo: repeated helper method from dashboards controller

app/helpers/users_helper.rb
```diff
@@ -6,4 +6,17 @@ def user_role_options
...
+    raise 'Bad datatype' unless %w(string string-ins int float).include? type
+    content_tag :th, data: { sort: type } do
+      safe_join [column_name, autosort_arrow_span(local_assigns)], ' '
+    end
+  end
+
+  def autosort_arrow_span(local_assigns)
+    return '' unless local_assigns[:autosortable]
+    content_tag(:span, '[-]', class: 'arrow')
+  end
```
tingaloo: repeated from dashboards_helper,

app/models/concerns/searchable.rb
```diff
@@ -20,8 +20,35 @@ def search(name_or_phone_str, lines = LINES)
...
     end
 
+    def search_users(name_str)
+      name_regexp = /#{Regexp.escape(name_str)}/i
+      all_matching_names = find_name_matches_user name_regexp
+
+      sort_and_limit_user_matches(all_matching_names)
+
+    end
+
```
tingaloo: I can refactor this so that #search takes a parameter of either ```Patient``` or ```User```, this will change some tests. But it should be worth it.

colinxfleming: I'm interested in this! I feel like in some ways User and Patient are such radically different models, where we care about much different things, that I'm okay not having them share a search method. I think that if this winds up being a big pain in the neck it would be 100% acceptable to move this into `PatientSearchable` and have another `UserSearchable` concern. 

A thing I **would** like is to continue to invoke `Patient.search` in controllers and models. So let's make sure whatever solution we implement, we're still doing that where appropriate.

app/models/concerns/searchable.rb
```diff
@@ -20,8 +20,35 @@ def search(name_or_phone_str, lines = LINES)
       sort_and_limit_patient_matches(all_matching_names, all_matching_phones, all_matching_identifiers)
     end
 
+    def search_users(name_str)
```
colinxfleming: I'd love to be able to search on email too, if that's something that can be in the cards! But I'm okay with rolling that out separate since there's already a lot swirling in this pr.

app/models/user.rb
```diff
@@ -3,7 +3,13 @@
 class User
   include Mongoid::Document
   include Mongoid::Enum
+  include Mongoid::Timestamps
```
colinxfleming: whoops! haha. thanks for fixing this!

app/views/users/_search_form.html.erb
```diff
@@ -0,0 +1,4 @@
+  <%= bootstrap_form_tag url: '/users/search', remote: true, layout: :inline do |p| %>
```
colinxfleming: This is in part my own weird junk but I vastly prefer the the route syntax here. Can you swap this to use `users_search_path` instead of `url`?

app/views/users/_users_table.html.erb
```diff
@@ -0,0 +1,33 @@
+<div id="user-list" class="margin-bottom">
+  <table class="table border-side">
+    <thead>
+      <tr>
+        <% if local_assigns[:sortable] %>
+          <th><!-- handle --></th>
+        <% end %>
+        <th><input type="checkbox" id="select-all-users"></th>
+        <th>FULL NAME</th>
```
colinxfleming: Pop these into regular case rather than upcase?

app/views/users/_users_table.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    </thead>
+    <tbody>
+      <% users.each do |user| %>
+        <tr>
+          <td><input type="checkbox" id="select-<%= user.id %>"></td>
+          <td><%= link_to user.name, edit_user_path(user) %></td>
+          <td><%= user.role %>
+          <td><%= user.last_sign_in_at %>
+          <td><%= user.created_at %>
+          <% if !user.locked_at.nil? %>
```
colinxfleming: This condition is a pretty good candidate to move to its own method in `users_helper`! Esp since it looks like it's used in edit too.

config/environments/development.rb
```diff
@@ -14,7 +14,7 @@
   config.public_file_server.enabled        = false
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
```
colinxfleming: Why the swap on this? I assume sending the unlock token?

config/routes.rb
```diff
@@ -4,7 +4,12 @@
     get 'dashboard', to: 'dashboards#index', as: 'dashboard'
     get 'reports', to: 'reports#index', as: 'reports'
     post 'search', to: 'dashboards#search', defaults: { format: :js }
-    resources :users, only: [:new, :create, :index]
+
+    post 'users/search', to: 'users#search', defaults: { format: :js }
```
colinxfleming: Slap an `as` on this so we can refer to the named route elsewhere!

config/routes.rb
```diff
@@ -4,7 +4,12 @@
...
+    get 'users/reset_password/:user_id', to: 'users#reset_password', as: 'reset_password'
```
colinxfleming: So this is a little more rest-y, can you swap the placement of reset_password and :user_id here? So have it end up as `users/:user_id/reset_password`, and below, `users/:user_id/toggle_lock`.

test/controllers/users_controller_test.rb
```diff
@@ -99,25 +99,25 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
-    it 'should respond success' do
-      assert_response :success
-    end
-
-    it 'should populate the user call order field' do
-      assert_not_nil @user.call_order
-      assert_equal @ids, @user.call_order
-    end
-  end
+  # describe 'reorder call list' do
```
colinxfleming: I promise I will fix this stupid test soon :(

app/assets/javascripts/calls.coffee
tingaloo: Good idea, will do.

app/controllers/users_controller.rb
tingaloo: You're right, I got it confused with ```searchable.rb```

app/helpers/users_helper.rb
tingaloo: The goods news is you're not getting old, but we need to move this code out to a shared helper. I was thinking ```tables_helper.rb``` that contains all table logic. Thoughts?

app/models/concerns/searchable.rb
tingaloo: Got it, i'm going to move all my helpers into concerns ```UserSearchable``` and keep original ```Searchable``` untouched. I'll try to remember to migrate into ```PatientSearchable``` after this PR is merged.

config/environments/development.rb
tingaloo: I was testing out the password reset email, it wasn't working in dev so I was doing some random troubleshooting. I'll leave this for future PR.

colinxfleming: Mysterious! Usually actionmailer emails log the full content to the rails console in dev, but if you give me steps to recreate I can look into this.

app/models/concerns/searchable.rb
colinxfleming: Sounds like a plan. Thanks bud!

app/helpers/users_helper.rb
colinxfleming: `tables_helper` sounds A+++, good call!

config/initializers/devise.rb
```diff
@@ -168,7 +168,7 @@
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  config.unlock_strategy = :time
+  config.unlock_strategy = :both
```
colinxfleming: I think from a business perspective, the only way we want someone to be able to unlock is if an admin unlocks them from this panel, and I think this might actually interfere with 'failed login attempts', which results in like a 15 minute 'chill out' ban.

This might be a thing where really what we're looking for is a `disabled` field rather than `locked`, since `locked` seems to mean something in the context of devise. (If that's the case let's do it in a separate PR and just leave the 'lock account' button nonfunctional for now)

app/views/users/_users_table.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <th>Role</th>
+        <th>Last Login</th>
+        <%= th_autosortable 'Date Created', 'string', local_assigns %>
+        <th>Status</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% if users.nil? %>
+      <% else %>
+        <% users.each do |user| %>
```
colinxfleming: Does `users.each` without the nil guard not work here? Usually you can `each` on empty arrays, which is what users should probably be. Let me know if that's a problem and we can write a workaround. $$

app/models/concerns/patient_searchable.rb
```diff
@@ -1,5 +1,5 @@
 # Methods pertaining to patient search
-module Searchable
+module PatientSearchable
```
colinxfleming: Looks like there mightbe something to sort out here still -- `search_users` looks like it's still in there?

app/helpers/users_helper.rb
```diff
@@ -6,4 +6,12 @@ def user_role_options
       ['Admin', 'admin']
     ]
   end
+
+  def status user
```
colinxfleming: Let's revisit this after we figure out the lock strategy stuff.

config/initializers/devise.rb
tingaloo: Hm, I think a manual unlocking strategy would be best. If you can write an issue with the unlock workflow, that would be awesome.

app/controllers/users_controller.rb
```diff
@@ -2,22 +2,74 @@
...
+        @user.lock_access!
+      end
+      redirect_to edit_user_path @user
+    end
+    # TODO multiple locks?
+    # respond_to { |format| format.js }
+  end
+
+  def reset_password
+    @user = User.find(params[:user_id])
```
tingaloo: I think i'm missing something but ```User.find(params[:id])``` is different from ``` User.find(params[:user_id])```?

colinxfleming:
I feel pretty good about this. I think we can spin some of the peripheral stuff into issues, because this is pretty gigantic. I'm going to do a pass of suggested lint edits and things to spin into issues, and then let's merge.

app/assets/javascripts/users.coffee
```diff
@@ -0,0 +1,9 @@
+ready = ->
```
colinxfleming: I gotta be honest, being able to lock the entire set of users all at once kinda skeezes me out - worried about a bad actor worst case scenario, and I can't really think of a time when someone would need to mass reset passwords like that. Can we comment it out for now and create an issue on how to deal with it?

app/controllers/users_controller.rb
```diff
@@ -1,23 +1,76 @@
...
+    #     @user.lock_access!
+    #   end
+    #   redirect_to edit_user_path @user
+    # end
+    # # TODO multiple locks?
+    # # respond_to { |format| format.js }
+  end
+
+  def reset_password
+    @user = User.find(params[:user_id])
```
colinxfleming: I see what you're saying about the id thing. Set `find_user` as a before_action here, and we'll make a routing change to correct this.

app/controllers/users_controller.rb
```diff
@@ -1,23 +1,76 @@
...
+
+    # TODO doesn't work in dev
```
colinxfleming: Create an issue for this, if you'd be so kind!

app/controllers/users_controller.rb
```diff
@@ -1,23 +1,76 @@
...
+    @user.send_reset_password_instructions
+
+    flash[:notice] = 'Successfully sent password reset instructions to ' + @user.email
```
colinxfleming: Can we use string interpolation here? `"blah blah blah #{@user.email}"`. More rails-y.

app/controllers/users_controller.rb
```diff
@@ -1,23 +1,76 @@
...
+    if @user.update_attributes user_params
+      flash[:notice] = 'Successfully updated user details'
+    else
+      flash[:alert] = 'Error saving user details'
+    end
+    redirect_to edit_user_path @user
+  end
+
   def create
+    # should confirm admin user and remove runtime error?
```
colinxfleming: Not sure I follow, can you spin up an issue for this?

app/controllers/users_controller.rb
```diff
@@ -1,23 +1,76 @@
...
+    redirect_to edit_user_path @user
+  end
+
+  def update
+    if @user.update_attributes user_params
+      flash[:notice] = 'Successfully updated user details'
+    else
+      flash[:alert] = 'Error saving user details'
+    end
+    redirect_to edit_user_path @user
```
colinxfleming: Can we have this redirect to `users_path` on success, and `render 'edit'` on failure? See `PatientsController#data_entry_create` for an example of this.

app/models/user.rb
```diff
@@ -3,7 +3,13 @@
...
   include Mongoid::Userstamp::User
+  include Mongoid::History::Trackable
+
+  include UserSearchable
+  track_history   :track_create   =>  true    # track document creation, default is false
```
colinxfleming: Let's use the boilerplate track code here:

```
  track_history on: fields.keys + [:updated_by_id],
                version_field: :version,
                track_create: true,
                track_update: true,
                track_destroy: true
```

app/models/user.rb
```diff
@@ -59,7 +65,7 @@ class User
 
   ## Lockable
   field :failed_attempts, type: Integer, default: 0
-  # field :unlock_token,    type: String # Requires unlock strategy
+  field :unlock_token,    type: String
```
colinxfleming: Let's turn this off for now, if that's okay. I want to do this in its own PR

app/views/users/index.html.erb
```diff
@@ -1,36 +1,15 @@
 <h1>User Account Management</h1>
 <div class="row" style="padding-top: 10px; padding-bottom: 20px;">
-  <div class="col-md-8">
+  <div class="col-md-12">
     <%= link_to "Add New User", new_user_path, class: "btn btn-primary" %>
-  </div>
-  <div class="col-md-4">
+    <%= link_to "Reset Password (WIP)", "#", class: "btn btn-primary" %>
```
colinxfleming: throw a `disabled: true` in here until we have this set?

app/views/users/index.html.erb
```diff
@@ -1,36 +1,15 @@
...
+    <button type="button" class="btn btn-primary">Toggle Lock (WIP)</button>
```
colinxfleming: throw a `disabled` attribute in here until we have this set?

app/views/users/index.html.erb
```diff
@@ -1,36 +1,15 @@
...
     <div style="float: right;">
-      <%= bootstrap_form_tag url: '/search', remote: true, layout: :inline do |p| %>
-        <%= p.text_field :search, placeholder: 'search', hide_label: true, class: 'search_form_input' %>
-        <%= p.submit 'Search', class: 'btn btn-primary' %>
-      <% end %>
+      <%= render partial: 'users/search_form', locals: { autosortable: true } %>
```
colinxfleming: Thing to add as an issue: the stupidtable sort doesn't work after a search. Let's fix in a separate PR.

app/views/users/search.js.erb
```diff
@@ -0,0 +1,6 @@
+
+<% if @results.any? %>
+  $('#user-list').html("<%= escape_javascript(render partial: 'users/users_table', locals: { users: @results }) %>");
+<% else %>
+  $('#user-list').html("<%= escape_javascript(render partial: 'users/users_table', locals: { users: nil }) %>");
```
colinxfleming: Not sure this path fires on no users, let's file as an issue

config/initializers/devise.rb
```diff
@@ -168,7 +168,7 @@
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  config.unlock_strategy = :time
+  # config.unlock_strategy = :none
```
colinxfleming: Swap this back to `:time` for now please!

config/routes.rb
```diff
@@ -4,7 +4,12 @@
     get 'dashboard', to: 'dashboards#index', as: 'dashboard'
     get 'reports', to: 'reports#index', as: 'reports'
     post 'search', to: 'dashboards#search', defaults: { format: :js }
-    resources :users, only: [:new, :create, :index]
+
+    post 'users/search', to: 'users#search', as: 'users_search', defaults: { format: :js }
+    get 'users/:user_id/reset_password', to: 'users#reset_password', as: 'reset_password'
```
colinxfleming: in `get 'users/:user_id/reset_password`, change the `:user_id` to just `:id`. That follows the rails / rest pattern a bit closer, and will let us take advantage of the `find_user` before_action in the controller. Sound good?

config/routes.rb
```diff
@@ -4,7 +4,12 @@
...
+    get 'users/:user_id/toggle_lock', to: 'users#toggle_lock', as: 'toggle_lock'
```
colinxfleming: Same deal here, but let's comment this out until we roll this out.

test/integration/user_management_test.rb
```diff
@@ -0,0 +1,149 @@
...
+
+  end
+
+  describe 'case manager' do
+    before do
+      log_in @user
+    end
+
+    it 'should not be able to access Admin links' do
+      assert has_no_link? 'Admin'
```
colinxfleming: can you throw a `visit users_path` in here and the data vol test as well and confirm that they're redirected back to the main dashboard, to emulate someone trying to get at this with just the url?

test/integration/user_management_test.rb
```diff
@@ -0,0 +1,149 @@
...
+    end
+
+    it 'allows email editing' do
+      fill_in 'Email', with: 'johan@gmail.com'
+      click_button 'Save'
+      wait_for_element 'Successfully updated user details'
+      # assert_text 'johan@gmail.com'
+    end
+
+    # TODO figure out a way to show error
```
colinxfleming: Spin this into an issue?

test/integration/user_management_test.rb
```diff
@@ -0,0 +1,149 @@
...
+      click_button 'Save'
+      wait_for_element 'Successfully updated user details'
+      assert_text 'johan'
+    end
+
+    it 'allows email editing' do
+      fill_in 'Email', with: 'johan@gmail.com'
+      click_button 'Save'
+      wait_for_element 'Successfully updated user details'
+      # assert_text 'johan@gmail.com'
```
colinxfleming: Spin this into an issue so we don't accidentally let it slip?

test/integration/user_management_test.rb
```diff
@@ -0,0 +1,149 @@
...
+    before do
+      log_in @admin
+    end
+
+    it 'should be able to access user management' do
+      click_link 'Admin'
+      assert_text 'User Management'
+      click_link 'User Management'
+      wait_for_element 'User Account Management'
+      click_link 'Add New User'
```
colinxfleming: Can you throw some `assert has_content? 'Whatever text'` checks in here, so it yields more helpful error messages if it ever chokes?

mebates:
Hey! Would love to see some fixes to make this align to the designs a bit more. (this can definitely happen later). Mainly, we want to keep one primary action (purple button) on each page. Secondary actions shouldn't also be purple buttons. Please review the mocks:

[DCAFwireframe110316_user_account_management (1).pdf](https://github.com/DCAFEngineering/dcaf_case_management/files/1053501/DCAFwireframe110316_user_account_management.1.pdf)

colinxfleming:
Per @tingaloo we've spun up stuff into issues at this point. We should unlink this from the panel, merge this, and then relink when we've got it in a working state. I'll take care of that.

I'll spin up 'hew to design' as an issue before I merge, per @mebates good flag here.

tingaloo:
You want me to remove the comments or do you mean just removing the lock/reset_password buttons? I think the features and tests are cold enough to not be invasive.

colinxfleming:
@tingaloo I meant taking it out of the navbar, but even that's probably not necessary -- we can just hide the lock and reset stuff. I'll handle this in the next day or two, I think you can consider this officially good to go. Thanks again for all your hard work on this -- it's looking great and I'm stoked to see it become a reality.

tingaloo:
@colinxfleming Thanks, I am scared of touching this again because I broke the search functionality while doing backwards fixes. I'll come back to this later if nobody wants to fix them.

PR #1079, opened 2017-05-28
Stop incrementing LMP after a certain point ctd

### LINKED ISSUE #1008: Stop incrementing LMP after a certain point

colinxfleming:

Very small syntactic bump to #1031 to get tests to pass.

* remove `patient` from initial call date ref

* stems from  #1031 
* Fixes #1008


colinxfleming:
@DarthHater can I get a quick review on this?

colinxfleming:
(or @KatSDC if you're on the grid this weekend)

app/models/concerns/last_menstrual_period_measureable.rb
```diff
@@ -30,7 +30,11 @@ def _last_menstrual_period_on_date(date)
     return nil unless initial_call_date && last_menstrual_period_weeks
     weeks = 7 * (last_menstrual_period_weeks || 0)
     days = (last_menstrual_period_days || 0)
-    (date - initial_call_date) + weeks + days
+    if (date - initial_call_date) + weeks + days > 280
```
DarthHater: Might want to just dry this out into another quick helper method since you repeat it.

colinxfleming: Good call! (Also good call on the test, I'll whip one up now)

PR #1080, opened 2017-05-29
Config the external_pledges, add test for all of them, fun fun fun

### LINKED ISSUE #1072: Turn external pledges list into a config

The list of places you can log an external pledge is hardcoded. We should move this to a config of some sort.

Abstracting DCAF out as the sole user.

Either setting this as an env var or setting this in a config model somewhere.

* **Anything else?**
Approximately part of #999

DarthHater:

This pulls the external_pledges list out of "code" and into a config file. 

* Adds external_pledges.yml
* Loads the external_pledges into Rails.configuration.external_pledges
* Modifies method to return config values
* Adds test to ensure list values are tested (now that they are config)

* Fixes #1072


lomky:
@DarthHater by testing for the explicit list, don't we just reintroduce the external sources into our codebase then? What if we tested explicitly for the values "Clinic discount" and "Other funds (see notes)", and then generally for 3+ items?

DarthHater:
@lomky it walks a funny line between a Unit Test and an IT. Testing the unit would be like "Yep it returns an array", which we can totally do. I chose to be explicit at least for the time being because I also made this a part of a yaml file, therefore, some "code". It's more like an IT at that level, but you know, YOLO.

colinxfleming:
@DarthHater I brought this up to speed with master and extended this following the same pattern for insurance. (It was SUPER easy to extend, very nicely done!)

What are the feelings about pushing this out as the interim fix so that Baltimore can enter their data? Anything else we should do to this? This probably solves the problem at least until we add in another fund.

colinxfleming:
👍 from me too! Merging.

PR #1084, opened 2017-05-29
Sub in Fund env var for DCAF in calls helper

colinxfleming:

Sub in `FUND` for DCAF in the voicemail preference part of the calls helper.

* What it says on the label

@KatSDC could I get you to approve these changes?

* Bumps #999


KatSDC:
This looks good to me!

PR #1085, opened 2017-05-29
Add skylight monitoring

### LINKED ISSUE #1017: Add skylight

I'd like to add a little more monitoring to the application than new relic gives us. I'd like to add Skylight for this.

Monitoring app health, more easily identifying pain points.

Directions at https://www.skylight.io/

* **Anything else?**
Low priority

colinxfleming:

Adds the skylight gem and config file, so we have more rails specific monitoring! I think this will help us identify where the pain points are, where loads are slow, and what keeps triggering logentries alerts.

I also added the appropriate api keys to the prod environment(s). This is free up until 100,000 requests so it's rolled out to just dcaf and baf, but not the sandbox env.

* Sets up skylight for accessible monitoring

@DarthHater I added you (and @lwaldsc and @lomky) as a collaborator to this. Can you approve these changes? I don't plan on rolling them out until this Sunday during out deployment window either way. 

* Fixes #1017 (brick squad)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,4 +77,5 @@ end
 group :production do
   gem 'puma'
   gem 'letsencrypt-rails-heroku'
+  gem 'skylight'
 end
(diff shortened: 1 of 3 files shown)


colinxfleming:
thanks @lomky !

PR #1087, opened 2017-06-02
Adds Patient#pledge_generated_by field

### LINKED ISSUE #1051: Add Patient#pledge_generated_by field

Track what user last generated a pledge, in addition to timestamps.

Auditing/Verifying who generated the last pledge

<strike>Three</strike>Four steps:
* Add field `pledge_generated_by` to Patient, as a user id
* Display on `_fulfillment` partial, below pledge amount
* Add to PatientsController#download update statement, to set `pledge_generated_by` as `current_user`
* Add to PatientsController test to confirm that the action is triggering

* **Anything else?**
nope!

itroman:

Added ability to track what user last generated a pledge, by user id.  

* Adds field pledge_generated_by to Patient, as a user id
* Displays pledge_generated_by on _fulfillment partial, below pledge amount
* Adds to PatientsController#download update statement, to set pledge_generated_by as current_user
* Adds PatientsController test to confirm that the action is triggering

* Fixes #1051 

All tests coming up green.  Let me know if there's anything I need to do to clean up the Patient model.  My first time working with MongoDB really, so lemme know if things need fixing.


colinxfleming:
Awesome, thank you so much @itroman ! Taking a look at this shortly.

app/models/user.rb
```diff
@@ -25,7 +25,8 @@ class User
 
   # Relationships
   has_and_belongs_to_many :patients, inverse_of: :users
-
+  belongs_to :pledge_generated_by, class_name: 'Patient', inverse_of: 'current_user'
```
colinxfleming: Fun fact about document stores: They really, really, really suck at relational stuff like this :) I think we can take this out until we have a specific use case for needing to see all pledges generated by a particular user, go ahead and revert this change, if you don't mind.

app/views/patients/_fulfillment.html.erb
```diff
@@ -9,6 +9,7 @@
           <ul>
             <li>Clinic: <%= @patient.clinic.try(:name) %></li>
             <li><%= FUND %> Pledge Amount: $<%= @patient.fund_pledge %></li>
+            <li>Pledge generated by: <%= @patient.pledge_generated_by %></li>
```
colinxfleming: 👍

test/controllers/patients_controller_test.rb
```diff
@@ -194,6 +194,7 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
       end
 
       refute_nil @patient.reload.pledge_generated_at
+      refute_nil @patient.reload.pledge_generated_by
```
colinxfleming: and tests!!! 😻

app/models/patient.rb
```diff
@@ -35,6 +35,7 @@ class Patient
   embeds_many :calls
   embeds_many :external_pledges
   embeds_many :notes
+  has_one 'current_user', class_name: 'User', inverse_of: :pledge_generated_by
```
colinxfleming: If memory serves, this adds a `current_user` attribute to patient objects, which isn't *quite* what we're looking for, I don't think. 

I think there's a good analog here in how we connect `clinic` to `patient` objects. We can get to the `clinic` object by doing `@patient.clinic`, but the attribute it's actually storing is `@patient.clinic_id` (it's just smart enough to do the lookup on its own).

I think that if we follow that pattern and alter it to a `belongs_to :pledge_generated_by`, it'll follow that same pattern and store an id nice and clean-like. And since there won't be an `inverse_of` to reference, we can just set that to `nil`.

short version: I think if we swap this to a `belongs_to :pledge_generated_by` with the same class name and an `inverse_of: nil`, I think this will blend beautifully and we'll be able to retrieve the user with a `@patient.pledge_generated_by`, nice and clean!

app/controllers/patients_controller.rb
```diff
@@ -51,6 +51,7 @@ def download
                                     params[:case_manager_name].to_s)
                                .generate_pledge_pdf
       @patient.update pledge_generated_at: Time.zone.now
+      @patient.update pledge_generated_by: current_user.id
```
colinxfleming: We can save using an object here, it'll make it a little more readable (get them small gains!!). Go ahead and take off the `id` here and just save `current_user` and that should blend okay.

Could you also add this update to the previous one, so it updates both `pledge_generated_at` and `pledge_generated_by` at the same time? No need to do two db calls when we can do one!

app/models/patient.rb
```diff
@@ -81,6 +82,7 @@ class Patient
   field :pledge_sent, type: Boolean
   field :resolved_without_fund, type: Boolean
   field :pledge_generated_at, type: DateTime
+  field :pledge_generated_by, type: String
```
colinxfleming: After making the `belongs_to` change, we should be able to ditch this line (since the `belongs_to` will define this field instead), so go ahead and strip this, if that's okay.

itroman:
Alrighty...all the changes you suggested should be done.  Everything looks good to go vis-a-vis the tests.  Lemme know if I missed somethin'!

colinxfleming:
This is great work @itroman , nicely done! We'll push it into production later this week.

PR #1090, opened 2017-06-03
fix modal load

### LINKED ISSUE #1089: Turbolinks issue on pledge modal

It is with a heavy heart that I must inform us that our enemy turbolinks is at it again. The issue that had us stuck on version 2 or whatever, the pledge modal loading empty when nothing's happened on a page yet, has reared its ugly head again.

Steps to recreate:

* `rake db:seed`
* log in to app
* click on patient `Clinic and Appt - 2` in call log
* click on 'Submit Pledge'
* oh no!

screen shot 2017-06-03 at 11 34 37 am

* update the `Appointment date` field - increment a day, do something to trigger the onChange action/get some javascript to fire
* Click on 'Submit Pledge' again and see it show up now (it's cancel pledge for me, but the point is the modal has data):

screen shot 2017-06-03 at 11 40 23 am

Pretty core functionality of sending pledges, so we should prioritize resolution and not deploy this Sunday until this is fixed, if possible. 

Adjust the magic turbolinks switch, I assume. We should also add an integration test for this particular case, if possible, since it's bitten us in the behind a few times.

* **Anything else?**
@tingaloo you around to give this a fixer upper?

tingaloo:

Fix submit pledge modal, this was pushed before the turbolinks merge so that'd explain the failure of modal steps populating.

* Added turbolinks loader

* Fixes #1089

### DIFF
--- a/app/assets/javascripts/external_pledges.coffee
+++ b/app/assets/javascripts/external_pledges.coffee
@@ -4,6 +4,6 @@
 
 jQuery ->
   # use event that works with rails turbolinks
-  $(document).on 'ready page:load', ->
+  $(document).on 'turbolinks:load', ->
     # select all elements with id ending with "modal"
     $('[id$=modal]').modalSteps()
--- a/test/integration/submit_pledge_test.rb
+++ b/test/integration/submit_pledge_test.rb
@@ -7,13 +7,22 @@ class SubmitPledgeTest < ActionDispatch::IntegrationTest
     @clinic = create :clinic
     @patient = create :patient, clinic: @clinic,
                                 appointment_date: Time.zone.now + 14,
-                                fund_pledge: 500
+                                fund_pledge: 500,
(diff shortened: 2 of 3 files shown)


colinxfleming:
MY DUDE

Can you merge that failing test in #1091 into here so we can confirm it blends properly?

colinxfleming:
This fixes the issue in #1089 and the test failures are due to flakiness. I'm merging.

app/assets/javascripts/external_pledges.coffee
```diff
@@ -4,6 +4,6 @@
 
 jQuery ->
   # use event that works with rails turbolinks
-  $(document).on 'ready page:load', ->
+  $(document).on 'turbolinks:load', ->
```
colinxfleming: ALWAYS SOMETHING WITH THESE TURBOLINKS

PR #1092, opened 2017-06-04
Config model

### LINKED ISSUE #1074: Turn insurance options list into a config

The list of possible insurances is hardcoded. We should move this to a config of some sort.

Abstracting DCAF out as the sole user.

Either setting this as an env var or setting this in a config model somewhere.

* **Anything else?**
Approximately part of #999 and #1072

colinxfleming:

Attempt to whip up a generalizable config model and smash insurance options into it.

TK: Figure out how to handle update assignment, figure out how to handle defaults, figure out how to mash it into the existing helper, determine whether this is the proper model structure, write controller tests, add to navbar dropdown.

* Create config model and tests
* Create configs controller and update-y forms

* Fixes #1074 ... in theory


colinxfleming:
After some chatter I'm kinda uncomfortable with rolling this out without a pretty extensive review from the likes of @lomky and @DarthHater and the rest of the corps, so switching efforts to getting #1080 in place for insurance and ext pledges.

colinxfleming:
Still a WIP, but this fleshes this out a LOT and turns it into something usable for insurance and ext pledges. Sorry, it got a little bulkier than I meant it to.

Here's what this does:

* creates two config objects for insurance and ext pledges
* makes a config route, accessible only by admins, to tweak them
* hooks them into existing helpers that populate dropdowns
* adds a rake task to generate stubbed objects in the model easily for orgs to fill in
* tesssssssts

Here's what's TK:

* add language to this
* get a second opinion on the model structure
* get a second opinion on names
* get tests to pass. A lot of stuff relies on these objects existing now and I need to come up with a good holistic way to handle that. 

PS We should not merge this before resolving the enumeration PR.

@lomky for review pls!

app/views/configs/_config.html.erb
```diff
@@ -1,15 +1,20 @@
 <div class="row">
   <div class="col-md-6">
     <%= bootstrap_form_for config, inline: true do |f| %>
-      <%= f.text_area :options, label: "#{config.config_key.capitalize} options", value: config.config_value[:options].join(', '), help: 'Please separate with commas.' %>
-      <%= f.submit "Update options for #{config.config_key}", class: 'btn btn-primary' %>
+      <%= f.text_area :options, label: "#{config.config_key.to_s.humanize} options",
```
lomky: `humanize` is cool!

app/views/patients/_patient_information.html.erb
```diff
@@ -74,7 +74,7 @@
           </div>
 
           <%= f.select :insurance,
-                       options_for_select(insurance_options,
+                       options_for_select(insurance_options.push(patient.insurance).uniq,
                                           patient.insurance),
```
lomky: Are these lines redundant?

test/controllers/configs_controller_test.rb
```diff
@@ -1,9 +1,42 @@
...
+      it "should deny access as a #{role}" do
+        delete destroy_user_session_path
+        user = create :user, role: role
+        config = create :config
+        sign_in user
+
+        get configs_path
+        assert_redirected_to authenticated_root_path
+
+        patch config_path(config, params: { options: ['no'] })
```
lomky: Why does this test setup on the negative test look different that the setup on the positive test down on line 35 onwards? i.e. `config` vs `@config`. `params: { options: ['no'] }` vs `params: { config: { options: ['no']}}`

db/seeds.rb
```diff
@@ -137,6 +137,13 @@
 patient_in_completed_calls.calls.create! status: 'Left voicemail',
                                         created_by: user
 
+# Create insurance and external pledge source keysets
+Config.destroy_all
+Config.create config_key: :insurance,
```
lomky: Config has no create function?

lomky:
Brainstorming conclusion: this is awesome and clever. :100: 

Happy to hash through any reservations you may have, but this looks like a solid ++ to me.

db/seeds.rb
colinxfleming: It inherits stuff from Mongoid, so it basically counts as an activerecord object with all the trappings (like find, find_or_create, pluck, etc)! Kinda handy!

app/views/configs/_config.html.erb
colinxfleming: I strongly recommend digging thru activesupport stuff for http://api.rubyonrails.org/classes/String.html and http://api.rubyonrails.org/classes/Array.html -- all sorts of cool stuff and it's kind of a neat look into the thought process behind a lot of rails ('wouldn't it be cool if...'). String stuff like humanize is really, really nice and I miss it a lot out of rails.

app/views/patients/_patient_information.html.erb
colinxfleming: My thought here was that we don't want to accidentally overwrite an insurance option if it disappears from the dropdown and the form saves, so if a patient has an insurance option that isn't present anymore, it'll just append it to the end of the list. Not very clear from the code here, so could probably use an explanation, though!

test/controllers/configs_controller_test.rb
colinxfleming: this is a good flag, they should def be identical. I'll standardize them before I put this up for final review.

colinxfleming:
Since you can't really warm up the test suite with a rake task, I added a rails initializer that reads the Config key enum options and creates objects for them as part of boot. It adds some noise to the logs, but should work painlessly with the test suite. (he said, hopefully.)

colinxfleming:
Just kidding, still chokes on integration tests in the same way. Still an improvement in that we don't have to remember to manually bump rake test whenever we move something new to the configs, so I'm going to leave it, but I'll need to dig into that.

colinxfleming:
I think we have an honest working version of this in d435282 ! It's a little heavier than I'd like, and I really don't love DB queries in helpers, but the initializer is an absolute mess. Currently working on getting the test suite to pass.

colinxfleming:
(Note to self, make absolutely sure you bump the tests for ext pledges helper and the insurance helper methods, since those are now doing a lot more heavy lifting than they were.)

colinxfleming:
This turned into a real yak but I feel pretty great about where it ended up! We have a flexible, hopefully extendable model that we can pop into dropdown populator functions, and an easy adminuser-facing way to make it work.

The single remaining task here is to write tests for `Config#autosetup` and the new controller#index behavior. (And get CI to pass.) But after that I'll punch it up for ~ final review ~ and maybe we can ship it Sunday!

colinxfleming:
kaboom! @lomky for review please!

The transition process from our current yml to this is basically creating the config objects and mashing them into the config option text, separated by commas. Other than that, this should more or less be seamless.

This is wired up (with a few well-placed `find_or_create_by`s and an `autosetup` method) so that it shouldn't err out if anything doesn't exist.

lomky:
Will give this the review it deserves ASAP.  :100: 

Quick Thought: will this mean updates in any of our Docs?

colinxfleming:
Yeah, we'll want to bump the deploying or setup doc to include 'set
configs'. A++++ flag! I'll def bump that.

app/controllers/configs_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  private
+
+  def config_params
+    params.require(:config).permit(:options)
+  end
+
+  def format_config_params(params)
+    # should probably be a model method?
```
lomky: I don't mind it being here. Open to arguments though.

app/controllers/configs_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  before_action :confirm_admin_user
+
+  def index
+    Config.autosetup
+    @configs = Config.all.sort
+  end
+
+  def update
+    @config = Config.find params[:id]
+    @config.config_value = format_config_params(config_params) # this is not going to work
```
lomky: Does it work?

app/helpers/external_pledges_helper.rb
```diff
@@ -1,13 +1,14 @@
...
+    standard_options = ['Clinic discount', 'Other funds (see notes)']
+    Config.find_or_create_by(config_key: 'external_pledge_source').options
+          .push(*standard_options)
+          .uniq
   end
 
   def available_pledge_source_options_for(patient)
     used_sources = patient.external_pledges.map(&:source)
-    (external_pledge_source_options - used_sources).unshift(nil)
+    (external_pledge_source_options - used_sources)
```
lomky: Was the `unshift(nil)` placing nil on the array as the default value? What changed that we don't want that now?

lomky:
One suggestion: The order of the configs on the edit page isn't set, so on hitting "Update options for [Thing]" the page reloads and sometimes swaps them! It would be nice if the configs appeared in the same order each time.

app/helpers/external_pledges_helper.rb
colinxfleming: This is a great flag! Unlike the `insurance` field which does have a default value of nil, external pledges don't save until you enter in an amount and smash the create button. Though looking at this in app now, I'm going to switch it back because I think I like the safety layer. Thank you for holding me accountable in making deliberate choices!

app/controllers/configs_controller.rb
colinxfleming: I'm good with this too, on further reflection. Smell ya later comment.

app/controllers/configs_controller.rb
colinxfleming: haha it totally works. Definitely didn't when I wrote it though.

colinxfleming:
@lomky we should be all clear on this! If this looks good to you and CI passes, you should go ahead and Smash That Merge Button.

PR #1096, opened 2017-06-05
NONCE NONCE NONCE (dances to techNOOOOO)

### LINKED ISSUE #893: Update asset hash for CSRF

DarthHater:

Trying out a nonce using SecureRandom (http://daniel.fone.net.nz/blog/2014/12/07/generating-unique-random-tokens/)

* Minor changes to application_controller.rb (with large implications)

* Fixes #893

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,3 +1,5 @@
+require 'SecureRandom'
+
 # Sets a few devise configs and security measures
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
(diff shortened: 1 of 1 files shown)


DarthHater:
@colinxfleming so this kinda switches to nonces, although there's a few issues thus far that I've noticed:

- It only applies to inline scripts on self, not every script (guessing this was intentional)
- it needs to be applied to the inline script (which I can't find?)

DarthHater:
Closing in favor of going with something like #1102 , experimental PR is this one (Yoda would say, or someone like me with bad English)

PR #1102, opened 2017-06-05
Add in SecureHeaders, move config over to that

### LINKED ISSUE #412: Install secure_headers gem
https://github.com/twitter/secureheaders

### LINKED ISSUE #893: Update asset hash for CSRF

DarthHater:

An alternative to #1096 , adds in Twitter's SecureHeaders and moves config over to that, still a bit of investigation on hash stuff needed.

* Adds in newest SecureHeaders
* Configs it based on preexisting header in code

* Bumps #893 
* Fixes #412


colinxfleming:
I like this a lot. After an evening of self clowning I think it might be a good idea to roll with a CSP like this and progressively work on getting inline javascript out of the app as much as possible as a longer term goal. Either way we should merge this because it's definitely a value add -- I'll poke at it tomorrow and push if it passes muster, which I bet it does because our test suite is good $$$$

app/views/layouts/application.html.erb
```diff
@@ -5,9 +5,9 @@
     <title><%= content_for?(:title) ? yield(:title) : "DARIA" %></title>
     <meta name="description" content="<%= content_for?(:description) ? yield(:description) : "DCAF" %>">
     <meta name='ROBOTS' content='NOINDEX, NOFOLLOW'>
-    <%= stylesheet_link_tag 'application', media: 'all' %>
-    <%= favicon_link_tag 'favicon.ico' %>
-    <%= javascript_include_tag 'application' %>
+    <%= stylesheet_link_tag :application, media: 'all' %>
+    <%= favicon_link_tag %>
```
colinxfleming: Thanks for fixing this, I'd been meaning to do it for awhile 💯

colinxfleming:
I bumped the issue refs -- I think there's a little additional CSP work we want to do at this juncture: a) get any and all inline js out of the app where possible and b) follow that up by adjusting the CSP so it makes things choke rather than `reportOnly`, so I changed #893 to be a bump rather than a fix. 💯 💯

PR #1106, opened 2017-06-07
Change to new enum library

### LINKED ISSUE #866: Rails5 Technical Debt: Long Term MongoID Enum Solution

Find a long term solution to Enum support in MongoID. We have stop-gapped this by forking the [gem's main repo]( https://github.com/thetron/mongoid-enum/) into [ @DarthHater's space](https://github.com/DarthHater/mongoid-enum), but given the lack of updates on the main gem we should search for a better option long term.

 * Long term project maintainability.
 * Future Ruby & Rails upgrades. 
 * Making future-us not scowl grumpily at our decisions today.

It's possible that the main gem gets moved into new, actively-maintained hands, in which case we're good!

Otherwise, finding a new gem. One is [simple-enum](https://github.com/lwe/simple_enum) but there may be others.

A _worst case_ scenario would look at feasibility of staying on MongoID.

* **Anything else?**
Discovered during the great [Rails 5 move](https://github.com/colinxfleming/dcaf_case_management/pull/865)

tingaloo:

We can rescue ourselves from mongoid-enum. However any current pull requests that use current enum may be affected. Hopefully the tests cover all the bases.

* Change to new enum library

* Fixes #866


colinxfleming:
Thanks a ton @tingaloo ! Really appreciate your looking out on this front.

@DarthHater can you review this? Do you think this is a viable migration away from `mongoid-enum`?

DarthHater:
I'll take a gander at reviewing, moving away from mongoid-enum is great, seriously.

Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'mongoid', '6.1.0'
 gem 'mongoid-history', '0.6.1'
 gem 'mongoid_userstamp', git: 'https://github.com/DCAFEngineering/mongoid_userstamp.git', branch: 'master'
 gem 'mongo_session_store', '>= 3.1.0'
-gem 'mongoid-enum', git: 'https://github.com/DCAFEngineering/mongoid-enum', branch: 'master'
+gem 'enumerize'
```
DarthHater: This is awesome, do you know if they are going to support Rails 5.1 anytime soon? The dependency on their end has it pinned to 5.0.0 from what I can tell.

app/models/user.rb
```diff
@@ -2,8 +2,9 @@
 # Fields are all devise settings; most of the methods relate to call list mgmt.
 class User
   include Mongoid::Document
-  include Mongoid::Enum
   include Mongoid::Userstamp::User
+  extend Enumerize
+  extend ActiveModel::Naming
```
DarthHater: Slightly squeamish seeing ActiveRecord related stuff sneak in. Mind explaining it? I don't think it will cause a collision with MongoID but it's good to know in general.

app/views/layouts/_navigation_links.html.erb
```diff
@@ -6,18 +6,18 @@
     <% if ENV['CM_RESOURCES_URL'] %>
       <li><%= link_to 'CM Resources', ENV['CM_RESOURCES_URL'], target: '_blank' %></li>
     <% end %>
-  <% end %>  
-</ul>
```
DarthHater: Love to see the clean up, ftr. Maybe just make comments in the future akin to that "Did some cleanup of files while I was at it" otherwise you run the risk of having a reviewer go "WHY DID YOU REMOVE SPACES AND TABS?!?!"

app/controllers/application_controller.rb
```diff
@@ -36,10 +36,10 @@ def csp_headers
   end
 
   def confirm_admin_user
-    redirect_to root_url unless current_user.admin?
+    redirect_to root_url unless current_user.role_admin?
   end
 
   def confirm_user_has_data_access
-    redirect_to root_url unless (current_user.admin? || current_user.data_volunteer?)
+    redirect_to root_url unless (current_user.role_admin? || current_user.role_data_volunteer?)
```
DarthHater: Knowing absolutely nothing about enumerize, why do you have to do role_admin vs admin?

app/controllers/users_controller.rb
```diff
@@ -51,7 +51,7 @@ def reorder_call_list
   private
 
   def user_params
-    params.require(:user).permit(:name, :email, :_role)
+    params.require(:user).permit(:name, :email, :role)
```
DarthHater: Similar to earlier question, why from _role to role? I assume it's a difference between the two libs, just want to understand.

Gemfile
tingaloo: I believe it to be, they are quite active with >1000 stars (not sure if that's an accurate gauge).

app/models/user.rb
tingaloo: I think I just copied it there without thinking too hard about it. I removed it, ran a test, and it made no difference. Good catch!

app/controllers/application_controller.rb
tingaloo: Just went with docs, i like it because it actually shows the parent relationship.

app/controllers/users_controller.rb
tingaloo: I actually was stumped by this, so I'm going to copy/paste my notes.

Before with mongoid-enum:
```2.4.1 :001 > @user = User.first
 => #<User _id: BSON::ObjectId('5931ea4f2cde1152ba1ebcb5'), _role: :admin, call_order: ["5931ea512cde1152ba1ebcd1", "5931ea512cde1152ba1ebcd5", "5931ea512cde1152ba1ebccd", "5931ea512cde1152ba1ebcc9", "5931ea512cde1152ba1ebcc5", "5931ea512cde1152ba1ebcc1"], email: "test@test.com", line: nil, name: "testuser (admin)", patient_ids: [BSON::ObjectId('5931ea512cde1152ba1ebcc1'), BSON::ObjectId('5931ea512cde1152ba1ebcc5'), BSON::ObjectId('5931ea512cde1152ba1ebcc9'), BSON::ObjectId('5931ea512cde1152ba1ebccd'), BSON::ObjectId('5931ea512cde1152ba1ebcd5'), BSON::ObjectId('5931ea512cde1152ba1ebcd1')]>
```

with enumerize

```
=> #<User _id: BSON::ObjectId('5931ecbe2cde1156ca284844'), call_order: ["5931ecbf2cde1156ca284860", "5931ecbf2cde1156ca284864", "5931ecbf2cde1156ca28485c", "5931ecbf2cde1156ca284858", "5931ecbf2cde1156ca284854", "5931ecbf2cde1156ca284850"], email: "test@test.com", line: nil, name: "testuser (admin)", patient_ids: [BSON::ObjectId('5931ecbf2cde1156ca284850'), BSON::ObjectId('5931ecbf2cde1156ca284854'), BSON::ObjectId('5931ecbf2cde1156ca284858'), BSON::ObjectId('5931ecbf2cde1156ca28485c'), BSON::ObjectId('5931ecbf2cde1156ca284864'), BSON::ObjectId('5931ecbf2cde1156ca284860')], role: "admin”>
```

Long story short, mongoid uses symbols and aliases with _. enumerize explicitly states in the docs that it only deals with strings.

Gemfile
DarthHater: It's a great gauge, just wanted to make sure it's not going to be a road block on moving to 5.1 (which we should, when we can)

app/controllers/application_controller.rb
DarthHater: Yeah don't disagree, just curious.

app/controllers/users_controller.rb
DarthHater: Yeah it's interesting to me too. _ tends to just be like used to show something as internal or private, or to avoid usage of this, but the usage is wildly different across languages and such. That it was _role to begin with is more puzzling, since it's definitely not an internal or private variable. Will the switch have any impact on existing data?

test/controllers/users_controller_test.rb
```diff
@@ -10,7 +10,7 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
 
   describe 'create method' do
     it 'should raise if user is not admin' do
-      User::ROLE.reject { |role| role == :admin }.each do |role|
+      User.role.values.reject { |role| role == :admin }.each do |role|
```
DarthHater: I'm guessing this is just another difference in how you access this?

app/models/user.rb
DarthHater: This is only for feedback (we do this at Sonatype, I find it extremely useful), is when you make the change, comment on the thread where it originated with the commit sha so that the reviewer can find your change (for one), and for traceability in the future. We do something like:

In: e561096

app/controllers/users_controller.rb
tingaloo: It only impacts the search methods and tests IIRC. Since this codebase is extremely well-tested, it shouldn't be an issue. thanks to @colinxfleming  and his nazi testing.

test/controllers/users_controller_test.rb
tingaloo: Yeah, I was really confused how the double colons worked in general. Enumerize just makes more sense.

DarthHater: For your edification, basically the difference is this (since I looked it up).

Since we were using import MongoID::Enum (which is how you'd bring in a class), I am guessing slightly here that access to Role before was via class, and not actual field, making it like ok, I want the User class, and then the ::Roles class on it. Now instead you have an actual field that has behavior added to it, so it's User.roles (field access) instead of User::Roles (class access). I'm typing this out mainly so someone can come in and go "No Jeffry you are totally wrong", too.

app/controllers/users_controller.rb
DarthHater: I'd actually say this might be kinda a bad assumption here. If we are changing the data model at all, a lot of existing records might end up with duplicate data in Mongo, and it could go unnoticed because we might only load certain fields, etc...

What I think would be good, for my sanity sake is for you and @colinxfleming to run this through staging or something with a larger data set, and look at records before and after. Just basically check for any gremlins left behind so we know if we need to write any scripts for cleanup, etc... It could very well be that there's nothing to be concerned about, I just get panicky anytime a data model shifts and you use a NoSQL db.

app/models/user.rb
tingaloo: Ha you made me realize a flaw in my coding is big commits. I usually don't change bad habits until somebody actually points it out. Thanks for the valuable feedback.

app/controllers/users_controller.rb
tingaloo: Actually scratch what I said earlier. You are 100% right. I had to reseed the database because enumerize actually mutates the Patient and User fields. I'm going to flag an extended issue on this.

app/models/user.rb
DarthHater: I have the same flaw, when you point something out there are usually four fingers pointing back at you. Habit that I'm breaking over time. You in no sense want to end up with superfluous commits but for feedback this stuff can be pretty valuable (if for example you bundle like ten fixes into one that says "Feedback", it's perfectly reasonable to just use that commit hash for all of them).

app/controllers/users_controller.rb
DarthHater: I think that you can probably provide any scripts for you know, migrating the data as a part of this PR rather than a separate one, unless you feel absolutely that you need to break it apart.

colinxfleming: If this switches enums from ints to strings, I'd very much appreciate commands to run in the rails console that will let us cleanly migrate things over to the new datatypes.

e.g. 
```
User.all.each do |user|
  if user[:_role] == 'admin'
    user.role = 'admin'
  elsif ...
end
```

That said having enums as strings is sort of annoying! Breaks the pattern from standard activerecord enums. Kinda sucks that every implementation of this is bad. I'm good with this if it means we get away from mongoid enum though.

test/controllers/users_controller_test.rb
DarthHater: Turns out I didn't fully understand this but I was close. It allows you to access classes and constants on modules. SO:

````
module Foo
  class Some
  CONST = 1
     def Thing
     end
   end
end
````
You'd call these things like such:
`Foo::Some`
`Foo::Some::CONST`
`Foo::Some.Thing`

There is a pretty good thing that explains more here: http://www.railstips.org/blog/archives/2009/05/15/include-vs-extend-in-ruby/

This is all of course if you are interested, and didn't know this to begin with. I certainly didn't :)

Comment (tingaloo) [inline test/controllers/users_controller_test.rb]: That's pretty neat, I remember mongoid-enum mentioning how it makes globals accessible everywhere.

And keep the textbook information coming! Normally I'd just ignore it, but there is actually context and I was curious what enumeration was all about. I'm probably more optimistic and think the solution is good enough. But something tells me I am not seeing the side effects that you're anticipating.

Comment (DarthHater) [inline test/controllers/users_controller_test.rb]: Enumeration is basically the way of doing a few things. It can bound a set of values, for one. For example, the data type `boolean` is an enum, it has two constant values, `true` and `false`. For most programming you'll ever see, enums will be bit strings, basically like:
```
0=THIS
1=THAT
2=THAR
```

These become useful if you want to deliberately limit the input to a system via code. It also becomes helpful because rather than save something like "1" to a DB via code, you can go "THAT" so it reads cleaner. 

Here's a fun read through on that: https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

Much easier to switch on MONDAY, TUESDAY, WEDNESDAY, when you could also just do 1,2,3. Reads cleaner for all parties involved.

**Comment by colinxfleming** [comment]
I'm interested in pushing on this hard after the scheduled deployment this Sunday. @tingaloo can you do me a solid and resolve merge conflicts on this so we can hit the ground running on this? As @DarthHater pointed out to me, mongoid enum is something that's holding us back from getting on rails 5.1.1, so once we get this in we can get that process started.

**Comment by colinxfleming** [comment]
I spent some time today poking around the enumerize docs -- @tingaloo this was a great find and I think it's a great move toward something much more sustainable for this project.

* I was really concerned about enums being saved as strings, but I think we might be able to rig it up so that it's functionally pretty similar to mongoid-enum, save-wise. From the docs:

```
class User < ActiveRecord::Base
  extend Enumerize

  enumerize :role, in: {:user => 1, :admin => 2}
end
```

So I think we can continue to save role data as ints (with CMs as 0, admins as 1, data vols as 2) and not have to do a database migration here, necessarily. It's kinda silly that saving them as ints isn't standard but I will happily accept this alternative.

* I was also kinda concerned about predicate methods but it looks like we can chuck a `predicates: true` on the end of an enum and be able to use them similarly. Also from the docs:

```
class User
  extend Enumerize

  enumerize :sex, in: %w(male female), predicates: true
end

user = User.new
user.sex = 'male'
user.male?   # => true
```

I think this means that we can skip doing `role_admin?` in a bunch of places, and more or less leave this the same. Thoughts @tingaloo ?

**Comment by tingaloo** [comment]
I changed the roles back to the initial code. I think there still needs to be a DB refresh because of how ```_line``` becomes ```line``` and ```_role``` becomes ```role```?

**Comment by colinxfleming** [comment]
Mongoid enum MIGHT store it in both places, with one as an alias, so we might luck out. I hope. In `rails console`:

```
irb(main):001:0> User.first
=> #<User _id: BSON::ObjectId('595432ac311c8884230e4ca1'), _role: :admin, call_order: [...], created_at: Wed, 28 Jun 2017 18:50:20 EDT -04:00, email: "test@test.com", line: nil, modifier_id: nil, name: "testuser (admin)", patient_ids: [....], updated_at: Tue, 04 Jul 2017 13:48:46 EDT -04:00, version: 1>

irb(main):002:0> User.first.role
=> :admin
irb(main):003:0> User.first._role
=> :admin
```

I bet you're right that we won't get that lucky though.

**Comment by colinxfleming** [comment]
It looks like most of the tests are passing -- since we have a pretty rigorous test suite, that's a real accomplishment. Nicely done @tingaloo !

File: app/models/patient.rb [hunk 7]
```diff
@@ -50,8 +50,12 @@ class Patient
   field :identifier, type: String
 
   # Contact-related info
-  enum :voicemail_preference, [:not_specified, :no, :yes]
-  enum :line, LINES # See config/initializers/env_vars.rb
+  field :voicemail_preference
+  enumerize :voicemail_preference, in: [:not_specified, :no, :yes], default: :not_specified
```
colinxfleming: I **think** we'll want to declare this an int by doing something like `in: [not_specified: 0, no: 1, yes: 2]` so we can avoid migrating the current data. I think.

app/models/user.rb
```diff
@@ -34,14 +34,17 @@ class User
...
-  
+
   # Fields
   # Non-devise generated
   field :name, type: String
   field :line, type: String
-  enum :role, [:cm, :admin, :data_volunteer]
+  field :role
+
+  enumerize :role, in: {:cm => 0, :admin => 1, :data_volunteer => 2}, predicates: true
```
colinxfleming: 👍

app/models/user.rb
```diff
@@ -153,7 +156,7 @@ def clear_call_list
   end
 
   def admin?
-    role == :admin
+    role == 'admin'
```
colinxfleming: This might be redundant to predicate stuff provided by the new enum library -- wonder if we can nuke it?

test/controllers/users_controller_test.rb
colinxfleming: 🔬

app/helpers/dashboards_helper.rb
```diff
@@ -16,9 +16,10 @@ def voicemail_options
     enum_text = { not_specified: 'No instructions; no ID VM',
                   no: 'Do not leave a voicemail',
                   yes: 'Voicemail OK, ID OK' }
-    vm_options = Patient::VOICEMAIL_PREFERENCE
+    vm_options = Patient.voicemail_preference.values
```
colinxfleming: Want to make sure I follow this -- we take voicemail pref options, change them to symbols (because they're being returned a strings), and then turn it into a tuple of the text and symbol. 

Wonder if we can ditch that second `to_sym`, or collapse lines 19 and 20?

tingaloo: From what I remember, the view depended on having symbols on the erb.html code to render 'MD','VA', etc. I tried using strings and could never get them showing up.

tingaloo:
The way I'm testing this is switching to master, rake db:seed, then switching back to this branch and seeing if index is populating. I'm not exactly sure what I am doing with setting 0,1,2 values. So an explanation of how we're migrating this new library <b>without altering the DB</b> would be awesome.

colinxfleming:
the plot thickens a little bit -- I popped into mongo and looked at the saved json, and it looks like it's saving stuff as strings now. Which, rude! But also simplifies moving over quite a bit.

```
> use dcaf_case_management_development
switched to db dcaf_case_management_development
> db.users.find({}).limit(1)
{ "_id" : ObjectId("595c47fa311c884ae5ee33bf"), "_role" : "cm", "email" : "test2@test.com", "encrypted_password" : "$2a$10$ECcGOwblCGQD0vVwP3I8he7qToD37wMCf6/4fUYx5N9bSB/Dr0sXa", "sign_in_count" : 0, "failed_attempts" : 0, "name" : "testuser2", "updated_at" : ISODate("2017-07-05T01:59:22.401Z"), "created_at" : ISODate("2017-07-05T01:59:22.401Z"), "version" : 1 }
```

Looks like `_role` is getting saved as a string, and it's getting translated from there. So probably we can ditch the number values. 

In the spirit of suffering for my sins I'm going to pull this into a branch and see what steps I have to take to get our enum fields to work, and doc them here.

colinxfleming:
ps sorry for the back and forth @tingaloo . I know that's it's a pain, thanks for bearing with me.

colinxfleming:
Here's what it took to get a call list to load:

bundle install

in user model: `enumerize :role, in: [:cm, :admin, :data_volunteer], predicates: true`

in patient model: `enumerize :voicemail_preference, in: [:not_specified, :no, :yes], default: :not_specified`

Running the following in rails console:
`User.each { |u| u.rename(_role: 'role')}`
`Patient.each { |pt| pt.rename(_line: 'line')}`
`Patient.each { |pt| pt.rename(_voicemail_preference: 'voicemail_preference')}`

I was hopeful that there was a way to avoid doing a field rename, but I think this is okay.

tingaloo:
Gotcha, I changed the code back to what you stated earlier. As for the rails console script, I guess you can run that on another PR?

colinxfleming:
We'll have to run both of those on the prod DBs after taking backups. (It's kind of a pain, which is why I try to avoid doing it.)

Here's our remaining test failures: 

```
ERROR["test_0003_disallows role editing", #<Class:0x007f906063c980>, 221.42816523701185]
 test_0003_disallows role editing#UserManagementTest::edit a user (221.44s)
Capybara::ExpectationNotMet:         Capybara::ExpectationNotMet: expected to find text "cm" in "DARIA - test Your current line: DC Admin tools admin × Successfully updated user details User Account Management Add New User Reset Password (WIP) Toggle Lock (WIP) Search FULL NAME ROLE LAST LOGIN DATE CREATED [-] STATUS john admin 2017-07-08 19:45:44 -0400 Active admin admin 2017-07-08 19:45:44 -0400 2017-07-08 19:45:44 -0400 Active volunteer data_volunteer 2017-07-08 19:45:44 -0400 Active Please send feedback or ideas for this case management app to DCAF.CMapp@gmail.com DC Abortion Fund -- 202-452-7464 National Abortion Federation -- 800-772-9100 onlinefaxes.com". (However, it was found 1 time using a case insensitive search.)
            test/integration/user_management_test.rb:151:in `block (2 levels) in <class:UserManagementTest>'

 FAIL["test_0002_should not cancel if not rescinded", #<Class:0x007f905ca557f0>, 250.45839508601057]
 test_0002_should not cancel if not rescinded#SubmitPledgeTest::cancelling a pledge (250.46s)
        Expected false to be truthy.
        test/integration/submit_pledge_test.rb:108:in `block (2 levels) in <class:SubmitPledgeTest>'

 FAIL["test_0001_should let you mark a pledge submitted", #<Class:0x007f905ca56740>, 355.312451247999]
 test_0001_should let you mark a pledge submitted#SubmitPledgeTest::submitting a pledge (355.31s)
        Expected false to be truthy.
        test/integration/submit_pledge_test.rb:37:in `block (2 levels) in <class:SubmitPledgeTest>'
```

colinxfleming:
That test itself looks kinda weird 

Once that's merged I think we get this to pass CI and then ship!

tingaloo:
I pulled your PR #1128 to let CircleCI to do its thing. Other than the DB migration, I think this is at the finish line.

colinxfleming:
Looks good to me! @DarthHater can you review this too and make sure it looks good to your eyes, since this is a pretty heavy hitting PR?

colinxfleming:
🥇 

going to ship this and get it out tonight hopefully!

PR #1110, opened 2017-06-09
Lint models

colinxfleming:

Had time on the plane so I resolved some rubocop violations in our models.

* Top level documentation for everything
* Some whitespace and line-too-long fixing

No issue.

@cfinucane  can you eyeball this for me? In particular I'm interested in whether or not the top level documentation comments make sense, and I think you're pretty well positioned to do that as someone who hasn't been swimming in this for the better part of a year 💯


colinxfleming:
@lomky can you review when you get a min?

PR #1111, opened 2017-06-09
Lint controllers

colinxfleming:

Had time on the plane, linted models.

* Top level doc comments for everything!
* A couple style changes/ whitespace eliminations/etc

No issue. 

@tingaloo can you eyeball this for me? Interested in particular on whether the top level documentation comments make sense. (We use rubocop for linting, it's basically the same as eslint.)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,9 +7,6 @@ Style/FrozenStringLiteralComment:
 Lint/AmbiguousRegexpLiteral:
   Enabled: false # this is a tremendously stupid rule
 
-Documentation: 
-  Enabled: false
-
 Rails:
   Enabled: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,9 @@ Style/ClassAndModuleChildren:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+Style/SymbolArray:
+  Enabled: false
+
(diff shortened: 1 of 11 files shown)


.rubocop.yml
```diff
@@ -4,12 +4,12 @@ Style/ClassAndModuleChildren:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+Style/SymbolArray:
```
colinxfleming: I went back and forth on this for a long time but I came down on 'array of symbol syntax is not super helpful'. (Ruby will let you do `%i[thing thing2 thing3]` and it'll be interpreted as `[:thing, :thing2, :thing3]` but I think it's non-intuitive, and if I were coming in cold I think I'd be like 'what on earth does that mean'. So, shutting it off, and moving it back in a couple places.

tingaloo: I like symbols more :100:

app/controllers/accountants_controller.rb
```diff
@@ -1,11 +1,10 @@
+# Controller pertaining to accountant functions, usually fulfillments.
 class AccountantsController < ApplicationController
-  def index
-  end
+  def index; end
```
tingaloo: One liners for method definitions look funny, but it looks cool in a JS pov.

PR #1113, opened 2017-06-21
Draft user management panel, ctd

colinxfleming:

Makes test and some other misc fixes from #1073 so we can get the first iteration of the users index into production.

I haven't done it here yet but I think I'm going to shut off the new edit/update routes, so we can roll those out PR by PR. This is massive and I'm a little gunshy about it.

* mostly fixes to #1073  so the tests pass

@tingaloo for review/thoughts

* Bumps #686


test/controllers/users_controller_test.rb
```diff
@@ -198,10 +195,12 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     end
 
     it 'should respond success' do
+      # this is wrong. it's returning bad request
```
colinxfleming: yep. durf. fixed in 168b6db

colinxfleming:
Added a few comments for places where test coverage needs to crank up, merging after the build passes.

PR #1050, opened 2017-05-10
Email is no longer an edited field; added link to new_user_password_path

AlexandraCasser:
I've updated the user edit page to make the following changes: 

Email is no longer an editable field
Removed 'enter password to confirm'
Included link to `new_user_password_path'

It relates to the following issue #s: 
[Issue #951](https://github.com/DCAFEngineering/dcaf_case_management/issues/951)

PR #1116, opened 2017-06-24
User edit panel remix

### LINKED ISSUE #951: User edit page: make changes
TK, just an issue I noticed in talking to a user

colinxfleming:

Adapts @AlexandraCasser 's great work in #1050 , and works out an additional kink or two (the reset password link doesn't work if you are already logged in, whoops, sorry @AlexandraCasser !).

* email no longer an editable field (woo security)

screen shot 2017-06-24 at 1 23 16 pm

* Fixes #951

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,9 @@ class ApplicationController < ActionController::Base
   # whitelists attributes in devise
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:sign_up, keys: [:name])
-    devise_parameter_sanitizer.permit(:account_update, keys: [:name])
+    devise_parameter_sanitizer.permit(:account_update) do |user|
+      user.permit :name, :current_password
+    end
   end
 
   private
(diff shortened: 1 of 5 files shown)


colinxfleming:
A couple quick notes: 

* email is a de-whitelisted field, so it now takes admin action to edit. I think this is a security gain
* removing password editing was something that I hoped would be superseded by oauth, but we're unlikely to go full oauth ever, so I think we can punt on that
* adds a user's own role to the rollup

Has the feel of a 'here's my user panel' now, rather than just 'edit my information', but the important thing (making email uneditable) is good to go. 

Also added a test of the controller action, so we can confirm that this security enhancement doesn't regress.

colinxfleming:
@lomky would love your thoughts on this if you have a sec !

colinxfleming:
✨ 

we're working on role type changes done by users, but they're eventually going to live at `users/:user_id/edit`, which is an 'admins only' route (the route is set up, but saving changes doesn't work yet). If we have to do it right now, it happens in the rails console.

PR #1117, opened 2017-06-24
Pledge reload ajax

### LINKED ISSUE #490: Patient Info in pledge and status isn't reflected until page refresh

colinxfleming:

@tingaloo has solved a long running problem of this app in #1064 , which is that the patient edit page is not particularly responsive. I've added a little bit of business logic surrounding this. This now makes the following fields bump on a patient edit:

* status
* current LMP
* approx gestation at appt
* the change log
* cancel pledge / submit pledge button status
* whether or not the fulfillment partial shows up

There's a thing that my js skills are a little weak that I'd like to sort out -- refreshing the page makes you lose focus, so if you're tabbing from field to field on the dashboard, you lose whatever you're editing. @tingaloo do you know of a good way to stop that from happening? (Maybe by just re-racking particular pieces of text in specific IDs, rather than blowing away the entire dashboard partial? That's probably the most straightforward way, but I'd like your opinion.)

* update js action fires now and refreshes certain partials we want to be dynamic
* tidies up in a couple other places

* Fixes #490 !!!!!! ahhhhhhh

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -63,6 +63,7 @@ class PatientsController < ApplicationController
 
   def update
     if @patient.update_attributes patient_params
+      @patient.reload
       respond_to { |format| format.js }
     else
       head :internal_server_error
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -1,6 +1,6 @@
 # Create, edit, and update patients. The main patient view is edit.
 class PatientsController < ApplicationController
-  before_action :confirm_user_has_data_access, only: [:index]
+  before_action :redirect_unless_has_data_access, only: [:index]
   before_action :find_patient, only: [:edit, :update, :download]
   rescue_from Mongoid::Errors::DocumentNotFound,
               with: -> { redirect_to root_path }
@@ -63,15 +63,15 @@ class PatientsController < ApplicationController
 
   def update
     if @patient.update_attributes patient_params
-      if @patient.previous_changes.key?("pledge_sent")
+      puts @patient.previous_changes
+      # if @patient.previous_changes.key?("pledge_sent")
         respond_to { |format| format.js }
-      else
-        head :ok
-      end
+      # else
+        # head :ok
+      # end
     else
       head :internal_server_error
     end
-
   end
 
   def data_entry
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -63,12 +63,7 @@ class PatientsController < ApplicationController
 
   def update
     if @patient.update_attributes patient_params
-      puts @patient.previous_changes
-      # if @patient.previous_changes.key?("pledge_sent")
-        respond_to { |format| format.js }
-      # else
-        # head :ok
-      # end
+      respond_to { |format| format.js }
     else
       head :internal_server_error
     end
--- a/app/views/patients/update.js.erb
+++ b/app/views/patients/update.js.erb
@@ -5,7 +5,7 @@ $('#patient_dashboard').html(
 );
 
 // $('#change_log').remove();
-$('#patient_change_log').html(
+$('#change_log').html(
   '<%= escape_javascript(render partial: "patients/change_log", locals: { patient: @patient })%>'
 );
 
@@ -20,4 +20,6 @@ $('#patient_change_log').html(
   $('#menu').html(
     '<%= escape_javascript(render partial: "patients/menu", locals: { patient: @patient })%>'
   );
+<% else %>
+  $('#fulfillment').empty();
 <% end %>
--- a/app/views/patients/update.js.erb
+++ b/app/views/patients/update.js.erb
@@ -1,25 +1,37 @@
-// $('a[href$="#patient_information"]').click();
+// rerack approx gestation
+$('#patient_appointment_date')
+  .parent()
+  .find('span.help-block')
+  .text('Approx gestation at appt: <%= @patient.last_menstrual_period_at_appt %>')
 
-$('#patient_dashboard').html(
-  '<%= escape_javascript(render partial: "patients/patient_dashboard", locals: { patient: @patient })%>'
-);
+// rerack status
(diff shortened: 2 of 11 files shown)


app/controllers/patients_controller.rb
```diff
@@ -63,7 +63,12 @@ def edit
 
   def update
     if @patient.update_attributes patient_params
-      head :ok
+      puts @patient.previous_changes
```
colinxfleming: some cleanup here

app/views/patients/update.js.erb
```diff
@@ -0,0 +1,25 @@
...
+  '<%= escape_javascript(render partial: "patients/patient_dashboard", locals: { patient: @patient })%>'
+);
+
+// $('#change_log').remove();
+$('#change_log').html(
+  '<%= escape_javascript(render partial: "patients/change_log", locals: { patient: @patient })%>'
+);
+
+// $('#fulfillment').remove();
+<% if @patient.previous_changes.key?('pledge_sent') && current_user.allowed_data_access? %>
```
colinxfleming: the idea here is that if pledge sent changed, we rerack the particular partials dealing with it if a user can see them. That includes nuking or populating the fulfillment form if necessary.

test/controllers/patients_controller_test.rb
```diff
@@ -143,9 +143,13 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
     end
 
     it 'should respond success on completion' do
+      patch patient_path(@patient), params: { patient: @payload }, xhr: true
+      assert_response :success
+      patch patient_path(@patient), params: { patient: @payload }
```
colinxfleming: no need for this, since we'll only ever call update with js

test/integration/submit_pledge_test.rb
```diff
@@ -133,5 +133,32 @@ class SubmitPledgeTest < ActionDispatch::IntegrationTest
 
       assert has_link? 'Submit pledge'
     end
+
+    it 'should dynamically update page' do
```
colinxfleming: I'd like to spin this out into its own test since it's basically a feature in its own right $$

colinxfleming:
Rolled out a more surgical refresh in da984de

colinxfleming:
Tests to fix:

```
ERROR["test_0001_should alter the information", #<Class:0x0000000609c670>, 57.762274324995815]
 test_0001_should alter the information#UpdatePatientInfoTest::changing fulfillment information (57.77s)
Capybara::ElementNotFound:         Capybara::ElementNotFound: Unable to find select box "Weeks along at procedure"
            test/integration/update_patient_info_test.rb:199:in `block (2 levels) in <class:UpdatePatientInfoTest>'

 FAIL["test_0001_should respond success on completion", #<Class:0x0000000443b328>, 1.6821190850023413]
 test_0001_should respond success on completion#PatientsControllerTest::update method (1.68s)
        Expected response to be a <2XX: success>, but was a <406: Not Acceptable>
        test/controllers/patients_controller_test.rb:149:in `block (2 levels) in <class:PatientsControllerTest>'

 FAIL["test_0004_should dynamically update page", #<Class:0x000000036a3cb0>, 36.95332306199998]
 test_0004_should dynamically update page#SubmitPledgeTest::cancelling a pledge (36.96s)
        Expected false to be truthy.
        test/integration/submit_pledge_test.rb:156:in `block (2 levels) in <class:SubmitPledgeTest>'

ERROR["test_0002_should uncheck when all fields are empty", #<Class:0x00000005a6f3d8>, 61.03777930600336]
 test_0002_should uncheck when all fields are empty#PledgeFulfillmentTest::pledge fulfilled autochecking on change (61.04s)
Capybara::Poltergeist::ObsoleteNode:         Capybara::Poltergeist::ObsoleteNode: The element you are trying to interact with is either not part of the DOM, or is not currently visible on the page (perhaps display: none is set). It's possible the element has been replaced by another element and you meant to interact with the new element. If so you need to do a new 'find' in order to get a reference to the new element.
            test/integration/pledge_fulfillment_test.rb:105:in `block (2 levels) in <class:PledgeFulfillmentTest>'
```

colinxfleming:
the core of this is in place and ready for poking at, I'll do the ugly test stuff a bit later.

app/views/patients/update.js.erb
```diff
@@ -0,0 +1,37 @@
+// rerack approx gestation
+$('#patient_appointment_date')
+  .parent()
+  .find('span.help-block')
+  .text('Approx gestation at appt: <%= @patient.last_menstrual_period_at_appt %>')
+
+// rerack status
```
tingaloo: This is a cool way to preserve fields from before without recalling the DB.

colinxfleming: Yeah. It's calling for the editing user on the audit trails still, so there's that n+1 we'll have to sort out, but I don't think that's terribly problematic at this stage (and iirc you've got an open PR that's a stab at the problem).

test/integration/submit_pledge_test.rb
colinxfleming: come to think of it, just rolling it up into submit pledge and update patient and etc is a better strategy

colinxfleming:
kaboom. @tingaloo can you review?

test/controllers/patients_controller_test.rb
```diff
@@ -169,7 +170,7 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
-  #confirm get :data_entry returns a success code
+  # confirm get :data_entry returns a success code
```
tingaloo: Maybe remove this comment?

app/views/patients/update.js.erb
tingaloo: It'd be awesome to get that PR merged, I think it would make things even more snappy.

test/controllers/patients_controller_test.rb
colinxfleming: good call!

PR #1118, opened 2017-06-26
Render Report Graphs Asynchronously

### LINKED ISSUE #1095: Load reports asynchronously

Actually using reports.

* **Anything else?**
See also #1009

tingaloo:

Report graphs now display, async rendering of partials. If unsafe-inline is an issue, we need to find a workaround.

* Allows graphs to render by allowing unsafe inline

* Fixes #1095


config/routes.rb
```diff
@@ -23,6 +23,10 @@
       resources :external_pledges,
                 only: [ :create, :update, :destroy ]
     end
+    get :weekly_report, :controller => :reports
```
colinxfleming: kinda hinted at this above, but I wonder if there's a way we can roll these up into a single endpoint, since they're all basically doing the same thing? Could take a date as an argument, like `time_report/:start_date` or something like that. Would have the benefit of making it more testable and maintainable going forward. What do you think?

app/views/layouts/application.html.erb
```diff
@@ -12,6 +12,8 @@
     <%= Gon::Base.render_data %>
   </head>
   <body>
+    <%= content_for :render_async %>
```
tingaloo: Let me know how to go about this.

config/initializers/csp.rb
```diff
@@ -1,11 +1,11 @@
-SecureHeaders::Configuration.default do |config|  
+SecureHeaders::Configuration.default do |config|
   config.csp = {
     preserve_schemes: true, # default: false.
     default_src: %w('self'),
-    script_src: %w('self' 'unsafe-eval'),
+    script_src: %w('self' 'unsafe-eval' 'unsafe-inline'),
```
tingaloo: I'm not familiar with CSP or its workarounds. Is there a doc or should I let you and darth handle it?

app/views/layouts/application.html.erb
colinxfleming: If you flip off `allow-inline` and check out the javascript console, you'll see errors like this:

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-+PC+Kf6qxwFk9MeuRMDmsALBNWNX473zplWkiVYhJgY='), or a nonce ('nonce-...') is required to enable inline execution.
```

If those hashes have a consistent value, you can paste that `sha256-ihgwphiugwrhpiurgphiurwgphiu` value into the CSP script-src array. If they don't have consistent values, let's punt on this and we'll figure something out.

app/views/reports/index.html.erb
```diff
@@ -1,40 +1,13 @@
...
+<div class="row">
+  <h4 class="col-sm-8"><strong>Budget bar, design TK</strong></h4>
+</div>
+<!-- UNCOMMENT WHEN BUDGET BAR IS READY
+<div id="budget_bar">
+  <div class="progress"></div>
+  <div class="progress-bar progress-bar-striped"</div>
+</div>
+ -->
+<%= render_async patients_report_path(timeframe: 'weekly') %>
```
colinxfleming: 👍

config/routes.rb
```diff
@@ -23,6 +23,9 @@
       resources :external_pledges,
                 only: [ :create, :update, :destroy ]
     end
+
+    get 'reports/:timeframe', to: 'reports#report', as: 'patients_report'
```
colinxfleming: 👍

test/controllers/reports_controller_test.rb
```diff
@@ -19,4 +19,21 @@ class ReportsControllerTest < ActionDispatch::IntegrationTest
...
+      assert_response :success
+      get patients_report_path(timeframe: 'yearly')
+      assert_response :success
+    end
+
+    it 'should return 406 on bad timeframe' do
+      get patients_report_path(timeframe: 'nonexistent')
+      assert_response 406
+    end
+  end
```
colinxfleming: This is excellent, thank you very much! We'll want to add an integration test too but we can file that as an issue for later, let's just make sure it doesn't slip

app/views/layouts/application.html.erb
colinxfleming: hmm, definitely not consistent, the js gets generated with a securerandom. https://github.com/kaspergrubbe/render_async/blob/0e11963c9d13df5d266b79044934cbf649dafb2a/lib/render_async/view_helper.rb#L6

Maybe nonces are the way to go here? (maybe I'll submit a PR to allow for that to the library?)

colinxfleming: stay tuned -- https://github.com/renderedtext/render_async/issues/8

colinxfleming:
I think since getting this to work with nonces is not a right now thing, and this is a major step in the right direction even if it isn't currently functional as such, we should just remove the inline-eval part of the security policy to re-break it and run with this as is, and get it to work with the CSP later. Does that sound good to you @tingaloo ?

tingaloo:
Getting this PR pushed and filing a separate issue for CSP is 👍 .

What do you mean by remove inline-eval?

config/initializers/csp.rb
```diff
@@ -1,11 +1,11 @@
-SecureHeaders::Configuration.default do |config|  
+SecureHeaders::Configuration.default do |config|
   config.csp = {
     preserve_schemes: true, # default: false.
     default_src: %w('self'),
-    script_src: %w('self' 'unsafe-eval'),
+    script_src: %w('self' 'unsafe-eval' 'unsafe-inline'), #TODO find workaround
```
colinxfleming: Can you remove the `unsafe-inline` here so we don't leave ourselves vulnerable while we sort out the nonce / csp stuff? After that we're good to merge!

tingaloo: Oh I misread your previous comment, it's all ready now.

colinxfleming: thanks much!

colinxfleming:
I popped a PR at `render_async`, which will solve our nonces problem pretty cleanly, I think, and we can do that in a separate PR. Getting CI to pass and then merging. Thanks again for sticking on this @tingaloo !

colinxfleming:
hooooooonk

PR #1122, opened 2017-07-03
Bump dev dependencies

colinxfleming:

Some gemnasium dev bumps.

Note that for our mac using friends, this might require some xcode wrangling. See https://stackoverflow.com/questions/40038953/how-to-install-nokogiri-on-mac-os-sierra-10-12 just in case.

* bumps dev dependencies
* bumps nokogiri :(

No issues.

@DarthHater can I get a review?


DarthHater:
Gave this a look, `bundle install` works fine, and rails s brings everything up. However, due to my natural curiosity I tried to use web console and I'm not sure if it's broken or working.

In `application.html.erb#L33`, I added `<% console %>` and when I load up the app I just get a white page with #. Something doesn't seem right to me with that, so I'm just trying to see if that's actually what's expected (I sorta expected an interactive ruby console). Otherwise, everything worked fine that I tested.

PS branch name RULZ

colinxfleming:
@DarthHater Weirdly, I can't get this to work with the old version of web console either. One of those things where I wonder how it ever worked. That said, it seems to break the same way -- white page with #. How do you think we should handle it, reverting or pushing forward? I'm good with either, so up to you.

DarthHater:
Push forward and file ticket to figure out that lib.

colinxfleming:
I figured out what `web-console`'s stupid problem was, turns out the answer is that it's running afoul of the inline-js block in our CSP. If we whitelist `'unsafe-inline'` it totally works like a charm. I'd rather not do this though so I'm going to remove it, given that I don't think anybody really leans on it (as evidenced by nobody having an issue with this until now, really).

colinxfleming:
Bumped gems again, going to get this merged shortly. The verdict:

* web console isn't good with our CSP
* everything else seems to upload pretty cleanly!

Once I get CI to pass I'm going to merge, since @DarthHater is still wrestling goats on mountaintops.

PR #1123, opened 2017-07-03
Bump runtime dependencies

### LINKED ISSUE #1131: `bundle update` causes app to begin to fail

  - Keep everything running smoothly!

  - being able to safely run `bundle update`

  - I'm very suspicious of the `platform-api` line in our `Gemfile` @colinxfleming says its a legacy gem from before Heroku gave us a cert, but somewhat hard to extract from our system. 

* **Anything else?**

Reproduce by checking out master, running rake test (tests should start, cancel out), run bundle update (tests will _fail to run_).

colinxfleming:

Bump runtime dependencies, except for rails, puma, google oauth, and mongoid (because bumping those is scarier and we should do that a bit more deliberately). Should bring us mostly up to speed.

* bundle update {runtime dependency}

fixes #1131 

@DarthHater can you review?


colinxfleming:
This is bumped and ready for 🦅 

@DarthHater at your leisure, probably best to rack this before getting going on rails 5

colinxfleming:
Got an all clear on this from @DarthHater in slack, so invoking admin powers and merging 💯

PR #1124, opened 2017-07-03
Pledge fulfillment now populating on submit pledge

### LINKED ISSUE #1120: Pledge fulfillment partial not clickable on hot reload

Very small bug here:
* as an admin, create a new patient
* fill in appointment date, clinic, and pledge amount
* submit the pledge
* at this point, the fulfillment link should show up on the left side

Clicking the fulfillment link doesn't do anything -- it should load the fulfillment content.

Hot refresh as rolled out by #1064 / # 1117

Probably just a bad partial id, or we need to refire some bootstrap js after the partial refreshes, or the actual fulfillment partial isn't getting appended.

* **Anything else?**
nope!

tingaloo:

On submit pledge, pledge fulfillment is now clickable.

* Bugfix

* Fixes #1120


colinxfleming:
thanks a ton @tingaloo ! Looks good to me!

PR #1128, opened 2017-07-09
Slight test adjustment, no more role editing for now

colinxfleming:

Temporarily turns off role changing until we can build some precise rules around it; adjusts the user mgmt integration test so that #1106 will fly.

* Removes `_role` from `user_params`
* Makes some precision adjustments to the user management integration tests

No view changes

* No issue, but is here so we can get #1106 to fly


colinxfleming:
@tingaloo can you review?

PR #1134, opened 2017-07-11
align Call List table with Navbar title

### LINKED ISSUE #1103: Increase max width of DCAF site
The call list is really squished. I think we should increase the max width of the DCAF website site-wide. I probably set a max width 18 months ago, and I set it too narrow. I'm not sure what it is now, but it could be set to 1600 as a max-width. After it hits max-width, content should center and margins should increase. 

LMK if you have questions!

screen shot 2017-06-05 at 8 22 24 pm

EricSchwartz7:

* change the `index.html.erb` wrapper's class from "container" to "row"

image

* Fixes #1103

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -13,6 +13,7 @@ $daria-color: #5D3E8D;
 $grey-color: #828282;
 $font-size: 17px;
 
+
 /* NAVBAR */
 .navbar-inverse{
   background-color: #f5f5f5;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/application.css.scss
```diff
@@ -301,7 +302,7 @@ $daria-color-lt : #825fb9;
       border-color: $daria-color;
     }
   }
-  
+
```
colinxfleming: appreciate you leaving the code in better shape than you found it! Especially important for the CSS since we generally touch it to fix things, and cleaner code makes it easier to find the stuff to fix.

PR #1136, opened 2017-07-11
[WIP] 654 - Event logging for use in an activity feed

### LINKED ISSUE #654: Feature req: Activity feed
A CM said it would be nice to have some representation of what she did over the course of the evening -- her point of reference was a facebook activity feed. This seems to stem from a concern that the call list being cleared would run the risk of people falling through the cracks. For all the things we hate about the spreadsheet it does make history accessible. 

Should keep in mind as an eventually thing.

Stems from #614

tedPen:

This change will log relevant events to the database, for use in a general activity feed.

* added event model to track events
* added a callback on the call model to create an event after save

* Fixes #654


app/models/events.rb
```diff
@@ -0,0 +1,22 @@
+# Object representing relevant actions taken by a case  manager
+class Event
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::History::Trackable
+  include Mongoid::Userstamp
+  # Fields
+  field :event_type, type: String
```
colinxfleming: May want to use our enum library `enumerize` for event type.

app/models/events.rb
```diff
@@ -0,0 +1,22 @@
+# Object representing relevant actions taken by a case  manager
+class Event
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::History::Trackable
```
colinxfleming: I think since these are disposable and immutable events, my inclination is to ditch historytrackable and userstamp here. Thoughts?

app/models/events.rb
```diff
@@ -0,0 +1,22 @@
...
+  include Mongoid::Userstamp
+  # Fields
+  field :event_type, type: String
+  field :cm_name, type: String
+  field :patient_name, type: String
+  field :patient_id, type: Integer
```
colinxfleming: Let's throw `line` in here too, the same way it shows up in `Patient`, if we can -- DCAF divides its work into three distinct buckets, and so we'll want to have these scopeable on line.

app/models/call.rb
```diff
@@ -27,6 +27,18 @@ class Call
                 track_destroy: true
   mongoid_userstamp user_model: 'User'
 
+  after_save :log_event, if: -> { status_changed? }
```
colinxfleming: We want to log an event on every call, so probably `after_create` and no need to make this conditional, right?

app/models/call.rb
```diff
@@ -27,6 +27,18 @@ class Call
...
+
+  def log_event
```
colinxfleming: @tedPen I'm interested in your thoughts -- we'll probably have to use this code in patient too, and there's totally a scenario where our users ask for more events. What do you think about throwing `log_event` into its own concern and having it take args that populate the object fields? I'm spitballing here, super open to other stuff, would love to know what you think.

colinxfleming:
This is shaping up wonderfully! Love the direction you're moving it in, it's perfect. Thanks for sticking on it!

app/views/dashboards/_activity_feed.html.erb
```diff
@@ -0,0 +1,6 @@
+<div id='activity_feed' class='margin-bottom'>
+  <h1 class='border-bottom title'>Activity Feed</h1>
+  <% Event.order_by(created_at: :desc).limit(10).each do |event| %>
```
colinxfleming: Let's move the db call into the controller and also filter on `where(line: current_user.line)`, if that's okay!

app/models/event.rb
```diff
@@ -0,0 +1,23 @@
...
+  enumerize :line, in: LINES, default: LINES[0] # See config/initializers/env_vars.rb
+  field :patient_name, type: String
+  field :patient_id, type: String
+
+  # Validations
+  event_types = [
+    'Reached patient',
+    "Couldn't reach patient",
+    'Left voicemail',
+    'Pledged'
```
colinxfleming: Change to `Pledge sent`

tedPen:
here's what the activity feed looks like right now: 
dcaf_activity_feed

If there's better ID data to pull in that would be good to update.

colinxfleming:
@tedPen that's looking real sharp. If that populates cleanly I'm comfortable just leaving the styling and cleanup to another PR (which we can do before we deploy this), and we can ship this. Thoughts?

colinxfleming:
Per discussion on Thursday, slapping approval on this and handling the frontend coding / any additional plug-in-to-the-architecture work in a separate PR. Thanks a ton for tackling this @tedPen , lots of folks are very excited about it!

PR #1137, opened 2017-07-12
Back out Lets Encrypt continued

colinxfleming:

Like #958 but after running `git pull origin master`.

* see #958


colinxfleming:
@DarthHater can you approve? I think we can ship this tomorrow with minimal pain, but want to make sure it passes CI first, and also not deploy

It's currently deployed in sandbox, if you want to poke at it some. Cert there should expire in Oct.

colinxfleming:
ps steps I took to swap this over:

* pushed the code changes
* pulled up authorizations with `heroku authorizations` and removed letsencrypt ssl 
* ran `heroku acm:auto:enable`
* removed env vars
* checked the cert end date
* ran `heroku acm:auto:disable` and `heroku acm:auto:enable` again for good measure

colinxfleming:
Pushing this thru. Confirmed it works on sandbox.

PR #1138, opened 2017-07-15
Support languages other than English and Spanish

colinmcglynn:

This PR replace the Spanish check box with a language drop down.

screen shot 2017-07-14 at 11 54 16 pm

* Fixes https://github.com/DCAFEngineering/dcaf_case_management/issues/1075


colinmcglynn:
I'll take a look at the failing tests over the weekend and fix the file conflict

app/helpers/patients_helper.rb
```diff
@@ -14,6 +14,10 @@ def race_ethnicity_options
      'Native American', 'Mixed Race/Ethnicity', 'Other']
   end
 
+  def language
+    [nil, 'English', 'Spanish', 'French', 'Korean']
```
colinxfleming: I think I'd rather store the nil instead of English for people who prefer english, since that's the bulk of our client base last I knew. I think that buys us a little bit of extra data space in that we won't have to save a default string 5000 times.

tldr: Can we remove the English option here, or better yet, have it display as 'English' but have the value of nil? I think that's the best of both worlds but if it winds up being a hassle, nbd.

fyi @alisonnjones -- please raise a flag NLT Wednesday afternoon if you'd prefer a different course of action.

app/models/concerns/exportable.rb
```diff
@@ -10,7 +10,7 @@ module Exportable
     "Has Alt Contact?" => :has_alt_contact?,
     "Voicemail Preference" => :voicemail_preference,
     "Line" => :line,
-    "Spanish?" => :spanish,
+    "Language?" => :language,
```
colinxfleming: No need for the question mark here, since this is no longer a boolean!

test/integration/update_patient_info_test.rb
```diff
@@ -114,7 +114,7 @@ class UpdatePatientInfoTest < ActionDispatch::IntegrationTest
       fill_in 'State', with: 'DC'
       fill_in 'County', with: 'Wash'
       select 'Voicemail OK', from: 'patient_voicemail_preference'
-      check 'Spanish Only'
+      select 'Spanish', from: 'patient_language'
```
colinxfleming: and tests!!! 😻

app/views/patients/_new_patient.html.erb
```diff
@@ -24,8 +24,8 @@
       </div>
 
       <div class="col-sm-2">
-        <%= f.form_group :spanish, label: { text: 'Spanish only' } do %>
-          <%= f.check_box :spanish, label: '' %>
+        <%= f.select :language, options_for_select(language, patient.language ),
+          label: 'Preferred Language', autocomplete: 'off' %>
         <% end %>
```
colinxfleming: I think this extra `end` is what's making the tests choke. (This is just a select instead of a `form_group ... do` block, so no need to close it anymore.) No promises but I bet this will get CI to either pass or mostly pass.

colinmcglynn: thanks! I'll give that a shot

app/helpers/patients_helper.rb
colinmcglynn: Two considerations I'll throw in the mix:
1) Is English the best default for all funds that will deploy the app?
2) Is it useful in the data analytics to use as a proxy for whether the question is being asked and answered as English vs the question wasn't asked?

I'm fine with either solution, I'll wait to hear back from @alisonnjones before I make the change

colinxfleming: I think English as the best default is a safe-ish assumption -- I don't know of any funds on our prospects list that. If it's possible it may be worth writing this so we can easily inject a default language into it at some point. (see #1092 for a direction we could feasibly take this.)

I defer to @alisonnjones on the analytics question -- she should be at the hacknight tomorrow so we can ask her point blank.

colinxfleming:
Looking good, thanks @colinmcglynn ! I think once we check in on the analytics question we're good to merge this. Thanks for wrenching this out so quick and clean!

We'll need to run a quick and dirty script on our databases to move the data over from the `language` -- something like `Patient.all.each { |pt| pt.update(language: 'Spanish') if pt[:spanish]; pt.unset(:spanish) }` that'll move our spanish records over and then strip the `spanish` field from the documents. I'll spin up an issue to handle that during our next deployment window after this gets in.

colinxfleming:
@colinmcglynn here's the implementation plan we developed with the analytics team -- `nil` can pretty much be a stand-in for english. So I think the remaining task here is as follows:

* change the dropdown so that the `nil` value shows up as 'English' on the dropdown
* pop into `app/models/concerns/exportable` and define a method `preferred_language`, which returns language if present, or English if it's nil or blank.

After that, we can :shipit: !

Sound like a plan? Let me know if there's anything I can clarify!

colinmcglynn:
@colinxfleming Works for me! I'll get on it this weekend

colinmcglynn:
@colinxfleming Those changes have been made I tested the export and it looks like it is working
screen shot 2017-07-22 at 9 47 59 pm

app/helpers/patients_helper.rb
```diff
@@ -14,6 +14,10 @@ def race_ethnicity_options
      'Native American', 'Mixed Race/Ethnicity', 'Other']
   end
 
+  def language
+    [['English', nil], 'Spanish', 'French', 'Korean']
```
colinxfleming: 😻

test/factories/patient.rb
```diff
@@ -3,7 +3,7 @@
     name 'New Patient'
     sequence(:primary_phone, 100) { |n| "127-#{n}-1111" }
     line 'DC'
-    spanish false
+    language 'English'
```
colinxfleming: I think we can get rid of this since we're assuming that nil == english, but I'll fix it direct since I think this is an important PR and I want to get it shipped.

Issue #1139, opened 2017-07-19
Remove duplicate fields from export


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

We rigged up a nice export for the analytics team, but it's got some duplicate fields in it that we should remove to make sure our analysts don't go crazy. 

* **What feature or behavior is this required for?**

Clean data!

* **How could we solve this issue? (Not knowing is okay!)**

Remove one of the `Race/ethnicity` duplicate fields and `employment status` duplicate fields from the `CSV_EXPORT_FIELDS` hash in `app/models/concerns/exportable.rb`. Your choice which one!

* **Anything else?**
 
Stems from @alisonnjones ' flag in #1127

SamuelLangenfeld:
Sent in a pull request for this issue.

PR #1146, opened 2017-07-19
remove duplicate line of race/ethnicity from csv field in exportable.rb

SamuelLangenfeld:

remove duplicate line of race/ethnicity from csv field in exportable.rb

* removes a field from the csv_export_field

* #1139


colinxfleming:
yusss @SamuelLangenfeld 

IIRC there's at least one other duplicate in there, employment status or something, could you make sure we've got that out of the way too?

PR #1155, opened 2017-07-23
Remove language from patient factory

colinxfleming:

Super small cleanup, removes language from the patient test factory.

* Removes language from the patient test factory, since nil == english

No view changes. 

No issues.


colinxfleming:
Minor commit + test suite, not production facing + test suite passes = smash that merge button

PR #1163, opened 2017-08-02
Render inline errors

### LINKED ISSUE #274: Render inline errors in forms with rails-bootstrap-forms gem

tingaloo:

Inline errors display on fail and success. Haven't wrote tests but this works well enough. No loading screen but we can work on that later.

Success and fail messages flash everytime, even after a user closes the flash notification box. If they find it annoying, we can stop it from opening after close.

* Add flash notifications.
screen shot 2017-08-01 at 10 53 15 pm

* Fixes #274


app/views/patients/update.js.erb
```diff
@@ -36,3 +36,14 @@ $('#change_log').html(
   // clear out fulfillment form
   $('#fulfillment').empty();
 <% end %>
+
+// flash success
+if ($( "#flash" ).length == 0) {
```
tingaloo: This conditional handles the case if the user deletes the flash notification, then another will pop up right after.

app/views/patients/update.js.erb
```diff
@@ -36,3 +36,14 @@ $('#change_log').html(
...
+  $('main > .container').prepend("<div id='flash'></div>")
+}
+
+$('#flash').removeClass().hide().fadeIn('slow')
```
tingaloo: Cool flashiness

app/views/layouts/application.html.erb
```diff
@@ -21,7 +21,11 @@
     <% end %>
     <main role="main">
       <div class="container">
-        <%= render 'layouts/messages' %>
+        <div id="flash">
```
tingaloo: Not exactly sure if we need this, I remembered following this https://stackoverflow.com/questions/23967390/rails-flash-notice-via-ajax/23967474

Then forgetting why I did it in the first place.

colinxfleming: If it turns out to be necessary, doable to just move it all to the layouts/messages partial so we can avoid editing the app layout?

app/views/patients/update.js.erb
```diff
@@ -36,3 +36,14 @@ $('#change_log').html(
...
+$('#flash').addClass("alert alert-success")
+$('#flash').html(
+  '<button type="button" class="close" data-dismiss="alert" aria-hidden="true">×</button><div id="flash_notice"> Patient data saved!</div>'
```
colinxfleming: I think we can straight up render a messages partial with something like `<%= j render 'layouts/messages' %>` and take advantage of the styling in there. (And the styling we'll enjoy when we finish off that long neglected alert styling formatting issue.) Thoughts?

app/views/patients/update_failed.js.erb
```diff
@@ -0,0 +1,9 @@
+// flash failure
+if ($( "#flash" ).length == 0) {
+  $('main > .container').prepend("<div id='flash'></div>")
+}
+$('#flash').removeClass().hide().fadeIn('slow');
+$('#flash').addClass("alert alert-danger")
+$('#flash').html(
+  '<button type="button" class="close" data-dismiss="alert" aria-hidden="true">×</button><div id="flash_notice"><%= @errors %></div>'
```
colinxfleming: Same deal, can probably render the partial here. Note that you might have it flash something in the controller action itself, see user creation for something similar I think.

test/controllers/patients_controller_test.rb
```diff
@@ -147,10 +147,10 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
       assert_response :success
     end
 
-    it 'should respond internal server error on failure' do
+    it 'should respond not acceptable error on failure' do
```
colinxfleming: good call!

app/views/patients/update_failed.js.erb
tingaloo: Hm, difference between the two is that the new calls render js. The flash does not persist after ```respond_to```. Are there any examples that handle that case?

app/views/patients/update.js.erb
tingaloo: Will this be in the ```update.js``` or in the controller action?

colinxfleming: It'd be in the update.js -- pretty much a drop in replacement for `'button type="button".....` as appears here.

From futzing, here's some drop in code to try out:

```
# patients controller
      @patient.reload
      flash.now[:notice] = 'Patient info successfully saved' # <- this saves the flash into a session variable that gets cleared out on next page load
      respond_to { |format| format.js }
# ...etc

# the flash success section of update.js.erb
$('#flash').html('<%= escape_javascript(render partial: "layouts/messages") %>'); // <- this just lets the existing template do the lifting of what to render for us
$('#flash').removeClass().hide().fadeIn('slow')
```

ps in futzing around with this I tried out the `fadeIn` effect and I really, really, really like it.

colinxfleming: to be clear, here's what the relevant part of `update.js.erb` looks like in here:

```js
// flash success
// if ($( "#flash" ).length == 0) {
  // $('main > .container').prepend("<div id='flash'></div>")
// }

// $('#flash').addClass("alert alert-success")
$('#flash').html('<%= escape_javascript(render partial: "layouts/messages") %>');
$('#flash').removeClass().hide().fadeIn('slow')
```

tingaloo: i feel dumb, im not sure how to pass the flash message into the messages partial

tingaloo: nevermind, read the first part of your code. Will push this soon!

colinxfleming: Definitely don't feel dumb because of rails flash stuff! It is a purer example than most of programmers making it up as they go -- tough for me to think of an analog elsewhere in web dev.

colinxfleming:
ps what are yr thoughts about trying to merge update and update_failed here? They're basically doing the same thing and if we have the flash set the good stuff I bet we can basically ditch update failed, yeah? Might take a slight further adjustment to update so it doesn't rerender all those templates on a failure, but I wonder.

app/views/layouts/application.html.erb
tingaloo: done

tingaloo:
I find that we still need a ```<div id='flash'>``` in ```application.html.erb``` for things to work, other than that this should be fine.

colinxfleming:
This is looking great to me! I might make a few place edits and revert the point release update of rails in a complementary PR to satisfy a few of my own neuroses, but the core here is really great.

Thanks a ton for your work on this @tingaloo ! This is something we've wanted for a long time (major quality of life improvement for CMs to know that their changes saved immediately!) and it's really impressive that you managed to craft it so cleanly. Very nicely done.

lwaldsc:
Also chiming in on a big thank you!!

PR #1164, opened 2017-08-03
Autoselect when 1 line

### LINKED ISSUE #1151: Autoselect line for single-issue funds

If a fund only has one line, we should autoselect that, so they don't have to bother to click thru the welcome page.

Not annoying case managers with extra clicks!

Probably adjust the `LinesController#new` method so that it checks on the number of lines available; if it's just one, autoselect it and redirect to the dashboard page.

* **Anything else?**

nah!

tingaloo:

Automatically selects and redirects to line if only 1 line count. Not sure how to test but it should be working.

Redirects to line if only 1 line.

* Fixes #1151

### DIFF
--- a/app/controllers/lines_controller.rb
+++ b/app/controllers/lines_controller.rb
@@ -1,6 +1,11 @@
 # Single-serving controller for setting current line for a user.
 class LinesController < ApplicationController
-  def new; end
+  def new
+    if LINES.count == 1
+      session[:line] = LINES[0]
+      redirect_to authenticated_root_path
+    end
+  end
 
   def create
     session[:line] = params[:line]
--- a/test/integration/select_line_test.rb
+++ b/test/integration/select_line_test.rb
@@ -24,6 +24,21 @@ class SelectLineTest < ActionDispatch::IntegrationTest
     end
   end
 
+  # TODO
+  describe 'line redirect on single line' do
+    before do
+      # TODO setup line array like so ['dc']
+      @user.line='DC'
+      # LINES=%w(DC).map(&:to_sym).freeze
+      puts @user.inspect
+      # how to change env var constants in this block?
+    end
+
+    it 'should redirect to the dashboard' do
+      assert_equal current_path, authenticated_root_path
+    end
+  end
+
   describe 'redirection conditions' do
     before { @patient = create :patient }
     it 'should redirect from dashboard if no line is set' do


lomky:
I would suggest an integration test. Set up a line and a user, log in the user and confirm the page you end up on is the dashboard and is not the line select?

tingaloo:
I'm not sure how to alter the global line variable, ```LINES=['md']``` in the before do block doesn't work.

app/controllers/lines_controller.rb
```diff
@@ -1,6 +1,11 @@
 # Single-serving controller for setting current line for a user.
 class LinesController < ApplicationController
-  def new; end
+  def new
+    if LINES.count == 1
```
colinxfleming: I'd like a controller test for new to go with this behavior too, but I'm happy to handle that since it's a little pedantic.

This is pretty much dead on, by the way 💯

test/integration/select_line_test.rb
```diff
@@ -24,6 +24,21 @@ class SelectLineTest < ActionDispatch::IntegrationTest
...
   end
 
+  # TODO
+  describe 'line redirect on single line' do
+    before do
+      # TODO setup line array like so ['dc']
+      @user.line='DC'
+      # LINES=%w(DC).map(&:to_sym).freeze
+      puts @user.inspect
+      # how to change env var constants in this block?
```
colinxfleming: hmm, this is tricky! `LINES` gets pulled from an env var `LINES` and gets set on app boot/in an initializer iirc, so I'd bet what we're looking to do here is, within the scope of this test case, set `LINES` to just 'DC' like you're doing in the commented out thing. 

I don't think rails boots up and down for every test, so I would bet that doing `ENV['LINES'] = [:DC].freeze` won't work. but worth a shot! If it's not a free basket I'll poke around a little.

app/controllers/lines_controller.rb
```diff
@@ -1,6 +1,11 @@
...
+      session[:line] = LINES[0]
```
colinxfleming: cool thing you can do with arrays in ruby/rails -- `Array.first` is equivalent to `Array[0]`. 'second', 'third' might work too, I forget. Cool either way, but a nice convenience method!

test/integration/select_line_test.rb
tingaloo: no free baskets here. on ```lines_path```, it always grabs from the env variable. Only way to get it to pass was to change ```%w(DC MD VA).map(&:to_sym).freeze``` in ```env_constants.erb``` to  ```%w(DC).map(&:to_sym).freeze```

Leaving this to you!

PR #1166, opened 2017-08-04
Inline errors nail in the coffin

### LINKED ISSUE #274: Render inline errors in forms with rails-bootstrap-forms gem

colinxfleming:

Playing the rest of the Cavs to @tingaloo 's LeBron, this takes his work in #1163 and sands off a rough edge or two.

The only real 'improvement' is skipping the re-render block if a patient wasn't changed (as in a failure to save, for example).

* flash messages on save!!!!!

@mebates does this look good to you?

Example success: 

screen shot 2017-08-03 at 11 24 20 pm

Example failure:

screen shot 2017-08-03 at 11 24 29 pm

@tingaloo can you confirm that this keeps with the spirit of your changes here? 

* Fixes #274


app/views/layouts/_messages.html.erb
```diff
@@ -1,8 +1,10 @@
-<% flash.each do |name, msg| %>
-  <% if name != 'timedout' && msg.length > 0 %>
-    <div class="alert alert-<%= name.to_s == 'notice' ? 'success' : 'danger' %>">
-      <button type="button" class="close" data-dismiss="alert" aria-hidden="true">&times;</button>
-      <%= content_tag :div, msg, :id => "flash_#{name}" %>
-    </div>
-  <% end %>
-<% end %>
+<div id='flash'>
```
tingaloo: I guess this solves about everything, nice!

mebates:
These look great!!!! 💜

PR #1169, opened 2017-08-05
More prominent Google SSO link ctd

### LINKED ISSUE #811: Phase out passwords in favor of google SSO
I'd like to move us entirely toward single sign on, so we can hop on google's security entirely and reduce the terror factor. Here are the hurdles we need to overcome to get to SSO-only:

* How do we get everyone google SSO keys to developers who need them without exposing them?
* How can we account for training needs, when a whole slew of people will show up and need accounts in a sandbox environment?

colinxfleming:

Some very slight adjustments to @rudietuesdays excellent and thoughtful work in #1149 to point people more toward google sign in.

* favor inline bootstrap classes over css
* some adjustments for my own neuroses
* follow 'one primary button per page' rule that @mebates has instilled unto me

@mebates does this look okay? We're trying to push people toward google sign in, primarily. The main diff is that `Sign in with Google` is now the button primary, and sign in with password is just a blank white button.

screen shot 2017-08-05 at 12 49 47 am

* Fixes #811

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -198,18 +198,6 @@ h5 {
   margin: 0 auto;
 }
 
-.authform a:first-of-type {
-  font-size: 1.625rem;
-  padding: 4px;
-  color: #FFFAF8;
-  background-color: $daria-color;
-  border-radius: 4px;
-}
-
-.authform a:first-of-type:hover {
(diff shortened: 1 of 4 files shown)


DarthHater:
LOVE THIS.

mebates:
"follow 'one primary button per page' rule that @mebates has instilled unto me"
👏 👏 👏 
my work here is done. looks great

PR #1170, opened 2017-08-07
downgrade omniauth-google-oauth2

colinxfleming:
Looks like this particular gem upgrade, despite being a point release, contains a breaking change of some sort.


colinxfleming:
Test failures not real, merging.

PR #1181, opened 2017-08-09
Deknapsack

### LINKED ISSUE #1177: What's the knapsack gem's problem?

Run tests in CI efficiently by splitting our tests onto parallel boxes in Circle. This used to be okay via something and then they started recommending the knapsack gem, but now THAT is causing some sketchy segfaults.

Running CI tests efficiently (meaningful, since we deploy to sandbox via CI)

Investigate the segfaults and come up with a solution that lets us continue splitting the test suite among boxes. Even is nice but not necessary. 

* **Anything else?**
nope!

colinxfleming:

Goal here is to get rid of knapsack and move back to regular parallelization without a sub-library, which I loathe.

Expect lots of back and forth and commits to circle.yml.

* smell ya later knapsack

* Fixes #1177


colinxfleming:
This seems to work, other than a failing test related to either upgrading to 5.1, or to something totally stupid like using `rails test` instead of `rake test`. @DarthHater other than the test failure, does this look okay to you?

DarthHater:
Have we tried raising this with knapsack and going hey why isn't this working? We could maybe do that to see if it's something easy.

colinxfleming:
tbh I was never wild about knapsack and threw it in because I was not willing to accept 6m builds, so I'm kind of inclined to take this opportunity to ditch it! This also lets us restore bubbling errors to the top with some junit xml noise, which knapsack didn't play nice with.

colinxfleming:
hm, looks like a host of other integration tests are also choking because of API changes. Wonder if this is fixable with systemtests?

colinxfleming:
well, that's most of em, anyway. remaining test failing in test/integration/user_management_test.rb:88. I'll deal with it later.

colinxfleming:
Had a hunch that a lonely operator / try would handle this and it seems to be right! Seeing if this passes CI.

colinxfleming:
boom. @DarthHater let me know if you want any changes, but this is all passing and we should be good to merge it into the 5.1 branch.

colinxfleming:
after, of course, our good for nothing CI build passes.

PR #1184, opened 2017-08-10
Alter CSP to allow for script tags to be executed

### LINKED ISSUE #1095: Load reports asynchronously

lapolonio:
Note that 'unsafe-inline' is ignored if either a hash or nonce value
is present in the source list. 

* csp script_src line should be: script_src: %w('self' 'unsafe-eval' 'unsafe-inline'),
* removal of nonce from render_async

* Fixes #1095

### DIFF
--- a/app/views/reports/index.html.erb
+++ b/app/views/reports/index.html.erb
@@ -8,6 +8,6 @@
   <div class="progress-bar progress-bar-striped"</div>
 </div>
  -->
-<%= render_async patients_report_path(timeframe: 'weekly'), nonce: content_security_policy_script_nonce %>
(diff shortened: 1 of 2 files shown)


colinxfleming:
thanks much @lapolonio ! If it's cool with you, going to let this sit for a minute while we sort out things in #1095 's discussion thread.

colinxfleming:
Per discussion in #1095 we're going to keep our CSP as is and pursue solutions to these problems that don't involve loosening CSP security. Marking closed. Thanks for helping us hash this out @lapolonio !

PR #1193, opened 2017-08-12
Add stub tests for single-line session set behavior

### LINKED ISSUE #1151: Autoselect line for single-issue funds

colinxfleming:

@tingaloo there are a few paths to testing this and none of them are fantastic. I propose that we punt and create an issue for this, since it's a little more involved and I'd rather get this in front of Baltimore on the sooner side.

* Adds commented out tests that we should get to pass, at some point (issue pending)

* Fixes #1151 
* launches off #1164


tingaloo:
Wrangling with env variables is tricky, but I can confirm single line redirect behavior is a go!

colinxfleming:
Thanks @tingaloo !

We're pushing a ton of stuff tomorrow during the scheduled deployment so I want to let this marinate a few extra days -- will merge Monday.

colinxfleming:
since this has hung out for awhile I thought I would push it over the line, and here's what I found, pretty much: 

* capybara tests are REALLY not down with swapping env vars around
* controller tests, on the other hand, are ready to party!

I think the controller test does introduce some testing around this behavior, and I think it's going to have to do because there straight up isn't a good rails solution to reload the app with different environment variables, test those, and then revert things to standard. Part of this is us cursing ourselves by setting these constants in a way that they're hard to stub, so if this continues to be a massive pain in the behind we should consider moving `config/initializers/env_var_constants` to a concern or helper, and including that in applicationcontroller. But anyway, fine for now, and I didn't think it was worth the overhaul at this point.

PR #1081, opened 2017-05-29
Fix dashboard N+1 queries

tingaloo:
I rule and have completed some work on Case Manager that's ready for review!

Optimized dashboard mongo queries.

This pull request makes the following changes:
call lists, etc should be snappy.

It relates to the following issue #s: 
* Fixes #1054

PR #914, opened 2017-03-05
Validate created_by_id instead of created_by

rivers:
Hopefully fixes #878

PR #1194, opened 2017-08-13
Fix dashboard n+1 queries

### LINKED ISSUE #1054: Eliminate n+1 query on call list

The top two pieces of the call list, `Your call list` and `Completed calls` (populated by `current_user.ordered_patients(current_line)` and `current_user.recently_called_patients(current_line)` respectively) are loading via individual finds. We should try to make this one query per partial, so it behaves more like the urgent patients partial (which is a simple, singular `{"find"=>"patients", "filter"=>{"_line"=>{"$in"=>[:DC]}, "urgent_flag"=>true}}`).

Page load optimizing, good db practice, keeping things feeling snappy, clean code.

Likely the answer here involves wrenching on `ordered_patients` and `recently_called_patients` to give those methods a sense of patients as a collection rather than a set of individual entries. (In other words, probably edit the methods to do the db call in there.)

* **Anything else?**
nope!

colinxfleming:

Jumps off @tingaloo 's sharp work in #1081 to eliminate an n+1 query.

Turns out we were both wrong! This wound up being a check in the view, which was way, way under. It was not my first, third, or fifth inclination, but I eventually traced it down to that, and it turns out that single check using the full `patients` set instead of `patient_ids` was causing our n+1 by re-getting `patients` over and over and over, instead of just checking the reference numbers.

looking back, could have probably guessed this -- this is the same thing that @rivers (hi! hope all is well back west 👋 ) fixed in #914 on a different model. But this was nested a little deeper, unfortunately, so it was pretty difficult to dig up. 

The more you know / the peril of rails magic, etc.

* Removes a reference that was causing an n+1 query

wooooo
screen shot 2017-08-12 at 8 33 42 pm

* Fixes #1054

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,7 +80,6 @@ class User
   validate :secure_password
 
   TIME_BEFORE_INACTIVE = 2.weeks
-  RECENT_TIMEFRAME = 8.hours.ago..Time.zone.now
 
   def secure_password
     return true if password.nil?
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -35,7 +35,7 @@
           <% if table_type == 'completed_calls' || table_type == 'urgent_patients' %>
             <td class="blank-td"></td>
           <% else %>
-            <% if current_user.patients.include?(patient) %>
+            <% if current_user.patient_ids.include?(patient.id) %>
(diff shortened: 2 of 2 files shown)


colinxfleming:
@tingaloo look good to you, at long last?

tingaloo:
Snapshot looks gold. I wonder what the logs look like with just your one
liner. Other than that, nails in the coffin!

PR #1207, opened 2017-08-17
Fix urgent flag bug

### LINKED ISSUE #1203: Patients still marked urgent after pledge sent

Need to revisit this because patients are still marked urgent after a pledge is sent. Refreshing the dash doesn't remove them.

My sanity

Looks like it has something to do with: https://github.com/DCAFEngineering/dcaf_case_management/pull/665

* **Anything else?**
screen shot 2017-08-15 at 8 57 36 pm

tingaloo:

On pledge submit, automatically removes urgent flag.
On pledge rescind, check if @patient.resolved_without_fund. If resolved, removes urgent flag.

I'm not really sure about the urgent flag flow, but i'm throwing this out there.

* Fix urgent flag

* Fixes #1203 
* Bumps #Y

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -59,8 +59,30 @@ class PatientsController < ApplicationController
     @external_pledge = @patient.external_pledges.new
   end
 
+  # turned off until update is resolved
+  # def update_urgent_flag patient
+  #   if patient.pledge_sent?
+  #     patient.update_attributes urgent_flag: false
+  #   else
+  #     if patient.resolved_without_fund
+  #       patient.update_attributes urgent_flag: false
+  #     end
+  #   end
+  #   return patient
+  # end
+
   def update
     if @patient.update_attributes patient_params
+      # @patient = update_urgent_flag(patient)
+      @patient.reload
+      if @patient.pledge_sent?
+        @patient.update urgent_flag: false
+      else
+        if @patient.resolved_without_fund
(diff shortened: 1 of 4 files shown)


app/views/patients/update.js.erb
```diff
@@ -27,6 +27,7 @@
     $('#sections').append(
       '<%= escape_javascript(render partial: "patients/fulfillment", locals: { patient: @patient })%>'
     );
+    <% @patient.update_attributes urgent_flag: false %>
```
colinxfleming: The idea here is dead on, but can we put this into the patients controller? Specifically, I'd like to see this happen as part of the update action -- we don't want to change object state in our views if we can possibly avoid it.

tingaloo:
I'm not sure if it is updating when I place it in ```patients#update```, can you verify?

app/controllers/patients_controller.rb
```diff
@@ -59,8 +59,30 @@ def edit
...
+  #     end
+  #   end
+  #   return patient
+  # end
+
   def update
     if @patient.update_attributes patient_params
+      # @patient = update_urgent_flag(patient)
+      @patient.reload
+      if @patient.pledge_sent?
```
colinxfleming: The core of this is pretty sound and looks good to me. I'm gonna hook this into an `after_save` callback on the Patient model -- seems dicey to me to put this in the controller layer, which is generally responsible for responding to http calls or whatever. (And probably add a test along the way!)

Issue #1201, opened 2017-08-14
Make reports queries more efficient


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Make reports at `/reports` work at scale!

* **What feature or behavior is this required for?**
Viewing reports and making informed judgments about the state of the program from that.

* **How could we solve this issue? (Not knowing is okay!)**
The reporting code is inefficient and involves a really heavy db load -- we should move the filters to the queries, move as much as possible to `pluck` instead of selects, and remove any other stuff that's causing memory brutality. 

The code lives in `app/lib/reporting/patient.rb` -- specifically, the first three methods in there. (The bottom three are more speculative/unimplemented shots at using mongoid's mapreduce.)

* **Anything else?**
nope!

lwaldsc:
![](http://www.wikihow.com/images/b/be/Complete-a-Mortal-Kombat-Finisher-Step-13.jpg)

PR #1212, opened 2017-08-17
Migrate to chartkick

### LINKED ISSUE #1095: Load reports asynchronously

lapolonio:

install chartkick and reproduce existing graphs

* install chartkick
* reformat JSON

* Fixes #1095


app/views/reports/_patient_report.html.erb
```diff
@@ -43,7 +10,9 @@
   <h2 class="border-bottom">Year to date</h2>
   <% end %>
   <div class="row">
-    <div id='chart-<%= timeframe %>_report'></div>
+    <div id='chart-<%= timeframe %>_report'>
```
colinxfleming: Love how clean this turned out!

lapolonio:
I didn't get it to work. Sorry I should have mentioned it was a work in progress. The payload is still an issue. I wanted an opinion on whether the manipulation should be in the query or frontend.

colinxfleming:
No worries! 

If the current output of the report method doesn't work with chartkick, probably we should change the reporting method, since those are intended to smoothly create a product that fits into whatever charting software we're using. There's a related ticket in #1201 to do more filtering in DB query side, since the current strategy of 'get em all and then mapreduce em' is tough at any scale besides small data, so we're going to have to do some wrenching in there anyway. 

Do you have any interest in taking a whack at that? Happy to assign it out if not.

colinxfleming:
We're going with a somewhat different reporting model now per @SerenaYudon 's recent work in #659 so closing this for the time being. My fault for the lousy guidance here, sorry bud.

PR #1215, opened 2017-08-19
Turn pledge limit tool tip into config

colinmcglynn:

Previously, the pledge limit dollar amounts were set as environmental variables, while the rest of the help text was hard coded in `patients_helper.rb`.  Now, each line of the help text as a config option.

The config model is only set up to accept drop down input options.  So the way I've done this is a little iffy, semantically speaking, but I'm not sure it's worth overhauling the model to fix it.

screen shot 2017-08-18 at 11 49 17 pm
screen shot 2017-08-18 at 11 49 28 pm

* Fixes https://github.com/DCAFEngineering/dcaf_case_management/issues/1119


app/views/patients/_abortion_information.html.erb
```diff
@@ -65,7 +65,7 @@
             <div id="pledge-contributions">
               <%= f.number_field :patient_contribution, label: 'Patient contribution', autocomplete: 'off', prepend: '$' %>
               <%= f.number_field :naf_pledge, label: 'National Abortion Federation pledge', autocomplete: 'off', prepend: '$' %>
-              <%= f.number_field :fund_pledge, label: "#{FUND} pledge", autocomplete: 'off', prepend: '$', label_class: 'tooltip-header', data: { 'tooltip-text': daria_limit_help_text() } %>
+              <%= f.number_field :fund_pledge, label: "#{FUND} pledge", autocomplete: 'off', prepend: '$', label_class: 'tooltip-header', data: { 'tooltip-text': pledge_limit_help_text_options().join("<br />") } %>
```
colinxfleming: Super minor change, but can you move the `join` to the helper instead of the view, to keep it consistent with how we handle this elsewhere? If it gives you a hard time about the html elements, try `safe_join` ( https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/safe_join )

colinxfleming:
ah, the config display is a loop. whoops. leaving that as is, but adding the integration test. my mistake!

PR #1216, opened 2017-08-19
change circle to chromedriver from phantom

colinxfleming:

Have a hunch that this will work better. Let's see.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+cache: bundler
 rvm:
   - 2.4.1
 services:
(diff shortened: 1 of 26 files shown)


colinxfleming:
A quick programming note: e3cee05 results from my making an improvement to the call log test reflecting our ability to hot reload, only to find that it was generating a bunch of divs with the same id. Solved this by moving the parent id to the menu and changing the ids of the partials to be `x_content`, for cleanliness' sake.

colinxfleming:
We don't generally need the purpose of this pr (a travis build, since it seems to eat mud at the same rate as circle; circle seems much better set up for systemtest than it did the first time we tried this out). But there are a few dom/html changes that we should definitely implement here, so I'm going to extract those into their own PR and close this one down.

Issue #998, opened 2017-04-14
Eliminate heroku memory errors


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
A long running issue that's been merely bad but not super problematic is our site's memory usage. It's consistently running on overdrive, causing all sorts of alerts, etc. Today we hit a 503 error which I think means we're starting to get into problematic territory.

* **What feature or behavior is this required for?**
Improve site reliability and reduce email alerts. (We hit that 503 error today because of 200% memory usage.)

* **How could we solve this issue? (Not knowing is okay!)**
Our goal here is to get us to consistently ride under 100% so the errors stop. I have no idea in the slightest but I bet someone does! Most likely there's some advice buried in heroku documentation or something else. 

* **Anything else?**
Here's our memory logs. They seem to start in the neighborhood of 90% and get more severe from there, then occasionally restart on the heroku sched.

<img width="1211" alt="screen shot 2017-04-14 at 1 19 27 pm" src="https://cloud.githubusercontent.com/assets/3866868/25050432/7170ad6a-2115-11e7-8020-35ceb53ea258.png">

colinxfleming:
Some blog posts and libs here: 

https://collectiveidea.com/blog/archives/2015/03/05/optimizing-rails-for-memory-usage-part-3-pluck-and-database-laziness

https://github.com/puma/puma/issues/342

https://github.com/arches/whacamole ?
https://github.com/schneems/puma_worker_killer ?

colinxfleming:
#1219 looking promising after a day.

colinxfleming:
I spent some time beating on this as hard as I could after #1219 and I think we can declare victory here. Closing.

PR #1219, opened 2017-08-20
decrease puma workers to see if it fixes those r14 errors

colinxfleming:

Reducing puma workers to 2 to see what effect it has on heroku R14 errors we're getting. See #998 

* Reduces puma workers from 3 to 2

* Bumps #998


colinxfleming:
Going to try this for a week and see how it goes. Invoking admin mode and merging.

PR #1088, opened 2017-06-02
Standardize most dates to MM-DD-YYYY (WIP)

tingaloo:
I rule and have completed some work on Case Manager that's ready for review!

Dug up old backend issues! I think this is a hack, because this formats the params of dates to YYYY-MM-DD before submitting to the DB (messes with frozen params). I get an odd feeling this is the wrong way to go around this issue, your thoughts @colinxfleming ?

Note, this fails on ```date_of_check```, since that uses some kind of date-picker. If i made it manual input, it will probably work. Manages to pass all tests and fixes all input/displays, but definitely needs a review.

This pull request makes the following changes:
* Formats user inputted dates MM-DD-YYYY to YYYY-MM-DD
* displays human dates
* Cleans up audit table a bit.

screen shot 2017-06-02 at 9 33 20 am

It relates to the following issue #s: 
* Fixes #860 
* Bumps #Y

PR #1220, opened 2017-08-20
Date format testing adjustments

### LINKED ISSUE #860: Standardize dates as MM-DD-YYYY

colinxfleming:

Test adjustments for the world's worst integration test failures. Poltergeist's driver has an interesting interpretation of `input type="date"` it seems. We'll probably need to fix when we move to a diff driver in #1216 / #1191 .

* Punts on some truly awful test problems with the integration test driver
* Everything in #1088 

Stems from #1088 , where all the good stuff is.

Tagging in @tingaloo for final thumbs up. Going to merge after the scheduled deploy today. 

* Fixes #860

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -74,8 +74,6 @@ class PatientsController < ApplicationController
     @patient = Patient.new
   end
 
-
-
   def data_entry_create
     @patient = Patient.new patient_params
     @patient.created_by = current_user
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -16,7 +16,6 @@ class PatientsController < ApplicationController
   end
 
   def create
-
     patient = Patient.new patient_params
 
     patient.created_by = current_user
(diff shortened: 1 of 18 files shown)


tingaloo:
Ah jeez, i guess this is the year we switch to chrome.

PR #1225, opened 2017-08-24
Fixing issue 1104 Search Functionality on Admin Panel Not Working

### LINKED ISSUE #1104: Search Functionality on Admin Panel not Working

We would like search functionality on the user management panel. Search button returns a POST error.

Required to search for users on the ```user_management``` panel.

Fix routes or some javascript search call needs tweaking.

* **Anything else?**
See #1073

KatSDC:

I added an if statement to check if the search results are nil in order to resolve the 500 error being thrown (which was actually a NoMethodError - undefined method `each' for nil:NilClass)

* app/views/users/_users_table.html.erb

* Fixes #1104

image

### DIFF
--- a/app/views/users/_users_table.html.erb
+++ b/app/views/users/_users_table.html.erb
@@ -13,6 +13,7 @@
       </tr>
     </thead>
     <tbody>
+      <% if users != nil %>
         <% users.each do |user| %>
           <tr>
             <td><input type="checkbox" id="select-<%= user.id %>"></td>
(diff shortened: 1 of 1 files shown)


app/views/users/_users_table.html.erb
```diff
@@ -13,6 +13,7 @@
       </tr>
     </thead>
     <tbody>
+      <% if users != nil %>
```
colinxfleming: Kind of an idiomatic think but can you swap this out to `if users.present?` Want to take advantage of that activesupport!

colinxfleming:
🦅

PR #1226, opened 2017-08-28
Shape frontend for events stream

### LINKED ISSUE #654: Feature req: Activity feed

colinxfleming:

Jumps off the structural work in #1136 to add a clean render stream for events as follows:

* Single serve events endpoint that renders a partial of events for that line
* Renders the activity feed async

* Refines some frontend stuff around #1136 

Screenshots coming when I get it to blend right

* Fixes #654

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -46,6 +46,12 @@ class Event
     end
   end
 
+  # Clean events older than three weeks
+  def self.destroy_old_events
+    Event.where(:created_at.lte => 3.weeks.ago)
+         .destroy_all
+  end
+
   private
 
   def pledged_type?
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -21,6 +21,7 @@ class Event
   enumerize :line, in: LINES, default: LINES[0]
   field :patient_name, type: String
   field :patient_id, type: String
+  field :pledge_amount, type: Integer
 
   # Validations
   validates :event_type, inclusion: { in: EVENT_TYPES }
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -25,7 +25,8 @@ class Event
 
   # Validations
   validates :event_type, inclusion: { in: EVENT_TYPES }
-  validates :cm_name, :patient_name, :patient_id, presence: true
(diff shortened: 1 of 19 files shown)


colinxfleming:
Here's a draft of what this looks like now: 

image

colinxfleming:
This is shaping up to be really usable and tested as of df6d6b8 . A few notes:

* @lapolonio identified an issue with turbolinks and `render_async`. I'm sure there's a turbolinks invocation we can use to get this to work right. I'm okay with punting on it for purposes of this PR though.
* I took an extremely back of the envelope approach to the icons. Would welcome feedback from @eheintzelman @serenayudon or @mebates on whether we should restyle that. Again, can happen outside the scope of this PR, but I want to make sure we don't let it slip.

I'll chase down any remaining test failures as well.

colinxfleming:
Tapping in @tedPen / @lomky for a review at yr leisure. No rush!

app/models/event.rb
```diff
@@ -0,0 +1,60 @@
...
+    when "Couldn't reach patient"
+      "called, but couldn't reach"
+    when 'Reached patient'
+      'reached'
+    end
+  end
+
+  # Clean events older than three weeks
+  def self.destroy_old_events
+    Event.where(:created_at.lte => 3.weeks.ago)
```
lomky: Huh. This is the first time Rails' readability has backfired for me. I read this the first ~3 times as "events created less than three weeks ago", instead of the correct "Events where created_at is before three weeks ago".

colinxfleming: This also bit me while I was writing it! There was a good five minutes where I was positive I was losing my mind.

colinxfleming:
Quick note so I don't lose it about formatting from #654 --

```
h1. [DC] Call Line Activity Log
h3. [Date]
bulleted body list
'Add' as an underlined blue link
```

PR #1227, opened 2017-08-30
Cancel pledge checkbox

### LINKED ISSUE #1046: What should the buttons say in the cancel pledge workflow?

If you go to cancel a pledge now, here's what you're going to see:

screen shot 2017-05-05 at 5 58 27 pm
screen shot 2017-05-05 at 5 58 16 pm

@mebates rightly pointed out that this veribiage is confusing in #925 -- " If my primary goal is to cancel, it's odd that the word "Cancel" is in the lower left corner of the modal, but that won't make it happen for me. I know that's the standard, but maybe we can use different words or just do an 'are you sure?'"

Clear, unambiguous direction if we have to cancel a pledge.

Two steps here: First, UX team should decide what we should change the copy to, then make a frontend change to the copy. 

* **Anything else?**
nope!

SamuelLangenfeld:

* removes checkbox from the cancel pledge modal
* changes _pledge.html.erb to only call views/shared/multi-modal for submitting a pledge, not cancelling. Cancelling a pledge is only one step.

Cancel pledge modal before:

image

Cancel pledge modal after:

image

* Fixes #1046

### DIFF
--- a/app/views/patients/_pledge.html.erb
+++ b/app/views/patients/_pledge.html.erb
@@ -1,6 +1,8 @@
-<%= render '/shared/multi_modal', disable_next: disable_continue?(patient), :captured => capture { %>
+
 
 <% if not patient.pledge_sent %>
+
+  <%= render '/shared/multi_modal', disable_next: disable_continue?(patient), :captured => capture { %>
   <div class="row hide" data-step="1" data-title="Confirm the following information is correct:">
     <% if patient.pledge_info_present? %>
       <% patient.pledge_info_errors.each do |error| %>
@@ -25,13 +27,13 @@
         <%= bootstrap_form_tag url: generate_pledge_patient_path(patient), method: 'get' do |f| %>
           <%= f.text_field :case_manager_name, label: 'Enter your name in this box to sign your pledge:' %>
           <%= f.submit 'Generate your pledge', class: 'btn btn-large btn-primary' %>
-        <% end %>      
+        <% end %>
       </div>
     <% end %>
   </div>
 
   <div class="row hide" data-step="3" data-title="Submit and send your pledge:">
-    <p class="col-sm-12">Awesome, you generated a DCAF Pledge! Thanks!</p>
+    <p class="col-sm-12">Awesome, you generated a <%= FUND %> Pledge! Thanks!</p>
     <p class="col-sm-12">Your pledge should appear in your computer's downloads. Find the document and go to <a href="http://www.onlinefaxes.com" target="_blank">onlinefaxes.com</a> to submit the pledge for your client.</p>
     <p class="col-sm-12"><strong>Once you've completed the online fax process, please use the check box below to indicate your pledge has been sent.</strong></p>
     <div class="col-sm-12">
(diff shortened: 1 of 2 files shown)


app/views/patients/_pledge.html.erb
```diff
@@ -1,6 +1,8 @@
-<%= render '/shared/multi_modal', disable_next: disable_continue?(patient), :captured => capture { %>
+
 
 <% if not patient.pledge_sent %>
+
+  <%= render '/shared/multi_modal', disable_next: disable_continue?(patient), :captured => capture { %>
```
colinxfleming: Looks like you swapped this around to just the multimodal on the submit pledge case. Not quite sure I follow why you did that, can you give me a sense of how this played out?

app/views/patients/_pledge.html.erb
```diff
@@ -40,19 +42,25 @@
...
+    <div class="row">
+      <div class="col-sm-12">
+        <p>Are you sure you want to cancel this pledge?</p>
+      </div>
     </div>
   </div>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-default js-btn-step pull-left" style="color:red;" data-dismiss="modal">No</button>
+
+    <%= form_for @patient, html:{id:'cancel-pledge-form'}, remote: true do |f|%>
```
colinxfleming: This is a little bit of a departure from what we were doing before, can you help me understand what's going on?

app/views/patients/_pledge.html.erb
SamuelLangenfeld: Since this is a one step modal process the multi_modal partial doesn't work very well with it. 

The multi_modal footer has two buttons, a cancel and a next-step/complete button. Based on Molly's design it seemed like we wanted a footer with just a 'No' and 'Yes' button for cancelling the pledge, with the 'No' button closing the modal and the 'Yes' button submitting the :pledge_sent value of false. 

The multi_modal design has the form in the body of the modal and the footer has buttons for going back and forth through the modal steps and for closing the modal. The cancel pledge modal is only going to be one step, so we don't need the multi_modal footers to get to other steps in the process.

app/views/patients/_pledge.html.erb
SamuelLangenfeld: Here I'm creating the footer for the cancel-pledge modal. For the submit pledge modal, we call the multi-modal partial and that generates the footer itself. The cancel pledge modal doesn't have multiple steps so we don't want to call the multi-modal partial. This means have to create a separate footer for the cancel pledge modal. Since we don't want a footer that takes us back and forth across steps, I made a footer that just has the two buttons we want without the step process. I also gave the two buttons the same classes that the multi-modal footer buttons would have to keep the same styling.

I changed the form from the bootstrap_form_for to just the form_for helper because of the way bootstrap handles dismissing modals. The normal way to close a bootstrap modal is to just throw data-dimiss:"modal" onto the button. If you do that bootstrap will call the js method preventDefault. This unintentionally stops the button from actually submitting the form. So instead I just made a button with the more basic form_for helper to handle the submission. 

Pledge.js.erb is the file that creates the modal and renders the pledge partial. I added on some javascript that lets the #cancel-form-pledge submit button close the modal as well as doing its normal job of submitting the form with the :pledge_sent attribute of false.

PR #1229, opened 2017-09-03
Changes to_csv to enumerable, streams CSV generation

### LINKED ISSUE #1224: Move CSV export to an async call

The only place in the app we're still getting memory errors or timeouts is the CSV export, which has become too big to pull. Rather than optimize the export I'd rather cheat and just move it to an async call, so it doesn't timeout in heroku after 30 seconds and we can just trust that it'll work no matter how huge it is. 

Export to CSV successfully, even after data gets massive.

Without knowing off the top of my head, here are two options:
* http://smsohan.com/blog/2013/05/09/genereating-and-streaming-potentially-large-csv-files-using-ruby-on-rails/ (streaming, probably requires more code)
* https://stackoverflow.com/questions/34729473/downloading-dynamically-generated-large-files-in-rails (activejob / background job processing -- probably requires more rigging)

* **Anything else?**

Tough one, but if you have experience doing this (especially on heroku), go for it!

valeriecodes:

Turns CSV generation into streaming request to avoid timeouts.

* Changes `to_csv` return enumerable (it only seemed to be called in one place)
* Turns patient csv generation into stream. Heroku won't time out on streaming requests unless it goes 55 seconds or more between packets, so this should allow for arbitrarily long CSVs (https://devcenter.heroku.com/articles/request-timeout)

* Fixes #1224

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -8,9 +8,7 @@ class PatientsController < ApplicationController
   def index
     respond_to do |format|
       format.csv do
-        now = Time.zone.now.strftime('%Y%m%d')
-        csv_filename = "patient_data_export_#{now}.csv"
-        send_data Patient.to_csv, filename: csv_filename, format: 'text/csv'
+        render_csv
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lwaldsc:
Also popping in to say thanks Valerie! This is fantastic!

PR #1230, opened 2017-09-04
Add doc explaining lines

### LINKED ISSUE #1070: Document how lines work

A core part of the app machinery is multiple lines, and this is not terribly well documented. We should add documentation about how it fits into the workflow of the app.

Bus test

Write documentation!

* **Anything else?**
Stems from #1069

colinxfleming:

Adds a doc explaining what lines are and how they work in the app. I hope!

* Adds `docs/HOW_LINES_WORK.md`

@lomky or @DarthHater can you review? Please def let me know if there's anything I can do to clarify anything in here, as I want to make sure this doesn't live mostly in my head.

* Fixes #1070

### DIFF
--- a/docs/HOW_LINES_WORK.md
+++ b/docs/HOW_LINES_WORK.md
@@ -11,6 +11,16 @@ own lines so patients in a given service area are easier to keep track of.
 A case manager can only be working on one line at a time; their line scopes their search
 results.
 
+Many (possibly most!) funds operate only one, primary line. The system functions perfectly
+well with this use case.
+
+## What this means for abortion funds
+
+For abortion funds, this means that adding a new division to how they field calls
+is as straightforward as changing the `DARIA_LINES` environment variable. If a
+fund decides to split their phone intake into two, just alter the environment variable.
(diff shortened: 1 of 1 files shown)


docs/HOW_LINES_WORK.md
```diff
@@ -0,0 +1,31 @@
...
+we set an array of the lines `LINES`. (We default it to the DCAF use case, which operates
+lines for DC, MD, and VA.) It reads from the environment variable `DARIA_LINES`.
+
+We bubble that up and make it accessible in the app via the [LinesHelper](../app/helpers/lines_helper.rb).
+We also have a convenience method, `current_line`, which aliases to the session variable.
+
+Case managers have their lines set as a session variable, and they have to log out to
+change it.
+
+## What this means for abortion funds
```
lomky: I suggest this section go above the engineer section, as technical users are interested in both but non-technical users only want to read through the intro and this one.

colinxfleming: clutch insight! Done

colinxfleming:
Thanks very much @lomky !

PR #1234, opened 2017-09-06
Move Patient Language to a config

colinmcglynn:

This PR switches the language drop down from a hard coded list to a config that can be set by admins like we have for insurance options.

screen shot 2017-09-06 at 11 55 32 am

* Fixes #1138


colinxfleming:
This looks good to me! Thanks for setting up tests too!

Separately I'm going to file a ticket for how we're rendering the 'how this displays', since language is kind of a special case what with the nil value being associated with English. Maybe we swap this to selects, who knows.

Beautiful clean work @colinmcglynn !

PR #1245, opened 2017-09-26
modify flash notice in patients contoller #update to give timestamp

### LINKED ISSUE #1179: Add timestamp to 'patient saved' flash message

When PatientsController#update saves some data, it flashes a 'Patient saved!' message in green, which is nice. Let's add a timestamp to that of `MM-DD-YYYY HH:MM A/PM` to give folks a heads up about what the current status is.

Patient saving enhancements.

Pretty much just like above!

* **Anything else?**
Nope!

SamuelLangenfeld:

Adds a timestamp to the success notice in patients_controller update method.

* Modifies patients_controller #update to give a timestamp in the flash notice.

* Fixes #1179

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -64,7 +64,7 @@ class PatientsController < ApplicationController
     end
     if @patient.update_attributes patient_params
       @patient.reload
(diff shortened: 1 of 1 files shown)


colinxfleming:
thanks @SamuelLangenfeld !

PR #1255, opened 2017-09-30
remove an encompassing form-group that caused inconsistent spacing

### LINKED ISSUE #1248: Design improvement: Consistent spacing on checkboxes

If you go to a patient edit page (by clicking on a patient name link) and then click on the `Abortion Information` link on the left sidebar, you'll see this: 

image

The spacing between the checkboxes on the left and checkboxes on the right are noticeably inconsistent! We'd like to straighten them out so the stuff on the right is spaced the same as on the left.

Nice, clean design!

The code for this is in `app/views/patients/_abortion_information.html.erb`. Most likely it involves applying whatever HTML structure is on the left to the checkboxes on the right.

* **Anything else?**

thanks!

CodingTea17:

On the abortion information form there are two sets of checkboxes. The set under the input field is wrapped in a checkbox class div whereas the second set to the right were wrapped in a checkbox class div within a form-group div. The margins on the nesting of those two divs was causing the inconsistency. My quick fix is to remove the encompassing form-group as the checkbox class provides better styling for the checkboxes.

* Removes two encompassing form-group divs from the `app/views/patients/_abortion_information.html.erb`

b4_after

* Fixes #1248

### DIFF
--- a/app/views/patients/_abortion_information.html.erb
+++ b/app/views/patients/_abortion_information.html.erb
@@ -30,12 +30,8 @@
 
           <div class="col-sm-6">
             <div class="info-form-right">
-              <%= f.form_group :resolved_without_fund do %>
-                <%= f.check_box :resolved_without_fund, label: "Resolved without assistance from #{FUND}" %>
-              <% end %>
-              <%= f.form_group :referred_to_clinic do %>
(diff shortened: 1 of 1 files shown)


CodingTea17:
@colinxfleming I found you through the hacktoberfest tag

PR #1257, opened 2017-10-01
Left align table headers

### LINKED ISSUE #1246: Design fix: Left align columns with column data

Left align column titles with column data in `table_content` elements, like so:

image

Consistent, clean design to keep our users happy.

Probably by slapping a class on the table in `app/views/dashboards/_table_content.html.erb` and then writing some CSS affecting the header row.

* **Anything else?**

You can see this element in action by logging in, going to the main dashboard at the authenticated root, and then looking at the table under Your Call List.

TraGicCode:
closes #1246

### DIFF
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -76,7 +76,7 @@ $font-size: 17px;
     text-transform: uppercase;
     letter-spacing: 1.5px;
     background: #f1f1f1;
-    text-align: left;
+    padding-left: 0px !important;
   }
   tr {
     td {
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -76,6 +76,7 @@ $font-size: 17px;
     text-transform: uppercase;
     letter-spacing: 1.5px;
     background: #f1f1f1;
+    text-align: left;
   }
   tr {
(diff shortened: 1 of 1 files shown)


colinxfleming:
Super elegant fix, @TraGicCode ! Can I get a screenshot if you get a chance?

TraGicCode:
Sure here you go
screen shot 2017-10-01 at 1 20 51 am

app/assets/stylesheets/pregnancies.scss
```diff
@@ -76,6 +76,7 @@ $font-size: 17px;
     text-transform: uppercase;
     letter-spacing: 1.5px;
     background: #f1f1f1;
+    padding-left: 0px !important;
```
colinxfleming: I'd love if we could avoid using `important` tags here, since that sets us up for potential future problems. Does this work without it?

TraGicCode: Yes unfortunately it requires the important.  I did not dig in any further as to what is conflicting with the precedence.

colinxfleming: Looks like this is duking it out with `.table > thead > tr > th` in bootstrap. Probably not winning that battle so let's merge as is.

colinxfleming:
Looks good, thanks @TraGicCode !

PR #1258, opened 2017-10-01
Take text-plus-sign out of its own column

### LINKED ISSUE #1247: Design improvement: Left-align 'show additional text' plus icon for notes

In our table content partial, we have a plus sign icon that displays additional text for the next note if it breaks grid. Right now it's in its own `td` cell, and we should have it inline with the text it's associated with rather than right aligning.

image

Nice and clean design, making it clearer what the button does.

In `app/views/dashboards/_table_content.html.erb`, put `plus_sign_glyphicon` in the same `td` cell as the note instead of its own. Make sure to remove the corresponding `th` in the table header.

* **Anything else?**

thanks!

TraGicCode:
This closes #1247

### DIFF
--- a/app/views/dashboards/_table_content.html.erb
+++ b/app/views/dashboards/_table_content.html.erb
@@ -16,7 +16,6 @@
         <th>Notes</th>
         <th></th>
         <th></th>
-        <th></th>
       </tr>
     </thead>
     <tbody id="<%= table_type %>_content">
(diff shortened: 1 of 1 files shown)


colinxfleming:
Thanks much @TraGicCode ! Appreciate your taking the time to grind this out. 

Can you do me a solid and post a screenshot of what this looks like on the page now in this PR thread?

TraGicCode:
Sure thing.  My Mac crashed so it will be a bit

PR #1261, opened 2017-10-03
WIP PR for Issue #1115

### LINKED ISSUE #1115: Document behavior of features within the app
Issue template TK, but I got an email this morning asking how the call list / completed calls worked, and I'm sure that's not the only part where the 'how it works' is squirreled away in docs and pull requests somewhere. There might be a UX thing here to make this kind of useful information a little more available to CMs in the thick of it.

KatSDC:
I rule and have completed some (emphasis on some) work on Case Manager that's ready for review!

These are my initial changes around adding tooltips to other areas of the applications. Honestly the stuff I was doing to the table contents in the index.html.erb is complete gibberish at this point. The important stuff is happening in _abortion_information.html.erb

* Well there is a (?) next to the checkbox for Resolved without assistance from DCAF now, so that's a start!

* Fixes #1115

### DIFF
--- a/app/assets/javascripts/patients.coffee
+++ b/app/assets/javascripts/patients.coffee
@@ -91,3 +91,15 @@ $(document).on 'turbolinks:load', ->
       placement: 'top',
       title: $(@).data( 'tooltip-text' )
     } )
+
+$( '.tooltip-header' ).parent().find('.has-tooltip, .form-control').each ->
+  $(@).parent().find( 'span.tooltip-header-help' ).tooltip( {
+    html: true,
+    placement: 'bottom',
+    title: $(@).data( 'tooltip-text' )
+  } )
+  $(@).parents('.form-group').find('.tooltip-header-help').tooltip( {
+    html: true,
+    placement: 'bottom',
+    title: $(@).data( 'tooltip-text' )
+  } )
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -72,4 +72,28 @@ module PatientsHelper
   def pledge_limit_help_text_options
     Config.find_or_create_by(config_key: 'pledge_limit_help_text').options
   end
+
+  def call_list_help_text
+    return "This sortable list is used to keep track of patients to call back during your shift. Use the search to populate it."
+  end
+
+  def completed_calls_help_text
+    return "This is a list of patients you have already called tonight."
+  end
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/patients.coffee
```diff
@@ -91,3 +91,12 @@ $(document).on 'turbolinks:load', ->
       placement: 'top',
       title: $(@).data( 'tooltip-text' )
     } )
+
+
+  $( '.tooltip-header' ).parent( '.checkbox' ).each ->
```
colinxfleming: Here's what I came up with after hacking on this for a bit -- this seems to work but it's pretty ugmo -- 

```
  $( '.tooltip-header' ).parent().find('.has-tooltip, .form-control').each ->
    $(@).parent().find( 'span.tooltip-header-help' ).tooltip( {
      html: true,
      placement: 'bottom',
      title: $(@).data( 'tooltip-text' )
    } )
    $(@).parents('.form-group').find('.tooltip-header-help').tooltip( {
      html: true,
      placement: 'bottom',
      title: $(@).data( 'tooltip-text' )
    } )
```

It looks like the DOM structure of `bootstrap_form_for`'s checkboxes and text inputs are different, so I split the difference, selected them all thru, and then just call `tooltip` on empty elements. I'm sure that's a javascript venial sin but I couldn't think of a good way to do `parents with class OR parent` and have it blend cleanly.

You'll also note that this depends on the checkboxes having a `has-tooltip` class. A more optimal way to do that would be to collapse the first line there to something like this:

```
  $( '.tooltip-header' ).parent().find('input').filter((x) -> x.attr('data-tooltip-text') != "").each -> ....
```

but I couldn't figure out the magic invocation.

tldr: This will work as is if you paste it in and slap 'has-tooltip' on the input elements in the views!

app/helpers/patients_helper.rb
```diff
@@ -72,4 +72,28 @@ def disable_continue?(patient)
...
+
+  def status_help_text
+    return true
+  end
+
+  def record_new_external_pledge_help_text
+    return "To record a pledge from another abortion fund, choose the fund from the dropdown menu, enter the amount of the pledge, and click Create External Pledge."
+  end
+
+  def resolved_without_DCAF_help_text
```
colinxfleming: downcase dcaf in this method name please!

Issue #1214, opened 2017-08-18
Customize BRAAF Information 


NerdyGirl537:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Get ready to bring BRAFF onto system.

* **What feature or behavior is this required for?**
Changing any tool tips, pledge caps, and fund specific information for BRAAF.

* **How could we solve this issue? (Not knowing is okay!)**
Change related fields per BRAAF's guidance

No cap for up to 13w6d. For 14 weeks the cap is $300.
Use BRAAF name for logo (currently no art work type logo available)
BRAAF website link: blueridgeaaf.com
BRAAF call line number: 434-963-0669
Currently no CM Resource link (hide for now?)

* **Anything else?**
Let me know if we missed anything.  I have a copy of their pledge form.  Will we be able to do auto-generate for them?

colinxfleming:
Down the line:

* There's a PR that I'm going to merge today #1215  that'll let them set their own text for cap guidance, so nothing much to do there. 
* The places I see the logo are the footer and the favicon. I think it's fine on both of those tbh, since this is a DCAF/Code for DC product, pretty much.
* The phone / website link are an oversight, I think! I'll spin up an issue to set those to env vars. 
* No resources link means the link just won't show up, so we're good on that. 
* Autogen is DCAF only for the immediate future, sad to say. They should plan on use their existing pledge shipping processes at launch.

colinxfleming:
@NerdyGirl537 when you get a sec could you send me the emails of anyone at BRAAF who should be on the updates distro list? 

If you wouldn't mind also emailing me that copy of their pledge form that would be 💯

colinxfleming:
Per an email, need to customize the fax link as well -- BRAAF has a different service they're using.

colinxfleming:
Talked to Toby tonight and she brought up zip codes -- rediscuss

colinxfleming:
I'm going to make a semi-executive decision and say no zip code for now, since I don't think it serves our project to continue going back and forth on this even if it does make them adjust their procs a little bit. If this is a pain for them on an ongoing basis then we should rediscuss, but for now they should think about just making a note about eligibility or zip in their notes.

The discrete tasks should be good to go here, so marking complete.

PR #1262, opened 2017-10-03
Allow fund information from env vars in footer

colinxfleming:

Per #1214 this makes it possible to set a DARIA_FUND_PHONE

* Uses an env var `DARIA_FUND_PHONE` to set a fund's phone number (defaulting to DCAF's #)
* Switches out the hardset DCAF to`DARIA_FUND_FULL` in the footer

* Bumps #1214 

@lomky for review please!

### DIFF
--- a/config/initializers/env_var_constants.rb
+++ b/config/initializers/env_var_constants.rb
@@ -11,4 +11,5 @@ LINES = if ENV['DARIA_LINES'].present?
 # Definition of fund
 FUND_FULL = ENV['DARIA_FUND_FULL'] || 'DC Abortion Fund'
 FUND_DOMAIN = FUND_FULL.gsub(' ', '').downcase
+FUND_PHONE = ENV['DARIA_FUND_PHONE'] || '202-452-7464'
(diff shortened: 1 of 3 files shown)


config/initializers/env_var_constants.rb
```diff
@@ -11,4 +11,5 @@
 # Definition of fund
 FUND_FULL = ENV['DARIA_FUND_FULL'] || 'DC Abortion Fund'
 FUND_DOMAIN = FUND_FULL.gsub(' ', '').downcase
+FUND_PHONE = ENV['DARIA_FUND_PHONE'] || '202-452-7464'
```
lomky: Do we want it to default to our number? It seems like ours is the _least_ likely not have the ENV vars setup :smile: 

Not sure what a good alternative would be, though. Is there any sort of national registry or AF info line?

colinxfleming: Haha, I definitely hear you. It kinda fits with the existing pattern of DCAF as the primary use case we work against, and other funds as 'eh, throw it in an env var if you need something different'. (Whether that is a good pattern is probably questionable.)

PR #1264, opened 2017-10-04
Fix Tablesort after Search Bug

### LINKED ISSUE #1099: Stupidtable Sort does not work after Search

In ```users/index.html.erb```, stupidtable sort does not work after search

It's a bug.

Something about javascript and turbolinks?

* **Anything else?**
See #1073

tingaloo:

Fix sorting by date after searching for a user.
A not so neat hack by calling the entire autosort function after a search completes. There might be a better way calling a javascript function elsewhere but I forget how to do so.

* Call autosorting table after search.js.erb dispatches

* Fixes #1099 
* Bumps #Y

### DIFF
--- a/app/assets/javascripts/table_sorting.coffee
+++ b/app/assets/javascripts/table_sorting.coffee
@@ -2,21 +2,19 @@
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
 
-ready = ->
+$(document).on 'turbolinks:load', ->
   # make the columns of all of the tables sortable
   sortTables()
-$(document).on 'ready page:load', ready
 
 sortTables = ->
-  autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table', '#user-list table']
-  autosortable_tables.forEach ( table ) ->
-      stupidtable = $( table ).stupidtable()
-
-      # when a column is clicked, make sure to reset all of the other columns back to [-]
-      stupidtable.on 'beforetablesort', ( event, data ) ->
-          stupidtable.find( 'th' ).filter( ( i ) -> return i isnt data.column ).find( '.arrow' ).html( '[-]' )
-
-      # when a column is clicked, show an up or down arrow appropriately
-      stupidtable.on 'aftertablesort', ( event, data ) ->
-          arrow = if data.direction is $.fn.stupidtable.dir.ASC then '[&uarr;]' else '[&darr;]'
-          stupidtable.find( 'th' ).eq( data.column ).find( '.arrow' ).html( arrow )
+  $('table').each (index) ->
+    stupidtable = undefined
+    stupidtable = $(this).stupidtable()
+    stupidtable.on 'beforetablesort', (event, data) ->
+      stupidtable.find('th').filter((i) ->
+        i != data.column
+      ).find('.arrow').html '[-]'
+    stupidtable.on 'aftertablesort', (event, data) ->
+      arrow = undefined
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/table_sorting.coffee
```diff
@@ -2,18 +2,21 @@
...
+ready = ->
+  # make the columns of all of the tables sortable
+  sortTables()
+$(document).on 'ready page:load', ready
 
-        # when a column is clicked, make sure to reset all of the other columns back to [-]
-        stupidtable.on 'beforetablesort', ( event, data ) ->
-            stupidtable.find( 'th' ).filter( ( i ) -> return i isnt data.column ).find( '.arrow' ).html( '[-]' )
+sortTables = ->
+  autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table', '#user-list table']
```
colinxfleming: Should probably make this a class rather than doing it by ids, no?

app/views/users/search.js.erb
```diff
@@ -1,6 +1,26 @@
 
 <% if @results.any? %>
-  $('#user-list').html("<%= escape_javascript(render partial: 'users/users_table', locals: { users: @results }) %>");
+  $('#user-list').html("<%= escape_javascript(render partial: 'users/users_table', locals: { users: @results, autosortable: true }) %>");
+
+  //TODO figure out way to import function from /javascripts/table_sorting.coffee
```
colinxfleming: So I checked and I think this is a problem in table content on the dashboard after we add or remove a user, too. I wonder if we can move the stupidtable invocation to its own partial (something like `app/views/shared/_activate_stupidtable.js.erb`) and drop in something like `<%= render partial: shared/activate_stupidtable %>`?

tingaloo: Don't know where the edge cases are but I made the partial for rendering.

app/assets/javascripts/table_sorting.coffee
```diff
@@ -2,18 +2,19 @@
 # All this logic will automatically be available in application.js.
 # You can use CoffeeScript in this file: http://coffeescript.org/
 
-jQuery ->
-  $(document).on 'ready page:load', ->
-    # make the columns of all of the tables sortable
-    autosortable_tables = [ '#call_list table', '#completed_calls table', '#urgent_patients table', '#user-list table']
-    autosortable_tables.forEach ( table ) ->
-        stupidtable = $( table ).stupidtable()
+$(document).on 'turbolinks:load', ->
```
tingaloo: This fixes a bug when you navigate between pages and the stupidtable no longer works.

app/assets/javascripts/table_sorting.coffee
```diff
@@ -2,18 +2,19 @@
...
 
-        # when a column is clicked, make sure to reset all of the other columns back to [-]
-        stupidtable.on 'beforetablesort', ( event, data ) ->
-            stupidtable.find( 'th' ).filter( ( i ) -> return i isnt data.column ).find( '.arrow' ).html( '[-]' )
-
-        # when a column is clicked, show an up or down arrow appropriately
-        stupidtable.on 'aftertablesort', ( event, data ) ->
-            arrow = if data.direction is $.fn.stupidtable.dir.ASC then '[&uarr;]' else '[&darr;]'
-            stupidtable.find( 'th' ).eq( data.column ).find( '.arrow' ).html( arrow )
+sortTables = ->
```
colinxfleming: Fun fact I learned after poking at coffeescript recently: You can export functions by slapping an `@` in front of em. Kinda nice! (This is good as is though.)

PR #1265, opened 2017-10-07
Make fax service in the footer settable by env var

colinxfleming:

BRAAF requested the ability to set their own fax service URL in the footer, so hooking that up for them. 

* Adds a constant `FUND_FAX_SERVICE` for use in places, defaulting to the DCAF option
* plugs it into the footer
* also makes clicking that link open in a new tab

* Bumps #1214 

@colinmcglynn  for review please!


colinxfleming:
Will do, thanks much @colinmcglynn !

Issue #858, opened 2017-02-19
ACCOUNTING: Implement accountant billing view


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Generate a billing report for accountants!

* **What feature or behavior is this required for?**
Easier accountant reconciliation data entry, since we tend to punch these in ~50 at a time!

* **How should we solve this issue?**
@eheintzelman has put together a slamming design, so we're mostly just writing backend and frontend code to match!

![accountant billing report - annotated](https://cloud.githubusercontent.com/assets/17989540/21289345/43d3ba04-c469-11e6-92d8-ff22f8b44660.png)

I think what we want here is:
* a `Patient` method that filters down to patients who have been marked as `fulfilled` (see the `Fulfillment` object)
* Revise the accountant table_content to match the designs, with a few changes: 
  * make abortion care $, weeks at proc, and pledge sent date editable
* On loading the accountants page, make the report just load for now. don't worry about the `reports` dropdown above the accountant billing section
* keep the search thing functioning the way it is now
* add an integration test to confirm that you can find a patient from this search bar, and edit a patient's fulfillment information (we can spin this off if needed)

The troublesome parts of this are probably more frontend than backend, so marking accordingly.

Stems from #643

lomky:
Tagging interest on this.

lomky:
Working on this for hacknight

lomky:
I have a bit done on this, but nothing substantial. Planning on picking it up again at hacknight Monday, but if someone wants to knock it out before then they're welcome to steal it :)

FeminismIsAwesome:
I'm gonna steal zing! @lomky do you have any code/branch that I might want to look at or should I go at it from scratch?

lomky:
Thank you for your theft @FeminismIsAwesome !

Here is my scratch work I put together, it's not much yet:

https://github.com/lomky/dcaf_case_management/compare/master...lomky:858_implement_accountant_billing_view

colinxfleming:
moving from the slack channel because I wrote kinda unclear directions here (sorry!):

* the code should live in `app/views/accountants/index.html.erb` and associated partials
* do the fields as follows (overruling what's in the designs):
- clinic (sourced from `patient.clinic_name` / `patient.clinic.name`)
- pt initials (source from `patient.name`)
[...]

PR #1267, opened 2017-10-09
WIP accountant billing page

tingaloo:

Have a basic accounting view that shows patients with fulfillments. It paginates by 25. 

BUG: On fulfillment update, if you previously had a search filter, it gets reset and loads the entire patient list again. Look very closely between 2nd and 3rd screenshot, on any update (in this case check number), it will reset to original set of patients without search filter.

TODOs
- [x] Implement pagination
- [x] Fulfillment partial on edit
- [x] Write tests
- [x] Search Functionality
- [ ] Additional ajax tests
- [ ] Possibly fix search bug on fulfillment update
- [x] Minor style changes

* Add table
* Add sorting 

screen shot 2017-10-13 at 10 19 22 am
screen shot 2017-10-13 at 10 19 39 am
screen shot 2017-10-13 at 10 19 59 am

* Bumps #858


colinxfleming:
@tingaloo I think this is pretty much a perfect use case for `will_paginate`!

colinxfleming:
I think this is 100% on the right track! Thanks for picking it up @tingaloo !

tingaloo:
Soo close getting this to work.

For tests, i can't use ```wait_for_ajax``` because phantom driver is missing?

And I discovered creating a form within a table is not allowed https://stackoverflow.com/questions/5967564/form-inside-a-table. So having the inline edits within the table cell won't work unless we wrap tables with ```form```. That means we have something that looks like so if we want inline editing to technically work. Otherwise the form will never wrap properly and no inline...

screen shot 2017-10-11 at 8 31 03 am

As of now, the form div looks like so with no children.
```<form role="form" class="edit_patient" id="edit_patient_59d6e4b52cde1127ba38f255" action="/patients/59d6e4b52cde1127ba38f255" accept-charset="UTF-8" data-remote="true" method="post"></form>```

**Comment by colinxfleming** [comment]
`wait_for_ajax` phantom error is a mystery to me -- is that in CI or in docker or your local? 

I didn't know that about forms inside `tr`s! Sorry for the scrub directions on my part. I think we have a few options here, and you're free to go with your favorite:

a) style the form so that it's stylistically identical to a table row
b) instead of the form, have a link in the `tr` that pops up a modal, which populates it with `_fulfillment` for that patient (this might be nice because it'd let us reuse the partial)
c) something else tbd?

File: Gemfile [hunk 1]
```diff
@@ -34,6 +34,9 @@ gem 'rack-test', '~> 0.6.3', require: 'rack/test'
 gem 'clinic_finder', '~> 0.0.1'
 gem 'geokit'
 gem 'secure_headers', '~> 3.6', '>= 3.6.4'
+gem 'kaminari-mongoid'
+gem 'kaminari'
```
colinxfleming: Haven't heard of this lib but it looks sweet! 👍

app/controllers/accountants_controller.rb
```diff
@@ -1,10 +1,14 @@
...
+  
+  def index
+    @patients = Patient.where('fulfillment.fulfilled' => true)
+    @patients = Kaminari.paginate_array(@patients).page(params[:page]).per(25)
+  end
 
   def search
-    @results = Patient.search params[:search]
-
+    @results = Patient.where('fulfillment.fulfilled' => true).where(:name => /#{params[:search]}/i)
```
colinxfleming: I'd suggest taking advantage of the `Patient.search` method here and chaining it against the `where`. e.g. `Patient.search(params[:search]).where('fulfillment.fulfilled' => true)` if memory serves. (s/o to @KatSDC for writing a slamming search algo!)

app/helpers/accountants_helper.rb
```diff
@@ -0,0 +1,13 @@
+# Functions primarily related to accountant billing view
+module AccountantsHelper
+
+  def patientInitials(patient)
```
colinxfleming: Might behoove us to move this to the `Patient` object as an instance method, so we can do `@patient.initials`? thoughts?

app/controllers/accountants_controller.rb
tingaloo: Can you tell me where that method definition is handled, i'd like to see if it supports cases.

Gemfile
tingaloo: I think ```will_paginate``` didn't support ```mongoid```, this looks cooler and provides cool pagination links.

tingaloo:
```wait_for_ajax``` error is from local. 

I'll probably do option b unless we REALLY want inline editing within a table. I'll close this PR up and start a new issue for that.

colinxfleming:
I think you can probably keep this PR cooking, right? Looks like you have most of the backend machinery in place already, and what's left is mostly just beating on views, I'd guess?

app/controllers/accountants_controller.rb
colinxfleming: The search internals for patients are here -- https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/models/concerns/patient_searchable.rb

colinxfleming:
cc @eheintzelman -- does that strategy (pop up modal rather than inline form) sound ok to you? Anything special we should do around behavior?

eheintzelman:
@colinxfleming That should work.  I'm imagining a pencil icon to the right of each row, indicating the data is editable. The pencil icon triggers a modal with all the fields. Something similar to what you see in LinkedIn when you edit a section of your profile.

colinxfleming:
@tingaloo anything else on this front?

tingaloo:
Ready to go. Bugs are minor and we can spin them off as individual issues.

colinxfleming:
Gonna in-house this so we can all push to it -- closing in favor of #1292

Issue #1233, opened 2017-09-05
Activity feed should live-ish update


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

The activity feed needs to be reloaded for it to show changes. Instead it would be awesome if it reloaded when a CM takes an action that would add something to it (making a call or a pledge).

* **What feature or behavior is this required for?**

Activity feed usability

* **How could we solve this issue? (Not knowing is okay!)**

* At a minimum, on an event creation it would be nice to reload the partial.

* If someone's dying to try out ActionCable, that's one way too!

* **Anything else?**

Stems from #1226

colinxfleming:
After a lot of reflection this is probably more trouble than it's worth and hitting reload won't kill anybody.

PR #1273, opened 2017-10-14
Live event stream enhancement

tingaloo:

Works, just needs to sort a few bugs.

Once that's resolved, this can go live. To test, start a redis server and call somebody.

- [x] Add ActionCable
- [ ] Fix link bug, see screenshot
- [ ] Fix CSP to allow ```ws://localhost:3000/cable```
- [x] Add Redis to heroku

screen shot 2017-10-14 at 2 14 32 pm 

Currently links are disabled. Referencing route paths by calling the event helper through a concern breaks this. Error looks like so ```NoMethodError - undefined method for #<ActionView::Base:0x00007fd274fcaa50>```

* Adds ActionCable for cool live feed

* Fixes #X
* Bumps #1233

### DIFF
--- a/app/assets/javascripts/cable.js
+++ b/app/assets/javascripts/cable.js
@@ -1,0 +1,13 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
+// You can generate new channels where WebSocket features live using the rails generate channel command.
+//
+//= require action_cable
+//= require_self
+//= require_tree ./channels
+
+(function() {
+  this.App || (this.App = {});
+
+  App.cable = ActionCable.createConsumer();
+
+}).call(this);
--- a/app/assets/javascripts/cable/subscriptions/event.coffee
+++ b/app/assets/javascripts/cable/subscriptions/event.coffee
@@ -1,0 +1,4 @@
+jQuery ->
+  App.cable.subscriptions.create "EventChannel",
+    received: (data) ->
+      $('#event-list').append '<p>'+data['event']+'<p>'
--- a/config/initializers/csp.rb
+++ b/config/initializers/csp.rb
@@ -7,7 +7,7 @@ SecureHeaders::Configuration.default do |config|
     default_src: %w('self'),
     script_src: ["'self'", "'unsafe-eval'", gon_gem_sha, popover_sha],
     font_src: %w('self' fonts.gstatic.com),
-    connect_src: %w('self'),
+    connect_src: %w(*),
     style_src: %w('self' 'unsafe-inline'),
     report_uri: ["https://#{ENV['CSP_VIOLATION_URI']}/csp/reportOnly"]
   }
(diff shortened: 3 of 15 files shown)


colinxfleming:
Oh sweeeeeeeeet. Thanks a ton for this, it is AWESOME.

I'm gonna have some questions as I dig in to this, but I'm super excited about this.

app/assets/javascripts/cable.js
```diff
@@ -0,0 +1,13 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
+// You can generate new channels where WebSocket features live using the rails generate channel command.
+//
+//= require action_cable
+//= require_self
+//= require_tree ./channels
+
+(function() {
```
colinxfleming: Per action cable's docs, you might need to add `<%= action_cable_meta_tag %>` to the applicaiton.html.erb layout: see https://github.com/rails/rails/tree/master/actioncable#consumer-configuration

app/assets/javascripts/cable/subscriptions/event.coffee
```diff
@@ -0,0 +1,4 @@
+jQuery ->
+  App.cable.subscriptions.create "EventChannel",
+    received: (data) ->
+      $('#event-list').append '<p>'+data['event']+'<p>'
```
colinxfleming: I think you mean `prepend` here -- we want to drop it on top rather than on bottom!

config/initializers/csp.rb
```diff
@@ -7,7 +7,7 @@
     default_src: %w('self'),
     script_src: ["'self'", "'unsafe-eval'", gon_gem_sha, popover_sha],
     font_src: %w('self' fonts.gstatic.com),
-    connect_src: %w('self'),
+    connect_src: %w(*),
```
colinxfleming: hm. I THINK we want to change this to `'self'` and whatever the env var for our future redis server is but not sure. I'll dig in on this.

colinxfleming:
PS if you're comfortable mucking around in docker-compose or heroku app.jsons, would you mind making the following changes?

* edit the docker-compose to require `redis` (should more or less be the same as 
* edit the `app.json` addons section by appending `plan: redis` with the other addons we have. (I assume there's a redis service there)

app/assets/javascripts/cable.js
tingaloo: this is to separate configurations between local and production, really not familiar with the production side of things but i could give it a shot.

tingaloo:
I looked at the redis services heroku offers. Generally they are paid, but I like [heroku-redis]( https://elements.heroku.com/addons/heroku-redis). Up to 20 connections free which I think is more than enough?

colinxfleming:
@tingaloo 20 connections ought to be plenty, I doubt we've ever hit more than ~10 concurrent users.

colinxfleming:
After a lot of thought on this I think I'm gonna nuke it -- I think adding a new system dependency for what's basically a very small enhancement is not quite worth the tradeoff, and I don't think we're ever going to be in a position where we need a perfect pubsub, so I'm going to put a pin in this for now and close the issue. New year, new approach of not trying to actively do everything, haha.

Definitely keeping this proof of concept around just in case.

tingaloo:
good call, close em.

PR #1274, opened 2017-10-19
Remove all patients button

### LINKED ISSUE #1202: Call List 'remove all' button for dashboard

Make a button to remove all calls from your call list. 

If the call list does not purge properly, it's a pain to remove each call by hand. 

`¯\_(ツ)_/¯ `

* **Anything else?**
I tried to put an ascii nyan cat here but markdown hates joy.

tingaloo:

Add button that removes all patients from user.

* Remove all button

screen shot 2017-10-18 at 7 37 38 pm

* Fixes #1202


lwaldsc:
Yeeeeeeeeeeeeessssssssssss

app/assets/stylesheets/dashboards.scss
```diff
@@ -33,3 +33,8 @@
   color: red;
   padding: 0px 10px 0px 10px;
 }
+
+.remove-all-patients-link {
```
colinxfleming: Can we do this with bootstrap styling rather than a one-off CSS class? Assigning the classes `text-danger` and `pull-right` should do the trick.

app/controllers/users_controller.rb
```diff
@@ -81,14 +81,21 @@ def new # TODO needs more rigorous testing
...
   def remove_patient
     current_user.remove_patient @patient
     respond_to do |format|
-      format.js { render template: 'users/refresh_patients', layout: false }
+      format.js { render template: 'users/refresh_patients', locals: {msg: 'Successfully removed Patient'}, layout: false }
+    end
+  end
+
+  def remove_all_patients
+    current_user.patients.destroy_all
```
colinxfleming: If memory serves, you want `clear` here rather than `destroy_all` (which I believe might actually actually destroy the patient records, unfortunately).

app/views/dashboards/_table_content.html.erb
```diff
@@ -47,4 +47,7 @@
       <% end %>
     </tbody>
   </table>
+  <% if table_type == 'call_list' && current_user.patients.count > 0%>
+    <%= link_to "Remove all patients from your call list", remove_all_patients_path(current_user), class:"remove-all-patients-link", data: { confirm: "Are you sure you want to remove all your calls?" }, method: :patch, remote: true %>
```
colinxfleming: Let's change link text to `Clear your call list` and the alert text to `Are you sure you want to clear your current call list?`

app/views/layouts/_call_list_messages.html.erb
```diff
@@ -0,0 +1,6 @@
+<div id='flash'>
```
colinxfleming: I think this might be redundant to `app/views/layouts/_messages` -- can we use regular `_messages` instead?

test/controllers/users_controller_test.rb
```diff
@@ -185,6 +185,23 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  describe 'remove_all_patients method' do
```
colinxfleming: Can you add a test here to confirm that it doesn't destroy patients in db? Something like `assert_no_difference 'Patient.count' do patch remove_all_patients(@user), ...` should do the trick.

app/controllers/users_controller.rb
```diff
@@ -81,14 +81,21 @@ def new # TODO needs more rigorous testing
...
 
   def remove_patient
     current_user.remove_patient @patient
     respond_to do |format|
-      format.js { render template: 'users/refresh_patients', layout: false }
+      format.js { render template: 'users/refresh_patients', locals: {msg: 'Successfully removed Patient'}, layout: false }
+    end
+  end
+
+  def remove_all_patients
```
colinxfleming: Thoughts about changing the name of this to `clear_call_list`? I think that might be a little closer to actual behavior -- your thoughts?

tingaloo:
Wow i forgot about the flash stuff we did, time flies.

app/controllers/users_controller.rb
```diff
@@ -80,15 +80,25 @@ def new # TODO needs more rigorous testing
 
   def add_patient
     current_user.add_patient @patient
+    flash.now[:notice] = "Successfully added call"
```
colinxfleming: I think we can do without the flash notices, would you mind stripping them?

test/controllers/users_controller_test.rb
```diff
@@ -185,6 +185,31 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
+
+    it 'should respond successfully' do
+      assert_response :success
+    end
+
+    it 'should remove all patients' do
+      assert_equal @user.patients.count, 0
+    end
+
+    it 'should not remove patients' do
```
colinxfleming: I'd like a vocabulary change so we don't confuse ourselves six months from now: can you call this test `it 'should not destroy patients'`? Remove is a word we use at a different endpoint and I'm worried I'm going to look at this in six months and be like 'wait, what?'

test/controllers/users_controller_test.rb
```diff
@@ -185,6 +185,31 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  describe 'clear_call_list method' do
+    before do
+      patch add_patient_path(@user, @patient_1), xhr: true
+      patch add_patient_path(@user, @patient_2), xhr: true
+      patch clear_call_list_path(@user), xhr: true
```
colinxfleming: Can you add some kind of check that our user has 2 patients here? Maybe an assert or raise something?

config/routes.rb
```diff
@@ -11,6 +11,7 @@
     patch 'users/:user_id/add_patient/:id', to: 'users#add_patient', as: 'add_patient', defaults: { format: :js }
     patch 'users/:user_id/remove_patient/:id', to: 'users#remove_patient', as: 'remove_patient', defaults: { format: :js }
     post 'users/search', to: 'users#search', as: 'users_search', defaults: { format: :js }
+    patch 'users/clear_call_list/:id', to: 'users#clear_call_list', as: 'clear_call_list', defaults: { format: :js }
```
colinxfleming: A few changes to the routing here:

* can you have it patch to `users/clear_call_list`, without an id, since you're using `current_user` in the controller method?
* for maximum clarity on what's going on, can you have it route `as: 'clear_current_user_call_list'`?

app/views/users/refresh_patients.js.erb
```diff
@@ -1,3 +1,6 @@
+$('#flash').html('<%= escape_javascript(render partial: "layouts/messages") %>');
```
colinxfleming: Probs no need for the flashes here per above, I think?

app/controllers/users_controller.rb
```diff
@@ -2,7 +2,7 @@
 class UsersController < ApplicationController
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
   before_action :confirm_admin_user, only: [:new, :index, :update]
-  before_action :find_user, only: [:update, :edit, :destroy, :reset_password]
+  before_action :find_user, only: [:update, :edit, :destroy, :reset_password, :clear_call_list]
```
colinxfleming: No need to do the `before_action` on `:clear_call_list`, since you aren't doing a user lookup on id.

app/views/dashboards/_table_content.html.erb
```diff
@@ -47,4 +47,7 @@
       <% end %>
     </tbody>
   </table>
+  <% if table_type == 'call_list' && current_user.patients.count > 0%>
+    <%= link_to "Clear your call list", clear_call_list_path(current_user), class:"pull-right text-danger", data: { confirm: "Are you sure you want to clear your current call list?" }, method: :patch, remote: true %>
```
colinxfleming: I think this will also flatten completed calls. Can you make a note of that in the alert text?

tingaloo:
Let's nail this up.

config/routes.rb
```diff
@@ -11,6 +11,7 @@
     patch 'users/:user_id/add_patient/:id', to: 'users#add_patient', as: 'add_patient', defaults: { format: :js }
     patch 'users/:user_id/remove_patient/:id', to: 'users#remove_patient', as: 'remove_patient', defaults: { format: :js }
     post 'users/search', to: 'users#search', as: 'users_search', defaults: { format: :js }
+    patch 'users/clear_current_user_call_list', to: 'users#clear_current_user_call_list', as: 'clear_current_user_call_list', defaults: { format: :js }
```
tingaloo: i think it should be this? ```users/:user_id/clear_current_user_call_list```

colinxfleming: I think we're looking for it to either hook into an `:id` (in which case we should do the db lookup to get whatever user corresponds to that id), or assume that it relies on `current_user` (which is whoever is logged in, and I think is a safe assumption). Works either way, but should be one or the other.

PR #1275, opened 2017-10-21
Add response headers

### LINKED ISSUE #1197: Fix cache-control and pragma http header set for a few endpoints

Prevent caching on certain controllers in the headers. Exact alert content is below:

screen shot 2017-08-12 at 11 41 20 pm

screen shot 2017-08-12 at 11 41 27 pm

Removing security flags, even very scoped and small ones!

Try to adjust our caching behaviors so that we prevent caching via headers at the endpoints in the screenshot. 

* **Anything else?**
Stems from ZAP scan in #394

### LINKED ISSUE #1198: Content-type header missing when hitting google sso endpoint

Set a content type header on our hits of the google oauth endpoint, to reduce a MIME security risk.

Exact content of the alert from ZAP as follows:

screen shot 2017-08-12 at 11 44 55 pm

Resolving security alerts, small ones included!

No idea! probably adjusting the oauth controller and adding a header.

* **Anything else?**
Stems from #394

valeriecodes:

Adds headers. Not sure how to test the content-type one or if adding that header on devise pages will cause issues (though I can't think why, but they did seem to be deliberately excluded before) so please check my work!

* Fixes #1198
* Fixes #1197

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery prepend: true, with: :exception
 
   before_action :configure_permitted_parameters, if: :devise_controller?
-  before_action :prevent_caching_via_headers, unless: :devise_controller?
+  before_action :prevent_caching_via_headers
   before_action :authenticate_user!
 
   # whitelists attributes in devise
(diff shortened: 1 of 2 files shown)


colinxfleming:
Test suite is passing, so I think this is okay, and if it chokes in sandbox I'll flag. I'm like 90% sure that we just copied boilerplate code to set up devise controller stuff, so 'deliberate' might be overstating it a little, haha.

Thanks a ton for taking care of this! 💯

PR #1276, opened 2017-10-21
Add noopener nofollow to new tab links

colinxfleming:

Was working on something unrelated and and discovered that `target='_blank'` in links is a minor security flaw. It's not really a problem for us since I don't think we're pulling in anything external, but, you know.

* Adds `rel='noreferrer nofollow'` to a pair of links
* Missed a spot on changing online faxes to an env var, fixes that as long as I'm wrenching around

No view changes.

No issue.

@valeriecodes could I get a review at your leisure, if you'd be so kind?


colinxfleming:
Thanks @valeriecodes !

PR #1281, opened 2017-10-24
Fix pledge generated by, pledge sent by

mdworken:
Hey all!

* properly displays name of the Case Manager that generated/sent a pledge (previously displayed the user object and it's pointer)


colinxfleming:
Test failure is a flaky test. Merging. Thanks a ton for spotting this @mdworken !

PR #1168, opened 2017-08-05
WIP integrating clinic finder frontend

tingaloo:
I rule and have completed some work on Case Manager that's ready for review!

Tried to make a frontend similar to the wireframes posted in #978. Wanted to put this out and maybe @colinxfleming can see if it's in the right direction. I am mighty rusty on my bootstrap. 

The cool part about this so far is clicking the Find clinics > search clinics, and clicking find clinics again.

This pull request makes the following changes:
* Attempt to make a frontend

screen shot 2017-08-06 at 1 06 29 pm

It relates to the following issue #s: 
* Fixes #978

PR #1287, opened 2017-10-29
Integrating clinic finder to frontend, ctd

### LINKED ISSUE #978: Integrate Clinic Finder with CM App

Integrate Clinic Finder with the CM App so CM's can find the closest or cheapest clinic for their PT based on location and gestation limit. 

Needs UX exploration on the Abortion Info>Clinic Details Section

colinxfleming:

Make clinic finder work.

* Adds a form for the clinic finder
* Implements gem changes

image

Continuation of PR #1168  by @tingaloo 

* Fixes #978


app/controllers/clinicfinders_controller.rb
```diff
@@ -4,10 +4,15 @@ class ClinicfindersController < ApplicationController
   def index; end
 
   def search
-    @abortron = Abortron::ClinicFinder.new("#{Rails.root}/clinics.yml")
-    @nearest = @abortron.locate_nearest_clinic patient_zipcode: params[:zip].to_i.to_s,
-                                               gestational_age: params[:gestation].to_i
-    @cheapest = @abortron.locate_cheapest_clinic(gestational_age: params[:gestation].to_i)
+    puts 'searching fasdlfndsl'
+    puts params
+    @abortron = ClinicFinder::Locator.new(Clinic.all,
```
colinxfleming: There are a few API changes here:

* Now takes `Clinic` objects
* the object is `ClinicFinder::Locator` instead of `Abortron` ( 😢  )
* Some changes in the args for everything

app/views/clinicfinders/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= bootstrap_form_tag url: clinicfinder_path do |f| %>
```
colinxfleming: Somewhat tempted to dump this into `clinicfinders#_new` to make it a bit more restful, and render that partial on page load instead of summoning it thru ajax. Going to think on this.

app/controllers/clinicfinders_controller.rb
colinxfleming: Also need to pass in API keys. Can't remember if I do em by env var in the gem or not.

colinxfleming:
Minimally functional thing is working, which is cool! We'll probably ship the small version and then figure out how to handle cost display first.

TODOS BEFORE MERGE
* closer design match
* better error handling? might have to do this gem-side too
* tests

LONGER TERM
* get clinic gestation limits/costs entered in!

colinxfleming:
I am in absolute disbelief that tests pass on this but I'll take it!

colinxfleming:
Taking aim at this next!

colinxfleming:
This is what we're going to end up with for v1.

screen shot 2018-01-13 at 12 49 58 am

We should keep the issue open, but this will yield a decent proof of concept and let us tell what the problems are, how useful it is, etc.

colinxfleming:
urrrrgh this got caught up in the dependency hell. resolving now by hand. BLEHHHH

app/views/clinicfinders/search.js.erb
```diff
@@ -1,9 +1,11 @@
...
+  $("#clinic-finder-results").append("<%= escape_javascript(render partial: 'clinicfinders/result_table', locals: { title: 'Nearest Clinics', clinic_fields: [:name, :location, :accepts_naf, :distance], clinics: @nearest }) %>");
 <% end %>
 
-<% if @cheapest.any? %>
-  $(".clinic-finder-results").append("<%= escape_javascript(render partial: 'clinicfinders/result_table', locals: { title: 'Most Affordable Clinics', labels: ['Name', 'Cost'], clinic_fields: [:name, :cost], clinics: @cheapest }) %>");
+<% if @cheapest&.any? %>
+  $("#clinic-finder-results").append("<%= escape_javascript(render partial: 'clinicfinders/result_table', locals: { title: 'Most Affordable Clinics', clinic_fields: [:name, :cost], clinics: @cheapest }) %>");
 <% end %>
+
+$("#clinic-finder-results").slideDown('fast');
```
colinxfleming: Hit em with the SLIDE DOWN

app/views/patients/_abortion_information.html.erb
```diff
@@ -9,7 +9,7 @@
     <div class="row">
       <div class="col-sm-12 info-form">
         <h3> Clinic details </h3>
-        <%= bootstrap_form_for patient, html: { id: 'abortion-information-form' }, remote: true do |f| %>
+        <%= bootstrap_form_for patient, html: { id: 'abortion-information-form-1' }, remote: true do |f| %>
```
colinxfleming: This is necessary because, as this project has driven home, forms-within-forms lead to some seriously evil magic. This splits the main patient information form into two instead.

db/seeds.rb
```diff
@@ -26,14 +26,14 @@
...
-Clinic.create! name: 'Sample Clinic 1 - DC', street_address: '123 Fake Street',
-               city: 'Washington', state: 'DC', zip: '20011'
-Clinic.create! name: 'Sample Clinic 2 - VA', street_address: '123 Fake Street',
-               city: 'Arlington', state: 'VA', zip: '22204'
-Clinic.create! name: 'Sample Clinic with NAF', street_address: '123 Fake Street',
-              city: 'Washington', state: 'DC', zip: '20011', accepts_naf: true
-Clinic.create! name: 'Sample Clinic without NAF', street_address: '123 Fake Street',
-              city: 'Washington', state: 'DC', zip: '20011', accepts_naf: false, accepts_medicaid: true
+Clinic.create! name: 'Sample Clinic 1 - DC', street_address: '1600 Pennsylvania Ave',
+               city: 'Washington', state: 'DC', zip: '20500'
```
colinxfleming: We have to give these real zip codes now so the clinic finder can find them.

colinxfleming:
I think we're in business! Some more stuff to do here (NAF filtering, etc), but we should push this out in the spirit of getting it in front of users sooner. I've labeled it experimental in the app to reflect that.

@tingaloo can I get a review on this when you get a chance?

colinxfleming:
Awwww yis.

app/controllers/clinicfinders_controller.rb
```diff
@@ -1,14 +1,17 @@
...
-  def index; end
-
   def search
-    @abortron = Abortron::ClinicFinder.new("#{Rails.root}/clinics.yml")
-    @nearest = @abortron.locate_nearest_clinic patient_zipcode: params[:zip].to_i.to_s,
-                                               gestational_age: params[:gestation].to_i
-    @cheapest = @abortron.locate_cheapest_clinic(gestational_age: params[:gestation].to_i)
+    return head :bad_request if params[:zip].blank?
+
+    @abortron = ClinicFinder::Locator.new(
```
tingaloo: Is abortron the right name for this?

colinxfleming:
Here's a gif of the interaction:

jan-13-2018 10-54-56

app/views/clinicfinders/search.js.erb
tingaloo: hitem

app/controllers/clinicfinders_controller.rb
colinxfleming: Probably not, haha. It also doesn't need to be an instance variable, so I'll rename it to something like `finder` and localize it.

colinxfleming:
Thanks @tingaloo !

PR #1290, opened 2017-11-04
Fix urgent flag bug - model edition

### LINKED ISSUE #1203: Patients still marked urgent after pledge sent

colinxfleming:

Moves @tingaloo 's work in #1207  out of the controller and into the model.

* Fixes nightly urgent flag cleanup bug
* Checks urgent status on pledge sent so our CMs don't go nuts
* a little bit of tidying

* Fixes #1203 
* stems from #1207 

@tingaloo can I get a review on this when you get a chance?


colinxfleming:
bleh there's a stack error here. I'm going to revise some of our urgent processor code and come back to this tomorrow.

app/models/concerns/urgency.rb
```diff
@@ -4,26 +4,26 @@ module Urgency
...
+    return false if recent_history_tracks.count.zero?
     return false if pledge_sent || resolved_without_fund
     recent_history_tracks.sort.reverse.each do |history|
       return true if history.marked_urgent?
     end
     false
   end
 
+  def confirm_still_urgent
+    update urgent_flag: false unless still_urgent?
```
tingaloo: Here's the 🐛

app/models/patient.rb
```diff
@@ -28,6 +28,7 @@ class Patient
   before_save :save_identifier
 
   after_create :initialize_fulfillment
+  after_save :confirm_still_urgent
```
tingaloo: I added this and i think it works, I'm guessing it is easier to control in the patient model than the module.
```
  before_save do
    self.urgent_flag = false unless still_urgent?
    true
  end
```

app/models/concerns/urgency.rb
colinxfleming: Oh, I think this was infinitely triggering the after_save callback somehow. Adding a conditional to check urgent status seems to fix it.

colinxfleming:
@tingaloo 24432c2 should fix this up. Take a look when you have a sec?

tingaloo:
Yup, it is that. 👍

colinxfleming:
This has a bunch of legit test failures that I will now go thru.

colinxfleming:
I think #1336 will fix the test failures here, so revisit after that

colinxfleming:
There's a rerender issue here where the active partial is getting out of sync with the menu when it re-racks. This is relatively trivial in a javascript framework but kind of a pain here, and is such a specific minor bug that I'm going to elect to just not deal with it in here because I don't think it's worth the hassle.

app/views/patients/_menu.html.erb
```diff
@@ -1,22 +1,19 @@
...
-    <li class='active'><%= link_to 'Patient Information', '#patient_information', data: { toggle: 'tab' } %></li>
-    <li><%= link_to 'Abortion Information', '#abortion_information', data: { toggle: 'tab' } %></li>
-    <li><%= link_to 'Notes', '#notes', data: { toggle: 'tab' } %></li>
-    <li><%= link_to 'Change Log', '#change_log', data: { toggle: 'tab' } %></li>
-    <li><%= link_to 'Call Log', '#call_log', data: { toggle: 'tab' } %></li>
-    <% if ( current_user.admin? || current_user.data_volunteer? ) && patient.pledge_sent == true %>
-      <li><%= link_to 'Pledge Fulfillment', '#fulfillment', data: { toggle: 'tab' } %></li>
-    <% end %>
+  <ul class="nav nav-pills nav-stacked" id='menu-list'>
+    <%= render 'menu_item', link_for: 'patient_information', active: true %>
```
colinxfleming: Moved this to a partial so we can more surgically rerender this, instead of flattening it.

app/views/patients/update.js.erb
```diff
@@ -1,44 +1,45 @@
-// rerack approx gestation
-<% if @patient.previous_changes.present? %>
-  $('#patient_appointment_date')
-    .parent()
-    .find('span.help-block')
-    .text('Approx gestation at appt: <%= @patient.last_menstrual_period_at_appt %>');
+// rerack fields
+$('#patient_appointment_date')
```
colinxfleming: The callback wipes `previous_changes`, so we had to rewire this a little (a lot). Now it just rerenders everything after save regardless.

PR #1298, opened 2017-11-13
Fix #1286 and change fax provider url

### LINKED ISSUE #1286: If pledge sent, pledge_sent_at updates every time a field changes
Not sure why, but if a patient is marked `pledge_sent`, then `pledge_sent_at` and `pledge_sent_by` refreshes every single time there's an update. This is a bug in the `PatientsController#update` behavior -- see L63. Task here is to resolve that. Gold medal version is to move this to an activerecord callback (probably an `after_save` or whatever the 'while you're in this transaction, also' callback is?) if we can set the current user as part of it.

The ideal behavior is that information this is set when `pledge_sent` is, and nulled out when pledge sent is false or nil.

mdworken:

-updated fax provider url
-Fixed #1286 .


app/controllers/patients_controller.rb
```diff
@@ -58,7 +58,7 @@ def edit
   end
 
   def update
-    if @patient.update_attributes params[:pledge_sent]
+    if @patient.update_attributes params[:pledge_sent] and !pledge_sent
```
colinxfleming: Looks like this change is causing some legit test failures:

```
$ rails test test/integration/update_patient_info_test.rb

 FAIL["test_0001_should alter the information", #<Class:0x0000000006ab3140>, 23.632488988999285]
 test_0001_should alter the information#UpdatePatientInfoTest::changing patient dashboard information (23.63s)
        Expected false to be truthy.
        test/integration/update_patient_info_test.rb:44:in `block (3 levels) in <class:UpdatePatientInfoTest>'
        test/integration/update_patient_info_test.rb:41:in `block (2 levels) in <class:UpdatePatientInfoTest>'

 FAIL["test_0001_should alter the information", #<Class:0x0000000006aaeaf0>, 34.88478609399954]
 test_0001_should alter the information#UpdatePatientInfoTest::changing fulfillment information (34.89s)
        Expected false to be truthy.
        test/integration/update_patient_info_test.rb:230:in `block (3 levels) in <class:UpdatePatientInfoTest>'
        test/integration/update_patient_info_test.rb:229:in `block (2 levels) in <class:UpdatePatientInfoTest>'
```
See https://circleci.com/gh/DCAFEngineering/dcaf_case_management/3262?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link for the rest of em. 


I think we're going to want to make a slightly more fine-tuned change here. Like we chatted about it in the issue thread, probably that looks like breaking some stuff out of this endpoint and either handling it in the model or a separate, pledge oriented endpoint. That said, I'm working thru a head cold right now, so I don't want to guess on how to do that right now, haha.

PR #1310, opened 2017-11-19
Add destroy patients route for admins (fixes #1279)

### LINKED ISSUE #1279: Add destroy route for patients

Once in a blue moon, CMs create duplicate patients, and a CM Director will reach out to me to ask to delete them. This doesn't happen often enough but it's a feature they should have in the unusual set of circumstances that they need it.

Colin not being the intermediary of CM business.

* Add `destroy` route for patients
* Make it usable by admin-only, and only if the patient in question does not have a `pledge_sent`
* Add a `Delete this patient from DARIA` button _that shows up for admins only and if the patient does not have a `pledge_sent`_ in the top patient dashboard - button should have the confirm popup text of 'Only do this if this patient is a duplicate of another patient. This is not reversible and this information is gone forever. Are you absolutely sure this patient is a duplicate and that you want to do this?'
* Add tests for this behavior to controller and integration

* **Anything else?**
 nope!

rudietuesdays:

I added a button to delete a patient to the front end that only appears for admins if the patient **does not** have a pledge sent. The button triggers an alert box, and if the user confirms, the patient will be deleted and the user redirected to their dashboard.

* The button appears for admins only, and if the patient does not have a pledge sent
> screen shot 2017-11-19 at 4 26 53 pm

* An alert box pops up to confirm that the user wants to delete the patient
> screen shot 2017-11-19 at 4 27 02 pm

* If the user confirms, and the patient is deleted from the database without errors, the user is redirected back to their dashboard with a flash notice that the patient was deleted successfully
> screen shot 2017-11-19 at 4 27 17 pm

* This PR **does not** add this change into the call activity log -- I wasn't sure if that was something that should be in there or not so I didn't add it.

Fixes #1279


colinxfleming:
You're right on that this shouldn't add patients to the activity log! And the general flow looks great. Leaving a few specific comments and suggestions that I think will make this a seamless addition.

app/controllers/patients_controller.rb
```diff
@@ -89,6 +89,18 @@ def data_entry_create
     end
   end
 
+  # allow ONLY admins to delete patients, and only then if patient does not have pledge_sent
+  def destroy
+    @patient = Patient.find params[:patient_id]
```
colinxfleming: There's a `before_action` we can take advantage of here -- on L4, `before_action :find_patient, only: [:edit, :update, :download]` basically does this! Add `:destroy` to the array and it'll do the same thing, pretty much, but keep this method clean.

app/controllers/patients_controller.rb
```diff
@@ -89,6 +89,18 @@ def data_entry_create
...
+    if current_user.admin? && !@patient.pledge_sent?
+      current_user.remove_patient(@patient)
```
colinxfleming: Rather than using `remove_patient`, let's use `@patient.destroy` so we can get this out of db. (Noted below, but `remove_patient` is totally an example of us naming something badly a year and a half ago, sorry about that! Def my fault)

app/controllers/patients_controller.rb
```diff
@@ -89,6 +89,18 @@ def data_entry_create
...
+      redirect_to dashboard_path
+      flash[:notice] = "Patient successfully deleted"
+    else
+      flash[:alert] = "Errors prevented this patient from being deleted: #{patient.errors.full_messages.to_sentence}"
```
colinxfleming: Two things here:

a) The `patient.errors` is more for model level errors, whereas this is more on the order of 'you can't do that for some reason'. I'd suggest changing this message to something along the lines of 'Can't delete patients with pledges; please correct the patient record and try again.'

b) Add a `render :edit` after this, so it re-shows the page with the flash message (and so it tells future us what we should expect to happen).

app/views/patients/_patient_dashboard.html.erb
```diff
@@ -1,3 +1,12 @@
+<div class="row">
```
colinxfleming: Can you stick this below the `Appointment date` input? Probably you can take out L2 and L8, and just copy-paste the rest of it onto L28 (so it's a new row below the appointment date div. We have a little bit of real estate to use over there.

app/views/patients/_patient_dashboard.html.erb
```diff
@@ -1,3 +1,12 @@
...
+  <div class="col-sm-4">
+    <% if current_user.admin? %>
+      <% if not patient.pledge_sent %>
+        <%= link_to delete_patient_button, delete_patient_path(patient),  class:'delete-patient-button', method: :delete, data: {confirm: "Are you sure you want to delete #{patient.name}?"} %>
```
colinxfleming: Might want to check out `button_to`, if memory serves -- might do both the `link_to` and `content_tag` in one.

Since we want to take advantage of the REST route we get from `resources`, go ahead and use `patient_path(patient)` rather than `delete_patient_path(patient)`.

lwaldsc:
Yo this is an awesome feature. Thanks, @rudietuesdays!

rudietuesdays:
Thanks for the great feedback @colinxfleming! I'll tackle these this weekend.

colinxfleming:
@rudietuesdays could you do me a solid and reopen this? Definitely still want it!

app/helpers/pledges_helper.rb
```diff
@@ -15,4 +15,12 @@ def cancel_pledge_button
       'Cancel pledge'
     end
   end
+
+  def delete_patient_button
+    content_tag :span, class: 'btn btn-primary btn-lg cancel-btn btn-block',
```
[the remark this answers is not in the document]
rudietuesdays: I'm not sure I follow the logic on updating the view, but I'd definitely like to update it with some guidance! Otherwise, moving relevant method to the `patients_helper`

app/helpers/pledges_helper.rb
```diff
@@ -15,4 +15,12 @@ def cancel_pledge_button
...
+                       aria: { hidden: true },
```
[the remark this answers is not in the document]
rudietuesdays: Not sure either haha, removing

rudietuesdays:
Reopening! Sorry I somehow missed your comments back in November and haven't updated since then >.<

PR #1317, opened 2017-11-23
Budget bar - preliminary

### LINKED ISSUE #692: SUNDAY Create functioning budget bar
Starting issue to develop budget bar

Weekly budgets:
DC: $1345
Virginia: $1345
Maryland: $1615

@colinxfleming @mebates @eheintzelman @NerdyGirl537 @vtmeghan

### LINKED ISSUE #759: BUDGET BAR: Budget bar styling
Placeholder to remind myself to finish budget bar styling

colinxfleming:

Hoo man do I ever have questions about this, but here's a very dirty implementation that reads live from DB without slowing down load time. TK figuring out the actual business logic for what goes where and figuring out whether or not we should add configs or not.

* Budget bar
* Budget bar
* Budget bar

Super dirty, but here's a proof of concept:

screen shot 2017-11-23 at 2 57 32 am

I'm going to take up my litany of questions in #692 .

* Fixes #692


colinxfleming:
After talking to @lwaldsc I walked away with the sense that the minimally useful thing here would be a one-week scope (covering just pledges for appointments within a given week), and then we can figure out the more precise business logic from there.

colinxfleming:
next up!

colinxfleming:
Here's what we got so far:

screen shot 2018-01-15 at 1 30 25 am

We have a patient method that returns a hash with the goods in it, and we can further reduce that.

Because it's passing along some additional data, we can spit out some extra information, like links to patient records. Kinda a mini-dashboard thing maybe. I'll see what I can spin up in a half hour.

To implement this, we're adding a `fund_pledged_at` field on patient, a privately set field which we can use to filter on. We're going to need to get a logic check on that from the CM team.

As a reminder, we're trying to get it to end up like this:

image

TODOs:

* Fix the colors
* Better dimension calculations
* Get overflow working
* Add a popover or slide (probably on the calculations) with links to individual patient records
* Tests!

app/models/patient.rb
```diff
@@ -132,19 +134,34 @@ class Patient
...
       else
-        outstanding_pledges += (patient.fund_pledge || 0)
+        summary[:pledged] << patient
       end
+      summary
     end
-    { pledged: outstanding_pledges, sent: sent_total }
+  end
+
+  def to_simplified_patient
```
colinxfleming: Note to self to shave out any unnecessary data.

app/views/dashboards/_budget_bar.html.erb
```diff
@@ -1,14 +1,29 @@
-<div id="overview" class="margin-bottom">
-  <h1 class="border-bottom title">Week of <%= week_range %></h1>
+<div class="col-sm-12" id='budget-bar'>
+  <div class="row">
+    <div class="progress" id='budget-progress-bar'>
```
colinxfleming: May be worth trying the aria-value stuff as detailed here: https://getbootstrap.com/docs/3.3/components/#progress-alternatives Way simpler than a bunch of obscured helper methods.

colinxfleming:
TK:

* tests
* budget bar is pledges MADE this week or pledges for appointments this week? I believe made, so assume that
* in a separate PR, config
* quick acid test with @NerdyGirl537 and @lwaldsc for a feedback round

colinxfleming:
also todo, colors, separate out $ remaining block

colinxfleming:
skipping colors and config for purposes of this PR. We can ticket those later.

app/helpers/budget_bar_helper.rb
```diff
@@ -5,15 +5,30 @@ def progress_bar_color(type)
     "progress-bar-#{color}"
   end
 
-  def progress_bar_width(value)
-    "width: #{to_pct(value)}%"
+  def progress_bar_width(value, total = 1_000)
```
lomky: Is the default `total` here meant to match `default_cash_ceiling` in the dashboard_controller (and will thus later be linked), or is this coincidence? If the former, I'd tag a comment here so we don't miss it later.

app/helpers/budget_bar_helper.rb
```diff
@@ -5,15 +5,30 @@ def progress_bar_color(type)
...
+
+  def budget_bar_expenditure_content(patient_hash)
+    link = link_to patient_hash[:name], edit_patient_path(patient_hash[:id])
+    appt_text = patient_hash[:appointment_date] ?
+      "appt on #{patient_hash[:appointment_date]&.display_date}" :
+      'no appt date'
+    safe_join([link, appt_text], ' - ')
+  end
+
+  def budget_bar_remaining(expenditures, limit)
```
lomky: This is some dense code in this here function! I feel like I'm back in Perl!

Some clarification comments on the logic behind this would be welcome, as I don't know how to check this does what is intended.

app/helpers/budget_bar_helper.rb
```diff
@@ -5,15 +5,30 @@ def progress_bar_color(type)
...
+    pledged_cash = expenditures[:pledged].map { |h| h[:fund_pledge] }.inject(:+) || 0
+    sent_cash = expenditures[:sent].map { |h| h[:fund_pledge] }.inject(:+) || 0
+    limit - pledged_cash - sent_cash
```
lomky: The ability to return a negative here is intentional, yes?

app/helpers/budget_bar_helper.rb
```diff
@@ -5,15 +5,30 @@ def progress_bar_color(type)
...
   end
 
   private
 
-  def to_pct(value)
-    budget_total = ENV['weekly_budget'] || 1_350
-    pct = (value.to_f / budget_total.to_f) * 100
+  def to_pct(value, total = 1_000)
```
lomky: Same question re:`default_cash_ceiling`

app/models/patient.rb
```diff
@@ -135,19 +137,25 @@ class Patient
...
+    start_of_week = Time.zone.today.beginning_of_week(:monday)
+    plucked_attrs = [:fund_pledge, :pledge_sent, :id, :name, :appointment_date, :fund_pledged_at]
+
+    patients = Patient.in(line: lines)
+                      .where(:fund_pledge.nin => [0, nil, ''])
+                      .where(:fund_pledged_at.gte => start_of_week)
+                      .pluck(*plucked_attrs)
+                      .map { |att| plucked_attrs.zip(att).to_h }
+
+    patients.each_with_object(sent: [], pledged: []) do |patient, summary|
```
lomky: todo self: lookup this function

test/controllers/dashboards_controller_test.rb
```diff
@@ -29,4 +29,20 @@ class DashboardsControllerTest < ActionDispatch::IntegrationTest
       end
     end
   end
+
```
lomky: More test to come? These don't distinguish the cases afaik.

test/helpers/budget_bar_helper_test.rb
```diff
@@ -3,14 +3,49 @@
...
+      assert_match @patient_hash[:name], content
+      assert_match 'no appt date', content
+    end
+  end
+
+  describe 'budget bar remaining' do
+    before { @expenditures = Patient.pledged_status_summary }
+
+    it 'should return an int' do
+      assert_equal 1000,
```
lomky: Same comment re:`default_cash_ceiling`

app/models/patient.rb
```diff
@@ -88,6 +89,7 @@ class Patient
   field :patient_contribution, type: Integer
   field :naf_pledge, type: Integer
   field :fund_pledge, type: Integer
+  field :fund_pledged_at, type: Time
```
lomky: Should this field follow onto archived patients?

app/helpers/budget_bar_helper.rb
colinxfleming: This is a fabulous catch, they're intended to match. I'll change these vars to match.

app/helpers/budget_bar_helper.rb
colinxfleming: Yep, can confirm that this is intentional. Going over budget is definitely a thing!

app/helpers/budget_bar_helper.rb
colinxfleming: haha, sorry, definitely my fault. I've added a few comments that I hope are helpful and adjusted the filter a little bit.

app/models/patient.rb
colinxfleming: It's basically `reduce`. I have no idea why they can't just be good with reduce.

app/helpers/budget_bar_helper.rb
colinxfleming: thumbs up emoji

app/models/patient.rb
colinxfleming: Not sure! My inclination is no because it's more of an app logic oriented field that we kind of obscure from people (a  la created_at/updated_at) but @mdworken @alisonnjones  let me know if we should take a diff tack.

test/helpers/budget_bar_helper_test.rb
colinxfleming: I think this one we can keep, since we have a todo to make this a settable config. But can talk further about how to handle for sure if you're differently inclined!

test/controllers/dashboards_controller_test.rb
colinxfleming: lol busted, thank you for catching this. I'll add fails for now and flesh out the test cases later.

colinxfleming:
The expenditures calculator was really tough to test as part of the controller, so I broke it into a concern. Kinda small but way easier to test this way.

colinxfleming:
HOOOOOOOOOOONK

PR #1329, opened 2017-12-10
Bugfix for pledge_sent_* updating at inappropriate times -- 1286 [WIP]

mdworken:

Fixes issues in #1286 

Adds logic in controller to prevent pledge_sent_by and pledge_sent_at from updating when they shouldn't.


app/controllers/patients_controller.rb
```diff
@@ -58,9 +58,13 @@ def edit
   end
 
   def update
-    if @patient.update_attributes params[:pledge_sent]
-       @patient.pledge_sent_at = Time.zone.now
-       @patient.pledge_sent_by = current_user
+    @patient.update_attributes params[:pledge_sent]
```
colinxfleming: I think this additional logic fixes the bug at hand, nicely done!

A quick rails-ism: `@patient` is stored independently of the patient in memory, so after running the update statement (which returns `true` or `false` based on whether the update transaction completes or not), we'll have a disagreement between the patient object we're working with and the actual representation in db. We can resolve this disagreement with a `@patient.reload`.

tldr: after this `update_attributes` , throw in `@patient.reload` and then this should be good to merge. (This should also fix your test failure, if I'm reading it right.)

Let me know if that makes sense! Thanks for knocking this out, really appreciate it.

mdworken: Great, thanks for the help! I'm headed to a holiday party right now, I'll fix it early tomorrow morning :)

PR #1335, opened 2017-12-17
Fix #1286 (this time in the model)

### LINKED ISSUE #1286: If pledge sent, pledge_sent_at updates every time a field changes

mdworken:

Corrects logic to fix bug 1286.

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -58,9 +58,13 @@ class PatientsController < ApplicationController
   end
 
   def update
-    if @patient.update_attributes params[:pledge_sent]
-       @patient.pledge_sent_at = Time.zone.now
-       @patient.pledge_sent_by = current_user
+    @patient.update_attributes params[:pledge_sent]
+    if @patient.pledge_sent and !@patient.pledge_sent_at
+      @patient.pledge_sent_at = Time.zone.now
+      @patient.pledge_sent_by = current_user
+    elsif !@patient.pledge_sent
+      @patient.pledge_sent_at = nil
+      @patient.pledge_sent_by = nil
     end
     if @patient.update_attributes patient_params
       @patient.reload
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -59,6 +59,7 @@ class PatientsController < ApplicationController
 
   def update
     @patient.update_attributes params[:pledge_sent]
+    @patient.reload
     if @patient.pledge_sent and !@patient.pledge_sent_at
       @patient.pledge_sent_at = Time.zone.now
       @patient.pledge_sent_by = current_user
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -58,15 +58,7 @@ class PatientsController < ApplicationController
   end
 
   def update
-    @patient.update_attributes params[:pledge_sent]
-    @patient.reload
-    if @patient.pledge_sent and !@patient.pledge_sent_at
-      @patient.pledge_sent_at = Time.zone.now
-      @patient.pledge_sent_by = current_user
-    elsif !@patient.pledge_sent
-      @patient.pledge_sent_at = nil
-      @patient.pledge_sent_by = nil
-    end
+    @patient.last_edited_by = current_user
     if @patient.update_attributes patient_params
       @patient.reload
       flash.now[:notice] = "Patient info successfully saved at #{Time.zone.now.display_timestamp}"
@@ -124,7 +116,7 @@ class PatientsController < ApplicationController
                   :procedure_cost, :check_number, :date_of_check]
   ].freeze
 
-  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent].freeze
+  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent, :pledge_sent_by, :pledge_sent_at, :last_edited_by].freeze
 
   def patient_params
     params.require(:patient).permit(
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -116,7 +116,7 @@ class PatientsController < ApplicationController
                   :procedure_cost, :check_number, :date_of_check]
   ].freeze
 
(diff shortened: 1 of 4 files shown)


app/controllers/patients_controller.rb
```diff
@@ -119,7 +116,7 @@ def find_patient
                   :procedure_cost, :check_number, :date_of_check]
   ].freeze
 
-  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent].freeze
+  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent, :pledge_sent_by, :pledge_sent_at, :last_edited_by].freeze
```
colinxfleming: hmm, I don't think you need the `pledge_sent_by`/`pledge_sent_at` here -- the `patient_params` deals explicitly with the payload sent to the controller for mass assignment (basically saying 'it's okay to change all of these fields at once, but only these fields'), and since I don't think we're setting them from the form, we can probably skip changing this set. Does that sound right to you?

app/models/patient.rb
```diff
@@ -179,4 +182,14 @@ def clean_fields
   def initialize_fulfillment
     build_fulfillment(created_by_id: created_by_id).save
   end
+  
+  def update_sent_by_sent_at
+    if self.pledge_sent  && !self.pledge_sent_by
```
colinxfleming: This looks good! The following feedback is totally in the domain of style / rails-ifying things, so implement what makes sense to you:

* I don't believe we need the `self`s here -- the object is smart enough to know that it's referring to its own attributes, pleasantly enough. You can just write `pledge_sent` and it'll work like a charm, I do believe.
* One of the sickest features of rails is that it throws a ton of convenience methods which help make your programming look a lot more like readable english -- rather than do `!pledge_sent_by`, or `!pledge_sent` we can do `pledge_sent_by.blank?` to the same effect. [Docs here](http://api.rubyonrails.org/v4.2/classes/Object.html#method-i-blank-3F).

app/models/patient.rb
```diff
@@ -38,7 +38,8 @@ class Patient
   embeds_many :notes
   belongs_to :pledge_generated_by, class_name: 'User', inverse_of: nil
   belongs_to :pledge_sent_by, class_name: 'User', inverse_of: nil
-
+  belongs_to :last_edited_by, class_name: 'User', inverse_of: nil
```
colinxfleming: I was initially on the fence about this but the more I think about it the more I like it. Good call on this! It's a good workaround to the problem of not having `current_user` in a model, it updates the stuff in its own db transaction, and we get a fringe benefit of having this in one convenient place if we need it. (We store something similar now but it's buried in the audit history, and wouldn't be available at the time the before_update callback fired.)

app/models/patient.rb
```diff
@@ -26,7 +26,7 @@ class Patient
 
   before_validation :clean_fields
   before_save :save_identifier
-
+  before_update :update_sent_by_sent_at
```
colinxfleming: I think this name is close but doesn't quite capture the important thing, which is the pledge! Maybe `update_pledge_sent_by_sent_at`?

app/models/patient.rb
```diff
@@ -160,6 +161,8 @@ def event_params
     }
   end
 
+
```
colinxfleming: Kill this whitespace please!

app/models/patient.rb
```diff
@@ -179,4 +182,14 @@ def clean_fields
   def initialize_fulfillment
     build_fulfillment(created_by_id: created_by_id).save
   end
+  
+  def update_sent_by_sent_at
+    if self.pledge_sent  && !self.pledge_sent_by
+      self.pledge_sent_at = Time.zone.now
+      self.pledge_sent_by = last_edited_by
+    elsif !(self.pledge_sent)#       self.pledge_sent_at = nil
```
colinxfleming: I assume we can ditch this comment, yeah?

app/controllers/patients_controller.rb
mdworken: Yes, I think you're right!

app/models/patient.rb
mdworken: Sounds good, will update it

app/models/patient.rb
mdworken: Oops, yeah, that was in there from when I was debugging, haha

mdworken:
Hi Colin!
Regarding the test failure, I would definitely expect it to have failed as the setting of pledge _sent_by and pledge_sent_at is no longer in the controller. I'll see if that change fixes it.

Re: model test, I haven't actually written model tests before, but I'll give it a crack late today after I merge in the other requested changes.

app/models/patient.rb
mdworken: Upon testing, we don't need the self in the if and elsif statements, but we do when setting the variables. Something something scoping something?

mdworken:
I sent slack messages, but the test that was previously passing seems to have had some issues. 

I suspect what was happening was:
1) the controller was updating the pledge_sent_at for the @patient, which led to the assert_kind_of Time to return true
2) the update failed validation for the database because it was missing clinic and fund pledge
3) the test that update works did NOT include a payload of pledge_sent, so no one noticed the database failing.

app/controllers/patients_controller.rb
```diff
@@ -119,7 +116,7 @@ def find_patient
                   :procedure_cost, :check_number, :date_of_check]
   ].freeze
 
-  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent].freeze
+  OTHER_PARAMS = [:urgent_flag, :initial_call_date, :pledge_sent, :last_edited_by].freeze
```
colinxfleming: `last_edited_by` we draw from the `current_user` variable in the controller, not the payload params, so we can revert this change.

mdworken: Good point!

PR #1336, opened 2017-12-17
Remove all btn

### LINKED ISSUE #1202: Call List 'remove all' button for dashboard

colinxfleming:

A couple additional things from @tingaloo 's work in #1274 --

* renames existing `clear_call_list` method to be clearer about intent, and adds a `clear_call_list` method that does what this is supposed to
* removes need for associated user id in params, running off current user... OR DOES IT??? (test failure)

UserControllerTest is currently choking, so that sucks. Not sure why and am not in a mind to fix it tonight. 

* Adds remove all patients from your call list btn
* some other assorted cleanup

TK integration test, model test.

image

* Fixes #1202

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -44,7 +44,6 @@ class User
   enumerize :role, in: [:cm, :admin, :data_volunteer], predicates: true
   field :call_order, type: Array
 
-
   ## Database authenticatable
   field :email,              type: String, default: ''
   field :encrypted_password, type: String, default: ''
@@ -143,7 +142,7 @@ class User
     ordered_patients
   end
 
-  def clear_call_list
+  def clean_call_list_between_shifts
     if last_sign_in_at.present? &&
        last_sign_in_at < Time.zone.now - TIME_BEFORE_INACTIVE
       patients.clear
@@ -153,6 +152,10 @@ class User
         patients.delete(p) if recently_reached_by_user?(p)
       end
     end
+  end
+
+  def clear_call_list
+    patients.clear
   end
 
   def admin?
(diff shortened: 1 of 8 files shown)


tingaloo:
Man you really want that associated user id gone lol

colinxfleming:
@tingaloo haha sure do. I'd like to get it out of the add patient to call list stuff too but that's a more involved change that touches a lot of stuff, so maybe later.

I think this is good to go, would you mind reviewing if you get a chance? It's pretty much everything you did with the slight routing change, but would love a second set of eyes.

app/models/user.rb
```diff
@@ -143,7 +142,7 @@ def ordered_patients(lines = LINES)
     ordered_patients
   end
 
-  def clear_call_list
+  def clean_call_list_between_shifts
```
tingaloo: I'm trying to understand the difference between ```#clear_call_list``` and ```#clean_call_list_between_shifts```.

The only time ```#clean_call_list_between_shifts``` is called is during the test cases is because there is additional logic dealing with last sign in?

colinxfleming:
Yep, clear call list is the frontend action, and clean call list between
shifts empties things out when someone hasn't logged in in a bit.

I couldn't really come up with better names on the spot, haha -- do you
think there's a better option here?

colinxfleming:
@tingaloo yeah, 'nuke an associated set' doesn't really have a REST verb associated with it, so this kinda thing is a little tough! Thanks for your eyes on this.

PR #1339, opened 2018-01-01
Bump development dependencies

colinxfleming:

Has been a minute since we updated dependencies, so we're missing out on a lot of good stuff (particular to bundle-audit, brakeman, and ruby-audit). This runs a bundle update on development and test requirements.

* Bumps development gems

No view changes.

No issue #s.


colinxfleming:
Whatever the point release in `minitest-ci` was, it busted up something. Removing it gets the build back to green, and we weren't really using its reporting functionality anymore anyway, so one less dep.

colinxfleming:
Merging since this doesn't touch prod code.

PR #1345, opened 2018-01-07
Allow admins to change roles ctd

### LINKED ISSUE #1271: Not able to upgrade users

Bump new users in BRAAF to admin. Currently new users are created as CM and a change to any other class of user doesn't save.

Fund's ability to manage users.

Make it so any admin can bump users to another class.

* **Anything else?**
Our engineering team rocks!

colinxfleming:

Continues #1272 - Made a quick change to allow roles to be permitted on update.

* Whitelists user role so an admin can change it
* Adds a safety so admins cannot demote themselves
* Overhauls the users controller test

* Fixes #1271

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,6 +29,10 @@ class ApplicationController < ActionController::Base
     redirect_to root_url unless current_user.admin?
   end
 
+  def confirm_admin_user_async
+    head :unauthorized unless current_user.admin?
+  end
+
   def redirect_unless_has_data_access
     redirect_to root_url unless current_user.allowed_data_access?
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,9 +1,14 @@
 # Additional user methods in parallel with Devise -- all pertaining to call list
 class UsersController < ApplicationController
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
-  before_action :confirm_admin_user, only: [:new, :index, :update]
+  before_action :confirm_admin_user, only: [:new, :index, :update,
+                                            :change_role_to_admin,
+                                            :change_role_to_data_volunteer,
+                                            :change_role_to_cm]
   before_action :confirm_admin_user_async, only: [:search]
-  before_action :find_user, only: [:update, :edit, :destroy, :reset_password]
+  before_action :find_user, only: [:update, :edit, :change_role_to_admin,
+                                   :change_role_to_data_volunteer,
+                                   :change_role_to_cm]
 
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :not_found }
   rescue_from Exceptions::UnauthorizedError, with: -> { head :unauthorized }
@@ -29,7 +34,7 @@ class UsersController < ApplicationController
   end
 
   def update
-    if user_not_demoting_themself(@user, user_params) && @user.update_attributes(user_params)
+    if @user.update_attributes(user_params)
       flash[:notice] = 'Successfully updated user details'
       redirect_to users_path
     else
@@ -50,6 +55,21 @@ class UsersController < ApplicationController
     else
       render 'new'
     end
+  end
+
+  def change_role_to_admin
+    @user.update role: 'admin'
+    render 'edit'
+  end
+
+  def change_role_to_data_volunteer
+    @user.update role: 'data_volunteer' if user_not_demoting_themself(@user)
+    render 'edit'
+  end
+
+  def change_role_to_cm
+    @user.update role: 'cm' if user_not_demoting_themself(@user)
+    render 'edit'
   end
 
   # def toggle_lock
@@ -114,7 +134,7 @@ class UsersController < ApplicationController
   end
 
   def user_params
-    params.require(:user).permit(:name, :email, :role)
+    params.require(:user).permit(:name, :email)
   end
 
   def retrieve_patients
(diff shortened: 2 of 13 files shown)


colinxfleming:
I think I'm going to use this as an opportunity to overhaul the users controller tests. Right now they're sort of all over the place (my fault!) and don't have fantastic coverage.

colinxfleming:
I'm going to file a ticket to split up the call list management functionality and the admin user management functionality. The users controller is kinda heavy as it is, the call list stuff needs a refactor, and the call list stuff doesn't need to inherit a bunch of confusing junk from devise.

@tingaloo can I get a review? Sorry this is so heavy.

colinxfleming:
We have a brakeman failure on role.

```
Confidence: Medium
Category: Mass Assignment
Check: PermitAttributes
Message: Potentially dangerous key allowed for mass assignment
Code: params.require(:user).permit(:name, :email, :role)
File: app/controllers/users_controller.rb
Line: 117
```

colinxfleming:
Potential fix is making 'set role' to be a button with its own endpoint. Not restful but likely won't trigger the warning.

colinxfleming:
Hot damn did this ever go off the rails. This moves the endpoints to change roles out of update, throws them behind an auth wall, adds buttons on the user edit form, and throws in a systemtest for good measure. I think that'll squelch the brakeman warning, so after that it'll be whatever else.

colinxfleming:
We're getting incidental failures, so I think this is ready for Serious Review.

app/controllers/application_controller.rb
```diff
@@ -29,6 +29,10 @@ def confirm_admin_user
     redirect_to root_url unless current_user.admin?
   end
 
+  def confirm_admin_user_async
```
tingaloo: Not sure if ```async``` is necessary.

app/controllers/users_controller.rb
```diff
@@ -121,4 +141,14 @@ def retrieve_patients
     @patient = Patient.find params[:id]
     @urgent_patient = Patient.where(urgent_flag: true)
   end
+
+  def user_not_demoting_themself(user)
```
tingaloo: Can add ```?``` at end of method name

app/helpers/users_helper.rb
```diff
@@ -15,4 +15,11 @@ def user_lock_status(user)
       'Active'
     end
   end
+
+  def role_toggle_button(user, new_role)
```
tingaloo: This is pretty cool, does this dynamically creates the button text based on roles?

app/models/concerns/user_searchable.rb
```diff
@@ -6,13 +6,12 @@ module UserSearchable
   VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
 
   class_methods do
-
     def search(name_or_email_str)
-      name_regexp = /#{Regexp.escape(name_or_email_str)}/i
-      email_regexp = (VALID_EMAIL_REGEX).match(name_or_email_str)
+      regexp = /#{Regexp.escape(name_or_email_str)}/i
+      # email_regexp = (VALID_EMAIL_REGEX).match(name_or_email_str)
```
tingaloo: can remove this?

test/controllers/users_controller_test.rb
```diff
@@ -2,49 +2,180 @@
...
+      end
+    end
+
+    it 'should show errors if creation fails' do
+      attributes = attributes_for(:user)
+      attributes[:name] = nil
+      assert_no_difference 'User.count' do
+        post users_path, params: { user: attributes }
+      end
+      assert_includes response.body, 'can&#39;t be blank'
```
tingaloo: What an odd assertion

colinxfleming: haha. it escapes the apostrophe like a real jerk.

app/controllers/application_controller.rb
colinxfleming: We already have a `confirm_admin_user` synchronous that redirects to root, sad to say.

app/controllers/users_controller.rb
colinxfleming: Good call! 61587d11

app/helpers/users_helper.rb
colinxfleming: yep! Spits out a button that, when you click it, patches to those endpoints.

app/models/concerns/user_searchable.rb
colinxfleming: Yep! 1013abd0

test/controllers/users_controller_test.rb
```diff
@@ -2,49 +2,180 @@
...
   end
 
-  describe 'create method' do
-    it 'should raise if user is not admin' do
+  describe 'authentication for restricted endpoints' do
+    %w[users_path new_user_path].each do |endpoint|
+      it "should redirect if user is not admin - #{endpoint}" do
+        User.role.values.reject { |role| role == :admin }.each do |role|
+          @user.update role: role
+          get send(endpoint.to_sym)
```
tingaloo: I've never seen send() used before, is it being used as a rest call here?

Kevin-Wei: Send is being used here to call methods :)

colinxfleming:
There's an alarming set of flaky tests here but YOLO. Thanks @tingaloo and @Kevin-Wei !

PR #1349, opened 2018-01-11
Cancel checkbox 2

### LINKED ISSUE #1046: What should the buttons say in the cancel pledge workflow?

colinxfleming:

Picks up changes in #1227 and resolves test failures/merge conflicts, pretty much. 

This pull request makes the following changes (copied from @SamuelLangenfeld 's):
 * removes checkbox from the cancel pledge modal
* changes _pledge.html.erb to only call views/shared/multi-modal for submitting a pledge, not cancelling. Cancelling a pledge is only one step.

image

Fixes #1046

### DIFF
--- a/app/views/patients/_pledge.html.erb
+++ b/app/views/patients/_pledge.html.erb
@@ -1,58 +1,62 @@
-<%= render '/shared/multi_modal', disable_next: disable_continue?(patient), :captured => capture { %>
-
 <% if not patient.pledge_sent %>
-  <div class="row hide" data-step="1" data-title="Confirm the following information is correct:">
-    <% if patient.pledge_info_present? %>
-      <% patient.pledge_info_errors.each do |error| %>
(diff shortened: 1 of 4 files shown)


colinxfleming:
The code changes are good with me, and the stuff in here is test changes, so invoking godmode and merging.

PR #1356, opened 2018-01-15
Mkd fix change log

### LINKED ISSUE #1330: Change log does not list updated by properly
Hi all! I'll fix it myself, but the change log is not listing the name properly. Probably needs a .name at the end to make it the user's name instead of the user's user object.

wrong user

mdworken:

Removes pledge_sent_by, updated_by, and last_edited_by from changelog, as this information can be otherwise inferred.

* Fixes #1330


colinxfleming:
Thanks @mdworken !

PR #1357, opened 2018-01-15
Squelch new_framework_defaults and remaining depreciation warnings

### LINKED ISSUE #984: Clean up Rails 5.1 deprecation warnings
In moving to Rails 5.0.2 we cleaned up a lot of deprecations, but some remain.

When you run `rails test` you should see the following warnings:

```
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead. (called from <top (required)> at /code/dcaf_case_management/config/environment.rb:5)
DEPRECATION WARNING: after_filter is deprecated and will be removed in Rails 5.1. Use after_action instead. (called from <top (required)> at /code/dcaf_case_management/config/environment.rb:5)
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead. (called from <top (required)> at /code/dcaf_case_management/config/environment.rb:5)
DEPRECATION WARNING: after_filter is deprecated and will be removed in Rails 5.1. Use after_action instead. (called from <top (required)> at /code/dcaf_case_management/config/environment.rb:5)
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <top (required)> at /code/dcaf_case_management/config/environment.rb:5)
DEPRECATION WARNING: #original_exception is deprecated. Use #cause instead. (called from initialize at /.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/better_errors-2.1.1/lib/better_errors/raised_exception.rb:7)
DEPRECATION WARNING: #original_exception is deprecated. Use #cause instead. (called from initialize at /.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/better_errors-2.1.1/lib/better_errors/raised_exception.rb:8)
```

Most of these are likely in gems since a cursory grep turned up no results, so we will have to figure out which gems they are in, update them, or figure out another alternative. This is currently causing no issues, but will prevent us from moving to Rails 5.1 if the time comes.

colinxfleming:

We're using the default settings everywhere but on the new `callbacks have to throw :abort to stop` setting. We have written callbacks in such a way (they mostly have side effects, or don't need to stop if they fail) that we can cleanly toggle this over.

s/o to my dude 'rigorous test suite'.

* Removes the `new_framework_defaults.rb` bridge file initializer

No view changes

* Fixes #984


colinxfleming:
Hey @Kevin-Wei could I get a review on this when you get a minute?

DarthHater:
Looked ok to me in a quick fly by. Ran `rails test` and all gravy. 

A side note, I did #1367 because looking at the logs made my eyes strain. Total optional PR but might make this stuff a bit easier in the future (since you are basically just looking at logs).

PR #1360, opened 2018-01-15
Use DARIA_FUND env var instead of FUND

### LINKED ISSUE #1259: Merge FUND and DARIA_FUND env vars
note to self

colinxfleming:

We don't set `FUND` in app.json, so we should depreciate it. This takes out all refs to it in favor of either `DARIA_FUND` or `DARIA_FUND_FULL` env vars. This should prevent us from accidentally blowing something up by accident.

* FUND -> DARIA_FUND

* Fixes #1259


colinxfleming:
thank you @lomky !

PR #1162, opened 2017-08-01
[WIP] Patient archive

lomky:
I rule and have ~*completed*~ started some work on Case Manager that's ~*ready for review*~ getting hard to track!

Work on Patient archiving.

Process so far: 
   - [x] lots of thoughts and working out the data processes on the ticket 
   - [x]  roughed out fields for archiving
   - [x] Basic archive: given a patient X days since last contact: hook them into archiving
     - [x] basic `archived?` validator
     - [x] basic `archive` function 
     - [x] test basic `archive` process
   - [x] Actual archive: erasing
     - [x] validate sensitive, to-delete fields are empty on archived patients
     - [x] `archive` function actually erases sensitive info
     - [x] test `archive` process erases
   - [ ] Actual archive: scrambling
     - [ ] Get final answer on linking repeat patients
     - [x] validate sensitive, to-scramble fields are scrambled on archived patients
     - [x] `archive` function actually scrambles sensitive info
     - [x] test `archive` process scrambles
   - [ ] Config
     - [ ] Create config flip to turn on archive
     - [ ] Create config flip to turn on wipe (for funds not interested in analysis / super paranoid)
     - [ ] Create config for times (dropoff time, post-fulfill time)
   - [ ] Final cleanup
     - [ ] Viewable only by admins (?)
     - [ ] hook in `archive` function 
     - [ ] don't re-archive patients
     - [x] test subfunctions
   

It relates to the following issue #s: 
* *Eventually* Fixes #812

PR #1362, opened 2018-01-22
WIP: Patient archive - Take 2

lomky:
Replacement branch for #1162, work for ticket #812.

As development was progressing on 1162, I felt more and more uncomfortable with the architecture of morphing hte existing patient object upon archiving. This rebuild addresses that with a distinct archive_patient object, which after inheriting the whitelisted fields from a patient then nukes that PII laden object to space.

This also makes several other things much easier.

 WIP:
  - [ ] Model out ArchivedPatient
     - [x] Fields
     - [x] Validations
     - [ ] Fancy creation process that reads from & deletes the Patient object
  - [x]  Controller, as needed, for ArchivedPatient
     - [x] Index
     - [x] Show
     - [x] Destroy
  - [x] Basic Views
  - [x] Restrict access to ArchivePatient to Data & Admins
  - [ ] Tests!
    - [ ] Models
    - [ ] Controllers
    - [ ] System - maybe followup?
  - [x] Nightly tasks
    - [x] Archive fulfilled patients
    - [x] Archive inactive, dropoff patients

Follow On PRs:
  - [x] Cherry pick DBSeeds from previous PR - still good. Maybe in a standalone PR?
  - [ ] Configs
    - [ ] Archive switch (vs plain deletion, for more paranoid / non data focused funds)
    - [ ] Days since creation to consider inactive
    - [ ] Days since fulfillment to consider complete
  - [ ] Better views 
  - [ ] Update Exportable to handle both Patients and ArchivedPatients


test/fixtures/archived_patients.yml
```diff
@@ -0,0 +1,7 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
colinxfleming: We generally use factories (via the FactoryBot gem) rather than fixtures, so you can drop this file. Check out `test/factories/patient.rb` for an example of what you can use instead.

colinxfleming:
@lomky this is a great model implementation! Really feeling how clean it is, I think this proves your point pretty hard.

I'm a little unclear on the utility of the controller -- my guess would be that we could do a lot of this in models, but you've spent a lot more time thinking about this than I have at this point, haha. What's yr game plan there?

Also really appreciate the roadmap! I think we can punt on the config stuff for the time being and just use a single constant # for the purposes of this PR.

All told this is looking like a really slamming path and I'm stoked about it! Also, again, deeply appreciative of your continuing to push on this 💯

lomky:
@colinxfleming  a lot of files are just what pops out of the `rails generate scaffold ...` command, so a lot of the files will be cast off as it moves forward, especially in the areas where we peel away from base rails.

My larger idea is to have this look a lot like the patient in terms of model and routes, as eventually they will both feed into reports. 

Re:controller gameplan  
 - Index I just made match Patient, figuring it probably feeds into reporting, but I haven't actually gone digging on that yet.
 - I don't see a need for any update paths
 - I like the idea of leaving destroy as an option - though probably restricting it to admins only.
 - Create is just there in case that path is useful, it may drop off as well if I move the logic all to the model. I'm mostly mulling over how to pass in the info. Just passing in a patient ID and let the archive_patient model do the heavy lifting is my current thought, and I wasn't clear at the point of this commit where the patient id will blow out into the fields, so I just left it alive as a controller call, for now.

I will focus on getting this to minimum viable product, to go full business jargon :business_suit_levitating: and then file tickets for parts moving forward re: configs and reports. Better to free up work!

test/fixtures/archived_patients.yml
lomky: Thanks! I haven't even looked at the test files yet, but I'll be pinging you when I do, since I haven't built one from scratch with FactoryBot before!

lomky:
Super basic views in. Controllers paired down to bare minimum needed. Model can now duplicate most of the needed patient fields into itself.

Considering what meta-info might be useful. Would it be helpful to have a field with which archive method caught the patient, i.e. whether they were archived after a successful pledge fulfillment vs dropoff?

To do next time - copying over the sub-objects. Cleaning the fulfillment. Tests

app/models/archived_patient.rb
```diff
@@ -0,0 +1,132 @@
...
+  # History and auditing
+  track_history on: fields.keys + [:updated_by_id],
+                version_field: :version,
+                track_create: true,
+                track_update: true,
+                track_destroy: true
+  mongoid_userstamp user_model: 'User'
+
+  # Archive & Delete patients who have fallen out of contact
+  def self.process_dropped_off_patients
```
colinxfleming: Since 'process' means archive here, thoughts?

Also, a ruby convention is that for methods considered 'dangerous' or irreversible, you slap a `!` on the end of em. Since we're considering archiving to be an irreversible process, let's pop a bang on the end of these class methods.

app/views/archived_patients/show.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.partial! "archived_patients/archived_patient", archived_patient: @archived_patient
```
colinxfleming: no json endpoint, so can ditch this too!

config/routes.rb
```diff
@@ -52,6 +52,7 @@
     resources :clinics, only: [:index, :create, :update, :new, :destroy, :edit]
     resources :configs, only: [:index, :create, :update]
     resources :events, only: [:index]
+    resources :archived_patients, only: [:index, :show, :destroy]
```
colinxfleming: Flagging an inconsistency: no `new` route here, but there's views for a `new`

lib/tasks/nightly_cleanup.rake
```diff
@@ -8,4 +8,11 @@ task nightly_cleanup: :environment do
 
   Event.destroy_old_events
   puts "#{Time.now} -- destroyed old events"
+
+  ArchivedPatient.process_dropped_off_patients
+  puts "#{Time.now} -- archived dropped off patients"
+
+  ArchivedPatient.process_fulfilled_patients
```
colinxfleming: 🦅

test/models/archived_patient_test.rb
```diff
@@ -0,0 +1,86 @@
+require 'test_helper'
+
+class ArchivedPatientTest < ActiveSupport::TestCase
```
colinxfleming: 😻

app/controllers/archived_patients_controller.rb
```diff
@@ -0,0 +1,69 @@
+class ArchivedPatientsController < ApplicationController
+  before_action :redirect_unless_has_data_access, only: [:index, :show, :destroy]
+  before_action :set_archived_patient, only: [:show, :destroy]
+
+  rescue_from Mongoid::Errors::DocumentNotFound,
+              with: -> { redirect_to root_path }
+
+
+  # GET /archived_patients
+  def index
```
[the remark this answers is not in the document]
lomky: if we're not going to do reports on archived patients, why keep them in the first place?

lomky:
My initial and preferred thought for the sub objects was to morph the call, fulfillment, and external pledges into [polymorphic](https://docs.mongodb.com/mongoid/master/tutorials/mongoid-relations/#polymorphism) `embedded_in`s instead of directly `embedded_in` `patient`. 

However, it looks like that would require a lot of tweaking throughout the codebase, because these objects could previously take for granted their parent object was a parent. 

I have slept on it and I think I'll pursue polymorphism a bit further, as it is ultimately much more elegent than the alternative, which is creating new objects for each time, i.e. `archived_call`, `archived_fulfillment`, `archived_external_pledge`. That will work, but just feels a bit ugly and redundant, as those objects won't actually change when archived.

Thanks for the great lookover, @colinxfleming!

app/controllers/archived_patients_controller.rb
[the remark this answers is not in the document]
colinxfleming: haha fair point. Let me rephrase -- should we just universally roll up archived patients into the current reports endpoint and have that spit out archived and current patients both, or should we keep the current reports endpoint as just active patients and then have this one be 'everyone ever, including archived patients'?

lomky:
Polymorphism is complete as of #1391! Moving forward on this now.

lomky:
I know of no way to revert this wrong way merge that plays nice with github :( Gunna close this PR and make a new branch.

PR #1366, opened 2018-01-28
Remove dropdown row UI

### LINKED ISSUE #1353: Viewing clinic details is slow

If you go to the clinic management page and click the expand chevron, it has a clunky feel. We should speed that up and put in a jquery `slideDown('fast')`.

Being able to easily see clinic details in the management list.

a) prerender, hide the dom element, and change it to a jquery slidedown or something
b) fix the test in `test/system/clinic_management_test.rb`

* **Anything else?**
nope!

geoffreyyip:

Clinics view used a dropdown row to show/hide additional info about a
given clinic. This UI was buggy. The dropdown row was implemented
through Bootstrap toggle and defined on a `<tr>` tag.

Bootstrap's toggle does not play well with the HTML `<table>` tag. A
workaround is possible but difficult. Instead, this commit takes the
three extra fields from the dropdown row and adds them as additional
columns within the table.

Before this commit, there were 5 columns of data and a chevron toggle
that would reveal 3 more columns of data in a dropdown row. Now
there is no chevron toggle, and the table displays 8 columns of data at
all times.

Old screen:

screenshot

New screen: 
screenshot-2018-1-28 daria

* Fixes #1353

### DIFF
--- a/test/system/clinic_management_test.rb
+++ b/test/system/clinic_management_test.rb
@@ -19,14 +19,6 @@ class ClinicManagementTest < ApplicationSystemTestCase
       within :css, '#clinics_table' do
         assert has_text? @clinic.name
       end
-
-      # TODO revisit this after viewing clinic details is less clunky
-      # find("#clinic-toggle-#{@clinic.id}").click
-      # wait_for_ajax
-      # within :css, "clinic-#{@clinic.id}-detail" do
-      #   assert has_text? 'Accepts Medicaid?'
-      #   assert has_text? 'Yes'
-      # end
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/system/clinic_management_test.rb
```diff
@@ -19,14 +19,6 @@ class ClinicManagementTest < ApplicationSystemTestCase
       within :css, '#clinics_table' do
         assert has_text? @clinic.name
       end
-
-      # TODO revisit this after viewing clinic details is less clunky
```
colinxfleming: 🔥

PR #1373, opened 2018-01-30
Rip out Gemnasium

DarthHater:

Removes Gemnasium:

https://gemnasium.com/blog/gemnasium-is-acquired-by-gitlab/

* Removes badge
* Removes docs related to gemnasium


colinxfleming:
RIP Gemnasium

Is there something we can sub in here? Maybe the in house github option or something? We get a layer of safety from `bundle-audit` and `ruby_audit` but it's a nice early warning system generally.

DarthHater:
GitHub just shows up in the ui now if you go into settings or whatever.
Kinda the same deal as gemnasium but since only contributors have access I
wasn’t going to add it to the readme.

colinxfleming:
If it's cool with you I'm going to keep this trucking for a little while because gemnasium seems to have a yellow status for when something isn't alerting, but is still out of date, which github doesn't have. Also I'm sad we never got to hit the auto update :(

colinxfleming:
D:

PR #1374, opened 2018-01-30
Reorganize gemfile

colinxfleming:

Been meaning to do this for a little while -- this reorgs our gemfile so that for the less common libraries, it's clear what's doing what.

I'm going to see if we can take out some of the stuff in `????` sections before I merge, but keeping these here for now.

* reorganizes gemfile, adds mad comments

No view changes.

No issues.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,77 +1,104 @@
 source 'https://rubygems.org'
 ruby '2.4.3'
 
+# Standard rails
 gem 'rails', '>= 5.1'
+gem 'puma' # roar
+gem 'turbolinks', '~> 5.0.0'
+gem 'jbuilder', '~> 2.0'
+gem 'sdoc', '~> 0.4.0', group: :doc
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,77 +1,102 @@
 source 'https://rubygems.org'
 ruby '2.4.3'
 
+# Standard rails
 gem 'rails', '>= 5.1'
+gem 'puma' # roar
```
lomky: Just noting that puma moved from prod only to all. If that change was intentional, then all looks GTG!

colinxfleming: Yep, this is intentional! There was an effort awhile back to test in an environment as close to actual production as possible in rails, and part of that is 'your prod server gem should be your dev server gem too', so now we're using puma in dev instead of webbrick.

colinxfleming:
thanks @lomky !

PR #1375, opened 2018-02-02
Adding in build command for docker-compose

DarthHater:

Adds in build command for `docker-compose`

I noticed we were missing an explicit command for `docker-compose build` so I'm adding this in for people who use `docker-compose` such as @geoffreyyip 

* Small change to `docker-compose`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,6 @@
 web:
   image: colinxfleming/dcaf_case_management
+  build: .
   command: rails s -p 3000 -b 0.0.0.0
   volumes:
     - .:/usr/src/app


docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 web:
   image: colinxfleming/dcaf_case_management
+  build: .
```
lomky: I assert this is indeed a `.` :tada:

DarthHater:
In testing this my local `docker-compose` didn't like it, so I played around a bit with #1376 and superseded this PR, I think.

colinxfleming:
Looks like this is overtaken by #1376, so I'll let this hang out until that's merged.

Thanks @lomky and @geoffreyyip for eyeballing this and poking at it with @DarthHater !

DarthHater:
Closed in favor of #1376

PR #1376, opened 2018-02-02
Docker Compose Upgrades

DarthHater:

This makes some changes that I think will make the development process smoother for people:
* On `docker-compose up` it will run an `entrypoint.sh` file that checks
  * `bundle check` and if it says yep, all gems are here, it runs `rails`
  * If it doesn't, it runs `bundle install` and then runs `rails` after the gems are installed

End result is no one should need to run `bundle install` anymore barring some disaster.  

* Adds `entrypoint.sh` and sets it as the entrypoint for the `Dockerfile`
* Removes command from `docker-compose.yml` since now the entrypoint dictates how the image will run
* Upgrades the `docker-compose.yml` file to a newer spec, 3.1 in this case

Pseudo related to #1375 and sort of came up with this thinking about the fun @geoffreyyip has been having with Docker overall.


docker-compose.yml
```diff
@@ -1,17 +1,19 @@
...
-    - db
-  environment:
-    mongohost: db
-  env_file:
-    - web-variables.env
-db:
-  image: mongo
-  ports:
-    - "27017:27017"
+version: '3.1'
```
colinxfleming: 🔥 🔥 🔥

docker-compose.yml
```diff
@@ -1,17 +1,19 @@
...
+services:
+  web:
+    build: .
```
colinxfleming: What does having both the `image` and `build` do? Normally `docker-compose build` refs the `:latest` image and then `docker-compose up` pulls the layers. (My own finger muscle memory when using docker tends to be `git pull origin master && docker-compose build && docker-compose up` or some form of that.) It looks like it falls back on a `build`, but wanted to make sure that's what's supposed to happen.

entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
colinxfleming: I think I understand why this comes up (there was a tough-to-diagnose mismatch between the docker image and a local repo). Depreciating our remote build makes sense to me, given the big hassle it can be when we update gems (and as we continue to move toward maintenance, we're going to be doing a lot more). That said, if people are just doing `docker-compose up` with the build local, they're going to run `bundle install` as part of the regular build, which I think eliminates the need for the `bundle check`, right?

I'm interested in your thoughts here (and a gut check on my assumptions!), but my solution to this would be:

* move the `command` out of docker-compose and into `dockerfile` 
* rather than the entrypoint shell script, have the command be `rails s -p 3000 -b 0.0.0.0`, or whatever the docker version of that is (I think `RUN ['rails', 's', '-p', '3000', '-b', '0.0.0.0']` but I forget exactly)
* keep the local build instead of the image, so that getting the code synced just takes a `git pull` instead of 'and also fight with docker'.

docker-compose.yml
DarthHater: Yeah, a few versions back from "tip". Probably good enough for anyone needing to upgrade Docker.

docker-compose.yml
DarthHater: In the 3.1 version, if you specify build and image it:

* Tells the `docker-compose build` command what image name to use when building the image, translates to basically `docker build -t colinxfleming/dcaf_case_management:latest .`

The main reason I did that was to just ensure that if someone builds it knows to a: create the image under the same namespace (avoid a bunch of odd images/layers being created) b: that it for sure would build, and not just always pull the remote from Dockerhub

entrypoint.sh
DarthHater: This doesn't really deprecate a remote build, it just gives someone an ability to build the image if need be via `docker-compose`. If for instance someone types in `docker-compose up` it will still go remote at first, but this gives someone an opportunity to build locally if they need to.

The entrypoint is meant to just do the following:

* When you type `docker-compose up` just really quickly check with `bundler` if the dependencies are golden. If they aren't, pull them, then start rails. 

Because `docker-compose up` will go remote, that's the reason I setup the entrypoint. You could pull an image with outdated code (since it's copied in as a part of the current build process) and if you've made gemfile updates, this just makes sure that the dependencies are fresh no matter what.

docker-compose.yml
geoffreyyip: Trying to follow this convo...

Are you saying that `docker-compose` uses the `image` parameter as a name (and nothing else) when you have both `image` and `build` params?

B/c it seems really weird if...

*Only `image` tag*: Build from remote, using `colinxfleming/dcaf_case_management:latest` ans a reference.
*`image` and `build` tag*: Build from local, using `colinxfleming/dcaf_case_management:latest`as a name.

DarthHater:
With only image tag it will pull, not build, from remote. With the build
tag you are saying if someone says build, use this context (which is set to
root by using .).

docker-compose.yml
geoffreyyip: @colinxfleming
[Docker docs (ha, Docker docs) say this about `dockercompose.yml` files](https://docs.docker.com/compose/compose-file/#image):

> 
> image
> 
> Specify the image to start the container from. Can either be a repository/tag or a partial image ID.
> ```
> image: redis
> image: ubuntu:14.04
> image: tutum/influxdb
> image: example-registry.com:4000/postgresql
> image: a4bc65fd
> ```
> If the image does not exist, Compose attempts to pull it, unless you have also specified build, in which case it builds it using the specified options and tags it with the specified tag.

So as I understand it,
`docker-compose pull` will remotely pull any service that has an `image` property.
`docker-compose build` will locally build any service that has a `build` property.
`docker-compose up` will attempt to use existing images. If images do not exist, then it will build images when a `build` property is specified. And it will pull images when *only* a `pull` property is specified.

And as @DarthHater mentioned, if `build` and `image` property are specified together, then the `build` property uses `image` tag for naming purposes. It doesn't seem like there's any fallback behavior anywhere...

DarthHater: As with all things, it's always more complex than a simple example BUT:

* `docker-compose up` will only build an image if it has to, if existing images are there that satisfy it's requirements, it'll just use them or the containers it's previously created from them
* `docker-compose pull` will pull from the Docker hub, remote, so if for some reason you want whatever is out there it's entirely possible to get it
* In most cases you don't absolutely need what's out on the hub BECAUSE:
  * The biggest advantage the remote image has it's got Gems already installed. I wouldn't say 9 times out of 10 but fairly often you are either introducing a new Gem or something akin and likely need to do `bundle install` to get things up and running

Build first would actually be better in most cases, because of the aforementioned Gem updates. That'll happen as you are building. However say you have an image locally already (lots of people do from pulling it in the past, and would in the future), the entrypoint.sh addition just makes it sort of seemless in terms of making sure your Gems are all there.

colinxfleming: It's news to me that you can have both `image` and `build` commands in a dockerfile and that it won't throw it back in your face! A nice-to-have feature for sure. I'm good with this part. Note to self to start internalizing `docker-compose pull` I guess.

entrypoint.sh
colinxfleming: I thought about this some and I think I'm good with it, but we're starting to get a lot of root directory docker overhead. Can we move this and `web-variables.env` to their own `docker_util` dir? I think at that point I'm good with this and we can smash merge.

Yr thoughts?

DarthHater: In: 294cb8d

DarthHater: You can't really move docker-compose (since you need it in root for people to run the command). Everything else is fair game.

DarthHater: More in: 3a657f4

PR #1379, opened 2018-02-03
Fix hardcoded dcaf URL in footer

### LINKED ISSUE #1378: BRAF Website Link is for DCAF

I was on the BRAF App today and noted that the BRAF Website link at the bottom of the page goes to the DCAF website. I checked heroku and that references BRAF's website so I'm not sure where it needs changed.

To access BRAF's website from the CM app.

find where the app is referencing the DCAF website

* **Anything else?**
Nope. Thanks for all the great work.

lomky:

Corrects the footer fund link and calls out for checking several other possible places needing the DCAF domain swapped for FUND_DOMAIN

* `FUND_DOMAIN` is now set to the value in the heroku env `FUND_MAILER_DOMAIN`
* flagged several other locations for possible changes

Domain still works in testing:
screenshot from 2018-02-03 16-13-44

* Fixes #1378 

Would need to be preceeded by a review of each running instance to confirm the `FUND_MAILER_DOMAIN` is set to the correct value (i.e. `dcabortionfund` not `dcabortionfund.org`)


colinxfleming:
Just checked -- In sandbox, `FUND_MAILER_DOMAIN` is set to `dcabortionfund.org`, and the fund production environments share that pattern (e.g. `whatever.org`). They're all .orgs at this time, and every fund I can think of is an .org, but I'm iffy on that being a safe assumption, haha.

ps great catches from poking around and seeing where we were using the mailer!

app/services/pledge_form_generator.rb
```diff
@@ -78,8 +78,8 @@ def build_header(pdf)
       pdf.text 'DC Abortion Fund'
       pdf.text 'P.O. Box 65061'
       pdf.text "Tel: #{FUND_PHONE}"
-      pdf.text 'E-mail: info@dcabortionfund.org'
-      pdf.text 'Web: www.dcabortionfund.org'
+      pdf.text 'E-mail: info@dcabortionfund.org' #should this be generalized?
```
colinxfleming: I don't think that generalizing this will do any harm, so I'm inclined to say yes.

config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = 'no-reply@dcabortionfund.org'
+  config.mailer_sender = 'no-reply@dcabortionfund.org' #should this be generalized?
```
colinxfleming: Almost definitely. Use the env var and fallback to dcabortionfund.org I think?

lomky:
Shall I go into the code and change to make the code expect `FUND_MAILER_DOMAIN` to have it's top level domain (e.g. `.org`), since that's safer and what the environments expect?

colinxfleming:
I think that's a good idea, if you don't mind making the fixes!

I think BRAF will be stoked about this, thanks for spinning this up so quick. Lots of power in being a responsive team. 🏋️‍♀️ 🏋️

lomky:
@colinxfleming for a second pass review to the new FUND format to include TLD. I'd appreciate a careful lookover since we're messing with the ENV stuff here, but I think everything is GTG.

config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  config.mailer_sender = 'no-reply@dcabortionfund.org'
+  config.mailer_sender = "no-reply@#{FUND_DOMAIN}" || 'no-reply@dcabortionfund.org'
```
colinxfleming: oh, oops, my fault! This one's going to break CI, because of how initializers load (in alphabetical order. ask me what deep, evil hole I had to go down to learn this.)

Recommend setting it to `no-reply@#{ENV['FUND_MAILER_DOMAIN']}` instead. A little dirty, but sometimes you can't fight city hall or whatever.

lomky:
@colinxfleming gtg! Let me know if I need to increase the patch coverage :)

colinxfleming:
@lomky nah you are good! Thanks for keeping tabs on this and bringing it home!

PR #1380, opened 2018-02-06
Tooltip popups

### LINKED ISSUE #1115: Document behavior of features within the app

colinxfleming:

Continues @KatSDC 's heroic fight with coffeescript in #1261 .

This works, but we're straight up using three different methods on tooltips -- one for checkbox labels, one for regular inline text, and one for text input labels. That isn't a great pattern. I'm not sure it's worth fixing now.

* Adds a tooltips helper to store the text constants
* Puts tooltips on status, call list, completed calls, urgent cases, external pledges, and resolved without assistance
* Moves tooltip specific js stuff to its own js file
* Unit tests

TK system tests for each type of tooltip. But for now.

screen shot 2018-02-06 at 12 23 45 am

* Fixes #1115


colinxfleming:
Still some mystery problems with this (it may have wrecked the call list?) but it's materially closer! Next steps are to get the test suite to pass, then we can ship.

colinxfleming:
We're looking good here!

test/system/tooltips_test.rb
```diff
@@ -0,0 +1,51 @@
...
+        find('.tooltip-header-help').hover
+        assert has_content? 'This is used to indicate that'
+      end
+    end
+  end
+
+  # Failing test. For some reason the data attributes don't show up?
+  # But only in chromedriver, and not firefox or real chrome?
+  # Stashing in an issue.
+  # describe 'an input label tooltip' do
```
colinxfleming: Going to see if I can fix this.

colinxfleming:
I think this is where it's going to end up, more or less. @KatSDC could I get a sign off from you on our changes here when you get a sec?

lomky:
I do think this one tooltip looks a little overwhelming, but if fixing it is a ton of work, nbd. :tada: 

tooltip

lwaldsc:
Hahaha that is certainly a tool tip huh.

Let me know if and how I can help on this @lomky. Writing appropriate tool
tip copy? Is there a way we can make the tool tip match the status or is
that way too complicated/dynamic for a quick fix? Seems too complicated.

colinxfleming:
Tooltip matching the status shouldn't be too tricky! Let me see what I can do.

colinxfleming:
This is maybe a little less buckwild:

image

@lwaldsc can you approve?

lwaldsc:
This is perfect. I think we could possibly make them a little more user friendly (on my to-do) but for right now these look awesome.

colinxfleming:
@lomky haha, they were passing before! This was just pre-polymorphic, so there was a little tweak to get it up to speed.

colinxfleming:
@lwaldsc I am 100% all ears for how this might work a little cleaner or better! Pop it in an issue when you get a sec. Merging this after tests pass though.

colinxfleming:
aw yeah passing tests. Merging. Thanks all!

PR #1384, opened 2018-02-21
updated rails to 5.1.5 and fixed docker error

rudietuesdays:

* updates rails to 5.1.5
* fixes a docker bug identified and fixed by @colinxfleming for starting up virtual box


.docker/Dockerfile
```diff
@@ -33,8 +33,4 @@ WORKDIR ${DCAF_DIR}
 
 COPY . ${DCAF_DIR}
 
-RUN chmod +x ${DCAF_DIR}/.docker/entrypoint.sh
-
 EXPOSE 3000
-
-ENTRYPOINT [ "/usr/src/app/.docker/entrypoint.sh" ]
```
colinxfleming: @DarthHater we were running into some trouble with this -- specifically, we couldn't run commands in `web`, they'd all default to the entrypoint script. We took it out as a fix and moved the command to `docker-compose.yml` which seemed to fix it.

I'll follow up with you separate but I think I'd like to simplify the docker stuff a little bit -- I don't know that we need a remote image in this era of being able to build in 5m start to back.

colinxfleming: (Thanks for working thru this with me @rudietuesdays !)

.docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM ruby:2.4.3-alpine
-MAINTAINER Colin Fleming <c3flemin@gmail.com>
```
colinxfleming: 😓

Issue #825, opened 2017-01-30
How should search handle patients on different lines?


colinxfleming:
Problem description TK

colinxfleming:
* **What are we trying to do?**
When CMs try to create a patient who's already on a line, right now they're getting an absolutely useless error message, and patient creation fails. The failure is right, but the useless error message sucks. (Steps to recreate - try logging into the VA line and creating a patient with phone number `123-123-1231`.)

* **What feature or behavior is this required for?**
Reducing frustration point for CMs.

* **How could we solve this issue? (Not knowing is okay!)**
Probably two things we should hammer out: What would a helpful error message look like? (Starting point: `That patient already exists on the VA line.`) And should we display them anyway with some kind of special flag that they're on a different line?

* **Anything else?**
nah

eheintzelman:
How does this play out?  If a DC line patient calls the VA line, does the VA line CM just pass on the vm message to a DC line CM?  Or are there cases where the CM would want the ability to change the line for that patient? Is this a functionality that currently exists in the system or is it only enabled for admin? 

Once we have a handle on this, we can create an error message that includes some direction for the CM.

How about the 'warning status' style for this? 
https://standards.usa.gov/components/alerts/

colinxfleming:
@eheintzelman I'm not sure how this plays out on the business side, actually. @lwaldsc @NerdyGirl537 what do CMs do when a patient hits the wrong voicemail when they're already going thru the process? 

There are def cases where we'd want to change the line of a patient, but that's admin only right now. We can loosen that to data volunteers and/or CMs @vtmeghan or @NerdyGirl537 think it's appropriate.

Warning status sounds great for this! 💯

lwaldsc:
Yeah, they'll simply email the case manager on the correct line with
something like "Oh hey Eva (#1-2345) called me on the DC line, but she's
actually yours." and then the correct CM would either start or continue
case management.

It really depends if the CM directors want to keep the added work of
changing a line for a patient. I don't think/know it happens often enough
to warrant letting all CMs do it, *but* it happens enough that a warning is
suuuuuuuuuuuuuper useful.

NerdyGirl537:
[...]

PR #1390, opened 2018-03-04
Handle search across lines

geoffreyyip:
Resolves: #825 

This change resolves a frustration point for CMs. Whenever they try to
register a patient whose phone number already exists in the system, they
will receive an error message stating the number is a duplicate. This
should not be a problem since CMs search for the number in the database
before creation. However, the search functionality is limited to the
CM's existing line, while the uniqueness check works across the entire
database.

For example, if the number 123-123-1231 exists on the DC line and the CM
is logged into the VA line, then they will receive no results when
searching for 123-123-1231. This implies a new patient. Then when the CM
creates a new patient with 123-123-1231, they will get an error saying
the number already exists.

To address this issue, I added a custom validator that tells the CM when
the patient already exists on a different line.

_(Additional side effect: I changed one of the phone numbers in `seed.rb`. 
Patient models are validated for unique phone numbers. Two of the
creation methods in seed.rb use the same template string to generate
phone numbers, which means there's a chance for the same number to be
generated twice.)_

If these changes look good, I can add some unit tests.

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -110,8 +110,10 @@ class Patient
             :line,
             presence: true
   validates :primary_phone, format: /\d{10}/,
-                            length: { is: 10 },
-                            uniqueness: true
+                            length: { is: 10 }
+
+  validate :confirm_unique_phone_number, on: :create
+
   validates :other_phone, format: /\d{10}/,
                           length: { is: 10 },
                           allow_blank: true
@@ -175,6 +177,28 @@ class Patient
     Event.where(patient_id: id.to_s).destroy_all
   end
 
+  def confirm_unique_phone_number
+    ##
+    # This method is preferred over Rail's built-in uniqueness validator
+    # so that case managers get a meaningful error message when a patient
+    # exists on a different line than the one the volunteer is serving.
+    #
+    # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
+    ##
+    already_taken_patient = Patient.where(primary_phone: primary_phone)
+    if already_taken_patient.exists?
+      # phones are unique, so there should only be one match in array
+      patients_line = already_taken_patient[0][:line]
+      volunteers_line = line
+
+      if volunteers_line == patients_line
+        errors.add(:this_phone_number_is_already_taken, "on this line.")
+      else
+        errors.add(:this_phone_number_is_already_taken, "on the #{patients_line}. Please contact the case manager on duty at the #{volunteers_line} line to confirm which line the patient should be under. If the line needs to be changed, please contact the CM Directors.")
+      end
+    end
+  end
+
   private
 
   def confirm_appointment_after_initial_call
@@ -194,7 +218,7 @@ class Patient
   def initialize_fulfillment
     build_fulfillment(created_by_id: created_by_id).save
   end
-  
+
   def update_pledge_sent_by_sent_at
     if pledge_sent  && !pledge_sent_by
       self.pledge_sent_at = Time.zone.now
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -185,16 +185,15 @@ class Patient
     #
     # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
     ##
-    # phones are unique, so there should be zero or one match
-    match = Patient.where(primary_phone: primary_phone).first
+    phone_match = Patient.where(primary_phone: primary_phone).first
 
-    if match
+    if phone_match
       # skip when an existing patient updates and matches itself
-      if match._id == self._id
+      if phone_match.id == self.id
         return
       end
 
-      patients_line = match[:line]
+      patients_line = phone_match[:line]
       volunteers_line = line
       if volunteers_line == patients_line
         errors.add(:this_phone_number_is_already_taken, "on this line.")
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -194,7 +194,7 @@ class Patient
       if volunteers_line == patients_line
         errors.add(:this_phone_number_is_already_taken, "on this line.")
       else
-        errors.add(:this_phone_number_is_already_taken, "on the #{patients_line}. Please contact the case manager on duty at the #{volunteers_line} line to confirm which line the patient should be under. If the line needs to be changed, please contact the CM Directors.")
+        errors.add(:this_phone_number_is_already_taken, "on the #{patients_line} line. If you need the patient's line changed, please contact the CM directors.")
       end
     end
   end
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -112,7 +112,7 @@ class Patient
   validates :primary_phone, format: /\d{10}/,
                             length: { is: 10 }
 
-  validate :confirm_unique_phone_number, on: :create
+  validate :confirm_unique_phone_number
 
   validates :other_phone, format: /\d{10}/,
                           length: { is: 10 },
@@ -185,12 +185,17 @@ class Patient
     #
     # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
     ##
-    already_taken_patient = Patient.where(primary_phone: primary_phone)
-    if already_taken_patient.exists?
-      # phones are unique, so there should only be one match in array
-      patients_line = already_taken_patient[0][:line]
-      volunteers_line = line
+    # phones are unique, so there should be zero or one match
+    match = Patient.where(primary_phone: primary_phone).first
 
+    if match
+      # skip when an existing patient updates and matches itself
+      if match._id == self._id
+        return
+      end
+
+      patients_line = match[:line]
+      volunteers_line = line
       if volunteers_line == patients_line
         errors.add(:this_phone_number_is_already_taken, "on this line.")
       else
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -248,7 +248,7 @@ end
   # initial create data from voicemail
   patient = Patient.create!(
     name: "Archive Dataful Patient #{patient_number}",
-    primary_phone: "321-0#{patient_number}0-004#{rand(10)}",
+    primary_phone: "321-0#{patient_number}0-005#{rand(10)}",
     voicemail_preference: "yes",
     line: 'DC',
     language: "Spanish",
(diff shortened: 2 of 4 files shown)


app/models/patient.rb
```diff
@@ -109,8 +109,10 @@ class Patient
             :line,
             presence: true
   validates :primary_phone, format: /\d{10}/,
-                            length: { is: 10 },
-                            uniqueness: true
+                            length: { is: 10 }
+
+  validate :confirm_unique_phone_number, on: :create
```
colinxfleming: Should do this on all state changes, not just create -- thoughts?

app/models/patient.rb
```diff
@@ -162,6 +164,28 @@ def event_params
...
 
+  def confirm_unique_phone_number
+    ##
+    # This method is preferred over Rail's built-in uniqueness validator
+    # so that case managers get a meaningful error message when a patient
+    # exists on a different line than the one the volunteer is serving.
+    #
+    # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
+    ##
+    already_taken_patient = Patient.where(primary_phone: primary_phone)
```
colinxfleming: Can change this `where` to a `find`, right? Since there will either be one or zero. That'll save us from having to do the index lookup on L178.

app/models/patient.rb
```diff
@@ -162,6 +164,28 @@ def event_params
...
+    if already_taken_patient.exists?
+      # phones are unique, so there should only be one match in array
+      patients_line = already_taken_patient[0][:line]
+      volunteers_line = line
+
+      if volunteers_line == patients_line
+        errors.add(:this_phone_number_is_already_taken, "on this line.")
+      else
+        errors.add(:this_phone_number_is_already_taken, "on the #{patients_line}. Please contact the case manager on duty at the #{volunteers_line} line to confirm which line the patient should be under. If the line needs to be changed, please contact the CM Directors.")
```
colinxfleming: I think we can collapse these two down to just something like `errors.add :this_phone_number_is_already_taken, 'on the #{patients_line} line. If you need the patient's line changed, please contact the CM directors.'

db/seeds.rb
```diff
@@ -254,7 +254,7 @@
   # initial create data from voicemail
   patient = Patient.create!(
     name: "Archive Dataful Patient #{patient_number}",
-    primary_phone: "321-0#{patient_number}0-004#{rand(10)}",
+    primary_phone: "321-0#{patient_number}0-005#{rand(10)}",
```
colinxfleming: 👍

app/models/patient.rb
```diff
@@ -162,6 +164,28 @@ def event_params
...
+      end
+    end
+  end
```
colinxfleming: I forget whether or not this is part of the callbacks behavior, but you MAY need to `throw(:abort`) on failure to prevent saving. Something to watch out for -- I forget exactly how this works.

app/models/patient.rb
geoffreyyip: `find` works by ids. There is `find_by`, which looks like this:

```ruby
Patient.find_by(primary_phone: primary_phone)
```
[Mongoid docs for #find_by](https://docs.mongodb.com/mongoid/6.3/api/Mongoid/Findable.html#find_by-instance_method)

Would that work? I dunno databases too well

app/models/patient.rb
geoffreyyip: done

app/models/patient.rb
geoffreyyip: You're right. Turns out I wrote my custom validator wrong. 

I set it to `on: :create` b/c the database seeding script kept yelling at me. My custom validator works for creating new items, but fails for updating existing items.

Say you wanted to do this to your data base:
```
- name: 'John', primary_phone: '123-456-7890'
+ name: 'Robert', primary_phone: '123-456-7890
```

You'd probably do this by searching for the John item, and then updating the name. Problem is, my custom validator sees your proposed change as a duplicate number. It doesn't know what an update is.

I need to change the validation function to handle updates. Uniqueness checking is trickiness than I first thought.

app/models/patient.rb
geoffreyyip: `find_by` has an tradeoff: code readability. `find_by` will throw an error when no match is found. 

There are two ways to handle this. Option 1 is something like this:
```ruby
begin
  already_taken_patient = Patient.find_by(primary_phone: primary_phone)
rescue
  return  // terminate the function since number is verified to be unique
end
```

(Syntax might be off.)

Option 2 is configuring `Mongoid.raise_not_found_error` to be `false`.

Oh, and option 3 is keeping the `where` syntax.

geoffreyyip: I've changed it to `Patient.where(primary_phone: primary_phone).first`, and rewrote the rest to match that.

Also, I'm just realizing as I type this that "index lookup on L178" is not a fancy obscure database perf issue, but just you wanting me to avoid `already_taken_patient[1]`.

app/models/patient.rb
geoffreyyip: done

geoffreyyip:
@colinxfleming, wanna check if there's a better way of doing this.

Right now, I'd using a `._id` to check uniqueness. This is something provided by Mongoid, and possibly hacky, since the leading underscore means "internal property". 

This is important for `on: update` checks. I want to validate unique phone numbers, so I search for records with duplicate numbers. Sometimes, the record with a duplicate number is the record I'm updating. In other words, the record is matching itself. In this case, I want to permit the update.

Here's what I'm doing now.

```ruby
match = Patient.where(primary_phone: primary_phone).first

if match
  # skip when an existing patient updates and matches itself
  if match._id == self._id
    return
  end
```

Is there a better way?

geoffreyyip:
On another note, I'm stumped as to why these system tests are failing. It seems that phone number updates on `_patient_dashboard.html` aren't being pushed thru in the system tests, but they seem to work fine when I try them out manually.

Even weirder is that the phone number is the only field that doesn't update. Every other field updates fine. Appointment date updates. First and last name updates. Etc.

Here's the main test that fails:
```ruby
describe 'changing patient dashboard information' do
    before do
      @date = 5.days.from_now
      fill_in 'First and last name', with: 'Susie Everyteen 2'
      select '5 weeks', from: 'patient_last_menstrual_period_weeks'
      select '2 days', from: 'patient_last_menstrual_period_days'
      fill_in 'Appointment date', with: @date.strftime('%m/%d/%Y')
      fill_in 'Phone number', with: '123-666-8888'
      fill_in 'First and last name', with: 'Susie Everyteen 2'
      click_away_from_field
      reload_page_and_click_link 'Patient Information'
    end

    it 'should alter the dashboard information' do
      within :css, '#patient_dashboard' do
        lmp_weeks = find('#patient_last_menstrual_period_weeks')
        lmp_days = find('#patient_last_menstrual_period_days')
        assert has_field?('First and last name', with: 'Susie Everyteen 2')
        assert_equal '5', lmp_weeks.value
        assert_equal '2', lmp_days.value

        assert has_field? 'Appointment date', with: @date.strftime('%Y-%m-%d')
        assert has_field? 'Phone number', with: '123-666-8888'

        assert has_content? 'Currently: 5w 4d'
        assert has_content? 'Approx gestation at appt: 6 weeks, 2 days'
      end
    end
  end
```

`assert has_field? 'Phone number', with: '123-666-8888'` is the failing line

app/models/patient.rb
```diff
@@ -162,6 +164,33 @@ def event_params
...
+    # exists on a different line than the one the volunteer is serving.
+    #
+    # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
+    ##
+    # phones are unique, so there should be zero or one match
+    match = Patient.where(primary_phone: primary_phone).first
+
+    if match
+      # skip when an existing patient updates and matches itself
+      if match._id == self._id
```
colinxfleming: Ah, I see what you mean here. You're right that the underscore is a mongoid internal thing and it's best not to use it. Fortunately, IIRC Mongoid also provides a regular `id` that you can use, I think. That said, see above for a suggestion that might make this a little straightforward?

app/models/patient.rb
```diff
@@ -162,6 +164,33 @@ def event_params
...
+  def confirm_unique_phone_number
+    ##
+    # This method is preferred over Rail's built-in uniqueness validator
+    # so that case managers get a meaningful error message when a patient
+    # exists on a different line than the one the volunteer is serving.
+    #
+    # See https://github.com/DCAFEngineering/dcaf_case_management/issues/825
+    ##
+    # phones are unique, so there should be zero or one match
+    match = Patient.where(primary_phone: primary_phone).first
```
colinxfleming: Ah, I see what you're doing! I think there a couple ways we can tidy this up.

I'd suggest renaming this variable to something that isn't a predefined method name in the ruby base library -- it should work fine, but this can be a total Zone of Pain kinda thing. Might suggest renaming it to 'phone_match' or something like that.

`.where` methods in activerecord/mongoid return an enumerable, as you've correctly noted with your `.first` here. You might be able to simplify this by changing it to a `.find`, which will either return nil or the single object.

You could also consider rewriting the query a little bit to explicitly exclude the patient's id -- it'll simplify your guard clause a little bit. Maybe something like the find version of `Patient.where.not(id: id.to_s)`, or the more mongoid version (I THINK) of `Patient.find_by(:id.ne => id.to_s)`? (you'll probably want to cast it to a string because the patient object considers it a BSON id, but it might work out of the box.)

In short, you can probably collapse these four lines into something similar to:

```
return true if Patient.find_by(primary_phone: primary_phone, :id.ne => id.to_s)
```

Don't feel married to this if it doesn't work, by the way! I'm pretty sleep deprived and you've spent more time thinking about this than I have!

colinxfleming:
I think this systemtest failure is a legit one, but I don't think it's you. Let me dig in.

colinxfleming:
Can you pull down from master?

geoffreyyip:
You mean merge changes from `master` into this branch? I'll have time this weekend

geoffreyyip:
@colinxfleming Fyi, system test passes now.

Still working on polishing changes and writing unit tests

colinxfleming:
awww yiss 👍

geoffreyyip:
Still working on this! April's been busy for me.
Working on it right now

geoffreyyip:
@colinxfleming It works! Ready for review

colinxfleming:
aw yus! Thanks @geoffreyyip , I'll review this wknd.

PR #1392, opened 2018-03-06
use number_field instead of text_field to improve usability

### LINKED ISSUE #1249: Design improvement: Use number inputs for numbers in patient_information partial

@eheintzelman has pointed out that we should use number fields for numbers, and she's def right!

Clean data!

In the patient information partial (`app/views/patients/_patient_information.html.erb`) we should change text fields to numbers. Specifically, we should do this for:

* Age
* Minors in household
* Adults in household

I THINK this is as simple as changing `text_field` to `number_field` -- check the docs of our form library here: https://github.com/bootstrap-ruby/rails-bootstrap-forms/tree/legacy-2.7 

* **Anything else?**

This started as something else. See below.

<strike>

@eheintzelman has very astutely pointed out that it's kinda clunky to us a toggle-spinner-deal for dropdowns (per below), and that a downward facing chevron would be better here.

A clean design where it's clear what certain inputs are looking for, datatype wise.

Conveniently, I have no idea! I'd like to avoid some jquery-ui plugin if possible but am not opposed if that's the path of least resistance.

* **Anything else?**

When we settle on a styling, we'll want to apply this in particular to input boxes on patient information, abortion information, and pledge fulfillment tabs, as well as the create patient form.

See this comment for origins. https://github.com/DCAFEngineering/dcaf_case_management/issues/1242#issuecomment-328695235

I'm tagging as UX/question because I don't have a terribly clear strategy/path to victory on this one, and would like people's opinions.
</strike>

mdworken:
Makes text forms for age and any $ amount (eg procedure cost) use number_field instead of text_field

* Fixes #1249

### DIFF
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -15,7 +15,7 @@
                                           patient.line),
                                           label: 'Line' %>
           <% end %>
-          <%= f.text_field :age, label: 'Age', autocomplete: 'off' %>
+          <%= f.number_field :age, label: 'Age', autocomplete: 'off' %>
           <%= f.select :race_ethnicity,
                        options_for_select(race_ethnicity_options,
                                           patient.race_ethnicity),
--- a/app/views/patients/_fulfillment.html.erb
+++ b/app/views/patients/_fulfillment.html.erb
@@ -26,7 +26,7 @@
         </div>
         <div class="info-form-left col-sm-6">
           <%= f.fields_for patient.fulfillment do |pt| %>
-            <%= pt.text_field :procedure_cost, label: 'Abortion care $', autocomplete: 'off', prepend: '$'%>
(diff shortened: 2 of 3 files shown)


mdworken:
No longer a WIP. Could someone review please?

lomky:
Screen cap of the cost fields. Age looks similar.

number fields

lomky:
@colinxfleming for final :+1: and :shipit:

app/views/patients/_patient_information.html.erb
```diff
@@ -113,4 +113,4 @@
       </div>
     </div>
   <% end %>
-</section>
+</section>
```
colinxfleming: Not a blocker here, but generally you want to have newlines at the end of files. This is the source of a lot of linux arguing/opinions, but it is kind of a good practice! Here's some of said arguing: https://unix.stackexchange.com/questions/23903/should-i-end-my-text-script-files-with-a-newline

colinxfleming:
Looking good, thanks a ton @mdworken !

PR #1398, opened 2018-03-21
Fix flaky patient update system test, hopefully

colinxfleming:

The patient update system test is causing us all sorts of consternation because it keeps failing to save phone number -- lately it's been failing about 9/10 times. This hopefully fixes it.

* adjusts the click body to click nav
* doesn't try to get too fancy with saving

No view changes, no issues.

### DIFF
--- a/test/integration_helper.rb
+++ b/test/integration_helper.rb
@@ -67,7 +67,7 @@ module IntegrationHelper
   end
 
   def click_away_from_field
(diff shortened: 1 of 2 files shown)


colinxfleming:
After observing some behavior in capybara that would be impossible for a human to replicate, I decided to go for a different tack and break this out into individual field tests instead.

PR #1404, opened 2018-03-31
Set up destroy route pt 2

### LINKED ISSUE #1279: Add destroy route for patients

colinxfleming:

Basically just @rudietuesdays pr but with a few place changes, and a few pattern changes here and there. (Pretty much works exactly right!) Next step is to write tests for:

* capybara deleting a patient success case, failure case, admin only
* patients controller route, success and failure case, admin only
* model callback
* new model method

* Add destroy patient route for dealing with dupes
* Add admin only destroy button

Here's the main view change, which will show up for admins only:

screen shot 2018-03-31 at 4 22 05 pm

* Fixes #1279 
* Stems from #1310


colinxfleming:
Tests are more or less done. Kicking up for review from @lomky and @rudietuesdays $$

PR #1405, opened 2018-04-03
Archive patient

### LINKED ISSUE #812: Patient archiving

lomky:
Creating a distinct archive_patient object, which after inheriting the whitelisted fields from a patient then nukes that PII laden object to space.

 WIP:
  - [x] Model out ArchivedPatient
     - [x] Fields
     - [x] Validations
     - [x] Fancy creation process that reads from & deletes the Patient object
       - [x] Patient
       - [x] sub objects
  - [x]  Controller, as needed, for ArchivedPatient
     - [x] Index
     - [x] Show
     - [x] Destroy
  - [x] Basic Views
  - [x] Restrict access to ArchivePatient to Data & Admins
  - [x] Tests!
    - [x] Model test
    - ~Controllers~
    - [x] Export test
    - [ ] System - maybe followup?
  - [x] Nightly tasks
    - [x] Archive fulfilled patients
    - [x] Archive inactive, dropoff patients

Follow On PRs:
  - [x] Cherry pick DBSeeds from previous PR - still good. Maybe in a standalone PR?
  - [x] Configs
    - [ ] Archive switch (vs plain deletion, for more paranoid / non data focused funds)
    - [ ] Days since creation to consider inactive
    - [ ] Days since fulfillment to consider complete
  - [x] Better views 
  - [x] Update Exportable to handle both Patients and ArchivedPatients

ReReplacement branch for #1162 & #1362, work for ticket #812.

Fixes #812!


colinxfleming:
This is looking great (and is way less complex since you extracted the polymorphism into its own thing)! Please tag me in if I can be of assistance on subobjects or testing, or when you're ready for a proper ~ code review ~ again.

I think that if it comes to it, we can skip the config work and just set a constant, and we can probs punt on better views if need be as well, but other than that I think you've nailed the scope here.

All in all, 💯

lomky:
I don't _remember_ writing the model tests, but git blame says I did it! /shrug

lomky:
@alisonnjones I'm looking to get accountant buy in for when it is okay to archive a fulfilled patient. We have it set at 120 days. How does that sound?

lomky:
@alisonnjones @mdworken Could I get your eyes on this PR to look at the Field list for the ArchivedPatient? What is missing from there compared to Patient that you will be sad to never have access to once this goes live?

FYI, the only _currently exported_ fields you will be losing are "Minors in Household" and "Adults in Household". Archived Patients will all report `nil` for those.

alisonnjones:
@lomky consulting with our treasurer now! if i don't hear back tonight i can definitely get you answer when i see them tomorrow.

lomky:
@alisonnjones Any update re: timeline on archiving fulfilled patients?

lomky:
Remaining tasks:

 - [ ] Debug the fulfillment test. 
    - possibly just the `before` objects are incomplete?
 - [ ] Create @mdworken requests fields:
    - `has_special_circumstances` - whether or not any special circumstance was flagged on the patient
    - `notes_count` - how many notes the patient had
  - [ ] Hear back re:fulfillment 120 days, adjust as requested.

alisonnjones:
@lomky @colinxfleming on her end she have to keep those physical pledges for seven years
per IRS guidelines, but she doesn't think there's a reason why we'd need to keep the data post 120 days. Can you link where the fields that are listed that will still be exportable from archived pts are?

mdworken:
@lomky what would your thoughts be on making an (offline) backup of the database and encrypting it before we archive patients?

lomky:
Tag in @lwaldsc, @colinxfleming, and @mdworken to give the above list another glare and see if there are fields we decide want to nuke after all.

lomky:
Colin and I discussed the config options, and decided against them.

config/initializers/env_var_constants.rb
```diff
@@ -14,3 +14,4 @@
 FUND_PHONE = ENV['DARIA_FUND_PHONE'] || '202-452-7464'
 FUND = ENV['DARIA_FUND'] || 'DCAF'
 FUND_FAX_SERVICE = ENV['DARIA_FAX_SERVICE'] || 'www.efax.com'
+FUND_ARCHIVE_PATIENTS = ENV['DARIA_ARCHIVE_PATIENTS'] || 'www.efax.com'
```
colinxfleming: we want this to be false instead of efax, I think?

app/models/patient.rb
```diff
@@ -175,6 +175,44 @@ def destroy_associated_events
...
+    notes.count
+  end
+
+  def has_special_circumstances
+    has_circumstance = 0
+    special_circumstances.each do |cir|
+      has_circumstance = 1 if cir
+      break
+    end
+    has_circumstance
```
colinxfleming: right on! Can we cast this as a boolean? We might also want to check that `cir.present?` since I think there are some Vintage Special Circumstances that might be empty strings?

app/models/concerns/exportable.rb
```diff
@@ -97,34 +116,18 @@ def reached_patient_call_count
   end
 
   def has_alt_contact?
-    other_contact.present? || other_phone.present? || other_contact_relationship.present?
+    if is_a?(Patient)
+      determine_has_other_contact
```
colinxfleming: I think this might confuse me a bit if we rely on the parent classes for this. I think it's alright if we keep the logic for this in here rather than delegating it back to the classes it's attached to.

e.g.

```
if is_a? Patient
  thing || other thing ||
else
  whatever the archived pt behavior is
```

Interested in your thoughts!

test/models/archived_patient_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  describe 'Archiving Methods' do
+    before do
+      @patient_old = create :patient, primary_phone: '222-222-3333',
+                                      other_phone: '222-222-4444',
+                                      initial_call_date: 600.days.ago
+      @patient_fulfilled = create :patient, primary_phone: '222-222-3334',
+                                   other_phone: '222-222-4443',
+                                   initial_call_date: 310.days.ago
+      @patient_fulfilled.fulfillment.fulfilled = true
```
colinxfleming: Might be slightly cleaner to do this with `update`, e.g.:

```
patient.update fulfillment: { fulfilled: true, date_of_check: 300.days.ago, ...}
```

That way you can be sure to get the state of the db in order before the tests start.

test/models/archived_patient_test.rb
```diff
@@ -0,0 +1,90 @@
...
+                                   appointment_date: 6.days.ago
+    end
+
+    it 'Should create an archive copy of a patient' do
+      @archived_patient = ArchivedPatient.convert_patient(@patient4)
+      @archived_patient.save!
+      assert_equal @archived_patient.line, @patient4.line
+      assert_equal @archived_patient.city, @patient4.city
+      assert_equal @archived_patient.race_ethnicity, @patient4.race_ethnicity
+      assert_equal @archived_patient.appointment_date, @patient4.appointment_date
```
colinxfleming: We should probably test the `destroy!` part of this process too, I think?

test/system/archived_patients_test.rb
```diff
@@ -0,0 +1,9 @@
+require "application_system_test_case"
```
colinxfleming: Can probably ditch this since there isn't any real user stuff, yeah?

app/models/archived_patient.rb
```diff
@@ -0,0 +1,146 @@
...
+      pledge_sent_by: patient.pledge_sent_by,
+      created_by_id: patient.created_by_id,
+
+      age_range: patient.age_range,
+      has_alt_contact: patient.determine_has_other_contact,
+      notes_count: patient.notes_count,
+      has_special_circumstances: patient.has_special_circumstances,
+
+    )
+    archived_patient.save! # Do I need this?
```
colinxfleming: I think I see your instinct here. Given that we want to roll this back if any part of this fails, maybe instead of `create`s we go with `new` and then do an `archived_patient.save!` at the end? This is kind of cheating our way into a database transaction but I think it will work. Thoughts?

app/assets/javascripts/archived_patients.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
colinxfleming: Can ditch the coffee and the scss file and the helper since we don't have the controller anymore, I think?

app/models/patient.rb
lomky: is there a nicer way to cast to boolean in rails other than `!!`?

app/models/concerns/exportable.rb
lomky: renamed the function on the Patient to `has_alt_contact` and removed the function from here completely. Now the call inside the export proper calls the correct one.

test/models/archived_patient_test.rb
lomky: threw in checks for Patient.count decrementing.

app/models/archived_patient.rb
```diff
@@ -0,0 +1,146 @@
...
+      pledge_sent_by: patient.pledge_sent_by,
+      created_by_id: patient.created_by_id,
+
+      age_range: patient.age_range,
+      has_alt_contact: patient.has_alt_contact,
+      notes_count: patient.notes_count,
+      has_special_circumstances: patient.has_special_circumstances,
+
+    )
+    archived_patient.save!
```
colinxfleming: What do you think about moving this down to L143, above where we return the archived patient? My reasoning there is that since mongo stores everything as a Big Blob O'JSON, if we delay the save action to the end we'll be sure to make sure that everything we copy over saves properly, and rolls back if not. This is kinda extremely defensive but I think be a helpful safety measure to prevent a patient from being created and then creation of associated objs failing quietly (since they don't use a bang! method). thoughts?

db/seeds.rb
```diff
@@ -322,46 +322,120 @@
...
 
 end
 
 
+# create patients who dropped off
+(1..2).each do |patient_number|
+  # initial create data from voicemail
+  patient = Patient.create!(
+    name: "Archive Dropoff Patient #{patient_number}",
+    primary_phone: "321-9#{patient_number}0-004#{rand(10)}",
```
colinxfleming: Lightning totally struck and this failed for me because of a primary phone conflict! (it ran fine the second time. Can we mix up the area code on this?

app/models/concerns/exportable.rb
```diff
@@ -48,7 +48,10 @@ module Exportable
     "Gestation at procedure in weeks" => :gestation_at_procedure,
     "Procedure cost" => :procedure_cost_amount,
     "Check number" => :check_number,
-    "Date of Check" => :date_of_check
+    "Date of Check" => :date_of_check,
+
+    # Notes
+    "Notes Count" => :notes_count,
```
colinxfleming: I have a left field question -- after taking the export for a test drive, I found myself sort of wishing for an indicator of whether a patient is archived or not. If you don't think it presents a PII or other problem, what are your thoughts about adding it?

test/models/archived_patient_test.rb
colinxfleming: 💯 this is pretty much exactly what I had in mind!

colinxfleming:
@lomky I gave some semi-bad advice -- `ArchivedPatient#convert_patient` needs those `creates` changed to `new`s, because you can only do create for embedded objects, which I guess makes sense but I had no idea until I tried to do it and mongoid told me to screw off. Sorry, my fault!

`ArchivedPatient.clinic` also seems to be acting a little weird. I think it might be assigning a random ID from whatever `clone` returns? I wonder if we can just assign `clinic_id: patient.clinic.id`. (We'll also want to add a `has_many :archived_patients` to `Clinic`.) The other objects all look perf though! I've replicated the clinics thing below:

```
irb(main):010:0> Clinic.each { |x| p x.id }  # the ids of our clinics
BSON::ObjectId('5ae3d056e3a47655b0fdcd96')
BSON::ObjectId('5ae3d056e3a47655b0fdcd98')
BSON::ObjectId('5ae3d056e3a47655b0fdcd9a')
BSON::ObjectId('5ae3d056e3a47655b0fdcd9c')
irb(main):018:0> ArchivedPatient.all.select { |x| x.clinic }.count.  # none of our archived patients have clinics
=> 0
irb(main):019:0> ArchivedPatient.all.select { |x| x.clinic_id }.count # but they do have clinic ids?
=> 2
irb(main):021:0> ArchivedPatient.all.select { |x| x.clinic_id }.last.clinic_id # clinic id doesn't seem to be one of the clinic ids currently in db
=> BSON::ObjectId('5ae3d090e3a47655ef5f886e')
```

Finally: Is there anything in this PR that you think could use some more rigorous testing or additional coverage? Anything you want to flag before we merge, or tests I can help with writing?

colinxfleming:
Just in case I ever need it again here's the one liner to import a mongodump into a local: `mongorestore -h 127.0.0.1:27017 -d dcaf_case_management_development exports/DIR_OF_EXPORT/ --drop --noIndexRestore`

PR #1406, opened 2018-04-05
Bump ruby and rails

colinxfleming:

A ruby vulnerability got reported, and it'll fail builds until we bump it, so! Bumping rails a point release while we're in here.

* Ruby to ~2.5.1~ 2.4.4
* Rails to 5.1.6

No views, no issues.


colinxfleming:
Circle is having an issue with rubygems 2.7.3, which has an incompatibility with 2.5.1, so upgrading to 2.4.4 to squelch the warnings instead of 2.5.1.  :(

colinxfleming:
This is dead in the water until circleci updates their docker images for 1.0, or possibly until we swap to 2.0 which I think lets you declare your own build image? idk. either way, grim scene.

PR #1414, opened 2018-04-15
Add ability to lock user accounts administratively

### LINKED ISSUE #1098: Add Locking Feature for Admins

Admins should be able to lock/unlock users for whatever reason.

User admin panel.

We are still discussing a workflow regarding how we handle locking patients.

* **Anything else?**
See #1073

colinxfleming:

Funds need a way to administratively and permanently shut off a user account, in case a CM moves or stops volunteering with the fund or something

* Adds perma-shutoff option and ability to toggle in the UI
* TK confirm not permashutoff before every action
* Tests for above

TODO: Add a before action to confirm that current user is not disabled by fund; verbiage check; tests.

Screenshot TK

* Fixes #1098


colinxfleming:
VROOM VROOM

app/controllers/application_controller.rb
```diff
@@ -34,6 +35,13 @@ def confirm_admin_user_async
   end
 
   def redirect_unless_has_data_access
-    redirect_to root_url unless current_user.allowed_data_access?
+    redirect_to root_path unless current_user.allowed_data_access?
+  end
+
+  def confirm_user_not_disabled!
+    if current_user && current_user.disabled_by_fund?
```
lomky: Currently this treats `current_user == false` the same as and enabled user, which seems like a bad plan. I don't think you can get here with current_user being false (authenticate would have kicked you out of the before_action stack and canceled the rest of em), but it might be good to clean that up.

app/controllers/application_controller.rb
```diff
@@ -6,7 +6,8 @@ class ApplicationController < ActionController::Base
```
lomky: Could you provide some written clarification about the distinction we're making between "locked" and "disabled" user accounts, and how we're communicating that outwards.

app/helpers/users_helper.rb
```diff
@@ -9,8 +9,10 @@ def user_role_options
   end
 
   def user_lock_status(user)
-    if user.access_locked?
-      'Locked'
+    if user.disabled_by_fund?
+      'Locked by admin'
+    elsif user.locked_at?
```
lomky: Why is this now `locked_at?` but before it was `access_locked?`?

colinxfleming: Turns out `access_locked?` is one of those things we get for free from devise, so I restored this.

app/controllers/application_controller.rb
colinxfleming: Took out the first current user. This should now choke if current user isn't set, which is good news I think!

app/controllers/application_controller.rb
```diff
@@ -34,6 +35,13 @@ def confirm_admin_user_async
   end
 
   def redirect_unless_has_data_access
-    redirect_to root_url unless current_user.allowed_data_access?
+    redirect_to root_path unless current_user.allowed_data_access?
+  end
+
+  def confirm_user_not_disabled!
+    if current_user && current_user.disabled_by_fund?
+      flash[:danger] = 'Account currently locked, check with your administrator.'
```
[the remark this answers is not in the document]
colinxfleming: Agree! changed.

app/controllers/application_controller.rb
```diff
@@ -6,7 +6,8 @@ class ApplicationController < ActionController::Base
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :prevent_caching_via_headers
-  before_action :authenticate_user!
+  prepend_before_action :authenticate_user!
+  prepend_before_action :confirm_user_not_disabled!, unless: :devise_controller?
```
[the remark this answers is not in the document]
colinxfleming: I was worried that changing this was going to be a walking talking nightmare, but it seems to work just by swapping it out, so I think we can pretty painlessly change this. Great catch.

app/controllers/application_controller.rb
colinxfleming: Did this in `app/models/user.rb` by the data field.

colinxfleming:
It's an absolute mystery to me why this is but I cannot get the flash messages to show up unless those filters fire as `prepend_before_action`s. I'm going to revert that change for the time being and follow up with @lomky later.

PR #1419, opened 2018-05-05
Delete flaky tests we are never going to fix

### LINKED ISSUE #1312: test/system/call_list_test.rb is still flaky after moving to systemtest

There is a part of the call list behavior test that depends on some time travel, and it seems to be incompatible with the systemtest.

Testing call list behavior, a major component of our app.

Either a) figure out how to test with time travel in systemtests or b) change the objects so that they're created db side or c) decide the unit test components of it are fine and we don't need to troubleshoot this particular behavior.

* **Anything else?**
Stems from #1309

### LINKED ISSUE #1313: test/system/navbar_links_test.rb has a flaw that's making it fail every time

Test the behavior around showing the link to a fund's CM resources, which depends on an environment variable, which systemtest doesn't play nicely with.

Relatively minor feature, but this gets a fund's resources closer to the CM's nose while they're working, which is a good thing. We rarely touch this part of the stack these days, but we don't want any regressions especially as we move into a more maintenance oriented mode.

Either we a) figure out how to change env vars and reload puma or b) move this part of it back to an integration test, where we CAN change env vars on the fly and have it be cool or c) move all this app code to a helper and test the helper.

* **Anything else?**
Stems from #1309

colinxfleming:

"Fixing" some tests that are going to soak up more time than they're worth. Know when you're beaten, etc.

* Remove call list time travel test
* Remove navbar test that requires subbing out env vars, which we can't do in puma

no view changes

* Fixes #1312 
* Fixes #1313 
* Bumps #Y

### DIFF
--- a/test/system/call_list_test.rb
+++ b/test/system/call_list_test.rb
@@ -76,24 +76,6 @@ class CallListTest < ApplicationSystemTestCase
         assert has_no_text? @patient.name
       end
     end
-
-    # TODO flaky test and I have no idea why
-    # it 'should time a call out after 8 hours' do
-    #   sign_out
-    #   travel(9.hours) do
-    #     log_in_as @user
-    #     wait_for_element 'Your completed calls'
-    #     sleep 5
-
-    #     within :css, '#completed_calls_content' do
-    #       assert has_no_text? @patient.name
-    #     end
-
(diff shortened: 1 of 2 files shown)


colinxfleming:
Invoking admin and merging.

PR #1421, opened 2018-05-26
Change order of call creation in seeding to avoid renaming thing

### LINKED ISSUE #1294: Patient Not Showing in Call List

Have a patient I reached and need to call back tonight. Used the activity feed option to add to the call list. Now, she doesn't show in the call list but when I search her, she only has the red X, not the Add option. I've X'd her and re-added her, and she still doesn't show. I've logged out of DARIA and closed the browser and then logged back in. Still no.

Having patients show in the call list.
No clue

* **Anything else?**
DARIA is awesome thanks to you!

colinxfleming:

Fun facts about our seeds:

* they're not relational

That means that the name sticks, and because a bunch of our fake patients get renamed in the seed process, that means a little bit of unhelpful confusion when events get created reflecting their old names. So we're reordering the call creation a little bit.

* Reorders the seed file so calls get created after renaming

* Fixes #1294

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -50,21 +50,6 @@ end
 
 # Create associated objects
 Patient.all.each do |patient|
-  # Create calls for patient
-  5.times do
-    patient.calls.create! status: 'Left voicemail',
-                          created_at: 3.days.ago,
-                          created_by: user2 unless patient.name == 'Patient 9'
-  end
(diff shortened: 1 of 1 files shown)


colinxfleming:
I'm invoking godmode after CI passes since this isn't a prod facing change.

PR #1422, opened 2018-05-26
Upgrade oauth gems

### LINKED ISSUE #1171: Upgrade omniauth-google-oauth2

Keep our oauth libraries up to date. (See the Gemfile for the particular two.)

Google SSO

Probably run back thru the gem installation and setup, and remove any other setup.

* **Anything else?**
The oauth library bump in #1123 busted SSO. Reverted in #1170 and d440abb . This here seems to be the root of the problem: https://github.com/intridea/omniauth-oauth2/issues/81

colinxfleming:

The oauth library ecosystem has resolved its weird hangups, so we can unpin a couple of our oauth libraries. And keeping oauth libraries up to date is important, because ~ authentication~ !

* bump `devise`
* bump `omniauth-google-oauth2`

I could get this to work locally but we should be sure to hammer on it on sandbox too.

* Fixes #1171

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,8 +29,7 @@ gem 'enumerize' # Mongoid doesn't have enum out of the box, so we get it here
 
 # Our authentication library is devise, with oauth2 for google signin
 gem 'devise', '~> 4.4'
-gem 'omniauth-google-oauth2', '0.2.1' # TODO upgrade
-gem 'omniauth-oauth2', '1.3.1' # TODO remove this pin
+gem 'omniauth-google-oauth2', '0.5.3'
 
 # We use `bootstrap_form_for` in views
 gem 'bootstrap_form'
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem 'mongo_session_store', '>= 3.1.0' # stores sessions in database for security
 gem 'enumerize' # Mongoid doesn't have enum out of the box, so we get it here
 
 # Our authentication library is devise, with oauth2 for google signin
-gem 'devise', '~> 4.3'
+gem 'devise', '~> 4.4'
 gem 'omniauth-google-oauth2', '0.2.1' # TODO upgrade
 gem 'omniauth-oauth2', '1.3.1' # TODO remove this pin
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,8 +24,7 @@ class User
           :validatable,
           :lockable,
           :timeoutable,
-          :omniauthable,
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -24,8 +24,7 @@ class User
           :validatable,
           :lockable,
           :timeoutable,
-          :omniauthable,
-          :omniauth_providers => [:google_oauth2]
+          :omniauthable, omniauth_providers: [:google_oauth2]
```
DarthHater: Inlining these seems unintentional, SPLAIN YASELF

config/initializers/devise.rb
```diff
@@ -235,7 +235,7 @@
   # ==> OmniAuth
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
-  config.omniauth :google_oauth2, ENV['DARIA_GOOGLE_KEY'], ENV['DARIA_GOOGLE_SECRET']
+  config.omniauth :google_oauth2, ENV['DARIA_GOOGLE_KEY'], ENV['DARIA_GOOGLE_SECRET'], {}
```
DarthHater: PREGUNTA: What is the {} for?, new options you can pass that we aren't?

Gemfile
```diff
@@ -28,9 +28,8 @@ gem 'mongo_session_store', '>= 3.1.0' # stores sessions in database for security
 gem 'enumerize' # Mongoid doesn't have enum out of the box, so we get it here
 
 # Our authentication library is devise, with oauth2 for google signin
-gem 'devise', '~> 4.3'
-gem 'omniauth-google-oauth2', '0.2.1' # TODO upgrade
-gem 'omniauth-oauth2', '1.3.1' # TODO remove this pin
```
DarthHater: PREGUNTA: I take it that devise folded omniauth-oauth2 into the main package and thus the drop?

config/initializers/devise.rb
colinxfleming: I'm not sure if omniauth now takes a set number of arguments or if this is basically the same as what we were doing, but I followed the setup docs just to be on the safe side since we're bumping a bunch of major versions. Anyway, this is in there, figured it was best to just go with it.

Gemfile
colinxfleming: We previously had to have this pinned because of a nasty incompatibility while some programmers argued about whose problem it was, or something? Either way, `omniauth-google-oauth2` doesn't break anymore if the `omniauth-oauth2` gem is too high, so we're good to just go with whatever the `omniauth-google-oauth2` gem would like.

PR #1423, opened 2018-05-26
Split call list endpoints into their own controller

### LINKED ISSUE #1346: Split out call list functionality from users controller

Our users controller is doing double duty by managing the call lists and also users. We should split them out so we don't bog it down.

Not driving our future selves crazy

Split em out and fix the routing until the tests pass.

* **Anything else?**
Nope!

colinxfleming:

Port call list specific functionality into its own controller, to not bog down the users controller (which mostly handles devise / auth junk).

This also fixes an unused arg in `add/remove patient`, since we're working with current user and not an actual user id.

s/o to our test suite for making this e-z!

* `rails g controller call_lists`
* copy and paste a couple of things around
* adjust a couple things to not take a second argument of user

* Fixes #1346

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -85,7 +85,7 @@ class User
   validates :name, presence: true
   validate :secure_password
 
-  TIME_BEFORE_INACTIVE = 2.weeks
+  TIME_BEFORE_DISABLED_BY_FUND = 9.months
 
   def secure_password
     return true if password.nil?


colinxfleming:
I hadn't planned on it here but I do think it's a good idea, and it shouldn't be terribly tough to do. I'll rack that before merge. Good call @mdworken !

colinxfleming:
hmm, it looks like the only actual call list field is `call_order`, and it probably doesn't make sense to have a one-field model. I'll move the methods to a concern and add a comment.

PR #1424, opened 2018-05-27
Lock inactive users after 9 months of inactivity

### LINKED ISSUE #713: Lock users out after 6 months
In order to control the number of users and access to the system, can CM App automagically lock users out after six months of inactivity? CMDs could unlock their accounts if they become active again (see #626). 

CMs are supposed to work at least one shift per quarter.

colinxfleming:

We talked about avoiding having too many inactive but usable accounts by locking after 6 months. This sets up disabling them after 9 months, to allow a little bit of wiggle room and to reduce threat vectors for funds with a lot of CMs.

TK is writing tests, and I also noticed that logging in from oauth doesn't update the `last_sign_in_at` field on User, so we'll want to make an adjustment to the `from_omniauth` method to handle that.

* Add a command to the nightly cleanup task
* Add a user command to get and lock/disabled_by_fund user accounts

No view changes.

* Fixes #713

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -19,7 +19,8 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def log_in_user_via_google
     flash[:notice] = I18n.t 'devise.omniauth_callbacks.success',
                             kind: 'Google'
-    sign_in_and_redirect @user, event: :authentication
+    sign_in @user, event: :authentication
+    redirect_to root_path
   end
 
   def reject_login
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -116,7 +116,7 @@ class User
     inactive_no_logins = where(:role.nin => [:admin],
                                :last_sign_in_at => nil,
                                :created_at.lt => cutoff_date)
-    (inactive_no_logins | inactive_has_logged_in).each do |set|
+    [inactive_no_logins, inactive_has_logged_in].each do |set|
       set.update disabled_by_fund: true
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -119,7 +119,6 @@ class User
     [inactive_no_logins, inactive_has_logged_in].each do |set|
       set.update disabled_by_fund: true
     end
-    fail 'until oauth login updates last_sign_in_at'
   end
 
   # ticket 241 recently called criteria:
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -111,12 +111,12 @@ class User
 
   def self.disable_inactive_users
     cutoff_date = Time.zone.now - TIME_BEFORE_DISABLED_BY_FUND
-    inactive_has_logged_in = where(:role.nin => :admin,
+    inactive_has_logged_in = where(:role.nin => [:admin],
                                    :last_sign_in_at.lt => cutoff_date)
-    inactive_no_logins = where(:role.nin => :admin,
+    inactive_no_logins = where(:role.nin => [:admin],
                                :last_sign_in_at => nil,
                                :created_at.lt => cutoff_date)
-    [inactive_no_logins, inactive_has_logged_in].each do |set|
+    (inactive_no_logins | inactive_has_logged_in).each do |set|
       set.update disabled_by_fund: true
     end
   end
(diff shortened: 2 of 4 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -19,7 +19,8 @@ def google_oauth2
   def log_in_user_via_google
     flash[:notice] = I18n.t 'devise.omniauth_callbacks.success',
                             kind: 'Google'
-    sign_in_and_redirect @user, event: :authentication
+    sign_in @user, event: :authentication
```
colinxfleming: `sign_in_and_redirect` doesn't fire off devise callbacks for some extremely obnoxious reason! So I had to make an edit to get google oauth to bump `last_sign_in_at`, `last_sign_in_ip`, etc.

colinxfleming:
The patch diff is because of an edit to the omniauth thing, which we aren't really testing because it's a pain in the ass.

Before we merge we'll need to account for the fact that google oauth login wasn't setting sign times via the following one liner (which is just going to set things to current day, so this is essentially a fix going forward for that population):

```
User.where(last_sign_in_at: nil).update(last_sign_in_at: Time.zone.now)
```

app/models/user.rb
```diff
@@ -85,6 +85,7 @@ class User
   validate :secure_password
 
   TIME_BEFORE_INACTIVE = 2.weeks
+  TIME_BEFORE_DISABLED_BY_FUND = 9.months
```
DarthHater: Suggestion/Bike Shedding: move these out of a model into something like config. Not the biggest deal at the end of the day but overall app config can benefit from being in one place rather than spread out.

lib/tasks/nightly_cleanup.rake
```diff
@@ -3,6 +3,9 @@ task nightly_cleanup: :environment do
   User.all.each { |user| user.clean_call_list_between_shifts }
   puts "#{Time.now} -- cleared all recently reached patients from call lists"
 
+  User.disable_inactive_users
+  puts "#{Time.now} -- locked accounts of users who have not logged in since #{9.months.ago}"
```
DarthHater: Suggestion: Replace `#{9.months.ago}` with the `TIME_BEFORE_DISABLED_BY_FUND` so that if it's updated this doesn't get out of date

test/models/user_test.rb
```diff
@@ -214,5 +214,28 @@ class UserTest < ActiveSupport::TestCase
       @user.reload
       assert @user.disabled_by_fund
     end
+
+    describe 'nightly cleanup task - disable_inactive_users' do
+      before do
+        @disabled_user = create(:user, last_sign_in_at: 10.months.ago)
```
DarthHater: Idea: given you have a couple different cases, look into doing this test with a `DataTable`?

test/models/user_test.rb
```diff
@@ -214,5 +214,28 @@ class UserTest < ActiveSupport::TestCase
...
+      it 'should find inactive users with logins before cutoff and disable' do
+        [@disabled_user, @no_login_user].each do |user|
+          user.reload
+          assert user.disabled_by_fund?
+        end
+
+        # No locking admins or users with recent activity
+        [@admin, @nondisabled_user].each do |user|
+          user.reload
+          refute user.disabled_by_fund?
```
DarthHater: Love the use of refute, but you could just do `assert !user.disabled_by_fund?` (pseudocode) probably, which might read easier to a newish Ruby person.

app/models/user.rb
colinxfleming: ah, I hear you. We've been doing this like so because of the rails convention to declare global stuff in initializers or `application.rb` and other constants more limited in scope in the module they're used in. Since this is specific to the user obj and not really used outside of here except in a print statement, probably best to follow convention and keep them closer to the place they're used.

lib/tasks/nightly_cleanup.rake
colinxfleming: great flag!

test/models/user_test.rb
colinxfleming: per discussion in slack, there isn't a fantastic model for this to glom onto in ruby rn, so punting.

test/models/user_test.rb
colinxfleming: Rubocop says we're both wrong as of 0.56: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/RefuteMethods

```
Use `assert_not` methods instead of `refute` methods.
```

:(

colinxfleming:
thanks @DarthHater !

PR #1426, opened 2018-06-02
Install sqreen's devise plugin to block sketch signups

### LINKED ISSUE #1372: Crank up sqreen with their devise stuff
https://github.com/sqreen/devise_sqreener for later

colinxfleming:

Installs a sqreen plugin to block iffy signups. More later because it's late. Still need to confirm that this doesn't just break signups via some local testing. 

* Installs sqreener devise gem

No view changes.

* Fixes #1372

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,4 +107,5 @@ end
 group :production do
   gem 'skylight' # our newrelic-style efficiency monitoring platform
   gem 'sqreen' # an active security monitoring platform
+  gem 'devise_sqreener' # A sqreen devise plugin
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -110,6 +110,7 @@ GEM
       railties (>= 4.1.0, < 6.0)
       responders
       warden (~> 1.2.3)
+    devise_sqreener (0.1.1)
     docile (1.1.5)
     easy_diff (1.0.0)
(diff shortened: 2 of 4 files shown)


colinxfleming:
I'm not sure if this is a DCAF problem or a Sqreen problem but making these calls even with an env var set seems to be causing us trouble. 

```
a = User.find_by email: 'c3flemin@gmail.com'
a.class.sqreener.sqreen_email('c3flemin@gmail.com')
=> nil
# inspecting the request response
=> "{\"code\":401,\"details\":null,\"errno\":4011,\"error\":\"Unauthorized\",\"info\":null,\"message\":\"No key was found in this request. You can retrieve your free API key at https://www.sqreen.io/developers\"}\n"

# and the class making the request
User.first.class.sqreener
=> #<DeviseSqreener::Sqreen:0x00007fae32a8e5e8 @sqreen_api_token="a790b0729e3b4779bebb07d6c2ec63d3931b5129fd1e4e23a24af5b7a5da09da">
a.class.sqreener
```

I'm not sure whether this is my fault because forkin' or not, but I kinda suspect not. I'll be spinning up  an issue on their side I think.

colinxfleming:
Opened an issue for this @ https://github.com/sqreen/devise_sqreener/issues/3

Issue #1097, opened 2017-06-05
Add Reset Password Functionality for Admins


tingaloo:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
We want to add reset password functionality for admins! This can be found at ```users#index```. 

* **What feature or behavior is this required for?**
To reset passwords.

* **How could we solve this issue? (Not knowing is okay!)**
We have a few TODOs lying around. See ```reset_password``` in the ```users_controller``` and ```lock_test``` in ```user_management_test```

* **Anything else?**
There WILL be a route change!
This spins off an issue from #1073

PR #1427, opened 2018-06-02
User view styling, password reset by admin shutoff

### LINKED ISSUE #626: Design administrative support screens for user management

### LINKED ISSUE #1100: Finish Writing Tests for User Management

There are a few TODOs lying in ```user_management_test``` that need finishing.

Being good people and writing tests.

Some tests require an understanding of the business logic of locking/resetting passwords. Bug @colinxfleming about this.

* **Anything else?**
See #1073

colinxfleming:

At long last I am finishing off the fabulous looking user screens.

After a lot of reflection, I am going to say we should pass on being able to administratively reset a user's password. The reason for this is that an individual user can already do this thru the frontend, so all this functionally does is open up an attack vector with no real benefit imo.

A big s/o to @nimiro whose beautiful designs were a pleasure to implement!!

* Clean up users index and user view
* Delete a bunch of not finished stuff on reset password in admin panel
* Some test cleanup since I was rooting around in there

Users index and edit page:

screen shot 2018-06-02 at 5 57 48 pm
screen shot 2018-06-02 at 5 57 54 pm

* Fixes #1100 
* Fixes #626 
* "Fixes" #1097 as a wontfix


colinxfleming:
thanks @tingaloo !

PR #1433, opened 2018-06-14
WIP clean up javascripts

colinxfleming:

Lots of cruft in the javascript. This leaves some comments and pushes things around so they're more cleanly organized.

* Delete empty coffee files
* Adds comments
* Does some shifting so things are a little more SRP

No views, no issues.


colinxfleming:
eh.

PR #1434, opened 2018-06-14
WIP Clean up CSS

colinxfleming:

Break up CSS a little bit so we don't drive ourselves nuts doing simple stuff.

* Clean up CSS dir

No view changes, no issues.


colinxfleming:
eh.

PR #1437, opened 2018-06-20
update sprockets

colinxfleming:

Sprockets has a vulnerability. We're upgrading to a safe version.

* `bundle update sprockets`

No views or issues


colinxfleming:
Invoking godmode and merging

PR #1440, opened 2018-06-22
Make CM Resources a config

### LINKED ISSUE #1268: Turn CM Resources link into a config

CM Resources link currently redirects to a production google drive link. Documents are currently stored in the drives of the individual line accounts.

Not sure. Need to consult CMDs on what they feel the best option is or how we make sure CMs are going to the right location.

* **Anything else?**
Anthony Clark's Wizard a Day is the most precious thing on the internet right now, but today's is especially precious: https://twitter.com/nedroid/status/917398416216518656

colinxfleming:

Adds a config for CM Resources link in the top navbar, and does some wiring to get the env var out of the way.

* New config value, plus the option to override it
* Tweak the view to pull from the config instead of the value

Shows up like so:

screen shot 2018-06-22 at 1 44 15 am

* Fixes #1268

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -10,6 +10,7 @@ fail 'No running seeds in prod' unless [nil, 'Sandbox'].include? ENV['DARIA_FUND
 Patient.destroy_all
 User.destroy_all
 Clinic.destroy_all
+Config.destroy_all
 
 # Create google SSO user
 # puts "Creating user with email #{ARGV[1]}..."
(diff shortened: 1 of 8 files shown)


db/seeds.rb
```diff
@@ -148,6 +148,8 @@
               config_value: { options: ['Pledge Limit Guidelines:', '1st trimester (7-12 weeks): $100', '2nd trimester (12-24 weeks): $300', 'Later care (25+ weeks): $600']}
 Config.create config_key: :language,
               config_value: { options: ['Spanish', 'French', 'Korean']}
+Config.create config_key: :language,
```
lomky: Should this be exercising the new one instead?

colinxfleming: 🤦‍♂️  oops

colinxfleming:
How's this look @lomky ?

screen shot 2018-06-27 at 10 29 09 pm

PR #1443, opened 2018-06-30
Clarify lines usage

### LINKED ISSUE #1385: Improve syntax in patient querying methods

Clean syntax up for clarity in patient querying methods. All the following methods take a `lines` parameter, implying an Array. In fact, the default parameter `LINES` defaults to `[:DC, :MD, :VA].

I didn't track down every instance where these methods were called... but it seems like these methods are always called with a String or Symbol. And as far as I could tell, the default `LINES` parameter of `[:DC, :MD, :VA]` was never used.

```ruby
# user.rb:106
  def recently_called_patients(lines = LINES)
    patients.in(line: lines)
            .select { |patient| recently_called_by_user? patient }
  end

  def call_list_patients(lines = LINES)
    patients.in(line: lines)
            .reject { |patient| recently_called_by_user? patient }
  end

# urgency.rb:22
  class_methods do
    def urgent_patients(lines = LINES)
      Patient.in(line: lines).where(urgent_flag: true)
    end

# patient_searchable.rb:54
    def find_name_matches(name_regexp, lines = LINES)
      if nonempty_regexp? name_regexp
        primary_names = Patient.in(line: lines).where name: name_regexp
        other_names = Patient.in(line: lines).where other_contact: name_regexp
        return (primary_names | other_names)
      end
      []
    end

# patient_searchable.rb:71
    def find_phone_matches(phone_regexp, lines = LINES)
      if nonempty_regexp? phone_regexp
        primary_phones = Patient.in(line: lines).where(primary_phone: phone_regexp)
        other_phones = Patient.in(line: lines).where(other_phone: phone_regexp)
        return (primary_phones | other_phones)
      end
      []
    end
```

This is not a new feature. This is a refactor proposal.

Remove the default `LINES` parameter. Then rename the `lines` argument to `line`. This conveys that the methods are meant to take a single String or Symbol.

* **Anything else?**

Maybe this method is meant to take a String, Symbol or Array? I'm not sure. I only scratched the surface of the Patient model. I could some context here. @colinxfleming and @tingaloo were the last contributors to modify these methods.

colinxfleming:

@geoffreyyip pointed out that we made a couple methods a little too loose, and it's confusing, so where possible we've tightened them down to just the locals.

* call list and urgent patients methods now take single lines
* fix tests to match

No view changes.

* Fixes #1385 

@geoffreyyip for review please!


app/models/concerns/urgency.rb
```diff
@@ -20,8 +20,8 @@ def confirm_still_urgent
   end
 
   class_methods do
-    def urgent_patients(lines = LINES)
-      Patient.in(line: lines).where(urgent_flag: true)
+    def urgent_patients(line)
+      Patient.where(line: line).where(urgent_flag: true)
```
DarthHater: TRIVIAL: Is it possible to combine the .where's into one? Not a big deal but figured I'd see.

app/models/concerns/call_listable.rb
```diff
@@ -8,13 +8,13 @@ module CallListable
   # AND they would otherwise be in the call list
   # (e.g. assigned to current line and in user.patients)
 
-  def recently_called_patients(lines = LINES)
```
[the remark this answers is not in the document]
colinxfleming: eh, this I think we should keep as in the PR. I think the core of @geoffreyyip 's point was that having default args like this are kind of confusing, esp given that we only ever think about call lists in the context of lines.

app/models/concerns/urgency.rb
colinxfleming: def!

app/models/concerns/call_listable.rb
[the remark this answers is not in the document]
DarthHater: They certainly are confusing, but changing method signatures is also confusing for code :)

Fine with either approach ultimately, but it's something that came to mind. When changing API's you generally take the least painful approach, or create a new method signature and deprecate the old one. The points mostly moot in this case, but you get what I'm after, I think.

app/models/concerns/urgency.rb
DarthHater: I think that'd be better, especially if it's doing something like && or ||, that way it's easier to read. Looking at it right now I wouldn't be able to tell that right off the bat.

PR #1444, opened 2018-06-30
Turn oauth test back on to see if it chokes

### LINKED ISSUE #1311: Make google sso integration test work with systemtest

Something about the `mock_oauth` function gets weird when it runs after a systemtest has started. That is bad news.

Testing SSO, a major component of our app and a notable part of our security best practices.

File a ticket with whoever wrote the `mock_omniauth` we're using? Put this test in its own container when running in CI?

* **Anything else?**
Stems from #1309

colinxfleming:

Does it blend?

* Turns oauth test back on

* Fixes #1311


colinxfleming:
Much to my surprise this didn't choke even after being rerun? I'm going to try reracking the knapsack segmentation to see if it still gets weird after three runs. if it doesn't, I'll merge this.

colinxfleming:
I'll take it! Invoking godmode because of tests.

PR #1445, opened 2018-07-07
Start bringing in Clinic Finder

DarthHater:

This starts to bring in the clinic finder gem, so that we can just manage that code as a part of DCAF

* Brings in files from gem that are needed
* Removes clinic_finder gem from Gemfile
* Party


colinxfleming:
Made some tweaks to get it into Pure Rails, tested locally and it seems to be working right. I have a plan for next steps. Punting on some of the test mocking in the name of expediency.

PR #1451, opened 2018-07-16
Budget bar tweak for deploy: comment out the Remaining figure

colinxfleming:

I think at first it will be more trouble than it's worth to set this up, until we have a config that does it, so commenting it out for now.

* Comment out the remaining div in budget bar


colinxfleming:
invoking godmode for this one.

PR #1468, opened 2018-08-16
expand exportable to include pledge data - first pass

mdworken:
This PR adds exportable info to csv export!

I've manually tested it a fair bit so far, and will add automated tests when I do the full data test.

@colinxfleming do you know when we might next be updating DARIA? This PR would definitely help for Data Team stuff.

### DIFF
--- a/app/models/concerns/exportable.rb
+++ b/app/models/concerns/exportable.rb
@@ -60,10 +60,9 @@ module Exportable
     # External Pledges
     "External Pledge Count" => :external_pledge_count,
     "External Pledges Sum" => :external_pledge_sum,
-    "Highest External Pledge Source" => :highest_external_pledge_source,
-    "Highest External Pledge Amount" => :highest_external_pledge_amount,
-    "Lowest External Pledge Source" => :lowest_external_pledge_source,
-    "Lowest External Pledge Amount" => :lowest_external_pledge_amount
+    "All External Pledges" => :all_external_pledges
+    
+    
     # TODO test to confirm that specific blacklisted fields aren't being exported
   }.freeze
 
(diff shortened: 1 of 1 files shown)


app/models/concerns/exportable.rb
```diff
@@ -56,7 +56,14 @@ module Exportable
     "Notes Count" => :notes_count,
 
     # TODO clinic stuff
-    # TODO external pledges
+
+    # External Pledges
+    "External Pledge Count" => :external_pledge_count,
```
colinxfleming: Since there's no PII here, what are your thoughts about taking a slightly different tack and just raw-dumping everything into a single concat'd column? Something like

```
def external_pledges
  external_pledges.map { |x| "#{x.source} - #{x.amount}" }.join('; ')
end
```

and then parse the raw result in whatever py script? This kind of approach might get clunky for calls since we're dealing with quite a few potentially, but for cases like this where almost all will have 0-2 it might work pretty cleanly. Thoughts?

colinxfleming:
@mdworken we usually ship sunday nights if there are outstanding changes but if this is blocking active work with a deadline then we can do a weekday deploy mid next wk if it's ready.

app/models/concerns/exportable.rb
mdworken: I actually like that idea a lot in general, but how would you feel about having pledge count, pledge total, and then the raw result? It would make my life slightly easier

colinxfleming: I'm good with that

mdworken: Updated to reflect decision!

mdworken:
@colinxfleming can I get another review on this?

app/models/concerns/exportable.rb
```diff
@@ -147,6 +153,27 @@ def get_field_value_for_serialization(field)
...
+  
+  def external_pledge_count
+    external_pledges.count
+  end
+  
+  def external_pledge_sum
+    sum = 0
+    all_pledges = external_pledges.all
+    
+    all_pledges.each do |pledge|
```
colinxfleming: An FYI -- I'll merge this as is, but next time you run into this you can do this immutably with `reduce`/`inject`, which is good for doing this with larger datasets. 

Something like:

```
def external_pledge_sum
  all_pledges.reduce { |accumulator = 0, pledge| pledge.try(:amount) || nil }
end
```

colinxfleming:
Looks good, thanks for working this out! Very glad that this change is being made and totally my mistake for the oversight.

PR #1471, opened 2018-09-14
Patch rubyzip CVE

colinxfleming:

Not something we were extremely affected by or at risk of, but nonetheless.

* `bundle update selenium-webdriver` (the parent library)

No issues


colinxfleming:
Not prod facing (we only use rubyzip as part of `selenium-webdriver`) so invoking godmode and merging.

PR #1475, opened 2018-09-23
Use patient timestamp for archiving rules

### LINKED ISSUE #1474: Web fulfillment form doesn't bump fulfillment timestamps, which is screwing up archiving
@verbingthenoun found a bug / surprise behavior with the mongo timestamp library where the parent object gets updated instead of the subobject. Steps to recreate below:

The archiver is specifically looking for the timestamp on the fulfillment to see if it is less than or equal to 120 days ago (e.g. have 120 days passed since the fulfillment was modified), OR cases where the initial call was less than or equal to a year ago (e.g. has a year passed since we logged their initial call). What we can do here is update a fulfilled patient from the UI to see if that changes the fulfillment timestamp.

From a fresh seed, let's look this up

irb(main):002:0> Patient.find_by(name: 'Reporting Patient 0').fulfillment.updated_at
=> Sun, 23 Sep 2018 11:11:03 EDT -04:00  # this is the time I ran the seed

I added a check number to this patient through the UI at about 11:19. When I run that command again:

irb(main):003:0> Patient.find_by(name: 'Reporting Patient 0').fulfillment.updated_at
=> Sun, 23 Sep 2018 11:11:03 EDT -04:00

No updated timestamp. But it looks like that DID update the patient object.

irb(main):004:0> Patient.find_by(name: 'Reporting Patient 0').updated_at
=> Sun, 23 Sep 2018 11:19:40 EDT -04:00

Action to take is to swap this to patient updated at.

colinxfleming:

People were getting archived early because the timestamp library worked differently than we thought it did (web forms don't update). This changes the archiving rules to reflect reality.

* use patient updated timestamp for archiving instead of fulfillment updated timestamp

* Fixes #1474

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -289,7 +289,7 @@ class Patient
   end
 
   def self.fulfilled_on_or_before(datetime)
-    Patient.where( 'fulfillment.fulfilled'=> true,
-      'fulfillment.updated_at' => { '$lte' => datetime } )
+    Patient.where('fulfillment.fulfilled' => true,
+                  updated_at: { '$lte' => datetime })
   end
 end
--- a/test/models/archived_patient_test.rb
+++ b/test/models/archived_patient_test.rb
@@ -109,38 +109,27 @@ class ArchivedPatientTest < ActiveSupport::TestCase
   describe 'archive_fulfilled_patients' do
     before do
       @patient_fulfilled = create :patient, primary_phone: '222-222-3334',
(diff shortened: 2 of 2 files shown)


colinxfleming:
@lomky can you gut check this? This is changing the behavior a little bit from 'no fulfillment action for 120 days' to 'patient record is dormant for 120 days', which I think is PROBABLY good, but I do want to check in on the founder's intent here. I think it does get @verbingthenoun out of the trouble spot.

PR #1477, opened 2018-09-23
Clear call lists based on actual last sign in instead of second to last sign in

### LINKED ISSUE #1476: Nightly cleanup of call list misfiring due to sign in times
CM Arona reported that after logging in for the first time in a few months, her call list had been mysteriously cleared the next morning.

I had a hypothesis here that we had a bug on setting last sign in timestamps, but that doesn't appear to be the case as Arona's last sign in is this morning in the db.

However, the 'it had been a few months' was a useful clue. It looks like we've set this rule according to last sign in timestamp, but there's also a current sign in timestamp which seems to reflect the actual most recent log in. Looking at a few people who haven't logged in in awhile, they also have current sign in timestamps from weeks or months ago.

Solution here is to do the nightly call list cleanup based on a max of either last login or current login timestamps.

colinxfleming:

See #1476 for an explanation, but in a nutshell current sign in at gets populated slightly differently than we thought it did.

* take the max of current sign in at and last sign in at when deciding whehter or not to clear a call list

* Fixes #1476

### DIFF
--- a/app/models/concerns/call_listable.rb
+++ b/app/models/concerns/call_listable.rb
@@ -55,7 +55,7 @@ module CallListable
        current_sign_in_at < Time.zone.now - TIME_BEFORE_INACTIVE
       clear_call_list
     else
-      patients.each { patients.delete(p) if recently_reached_by_user?(p) }
+      patients.each { |p| patients.delete(p) if recently_reached_by_user?(p) }
     end
   end
 
--- a/app/models/concerns/call_listable.rb
+++ b/app/models/concerns/call_listable.rb
@@ -50,9 +50,10 @@ module CallListable
   TIME_BEFORE_INACTIVE = 2.weeks
 
   def clean_call_list_between_shifts
-    if last_sign_in_at.present? &&
-       last_sign_in_at < Time.zone.now - TIME_BEFORE_INACTIVE
-      patients.clear
+    last_activity = [last_sign_in_at, current_sign_in_at].reject(&:nil?).max
+    if last_activity.present? &&
+       last_activity < Time.zone.now - TIME_BEFORE_INACTIVE
+      clear_call_list
     else
       patients.each do |p|
         # TODO: reexamine this behavior in awhile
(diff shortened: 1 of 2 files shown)


app/models/concerns/call_listable.rb
```diff
@@ -50,9 +50,10 @@ def ordered_patients(line)
...
-    if last_sign_in_at.present? &&
-       last_sign_in_at < Time.zone.now - TIME_BEFORE_INACTIVE
-      patients.clear
+    last_activity = [last_sign_in_at, current_sign_in_at].reject(&:nil?).max
+    if last_activity.present? &&
+       last_activity < Time.zone.now - TIME_BEFORE_INACTIVE
+      clear_call_list
     else
       patients.each do |p|
         # TODO: reexamine this behavior in awhile
```
lomky: Is now a good time to do this TODO?

app/models/concerns/call_listable.rb
```diff
@@ -50,9 +50,10 @@ def ordered_patients(line)
   TIME_BEFORE_INACTIVE = 2.weeks
 
   def clean_call_list_between_shifts
-    if last_sign_in_at.present? &&
-       last_sign_in_at < Time.zone.now - TIME_BEFORE_INACTIVE
-      patients.clear
+    last_activity = [last_sign_in_at, current_sign_in_at].reject(&:nil?).max
```
lomky: From chat:
> It turns out that `current_sign_in_at` is the most recent sign in timestamp, and `last_sign_in_at` is the timestamp before that
> So on login, `current_sign_in_at` becomes now, and `last_sign_in_at` becomes whatever `current_sign_in_at` was

Can we just look at `current_sign_in_at`?

lomky: Also, since these field names are unintuitive and we can't change them since they're devise, maybe a brief explainer comment?

app/models/concerns/call_listable.rb
colinxfleming: good call! I think this behavior is meeting our needs perfectly well so let's lock it in.

PR #1480, opened 2018-09-29
Library upgrades

colinxfleming:

a bunch of `bundle update` commands

* Upgrade kaminari and kaminari-mongoid, devise, turbolinks

no issues, just general health


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'jquery-ui-rails'
 # Our database is MongoDB
 gem 'mongoid', '>= 6.2.0', '< 7'
 gem 'bson_ext'
-gem 'mongoid-history', '0.6.1' # gives us object history
+gem 'mongoid-history', '< 1.0' # gives us object history
```
DarthHater: There isn't a version higher than `0.8.1` for this library, so you might need this level of safety on the semver. Nothing wrong with what you did, just figured I'd point that out.

DarthHater:
A lot of these bumps are a minor version bump up (not patch), I'm taking it you didn't run into any issues. I always get a bit nervous on devise since there are a lot of libs working with it.

PR #1483, opened 2018-09-29
Update factory-bot-rails

colinxfleming:

`factory_bot_rails` is adjusting how they set keys and asking that everything come through a block now. This upgrades the library and squelches the warnings.

* `bundle update factory_bot_rails`
* edit factories to reflect the change

no issues


colinxfleming:
test facing only so invoking godmode and smashing

PR #1484, opened 2018-09-30
Clinic finder: Cache clinic lat/lng on the object to avoid calling every time

### LINKED ISSUE #1355: Save geo information for clinics

The clinic finder lookup is slow and doesn't cache the lat/long. We should persist clinic l/l in db so we're only doing a lat/long lookup for the patient.

Making clinic finder not take 30 seconds every time to do 60 clinic lookups

Either a clinic finder change or a rake task

* **Anything else?**
nope!

colinxfleming:

A major thing making clinic finder hard to use is that, as it's set up now, it's making 1 + (clinics.count) API calls to google's geocoder API, which a) slows it down to the point of unusability and b) makes us probably hit the free api limit. To get around this, this PR stashes lat/lng on the clinic object, geocodes the patient zip thru an API call, and then calculates the distance without making an api call.

* `Clinic#coordinates` is a field, and it updates on change to an address field
* Adds a rake task to refresh all clinic coordinates
* Moves the logic to find a clinic's lat/lng into the clinic object

No view change.

* Fixes #1355


app/services/clinic_finder/locator.rb
```diff
@@ -60,18 +58,6 @@ def locate_cheapest_clinic(gestational_age: 999, limit: 5)
 
     private
 
-    # Set up mutable clinic structs.
-    def build_clinic_structs
```
lomky: I see this got demoted from a function up into a single line at 34 - can you talk me through how the one morphed into the other?

app/services/clinic_finder/locator.rb
```diff
@@ -85,27 +71,9 @@ def filter_by_params(clinics, gestational_age, naf_only, medicaid_only)
 
     def set_geocoder
```
lomky: What exactly is being returned by this function if `GOOGLE_GEO_API_KEY` isn't set? Just an empty object?

app/services/clinic_finder/locator.rb
```diff
@@ -85,27 +71,9 @@ def filter_by_params(clinics, gestational_age, naf_only, medicaid_only)
...
-    #   sorted_clinics
-    # end
-
-    # def three_cheapest(data)
-    #   data.sort_by { |name, information| information[@gestational_tier] }.first(3)
-    # end
-
-    # def available_clinics
-    #   @clinics.keep_if { |name, information| information[@gestational_tier] && @helper.within_gestational_limit?(information['gestational_limit']) }
-    # end
```
lomky: +:100: for removing dead code :tada:

app/services/clinic_finder/modules/geocoder.rb
```diff
@@ -6,49 +6,17 @@ module Modules
     # Nesting due to rails autoload
     module Geocoder
       # Get a patient's latitude and longitude.
-      def get_patient_coordinates_from_zip(zipcode)
+      def patient_coordinates_from_zip(zipcode)
         patient_location = @geocoder.geocode zipcode
-        @patient_context.location = patient_location.ll
+        patient_location.ll
```
lomky: I understand it might be out of scope - but I really hate `.ll` - its really short and its made out of indistinct letters. Could we maybe change it? :grinning:

app/services/clinic_finder/locator.rb
colinxfleming: yeah! It didn't used to be a safe assumption that we'd be loading ActiveRecord-ish objects (which respond to `attributes`) into this, so we'd have to handle the case of passing along OpenStructs. Since we've collapsed this into our rails app, it's now a pretty safe assumption, so we can simplify this handling down.

app/services/clinic_finder/modules/geocoder.rb
colinxfleming: haha I hear you. It looks like this works if instead of `ll` we do `[location.lat, location.lng]` which is a little more verbose but seems to blend okay. I'll swap that out where it's used.

app/services/clinic_finder/locator.rb
colinxfleming: It returns a `Geokit::Geocoders::GoogleGeocoder` without an API key set. The class is nice and malleable and will let you init it without a key, and (I think) google lets you have like one or two free calls per IP before it's like 'seriously, get a token pls'. But it's esp nice for us because it'll respond to everything we expect it to regardless of whether there's a key set.

PR #1488, opened 2018-10-10
remove gon gem

colinxfleming:

We were using gon on the old reports, but we aren't anymore, so this strips it in favor of render_async

tech debt!


colinxfleming:
dumping in favor of #1528

PR #1496, opened 2018-10-20
squelch a pair of cves via a ruby upgrade

colinxfleming:

ruby 2.5.1 => ~2.5.2~ 2.5.3

* in the circle config!
* in rubocop!
* in the app!


colinxfleming:
invoking godmode to fix the build

PR #1501, opened 2018-10-20
Fix add patient in activity log

### LINKED ISSUE #1494: Add to call list in event log doesn't work

The event log has a link to 'add to call list'. For some reason, it doesn't work! We've heard from several sources that this would be extremely helpful; let's do it.

CMs refer to this all the time and this is adding a somewhat obnoxious step to their workflow.

* Make the 'add to call list' link in the event log work
* Add a systemtest for it

* **Anything else?**

stems from #1465

tingaloo:

Removed the current user parameter, it messed with the route.

* Fix and test

* Fixes #1494


app/helpers/events_helper.rb
```diff
@@ -17,7 +17,7 @@ def log_entry(event)
     )
     add_link = link_to(
       '(Add to call list)',
-      add_patient_path(current_user, event.patient_id),
+      add_patient_path(event.patient_id),
```
colinxfleming: haha I had a feeling it was something like this. Thanks much, bud 💯

test/system/add_patient_test.rb
```diff
@@ -0,0 +1,43 @@
...
+      wait_for_element 'I left a voicemail for the patient'
+      click_link 'I left a voicemail for the patient'
+      wait_for_ajax
+      log_out && log_in_as(@user2)
+      wait_for_css '#activity_log_content'
+      wait_for_css '#event-item'
+      wait_for_ajax
+      wait_for_no_css '.sk-spinner'
+
+      within :css, '#activity_log_content' do
```
colinxfleming: This is a good test and the use of `within` is a great call. we can simplify it a little bit using `assert_difference`. Something like:

```rb
assert_difference '@user2.patients.count', 1 do
  within :css, ...
    assert has_content? ...
    click_link ...
    wait_for_ajax
  end
end
```

`assert_difference` is a nice little rails-ism. docs here: https://api.rubyonrails.org/v4.1/classes/ActiveSupport/Testing/Assertions.html

test/system/add_patient_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'application_system_test_case'
+
+class AddPatientTest < ApplicationSystemTestCase
```
colinxfleming: can we rename this test to reflect what it's doing a little better? maybe `EventLogInteractionTest` (and a renamed file to match)? `AddPatient` sounds like it's doing a patient create to me. Thoughts?

test/system/add_patient_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  before do
+    @user = create :user, role: 'admin', name: 'Billy'
+    @user2 = create :user, role: 'cm', name: 'Susie'
+
+    @patient = create :patient, name: 'tester',
```
colinxfleming: So events are created as a byproduct of other actions automatically, and I believe we're testing that behavior in the controller/model tests. I think we can slim this test down a little bit by doing the following after this patient create:

```
@patient.calls.create status: 'Left voicemail'
```

which should create a call via an unrelated stub user, and thenyou can ditch the part in L19-25 where you create a call, log out, and log back in. Might shave off some time on the test, and hone in on the specific behavior we want to test. thoughts?

tingaloo: I tried a few variations but the activity log didn't update after using that snippet. But that would be awesome to cut down on this chromedriver stuff.

colinxfleming:
hey @tingaloo can you bring this up to speed when you get a minute? would like to get it out in the next deploy if possible. thank you! 💯

PR #1502, opened 2018-10-20
Add textable option to voicemail preferences

### LINKED ISSUE #1498: Add indicator to tell if someone's phone is textable

BRAF texts. That's pretty sweet. A lot of funds are moving in this direction (I'm not sure if BAF texts or not). We should account for that

Better patient service

* Add a boolean field to patient: `textable`
* add it to `Patient Information`, under `Voicemail Preference`
* Bump the `update_patient_info` systemtest to confirm it works

* **Anything else?**

stems from #1465 

@alisonnjones do we need to worry about the DCAF case here? Should we hide it from DCAF/BRAF?

tingaloo:

Add text option to voicemail preferences.

screen shot 2018-10-20 at 14 04 48

* Add text field to voicemail options

* Fixes #1498

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -111,7 +111,7 @@ class PatientsController < ApplicationController
 
   PATIENT_INFORMATION_PARAMS = [
     :line, :age, :race_ethnicity, :language,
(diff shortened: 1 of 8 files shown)


colinxfleming:
we're gtg on this via this and so marking closed. Thank you very much for getting this cooking @tingaloo !

PR #1503, opened 2018-10-20
Add referred_by to config

### LINKED ISSUE #1500: Add additional referred by fields to config

While we have a pretty solid list of `referred_by` fields, we don't cover every case, and BRAF is interested in reshaping this list. So we should make it fund-configurable.

Better tracking of who a fund is working with and where their patients are coming from.

* Add it as an option in `app/models/config.rb`
* Have that select dropdown in `patient_information` use the stock configs + whatever else they're adding

This is pretty similar to what we're doing with Language, IIRC, so you can probably adopt a lot of it from that.

* **Anything else?**

Stems from #1465

tingaloo:

Vacate the code.

* Add referred_by to config

* Fixes #1500


app/helpers/patients_helper.rb
```diff
@@ -19,6 +19,11 @@ def language_options
     standard_options + Config.find_or_create_by(config_key: 'language').options
   end
 
+  def referred_by_options
+    standard_options = [[nil]]
```
colinxfleming: Let's port over the existing list as our standard options, and let funds staple on an additional set if they want via the config.

```suggestion
    standard_options = [nil, 'Clinic', 'Crime victim advocacy center',
     "#{FUND} website or social media",
     'Domestic violence crisis/intervention org', 'Family member', 'Friend',
     'Google/Web search', 'Homeless shelter', 'Legal clinic', 'NAF', 'NNAF',
     'Other abortion fund', 'Previous patient', 'School',
     'Sexual assault crisis org', 'Youth outreach']
```

vraravam:
Shouldn't this also have a UI change where the user can set this in the admin section?

colinxfleming:
I spruced this up some. @lomky or @DarthHater a review at yr leisure pls!

colinxfleming:
The UI thing is an A++ flag. rep'd that in #1559

Issue #1495, opened 2018-10-20
Set fax link in footer to an app config instead of an env var


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Our footer: 

![image](https://user-images.githubusercontent.com/3866868/47257375-c9a55180-d45a-11e8-924a-d9cf87bd4cb1.png)

It's nice, right? But sometimes people switch fax services, and we don't want to be on the hook for making a code change.

* **What feature or behavior is this required for?**

Getting engineers out of processes that should live with the funds themselves.

* **How could we solve this issue? (Not knowing is okay!)**

We have a working model for how to add this kind of app config in `app/models/config.rb`. We should:

* add fax service as an admin config option
* replace that footer link to either be nothing if the config isn't set or the config if so
* Figure out a way to accommodate both the name and the URL in one config

* **Anything else?**

stems from #1465

vraravam:
I'll sign up for this

colinxfleming:
hey @vraravam , I think @tingaloo  beat you to it, but could I point you toward #1499 instead?

vraravam:
@colinxfleming : @tingaloo had asked me to take over the PR, but as I had mentioned, I would much rather start afresh (attributing to him as well). Hope that is ok with you. Please ack before I can start the work.

colinxfleming:
resolved

PR #1504, opened 2018-10-20
Config change for fax

tingaloo:

I don't really like the design pattern I used here, primarily because there's both a service and url portion. Config is a little bloated, could use some thinning.

* Fax env var to config.

* Bumps #1495


app/views/layouts/_footer.html.erb
```diff
@@ -6,11 +6,12 @@
 		</div>
 	</div><!-- end row -->
 </div><!-- end container -->
-	  
+
 <div id="app_footer" class="container text-center margin-bottom-sm">
   <a href="http://<%= FUND_DOMAIN %>/"><%= FUND_FULL %></a> -- <a href="tel:<%= FUND_PHONE %>"><span class="glyphicon glyphicon-earphone" aria-hidden="true"></span> <%= FUND_PHONE %></a><br>
 	<a href="https://prochoice.org/">National Abortion Federation</a> -- <a href="tel:800-772-9100"><span class="glyphicon glyphicon-earphone" aria-hidden="true"></span> 800-772-9100</a><br>
-	<%= link_to FUND_FAX_SERVICE, "http://#{FUND_FAX_SERVICE}", target: '_blank', rel: 'noopener nofollow' %>
+	<a href="http://<%= fax_service %>/"><%= fax_service %></a> -- <a href="tel:<%= fax_number %>"><span class="glyphicon glyphicon-earphone" aria-hidden="true"></span> <%= fax_number %></a><br>
```
colinxfleming: Rails has a pretty sick helper for cleaning up this kind of thing - https://api.rubyonrails.org/v5.2.1/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

So we can effectively replace this with `link_to nil, fax_service, target:'_blank', rel: 'noopener nofollow'` and it'll save us having to do http etc.

```suggestion
	<%= link_to nil, fax_service, target: '_blank', rel: 'noopener nofollow' %>
```

app/models/config.rb
```diff
@@ -41,6 +41,14 @@ def options
     config_value[:options]
   end
 
+  def fax_service
+    config_value[:service]
+  end
+
+  def fax_number
```
colinxfleming: You know, I think you're absolutely right on how the fax number here makes things much more complex. Let's ditch it and just go with fax service, and ask them to make that a url like `http://whatever.com`. Hopefully that cuts down on a little bit of the complexity/ugmo-ness. That way we can make it handle more like `resources_url`. thoughts?

tingaloo:
@vraravam Hey, i'd like to hand this issue over to you if you'd like. I don't have as much domain knowledge on the performance aspects.

vraravam:
@tingaloo - I'm going to create a new PR with all of my changes - will attribute some of the work to you (but, I dont feel comfortable committing on top of yours due to lots of differences in opinion).

colinxfleming:
I think a takeaway I have is that our config model is starting to break a little bit. We were previously only really using it for drop down options, and it doesn't apply as well to singleshot stuff like this. Not something we have to immediately resolve but it's def a little too magical in a weird way right now.

PR #1506, opened 2018-10-22
Changed the label of the 'Referred by' dropdown #fixes-1499

### LINKED ISSUE #1499: Change label of `Referred by` to `How they heard about FUND`

We have a little bit of confusing

Keeping language not-confusing helps keep conversations flowing and prevents CMs from accidentally asking awkward questions, which is important to making sure everyone's feeling as comfortable as possible.

* Pop into `app/views/patients/_patient_information.html.erb` and add a label to the `referred_by` inputwith the text: `How patient heard about #{FUND}`. (`FUND` is a global constant we have set up that aliases to whatever the fund's abbreviation is.)
* You may also have to update a test in `test/system/update_patient_info.rb` or whatever it's called

* **Anything else?**

nope!

vraravam:

* Changed the label as per the issue #1499 

screen shot 2018-10-23 at 5 18 58 am

* Fixes #1499

### DIFF
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -84,6 +84,7 @@
           <%= f.select :referred_by,
                        options_for_select(referred_by_options,
                                           patient.referred_by),
+                                          label: "How patient heard about #{FUND}",
                                           autocomplete: 'off' %>
 
 


colinxfleming:
aw yus. thank you very much @vraravam !

the failure is related to some circleci stuff. there might be an easily correctable systemtest failure but I'm not super worried about it and we can fix it in post if necessary.

PR #1507, opened 2018-10-23
Updating Docs for Running tests locally by contributors

vraravam:

As a new contributor to this codebase, I had some difficulty in reading through the docs (links were referring to whole sentences(!) and also did not find any docs that outlined how to run tests locally before submitting PRs. This PR adds such documentation for future contributors. The content is taken from the `circle-ci` configuration file.

* Cleaned up README a little bit
* Documentation on how to run tests locally by contributors

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Feel free to hit any of us up with questions about the project, we're nice!
 * Our [list of issues in Github](https://github.com/colinxfleming/dcaf_case_management/issues) is our project's remaining TODO list
 * Post in an issue when you're starting work on something, so @colinxfleming can keep track of it and so we don't duplicate work
 * We <3 new people and beginners
+* We also would like to keep the codebase in a good state for all other contributors, and so have a few steps that would like contributors to [adhere](docs/TESTING.md) to before submitting PRs.
 
 ### If this is your first time, a good way to get oriented is to leave our users a nice note! Check out instructions [here](docs/YOUR_FIRST_CONTRIBUTION.md).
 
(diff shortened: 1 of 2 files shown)


colinxfleming:
hey @vraravam ! thank you very much for this thoughtful PR -- some of my favorite contributions to this project have been doc fixes because they encourage us to revisit assumptions the team has made.

That said, I'm going to close this without merging, and here's why. The central change here is guidance around testing, and I think we want to continue to play fast and loose on testing for the time being.

Contributors come to this project from all sorts of skill levels, from maintainers of the ruby ecosystem to people who are making their first public commits, and everywhere in between. As part of that, we do what we can to make sure a lot of stuff which might seem esoteric to new developers gets picked up by CircleCI.

Some of our contributors are super comfortable running tests, making atomic commits, etc. But for some, it's an insurmountable hurdle. We're all volunteers on this project, and my experience working with volunteers (and being a volunteer!) is that nothing is more of a bummer than a problem you don't have the conceptual tools to figure out on your own. At best you burn a lot of time trying to follow directions you're don't quite follow, at worst you give up and we never hear from you again. As it is, we do this all at once during the CI/PR review process, which is still a decent hurdle but lets us give some human-to-human guidance on the sooner side.

Thank you again for getting the leadership team's brains churning about this! Voices like this keep our minds moving, and it was definitely time to consider something like this, but I think we should maintain our recipe for now.

vraravam:
@colinxfleming - thanks for the detailed explanation.

But, I would like to respectfully disagree - Testing is something that seems intrinsic to this codebase. If not, then why do show the `CircleCi` badge that is currently green/passing - and also why hook up any CI tool in the first place for the codebase?

All that my PR does is to expose the same shell commands, in easily readable/understandable documentation format so that new contributors need not know the location/syntax of the `circleci` tooling to find out what to run before submitting a PR. I am sure every contributor would like to start with a green state and end on one as well - because that is the basic reference plane to give them succor that they have not broken anything for the next contributor.

As one of the gatekeepers for this codebase, its your right and responsibility to accept any PR or reject - so, that is not something I wanted to debate on - just wanted to state that the statement "we want to continue to play fast and loose on testing for the time being." seems to not be supported by the codebase itself.

PR #1509, opened 2018-10-24
#1420 add audited field to patient w validations and checks

### LINKED ISSUE #1420: Add Audited Field to Patient

Adding a boolean field to Patients that denotes whether the patient fulfillment has been audited.

Avoiding archiving fulfilled patients until they have been audited, up until a certain point.

Add the field and tie archiving into it. Add a failsafe to archive regardless after X days.

* **Anything else?**

We need to hash out what the timeline looks like for archiving fulfilled patients with audit and without. 
We strongly want to err towards code that supports best practices.

Tagging in @alisonnjones and @colinxfleming

rudietuesdays:

Related to #1420 , I added an `audited` boolean field to Patients, which translates to a check box on the front end for the case manager to check if the Patient has been audited. As per the issue, I also added a check to see if the Patient was created more than two years ago (chosen arbitrarily; let me know if I should change!), and if so, they will be archived regardless of if they've been audited.

screen shot 2018-10-23 at 8 42 54 pm

I haven't worked with Mongo very much, so plz let me know if I've jacked this up!

* Fixes #1420

### DIFF
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -7,6 +7,8 @@ class ArchivedPatient
   # Concerns
   include Exportable
 
+  before_save :check_if_audited
+
   # Relationships
   belongs_to :clinic
   embeds_one :fulfillment
@@ -142,5 +144,12 @@ class ArchivedPatient
 
     archived_patient.save!
     archived_patient
+  end
+end
+
+# check if a patient has been audited; allow them to be archived if they were NOT audited but were created more than two years ago
+def check_if_audited
+  unless self.audited
+    self.created_at > 2.years.ago ? true : false
   end
 end
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -21,7 +21,7 @@
                                           patient.race_ethnicity),
                                           label: 'Race / Ethnicity',
                                           autocomplete: 'off' %>
-                                          
+
           <%= f.form_group :language do %>
             <%= f.select :language,
                        options_for_select(language_options,
@@ -109,6 +109,8 @@
                               checked: patient.special_circumstances.include?('Prison'), namespace: 'Prison' }, 'Prison', '' %>
             </div>
           <% end %>
+
+          <%= f.check_box :audited, label: 'Has patient fulfillment been audited?' %>
         </div>
       </div>
     </div>
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -94,6 +94,7 @@ class Patient
   field :resolved_without_fund, type: Boolean
   field :pledge_generated_at, type: Time
   field :pledge_sent_at, type: Time
+  field :audited, type: Boolean, default: false
 
   # Indices
   index({ primary_phone: 1 }, unique: true)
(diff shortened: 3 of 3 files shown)


app/models/archived_patient.rb
```diff
@@ -7,6 +7,8 @@ class ArchivedPatient
   # Concerns
   include Exportable
 
+  before_save :check_if_audited
```
lomky: I'm a little worried having this hook in as a before action when it's being run in an automated fashion - if the Archive patient create fails will the patient still delete itself, or will the processing cut out?

lomky: I think I'd prefer to take the awesome new field, and rather than invoke the process in archive save, invoke it as a third run.

Right now we run:
 - Archive Dropoff Patients
 - Archive Fulfilled Patients

But we could turn that into:

 - Archive Dropoff Patients
 - Archive _Audited_ Fulfilled Patients
 - _Archive TimedOut Fulfilled Patients_

This would involve adjusting the function [Patient:archive_fulfilled_patients](https://github.com/DCAFEngineering/dcaf_case_management/pull/1509/files#diff-d213d69335ff3bc33b8ea8545f4c4459R84) to use our time out date, and then create another similar one that grabs 120 day old fulfilled _and audited_ patients to process.

lomky: @colinxfleming lemme know here - am I being too paranoid about the save hook?

app/models/archived_patient.rb
```diff
@@ -144,3 +146,10 @@ def self.convert_patient(patient)
     archived_patient
   end
 end
+
+# check if a patient has been audited; allow them to be archived if they were NOT audited but were created more than two years ago
+def check_if_audited
+  unless self.audited
+    self.created_at > 2.years.ago ? true : false
```
lomky: The date we want depends on how long a window they want for the auditing. @colinxfleming do you remember how long that was, or who we should ping to get it? My hunch is to check for the self.fulfillment date being six months ago.

app/views/patients/_patient_information.html.erb
```diff
@@ -109,8 +109,10 @@
                               checked: patient.special_circumstances.include?('Prison'), namespace: 'Prison' }, 'Prison', '' %>
             </div>
           <% end %>
+
+          <%= f.check_box :audited, label: 'Has patient fulfillment been audited?' %>
```
lomky: :heavy_check_mark:

app/models/archived_patient.rb
colinxfleming: @lomky I think I would favor your approach of doing them in isolated batches, yeah -- adds a layer of safety and is a little more explicit about this action. Can you work with @rudietuesdays to implement that?

app/models/archived_patient.rb
colinxfleming: It's currently 120 days. Let's ask @alisonnjones or @verbingthenoun  how long that window should be in practice.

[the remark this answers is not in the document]
colinxfleming: my possibly worst trait as a programmer is never remembering which way to point the greater than less than arrow, haha. But yeah, I have a hard time following the hook tbh. I would suggest spinning up a method like `archive_fulfilled_patients` and adjusting the activerecord query.

lomky:
@rudietuesdays & I pulled this work into #1524

colinxfleming:
Closing in favor of #1524 . thank you @rudietuesdays for the initial big push on this and kicking off the remix!

PR #1510, opened 2018-10-24
remove user searches from patient searchable

### LINKED ISSUE #1516: Tidy up dead functions from searchable

remove dead code

* **Anything else?**

astutely identified by #1510

ChrisBarthol:

I was looking through the test coverage and noticed that these two methods were never being called.  They are user search methods in the patient searchable module and probably dont belong.

* Remove unused private search methods that are not used.

* Fixes #1516 
* Bumps #Y

### DIFF
--- a/app/models/concerns/patient_searchable.rb
+++ b/app/models/concerns/patient_searchable.rb
@@ -22,25 +22,6 @@ module PatientSearchable
 
     private
 
-    def find_name_matches_user(name_regexp)
-      if nonempty_regexp? name_regexp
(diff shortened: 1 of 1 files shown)


colinxfleming:
thanks very much for this pr, @ChrisBarthol ! I'm a bit tied up with work rn but will look at this in the next day or two.

colinxfleming:
@ChrisBarthol thanks again for this! I've created a hacktoberfest issue for it so you get appropriate credit. thanks much for identifying this and cleaning it up!

PR #1518, opened 2018-10-29
Archive time

Kevin-Wei:
I grepped around and found that we have something that archives people after a year so that's what I went with. Let me know if I need to change the underlying logic at all


lomky:
Thanks! I've pulled this work into the pr #1524

colinxfleming:
Closing since this is pulled into #1524

Issue #1523, opened 2018-11-12
Revamp Archive to give logic to Patient object


lomky:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Surface archive issues earlier by giving over archive calculations to hte patient object.

* **What feature or behavior is this required for?**
Debugging issues with archive.

* **How could we solve this issue? (Not knowing is okay!)**
Move logic to patient creation, remove fulfillment-based early archiving.

* **Anything else?**

colinxfleming:
resolved via #1524

Issue #1485, opened 2018-10-09
Update "The Data Collection Practices" Post-Archive


lomky:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Update The Data Collection Practices doc's last section with finalized Archive practices!

https://github.com/DCAFEngineering/dcaf_case_management/blob/master/docs/OUR_DATA_COLLECTION_PRACTICES.md

* **What feature or behavior is this required for?**

Best Practice

* **How could we solve this issue? (Not knowing is okay!)**
Write it up!

* **Anything else?**

IOU a first pass. @colinxfleming Who else should we tag in for good verbiage here?

PR #1524, opened 2018-11-12
Archive restruct

lomky:

- Pulls in Auditing code #1509 
  - Patients have an auditing checkbox under the fulfillment
    - also available through the Accountant screen
  - Patients will delay being archived unless checked.
- Pulls in Surface Archive Date code #1518 
  - Patient fulfillment partial shows the archive date, which updates when audited is selected.
- Patient objects now can calculate their `archive_date` based on initial call date & audit status.
  - the date is visible under the audit box
 - Fulfillment based early archiving removed completely. 
    - All audited patients archive 3 months after `initial_call_date`
    - All patients archive a years after `initial_call_date`, regardless or audited!
    - /Timelines might be tweaked./
 - Rake task runs archive based on checking new patient `archive_date` method
 - Updates our Data Collection documentation with actual timelines.

Archive Date & Audit Checkbox on Fulfillment:
unadited
acctaudtiedd

 * Finishes #1420
 * Finishes #1518
 * Finishes #1523
 * Finishes #1485

### DIFF
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -73,7 +73,7 @@ class ArchivedPatient
 
   # Archive & delete audited patients who called a several months ago, or any
   # from a year plus ago
-  def self.archive_todays_patients!
+  def self.archive_eligible_patients!
     Patient.all.each do |patient|
       if ( patient.archive_date < Date.today )
         ArchivedPatient.convert_patient(patient)
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -71,7 +71,7 @@ class ArchivedPatient
 
   mongoid_userstamp user_model: 'User'
 
-  # Archive & delete audited patients who called a seveeral months ago, or any
+  # Archive & delete audited patients who called a several months ago, or any
   # from a year plus ago
   def self.archive_todays_patients!
     Patient.all.each do |patient|
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -7,8 +7,6 @@ class ArchivedPatient
   # Concerns
   include Exportable
 
-  before_save :check_if_audited
-
   # Relationships
   belongs_to :clinic
   embeds_one :fulfillment
@@ -73,19 +71,14 @@ class ArchivedPatient
 
   mongoid_userstamp user_model: 'User'
 
-  # Archive & Delete patients who have fallen out of contact
-  def self.archive_dropped_off_patients!
-    Patient.where(:initial_call_date.lte => 1.year.ago).each do |patient|
-      ArchivedPatient.convert_patient(patient)
-      patient.destroy!
-    end
-  end
-
-  # Archive & Delete patients who are completely through our process
-  def self.archive_fulfilled_patients!
-    Patient.fulfilled_on_or_before(120.days.ago).each do |patient|
-      ArchivedPatient.convert_patient(patient)
-      patient.destroy!
+  # Archive & delete audited patients who called a year or more ago, or any
+  # from two plus years ago
+  def self.archive_todays_patients!
+    Patient.where(:archive_date.lte => today()).each do |patient|
+      if (patient.audited_or_obsolete?) {
+        ArchivedPatient.convert_patient(patient)
+        patient.destroy!
+      }
     end
   end
 
@@ -144,12 +137,5 @@ class ArchivedPatient
 
     archived_patient.save!
     archived_patient
-  end
-end
-
-# check if a patient has been audited; allow them to be archived if they were NOT audited but were created more than two years ago
-def check_if_audited
-  unless self.audited
-    self.created_at > 2.years.ago ? true : false
   end
 end
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -75,7 +75,7 @@ class ArchivedPatient
   # from two plus years ago
   def self.archive_todays_patients!
     Patient.where(:archive_date.lte => Date.today).each do |patient|
-      if (patient.audited_or_obsolete?)
+      if (patient.audited? or patient.obsolete?)
         ArchivedPatient.convert_patient(patient)
         patient.destroy!
       end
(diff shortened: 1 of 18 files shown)


lomky:
TODO: run a script to set archive_date on all existing patients?

lomky:
thought - should audited go on the fulfillment object and partial?

app/models/patient.rb
```diff
@@ -294,4 +302,9 @@ def self.fulfilled_on_or_before(datetime)
     Patient.where('fulfillment.fulfilled' => true,
                   updated_at: { '$lte' => datetime })
   end
+
+  def calculate_date_to_be_archived
```
colinxfleming: Since this is setting `archive_date` to initial call date plus a year, should we just make it a method a la `obsolete?` rather than saving it as a db field and setting that with a callback?  e.g.:

```
def archive_date
  initial_call_date + 1.year
end
```

app/models/patient.rb
```diff
@@ -252,6 +255,11 @@ def has_special_circumstances
     !!has_circumstance
   end
 
+  # check if a patient is long past waiting for audit
+  def obsolete?
```
colinxfleming: love this pattern!

we can simplify the code itself. `initial_call_date < 2.years.ago` should produce the same result without needing the `self` or the ternary operator.

app/models/archived_patient.rb
```diff
@@ -71,19 +71,14 @@ class ArchivedPatient
...
-      patient.destroy!
-    end
-  end
-
-  # Archive & Delete patients who are completely through our process
-  def self.archive_fulfilled_patients!
-    Patient.fulfilled_on_or_before(120.days.ago).each do |patient|
-      ArchivedPatient.convert_patient(patient)
-      patient.destroy!
+  # Archive & delete audited patients who called a year or more ago, or any
```
colinxfleming: I have a question about this as related to patients who come back to dcaf for assistance a second time -- part of the purpose of archiving is handling this tough case. we don't have to hash this out here but before we merge let's talk thru the two year standard.

app/views/patients/_fulfillment.html.erb
```diff
@@ -42,6 +42,11 @@
           <% end %>
         </div>
       </div>
+      <div class="row">
+        <div class="col-sm-12">
+          <p>Patient will be archived on <%= patient.archive_date %>
```
colinxfleming: I would suggest slapping a `display_date` on this, e.g. `patient.archive_date.display_date`. (if it doesn't work immediately then it's because I spaced on something so nevermind, haha)

app/views/patients/_patient_information.html.erb
```diff
@@ -111,6 +111,8 @@
                               checked: patient.special_circumstances.include?('Prison'), namespace: 'Prison' }, 'Prison', '' %>
             </div>
           <% end %>
+
+          <%= f.check_box :audited, label: 'Has patient fulfillment been audited?' %>
```
colinxfleming: should we move this to the fulfillment partial? I'm not totally sure we need to surface this info for non-data volunteers and admins, but interested in your thoughts.

colinxfleming: ps let's make sure to add this to the strong params in `patients_controller`

app/views/patients/_fulfillment.html.erb
lomky: didn't work - but I don't actually know why not, as it works just a bit above here

lomky: my only sadness now with this is that when you click `audited` right nearby, the date won't update until a page reload.

rudietuesdays: would it be worthwhile or overkill (or bad UX?) to add some JS code here to trigger a submit on the checkbox click, so the user doesn't need to reload the page for the change to take effect?

lomky: Now it does!

app/views/patients/update.js.erb
```diff
@@ -49,6 +49,16 @@ if($('#row-<%= @patient.id %>')) {
   );
 }
 
+// Update the archive date on audit
+$('#patient_fulfillment_audited')
```
colinxfleming: Lots of parent scanning here! Does `$('p.archive-note').text('stuff');` not cooperate?

app/models/archived_patient.rb
```diff
@@ -71,19 +71,14 @@ class ArchivedPatient
...
-  end
-
-  # Archive & Delete patients who are completely through our process
-  def self.archive_fulfilled_patients!
-    Patient.fulfilled_on_or_before(120.days.ago).each do |patient|
-      ArchivedPatient.convert_patient(patient)
-      patient.destroy!
+  # Archive & delete audited patients who called a several months ago, or any
+  # from a year plus ago
+  def self.archive_todays_patients!
```
colinxfleming: kind of nitpicky so please forgive me but what are your thoughts on naming this something like `archive_eligible_patients!` or something similar? `today` as an indicator feels like it could mean a few things to me, and I think that's a little more precise.

app/views/patients/_fulfillment.html.erb
```diff
@@ -42,6 +42,19 @@
...
+      <div class="row">
+        <div class="col-sm-12">
+          <%= f.fields_for patient.fulfillment do |pt| %>
+            <%= pt.form_group :audited do %>
+              <%= pt.check_box :audited, label: 'Fulfillment audited?', autocomplete: 'off' %>
+            <% end %>
+          <% end %>
+        </div>
+        <div class="col-sm-12">
+          <p class="archive-note">Patient will be archived on <%= patient.archive_date %>
```
colinxfleming: Probs want to close this `p` tag!

app/models/patient.rb
```diff
@@ -252,6 +252,14 @@ def has_special_circumstances
     !!has_circumstance
   end
 
+  def archive_date
+    if fulfillment.audited?
+      initial_call_date + 3.months
```
colinxfleming: also, could we add a comment here from your email explaining the logic here? copying and pasting straight from the email is fine:

```
 - If a patient fulfillment is ticked off as audited, they will archive 3 months after the patient's initial call date
    - if we're already past 3 months later when the audit happens, it will archive that night
 - if a patient is waiting for audit they will not archive until a year after their initial call date, but then they will archive regardless of audit status.
```

PR #1528, opened 2018-11-22
Remove gon gem

colinxfleming:

We were using gon on the old reports, but we aren't anymore, so this strips it in favor of render_async

tech debt!

(the old PR #1488 was a little overtaken by events, so respinning it up.)

### DIFF
--- a/app/controllers/external_pledges_controller.rb
+++ b/app/controllers/external_pledges_controller.rb
@@ -3,7 +3,7 @@ class ExternalPledgesController < ApplicationController
   before_action :find_patient, only: [:create]
   before_action :find_pledge, only: [:update, :destroy]
   rescue_from Mongoid::Errors::DocumentNotFound,
-              with: -> { head :bad_request }
+              with: -> { head :not_found }
 
   def create
     @pledge = @patient.external_pledges.new external_pledge_params
@@ -17,7 +17,7 @@ class ExternalPledgesController < ApplicationController
   end
 
   def update
-    if @pledge.update_attributes external_pledge_params
+    if @pledge.update external_pledge_params
       respond_to { |format| format.js }
     else
       head :bad_request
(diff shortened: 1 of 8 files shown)


app/controllers/external_pledges_controller.rb
```diff
@@ -3,7 +3,7 @@ class ExternalPledgesController < ApplicationController
   before_action :find_patient, only: [:create]
   before_action :find_pledge, only: [:update, :destroy]
   rescue_from Mongoid::Errors::DocumentNotFound,
-              with: -> { head :bad_request }
+              with: -> { head :not_found }
```
colinxfleming: not sure what I was thinking here exactly!

app/controllers/notes_controller.rb
```diff
@@ -16,26 +15,11 @@ def create
     end
   end
 
-  def update
```
colinxfleming: We haven't used this endpoint since this app has existed. I'm stripping it.

test/controllers/external_pledges_controller_test.rb
```diff
@@ -57,9 +58,9 @@ class ExternalPledgesControllerTest < ActionDispatch::IntegrationTest
 
     it 'should produce an audit trail' do
       assert_equal @pledge.history_tracks.count, 2
-      @changes = @pledge.history_tracks.last
-      assert_equal @changes.modified[:updated_by_id], @user.id
-      assert_equal @changes.modified[:source], 'Edited Pledge'
+      last_changes = @pledge.history_tracks.last
+      assert_equal last_changes.updated_by_id, @user.id
```
colinxfleming: I'm not sure what about removing the `gon` gem commit makes `updated_by_id` not be included in the list of modified fields, but we can get this from a different spot so I'm not worried about it.

Also renamed `@changes` to `last_changes` to cut down on an unnecessary instance variable.

DarthHater:
Were we even using `gon`? I see a few references but it looks like this was mostly just "remove because it's not being used at all"

colinxfleming:
Nah. We were using it in an early reporting take but subbed it out for
plain old controller actions and async loads. This has been dead weight for
awhile now.

>

PR #1529, opened 2018-11-23
rename fulfillment.procedure_cost -> fund_payout for simplicity

### LINKED ISSUE #1402: Rename fulfillment.procedure_cost to something more reflective of what it is
What's the difference between abortion cost and procedure cost? No idea, but one of them is the quoted pricetag by the clinic and the other is what a fund paid out. (Procedure cost is the latter.) We should rename this so it doesn't drive me, personally, up the wall.

Steps are:

* write a ruby one liner that renames the field to something
* change it in the model
* update exportable
* update our docs
* flag for analytics team

cc @alisonnjones

colinxfleming:

`Patient` has a `procedure_cost`. `Fulfillment`also has a `procedure_cost`. I get these mixed up constantly, so we're changing it in the model to be clearer what the fulfillment one means.

@tingaloo for review on the code part

@alisonnjones and @mdworken this impacts your scripting, so please confirm that this a) makes sense and b) won't be terribly disruptive.

* fulfillment.procedure_cost -> fund_payout

Will need to run this to complete the migration:

```
Patient.all.each do |pt|
  pt.fulfillment.update fund_payout_amount: pt.fulfillment[:procedure_cost]
  pt.fulfillment.unset :procedure_cost
end
```

* Fixes #1402


colinxfleming:
The full migration wound up being: 

```
Patient.all.each do |pt|
  pt.fulfillment.update fund_payout: pt.fulfillment[:procedure_cost]
  pt.fulfillment.unset :procedure_cost
end

ArchivedPatient.all.each do |pt|
  pt.fulfillment.update fund_payout: pt.fulfillment[:procedure_cost]
  pt.fulfillment.unset :procedure_cost
end
```

PR #1530, opened 2018-11-23
move api key to an initializer, simplify clinicfinder code

### LINKED ISSUE #1337: Add Geokit config
In testing the /clinicfinder form, I was getting a 500 error due to a lack of Google Maps API Key. This is because the clinic_finder gem uses geokit, which requires some initialization.

To get it to work, I added https://github.com/geokit/geokit-rails/blob/master/lib/generators/templates/geokit_config.rb to config/initializers, and uncommented the Geokit::Geocoders::GoogleGeocoder.api_key line, adding my API key.

colinxfleming:

As pointed out by @dismantl in #1337 , the way we're handling google geo api keys is a little weird and not super railsy. We addressed the initial problem of not having the key anywhere in the codebase when we brought in the clinic finder code, but there's still some awkwardness and redundancy in how we're handling the key; this addresses most of it.

* set up geokit initializer
* move env var hits to the initializer

no view changes

* Fixes #1337

### DIFF
--- a/test/models/clinic_test.rb
+++ b/test/models/clinic_test.rb
@@ -75,11 +75,10 @@ class ClinicTest < ActiveSupport::TestCase
 
   describe 'callbacks' do
     describe 'updating coordinates when address changes' do
-      before { @cached_api_key = ENV['GOOGLE_GEO_API_KEY'] }
-      after { ENV['GOOGLE_GEO_API_KEY'] = @cached_api_key }
+      before { Geokit::Geocoders::GoogleGeocoder.api_key = '123' }
+      after { Geokit::Geocoders::GoogleGeocoder.api_key = nil }
 
       it 'should update coordinates on address change' do
-        ENV['GOOGLE_GEO_API_KEY'] = '123'
         fake_geo = OpenStruct.new lat: 38.8976633, lng: 77.0365739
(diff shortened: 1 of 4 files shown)


test/models/clinic_test.rb
```diff
@@ -75,11 +75,10 @@ class ClinicTest < ActiveSupport::TestCase
 
   describe 'callbacks' do
     describe 'updating coordinates when address changes' do
-      before { @cached_api_key = ENV['GOOGLE_GEO_API_KEY'] }
-      after { ENV['GOOGLE_GEO_API_KEY'] = @cached_api_key }
+      before { Geokit::Geocoders::GoogleGeocoder.api_key = '123' }
+      after { Geokit::Geocoders::GoogleGeocoder.api_key = nil }
```
lomky: By my read, at a high level this test has changed rather than just updated - is that the intent? 
Before we stashed and restored the value - now we set it and then unset it.

colinxfleming: this is a great flag! Yeah, this is an intentional change -- we handle api key setting better in the initializer now, and so we don't have to do a dance with env vars, but because this is a class attribute we want to reset it to the assumed state it was in. It's essentially the same dance of 'reset it to the state it was in' but we're not reading straight from the env var in the locator code anymore.

Issue #1515, opened 2018-10-28
A few unit and systemtests to add


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

In the sprint of hacktoberfest, we left a few things uncovered. So let's cover them per comments below.

* **What feature or behavior is this required for?**

Safety first!

* **How could we solve this issue? (Not knowing is okay!)**

Adding tests per comments below.

* **Anything else?**

colinxfleming:
Confirm `:textable` checkbox persists in `system/update_patient_info`

Kevin-Wei:
https://github.com/DCAFEngineering/dcaf_case_management/pull/1519 system test here

colinxfleming:
Controller tests for NAF and Medicaid clinics in clinicfinders per #1531

PR #1531, opened 2018-11-23
Add naf and medicaid only filters to clinic_finder

### LINKED ISSUE #1364: Clinic finder: Implement medicaid and NAF filters
TK

colinxfleming:

We've had the ability to filter by medicaid and naf-ness for a little while, so let's surface that and make it available.

* Add filter checkboxes to the clinic finder form
* A little bit of code reorganization

Not sure of a great way to handle testing this, so kicking it to the testing backlog in #1515 

image

* Fixes #1364


app/controllers/clinicfinders_controller.rb
```diff
@@ -5,10 +5,10 @@ def search
     return head :bad_request if params[:zip].blank?
 
     clinic_finder = ClinicFinder::Locator.new(
-      Clinic.where(:zip.nin => [nil, '']),
+      Clinic.where(:zip.nin => [nil, '', '99999']),
```
colinxfleming: doing this because this isn't a real zip code and gives us an out to have clinics not return. (This is the zip code of DCAF's catchall 'OTHER' option.)

lomky: Should we comment that in the code for future us's confusion alleviation?

app/views/clinicfinders/_search_form.html.erb
```diff
@@ -1,28 +1,33 @@
...
+      <div class="row">
+        <div class="col-sm-6 info-form-left">
+          <%= f.text_field :zip, label: "Patient's ZIP code" %>
+          <%= f.check_box :naf_only, label: 'Show only NAF clinics' %>
+          <%= f.check_box :medicaid_only, label: 'Show only Medicaid clinics' %>
+        </div>
+
+        <div class="col-sm-6 info-form-right">
+          This is an experimental feature.
+          <%#= f.select :gestation, weeks_options, label: "Patient's approx. gestation at appt" %>
```
lomky: You've kept this commented code, but not it's sibling comment line. Do we want to have this still? What gem is it waiting on?

PR #1536, opened 2018-11-29
bundle update rails

colinxfleming:

Squelch a rails cve in activejob and activestorage (which we aren't affected by, but hey).

* `bundle update rails`

no issues


colinxfleming:
godmode because a gem change to squelch a cve

PR #1538, opened 2018-12-02
Translate lines/new view to Spanish

colinxfleming:

Me and my good friend google translate had 20 minutes so I knocked out the lines/new view. Also took the opportunity to spruce up the language, since it was getting a little stale. (We're DARIA now, not case manager, for example.) 

@DarthHater can you confirm that the il8n keys I picked match the pattern you're rolling with here, and spot check the code?
@montanezp can you check the translations and confirm they make sense?

* translates the strings in the lines/new view in `es.yml`
* il8n's the `lines/new` view

image

* Bumps #1382


config/locales/es.yml
```diff
@@ -16,3 +16,8 @@ es:
       sign_in: "Registrarse"
       profile: "Mi Perfil"
       sign_out: "Desconectar"
+  lines:
+    new:
+      welcome_to_daria: "¡Bienvenido a DARIA (edición de %{fund})!"
+      what_line: "¿En qué línea estás trabajando?"
```
montanezp: Change "qué" to "cuál" in the line so it reads:

¿En cuál línea estás trabajando?

colinxfleming:
@DarthHater approved this in https://codefordc.slack.com/archives/C0E6APB36/p1543767015008100

colinxfleming:
got a slack approval from @DarthHater so mergin'

PR #1541, opened 2018-12-06
Add procedure_cost to fulfillment model

BintLopez:

* adds procedure_cost to Fulfillment model so that rake db:seed does not error out

* Fixes https://github.com/DCAFEngineering/dcaf_case_management/issues/1540

### DIFF
--- a/app/models/fulfillment.rb
+++ b/app/models/fulfillment.rb
@@ -16,7 +16,6 @@ class Fulfillment
   field :check_number, type: String
   field :date_of_check, type: Date
   field :audited, type: Boolean
-  field :procedure_cost, type: Integer
 
   # Validations
   validates :created_by_id,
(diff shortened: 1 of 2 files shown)


BintLopez:
Containers are failing to spin up at all it looks like?
image

colinxfleming:
yeah we're having a problem with builds from forks. I'll add you to the github group in a bit.

BintLopez:
cool coool! thanks @colinxfleming !

app/models/fulfillment.rb
```diff
@@ -16,6 +16,7 @@ class Fulfillment
   field :check_number, type: String
   field :date_of_check, type: Date
   field :audited, type: Boolean
+  field :procedure_cost, type: Integer
```
colinxfleming: this field is `fund_payout` now per #1529  -- `Fulfillment#procedure_cost` was conflicting with the field with the same name on `Patient`.

Let's change the field in `seeds` to be `fund_payout`, if that's cool?

sorry I missed this, thought I got it everywhere! my mistake.

BintLopez:
Closing in favor of https://github.com/DCAFEngineering/dcaf_case_management/pull/1542

PR #1544, opened 2018-12-07
Add mongoid_rails_migrations gem

### LINKED ISSUE #1535: Add mongoid migrations gem

There are scenarios like #1529 where we do want to actually do real data migrations, so we should start doing that. https://github.com/adacosta/mongoid_rails_migrations is a maintained library that looks like it'll do this -- let's install it.

Good data hygiene across multiple deployments

Installing https://github.com/adacosta/mongoid_rails_migrations

* **Anything else?**

nah

colinxfleming:

As noted in #1535 I am over doing potentially scary mongoid migrations like field renames through the rails console. We're a somewhat mature project with multiple clients, and so that needs to be reversible and checked into VC. So this adds a library that adds the ability for mongoid to run migrations.

* removes the migration file we made before we went to a nosql database
* adds the `mongoid_rails_migrations` gem

* Fixes #1535


db/migrate/20151222011145_devise_create_users.rb
```diff
@@ -1,42 +0,0 @@
-class DeviseCreateUsers < ActiveRecord::Migration
```
lomky: Was this old cruft?

colinxfleming: yeah, we made the devise migration before we fully swapped over to mongoid, so this (and possibly a few of the other generated files) aren't ever going to be used by anything again.

colinxfleming:
thanks for the approval @lomky ! I feel your point on the tech documentation and thought about it some last night. I think around where I personally landed is: data migration-y tasks like this are extremely few and far between -- I think this is the third one we've had in three years. I also think that mongoid's lack of migrations is a departure from the rails standard -- I remember when we plugged in mongoid it was a surprise to me that `rake db:migrate` and `rails g migration ____` didn't work. Since this gem essentially restores that functionality and brings it back into line with rails conventions, I think the way to handle the surprise factor here is to be clear when writing issues that require a data migration to solve.

Gonna merge this, but let's talk in slack to get it squared away if there are further questions for sure.

PR #1547, opened 2018-12-09
Translate the footer to Spanish

colinxfleming:

Adds il8n keys for the footer like so:

image

In that example, the DC Abortion Fund is set from an env var and so an in-place spanish translation would be a not-easy basket. but the stuff around it is hardset so let's il8n that.

* Translates the footer content

@DarthHater for review on the ruby, @montanezp for review on the spanish

* Bumps #1382

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -2,14 +2,16 @@
 	<hr class="hr_gradient">
 	<div class="row">
 		<div class="col-sm-12 text-center">
-			<small>Please send feedback or ideas for this case management app to <a href="mailto:DCAF.CMapp@gmail.com">DCAF.CMapp@gmail.com</a></small>
+			<small><%= t 'navigation.footer.please_send_feedback' %> <%= link_to 'DCAF.CMapp@gmail.com', mailto: 'DCAF.CMapp@gmail.com' %>.</small>
 		</div>
 	</div><!-- end row -->
 </div><!-- end container -->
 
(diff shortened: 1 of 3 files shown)


app/views/layouts/_footer.html.erb
```diff
@@ -2,14 +2,16 @@
 	<hr class="hr_gradient">
 	<div class="row">
 		<div class="col-sm-12 text-center">
-			<small>Please send feedback or ideas for this case management app to <a href="mailto:DCAF.CMapp@gmail.com">DCAF.CMapp@gmail.com</a></small>
+			<small><%= t 'navigation.footer.please_send_feedback' %> <%= link_to 'DCAF.CMapp@gmail.com', mailto: 'DCAF.CMapp@gmail.com' %>.</small>
```
colinxfleming: tried really hard to stick the link in the il8n key and couldn't get it to work in 5m of poking. Not ideal but I'm not stressing about it.

montanezp:
Translation looks good! My only flag is whether we should translate DCAF since NAF is translated. I know in the tri-fold that was translated into Spanish, DCAF was called "El Fondo para el Aborto del Distrito de Columbia (DCAF)"

colinxfleming:
hm, I think translating DCAF there would be a bit more trouble than it's worth, since we set that from an environment and not inline. Lots of work for not that much reward. That said, let's revisit this when we're getting ready to do the rollout and see if it's a pain point.

PR #1549, opened 2018-12-10
Translating the login page

lomky:
I rule and have ~completed~ started some work on Case Manager!

Started adjusting the login page and added some spanish welcome messages.

Tagging in @DarthHater for any devise-specific adjustments and @montanezp for reviewing my Spanish!

* Bumps #1382

### DIFF
--- a/app/helpers/welcome_message_helper.rb
+++ b/app/helpers/welcome_message_helper.rb
@@ -9,26 +9,29 @@ module WelcomeMessageHelper
   def welcome_messages
     [
       "Thank you! - Team CM",
+      "¡Gracias! - Equipo CM",
       "You are making choice a reality for the people you serve.",
       "Keep up the good work, you're doing great!",
+      "¡Sigan con el buen trabajo, lo están haciendo muy bien!",
       "We're all rooting for you!",
+      "Estamos todos apoyando para usted!",
       'We appreciate all that you do!',
-      "From San Francisco with love!",
-      "You're a real gem - the client_finder gem team",
-      "You're rad and your work is much appreciated!",
       "From San Francisco with love!",
       "You're a real gem - the client_finder gem team",
       "You're rad and your work is much appreciated!",
       "Thanks for rocking our socks!",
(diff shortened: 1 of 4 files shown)


app/helpers/welcome_message_helper.rb
```diff
@@ -9,26 +9,29 @@ def random_welcome_message
   def welcome_messages
     [
       "Thank you! - Team CM",
+      "¡Gracias! - Equipo CM",
```
DarthHater: It might be neat to have welcome_messages take a locale and spit out something related to it

config/locales/es.yml
```diff
@@ -21,3 +21,11 @@ es:
       welcome_to_daria: "¡Bienvenido a DARIA (edición de %{fund})!"
       what_line: "¿En cuál línea estás trabajando?"
       start: "Empezar"
+  devise:
+    sessions:
+      new:
+        google_signin: "Inicia sesión con Google (recomendado)"
+        password_signin: "O, inicie sesión con una contraseña:"
```
DarthHater: In some Spanish cultures I think it's preferred to use "clave" vs "contraseña" to refer to a typed password (at least from when I was living in Argentina). Not 100% sure here, but I figured I'd flag it for the expert @montanezp

app/helpers/welcome_message_helper.rb
```diff
@@ -9,26 +9,29 @@ def random_welcome_message
...
       "You are making choice a reality for the people you serve.",
       "Keep up the good work, you're doing great!",
+      "¡Sigan con el buen trabajo, lo están haciendo muy bien!",
```
colinxfleming: I legit love having welcome messages in spanish

config/locales/es.yml
montanezp: I've seen both used though I would suggest "contraseña" here as, at least anecdotally, it's been used for "password" more.

colinxfleming:
@lomky is this still WIP? Can we merge?

PR #1550, opened 2018-12-10
Patient Dashboard translation

lomky:

* Patient dashboard translated
* Status Keys and Helper text translated
* Delete confirmation translated
* Patient Week/Day helpers translated

screenshot from 2018-12-10 18-15-20

@colinxfleming and @DarthHater for some eyes on my code changes - especially where I chose to insert the i18n code for the patient helper & status concern.

@montanezp could you review my Spanish please? I'm especially interested if you have a shorter way to say "Weeks Along at Intake" as my Spanish version mucks up the UI a bit.

* Bumps #1382


app/views/patients/_patient_dashboard.html.erb
```diff
@@ -1,33 +1,33 @@
 <div id="patient_dashboard_content">
   <%= bootstrap_form_for patient, html: { id: 'patient_dashboard_form' }, remote: true do |f| %>
     <div class="col-sm-4" >
-      <%= f.text_field :name, label: 'First and last name', autocomplete: 'off' %>
-      <%= f.text_field :primary_phone, value: patient.primary_phone_display, label: 'Phone number', autocomplete: 'off' %>
+      <%= f.text_field :name, label: t('patient.dashboard.name'), autocomplete: 'off' %>
```
colinxfleming: Some of these keys we're going to reuse in a few different places. I wonder about having a form namespace explicitly for form labels and stuff like that which might be the same across different forms on different views? @lomky @DarthHater do you have any thoughts?

config/locales/es.yml
```diff
@@ -21,3 +21,39 @@ es:
       welcome_to_daria: "¡Bienvenido a DARIA (edición de %{fund})!"
       what_line: "¿En cuál línea estás trabajando?"
       start: "Empezar"
+  patient:
+    helper:
+      day: "día"
```
colinxfleming: You said you had some trouble with the pluralizing here, or am I remembering wrong?

app/views/patients/_patient_dashboard.html.erb
lomky: I'll create on called 'shared' and move name and status under there. we can shift other things into it as we realize they're needed in multiple places. :tada:

config/locales/es.yml
lomky: I was coming in from the wrong direction, coming from the pluralization we had been using. It turns out you can access pluralizing through I18n, though, of course!

lomky:
After we merge in this batch of Spanish translations, we should run through the `i18n-tasks health` checks. After that if we run it semi-reliably on new translation PRs it should be a clean process!

montanezp:
Everything looks good, I suggest making the weeks along at intake to be: "semanas de embarazo al inicio"

colinxfleming:
@lomky looks like there are some legit test failures here: 

image

I'm not 100% sure they're related to the work in here, though. should we merge as is or fix those in here?

lomky:
@colinxfleming Those are the new normalizer tests! Is this the last remaining open translate PR? If so I'll go ahead and run normalize and key checks on it.

colinxfleming:
@lomky yeah, this is the last one in this block of translations. Let's get the normalize stuff set up.

PR #1554, opened 2018-12-15
add tooltip to budget bar

### LINKED ISSUE #1532: Add tooltip for budget bar

The budget bar has complicated business logic that we don't surface at all, which means when there are questions, people are taking it more or less on faith.

A Better Budget Bar

Add a tooltip to the budget bar with the following explanatory text:

> The budget bar lists money tentatively set aside for a patient (by entering a value in the #{FUND} pledge field) and money sent to a clinic for a patient (by checking the I sent the pledge checkbox) over a given week. It resets on Mondays.

* **Anything else?**
Stems from #1458

tingaloo:

screen shot 2018-12-15 at 12 23 55

Adding the tooltip for budget bar, where should it go?

EDIT: moved tooltip.

* Add tooltip

* Fixes #1532

### DIFF
--- a/app/helpers/tooltips_helper.rb
+++ b/app/helpers/tooltips_helper.rb
@@ -24,6 +24,14 @@ module TooltipsHelper
     text.strip
   end
 
+  def budget_bar_help_text
+    text = <<-TEXT
+      The budget bar lists money tentatively set aside for a patient
+      (by entering a value in the #{FUND} pledge field) and money sent to a clinic for a patient (by checking the I sent the pledge checkbox) over a given week. It resets on Mondays.
+    TEXT
+    text.strip
+  end
+
   def your_completed_calls_help_text
     text = <<-TEXT
       This is a list of patients you have called within the last 8 hours.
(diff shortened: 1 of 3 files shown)


app/helpers/tooltips_helper.rb
```diff
@@ -24,6 +24,14 @@ def your_call_list_help_text
     text.strip
   end
 
+  def budget_bar_help_text
```
colinxfleming: Can we get a super simple test for this that asserts a match with the words `budget bar`, just so if we ever have to revisit this we don't accidentally muck it up? something like:

```
# in test/helpers/tooltips_helper_test.rb
assert_match /budget bar/, budget_bar_help_text
```

app/views/dashboards/_overview.html.erb
```diff
@@ -1,5 +1,6 @@
 <div id="overview" class="margin-bottom">
   <h1 class="border-bottom title">Week of <%= week_range %></h1>
+  <%= dashboard_table_content_tooltip_shell "budget_bar" %>
```
colinxfleming: Let's pop this within the h1 on line 2, after the `week_range`.

PR #1556, opened 2018-12-16
Initial Configs Changes for Spanish

DarthHater:

This is the start of some work on the Configs pages to translate them into Spanish

* Changes to config index template
* Changes to _config template
* Supporting keys in es, en locales

screen shot 2018-12-15 at 3 29 06 pm

@colinxfleming @lomky I got a few things started, but I'm kinda not sure what to do about the key stuff (that I think is loaded from the db). Open to ideas, thoughts before I move forward.

* Bumps #1382


app/views/configs/_config.html.erb
```diff
@@ -6,14 +6,14 @@
                                                           .try(:join, ', '),
                                 help: config.help_text,
                                 id: "config_options_#{config.config_key}" %>
-      <%= f.submit "Update options for #{config.config_key.to_s.humanize}", class: 'btn btn-primary' %>
+      <%= f.submit t('configs.config.update_options_for', option: config.config_key.to_s.humanize), class: 'btn btn-primary' %>
```
colinxfleming: does humanize work okay with spanish? I know @lomky had some problems with the pluralizer.

colinxfleming:
We can frankly probably live without config key names off the bat, so I think we can figure that out later and/or write a mapping of some kind down the road.

montanezp:
@colinxfleming hi! Can I get a screenshot of the english version of this translation, please? Just need a reference point since this one is a little longer. Thank you!

montanezp:
@colinxfleming made a couple of edits. Here yah go! [https://docs.google.com/document/d/1OswVVCbIao6PabVlSArhYDiihKUNRPRrYi6ayQbbs1A/edit?usp=sharing](url)

DarthHater:
screen shot 2018-12-26 at 8 40 22 am

@montanezp this is after changing the keys to what you put in the doc, I'm holding off on some of the config keys based on the convo with @colinxfleming in here (making those changes is a bit more involved than most of the other translation stuff). The button looks wonky to me (mainly because I kept the key in there). Do you want me to take the key out completely?

It would go from:

`Opciones para el compromiso externo %{option}, (por favor separe con comas)`

to:

`Opciones para el compromiso externo, (por favor separe con comas)`

colinxfleming:
I'm gonna make a call to keep the key in for now -- can always take it out. ty @DarthHater !

PR #1558, opened 2018-12-16
Filter and sort budget bar entries

### LINKED ISSUE #1533: Consistently sort budget bar entries

It can be kind of annoying to have to shuffle through patients in the budget bar, and they aren't always ordered consistently. We should explicitly do some sorting to avoid this surprise.

A Better Budget Bar

I THINK if you trace the budget bar code back, it turns you up in `Patient#pledged_status_summary`, and I think you'll add the sorts in there, after the where filters and before the pluck.

Let's sort by `pledged_sent` so that sent pledges show up first, then by `fund_pledged_at` so that earlier pledges show up first, then by name ascending. 

* **Anything else?**
Stems from #1458

### LINKED ISSUE #1534: Filter patients resolved_without_assistance from the budget bar

Patients marked as `Resolved Without Assistance` are showing up in the budget bar. Since the budget bar is supposed to represent live expenditures we're keeping track of, and people marked as such are saying they no longer wish to receive a fund's help, we should have them skip the budget bar.

A Better Budget Bar

Pop into `Patient#pledged_status_summary` and filter out patients marked as `resolved_without_assistance`.

* **Anything else?**
Stems from #1458

colinxfleming:

Addresses some points of CM feedback from the budget bar rollout -- consistent sorting of patients and filtering out patients who have elected not to get dcaf's help.

* order by `fund_pledged_at: :asc`
* `.where(:resolved_without_fund.nin => [true])`

* Fixes #1533 
* Fixes #1534


app/models/patient.rb
```diff
@@ -147,6 +147,8 @@ def self.pledged_status_summary(line)
     patients = Patient.in(line: line)
                       .where(:fund_pledge.nin => [0, nil, ''])
                       .where(:fund_pledged_at.gte => start_of_week)
+                      .where(:resolved_without_fund.nin => [true])
```
lomky: Any reason we're not just saying `in`? The negation term combined with `without` is a bit hard to parse, and I can see reading this line a bit too fast as the opposite with how close `nin` and `in` are. I might be being a bit nitpicky though. Thoughts?

test/models/patient_test.rb
```diff
@@ -126,23 +126,28 @@ class PatientTest < ActiveSupport::TestCase
...
     end
   end
 
   describe 'pledge_summary' do
-    it "should return proper pledge summaries for various timespans" do
+    it 'should return proper pledge summaries for various timespans' do
       @patient.update appointment_date: 2.days.from_now, fund_pledge: 300
       @patient2.update appointment_date: 4.days.from_now, fund_pledge: 500,
                        pledge_sent: true, clinic: create(:clinic)
+      # this pt should be filtered out because we don't worry about resolved without fund
```
lomky: I'm not clear in which test we're asserting this patient is filtered. Could you talk me through it?

PR #1570, opened 2018-12-30
Filter patients which resolved it without funds

### LINKED ISSUE #1533: Consistently sort budget bar entries

### LINKED ISSUE #1534: Filter patients resolved_without_assistance from the budget bar

nikobozi:

It filters out patients which have resolved the case without funds.
I decided to simplify the method by extracting it to a scope, and updated the test so that such patients are not included in the summary. 
Also the factory has been updated, since the tests were testing patient names, but the name appeared to be hard-coded in the factory

* Fixes #1534 
* Fixes #1533

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -143,6 +143,7 @@ class Patient
     where(:fund_pledge.nin => [0, nil, ''])
       .where(:resolved_without_fund.in => [0, nil, ''])
       .where(:fund_pledged_at.gte => Time.zone.today.beginning_of_week(:monday))
+      .order_by(fund_pledged_at: :asc)
       .in(line: line)
   end
 
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -138,15 +138,19 @@ class Patient
                 track_destroy: true
   mongoid_userstamp user_model: 'User'
 
+  # Scopes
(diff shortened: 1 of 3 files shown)


colinxfleming:
hey, thank you very much! I'm traveling right now for the holidays, so it'll take me a minute to get to this, thanks in advance for your patience!

ps the failure is probably a known problem with circleci, so don't worry about it.

Issue #1560, opened 2018-12-21
Update Oauth SSO


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Google is phasing out G+ and at least one of our oauth methods (`dcaf single sign on production (dcaf-single-sign-on-production)`) relies on this. We should bump this.

* **What feature or behavior is this required for?**

being able to continue to sign in after February 15 or so

* **How could we solve this issue? (Not knowing is okay!)**

Go into our GCP account, swap this key to use whatever they're replacing it with, and then replace that key wherever we use it. I think this only affects dcaf prod.

* **Anything else?**

PR #1588, opened 2019-01-24
add additional info to set up an instance docs

BintLopez:

I recently set up and deployed an instance of DARIA following the instructions in [docs/SETTING_UP_A_NEW_INSTANCE.md](https://github.com/DCAFEngineering/dcaf_case_management/blob/master/docs/SETTING_UP_A_NEW_INSTANCE.md). Adding some information based on my experience.

* adds to `docs/SETTING_UP_A_NEW_INSTANCE.md`


docs/SETTING_UP_A_NEW_INSTANCE.md
```diff
@@ -10,20 +10,38 @@ DCAF has a lot of the surrounding infrastructure set up in such a way that it's
...
-
-## Optional steps
-
-* Set `SKYLIGHT_AUTH_TOKEN` (get it from Skylight, this is a performance monitoring service)
-* Set `SQREEN_TOKEN` (get it from Sqreen, this is a security service)
+* Set up Google OAuth credentials
+  * Set up a google account to administer this app (or choose an existing account you own)
+  * Once logged in, navigate to https://console.developers.google.com/apis/dashboard
+  * Add a project (note -- credentials can be set up alongside existing creds for different environments)
+  * Enable the Google+ API (this is necessary to use google for sign in to the app)
```
BintLopez: I got hung up on this initially. The error messaging you receive from google when oauth doesn't work is also not particularly helpful in pushing toward this solution. Curious, do you all just have Google+ api enabled or do you also have Contacts and People APIs enabled? I found some conflicting sources on which apis were indeed necessary to have enabled for the creds to work for oauth sign-in and after testing manually found that Google+ was the minimum api to enable for login to work.

BintLopez:
@colinxfleming Adding my notes to the setting up and instance guide.

docs/SETTING_UP_A_NEW_INSTANCE.md
colinxfleming: yeah, you're right to call this out. Google is phasing out the google+ api (see https://developers.google.com/+/api-shutdown and #1560 ). Based on https://github.com/zquestz/omniauth-google-oauth2/pull/344 and some of the issues I think we can do without turning the G+ API on after we do a gem version bump. I'll fire up a PR for that. sorry you got burned by this!

docs/SETTING_UP_A_NEW_INSTANCE.md
```diff
@@ -10,21 +10,39 @@ DCAF has a lot of the surrounding infrastructure set up in such a way that it's
...
+    * Authorized Javascript Origins should be https://your-app.herokuapp.com 
+    * Authorized redirect URIs should be https://your-app.herokuapp.com/users/auth/google_oauth2/callback
+  * Fill in the `DARIA_GOOGLE_KEY` and `DARIA_GOOGLE_SECRET` env vars with your client id and client secret
+* While still in the google developers console, enable the Geocoding API, create an API key, and set the `GOOGLE_GEO_API_KEY` in heroku to that API key
+
+## Optional (but recommended) steps
+
+* Set `SKYLIGHT_AUTH_TOKEN` (get it from [Skylight](https://www.skylight.io/), this is a performance monitoring service)
+* Set `SQREEN_TOKEN` (get it from [Sqreen](https://www.sqreen.io/), this is a security service)
+* if you have a custom domain set up, set up heroku ACM ([Automated Certificate Management](https://devcenter.heroku.com/articles/automated-certificate-management))
```
colinxfleming: thank you for adding links to these!

docs/SETTING_UP_A_NEW_INSTANCE.md
```diff
@@ -10,21 +10,39 @@ DCAF has a lot of the surrounding infrastructure set up in such a way that it's
...
-
-* Set `SKYLIGHT_AUTH_TOKEN` (get it from Skylight, this is a performance monitoring service)
-* Set `SQREEN_TOKEN` (get it from Sqreen, this is a security service)
-* if you have a custom domain set up, set up heroku ACM
+* Set up Google OAuth credentials
+  * Set up a google account to administer this app (or choose an existing account you own)
+  * Once logged in, navigate to https://console.developers.google.com/apis/dashboard
+  * Add a project (note -- credentials can be set up alongside existing creds for different environments)
+  * Enable the Google+ API (this is necessary to use google for sign in to the app)
+  * Click to "Create OAuth client ID"
```
colinxfleming: I think these are configurations of an oauth client id, right? 

```suggestion
  * Click to "Create OAuth client ID"
    * Set your `Authorized Javascript Origins` to: https://your-app.herokuapp.com 
    * Set your `Authorized redirect URIs` to: https://your-app.herokuapp.com/users/auth/google_oauth2/callback
```

docs/SETTING_UP_A_NEW_INSTANCE.md
```diff
@@ -10,21 +10,39 @@ DCAF has a lot of the surrounding infrastructure set up in such a way that it's
...
+  * Configure your OAuth Consent screen
+    * Make your application type public
+    * Authorized Javascript Origins should be https://your-app.herokuapp.com
```
colinxfleming:
```suggestion

```

docs/SETTING_UP_A_NEW_INSTANCE.md
```diff
@@ -10,21 +10,39 @@ DCAF has a lot of the surrounding infrastructure set up in such a way that it's
...
+* Set up Google OAuth credentials
+  * Set up a google account to administer this app (or choose an existing account you own)
+  * Once logged in, navigate to https://console.developers.google.com/apis/dashboard
+  * Add a project (note -- credentials can be set up alongside existing creds for different environments)
+  * Enable the Google+ API (this is necessary to use google for sign in to the app)
+  * Click to "Create OAuth client ID"
+  * Configure your OAuth Consent screen
+    * Make your application type public
+    * Authorized Javascript Origins should be https://your-app.herokuapp.com 
+    * Authorized redirect URIs should be https://your-app.herokuapp.com/users/auth/google_oauth2/callback
```
colinxfleming:
```suggestion
```

docs/SETTING_UP_A_NEW_INSTANCE.md
BintLopez: All good! I'll take out that step about G+ api since it won't be relevant once your PR merges.

docs/SETTING_UP_A_NEW_INSTANCE.md
BintLopez: Yes... is OAuth client not what's needed for the google sign in? Or are you saying you'd rather not go into that level of detail?

colinxfleming: Oh, no, it's definitely needed! But I think setting these values is tied to creating the oauth client id, not configuring the consent screen. I might be wrong about that, let me check.

colinxfleming: yep, it looks like the consent screen is some different stuff, and the authorized origins and redirects are on the client id/client secret level:

Here's the oauth screen:

image

Also, this made me look at authorized javascript origins for the existing client ids we have set up for oauth, and none of them have authorized javascript origins, so we might be able to remove that line of setup. Do the keys still work if you empty out that line in your config?

colinxfleming:
ps the google oauth PR is merged, so we don't need the G+ API at all anymore! Next time you do a deploy you can pop into your GCP and shut that off.

colinxfleming:
bumping for @BintLopez -- lmk if there's anything I can do to help get this squared away!

BintLopez:
Welp! Sorry will get this over the hump this week 🐫 

Got kinda sidetracked with transition stuff and all the lawyerly concerns with that.

docs/SETTING_UP_A_NEW_INSTANCE.md
BintLopez: I think we should keep this line in here so that folks don't have to chase OAuth errors or check docs to figure out what this redirect should be set to.

PR #1589, opened 2019-01-25
The start (and end) of Rails 6

DarthHater:

Start of a PR that is the path to Rails 6, I suppose!

* Updated `Gemfile` to use Rails 6 beta 1
* `bundle update`


colinxfleming:
TODOs before we can merge:
* wait until mongoid's gemspec gets bumped (it currently excludes rails 6 activerecord)
* run rails upgrade or rails update or whatever the command is
* get tests passing

We don't need to worry about the new stuff in rails 6 yet.

colinxfleming:
Mongoid got bumped, but here are some downstream dependencies we need to update: 

```
    bootstrap_form-datetimepicker was resolved to 0.0.1, which depends on
      bootstrap_form was resolved to 4.2.0, which depends on
        actionpack (>= 5.0)

    mongo_session_store (>= 3.1.0) was resolved to 3.1.0, which depends on
      actionpack (>= 4.0)

    rails (~> 6.0.0) was resolved to 6.0.0, which depends on
      actionpack (= 6.0.0)

    devise (~> 4.6) was resolved to 4.6.1, which depends on
      responders was resolved to 2.4.1, which depends on
        actionpack (>= 4.2.0, < 6.0)
```

so I think devise and whatever datetimepicker is. we might not use it?

Gemfile.lock
```diff
@@ -279,37 +292,39 @@ GEM
...
-      actioncable (= 5.2.2.1)
-      actionmailer (= 5.2.2.1)
-      actionpack (= 5.2.2.1)
-      actionview (= 5.2.2.1)
-      activejob (= 5.2.2.1)
-      activemodel (= 5.2.2.1)
-      activerecord (= 5.2.2.1)
-      activestorage (= 5.2.2.1)
-      activesupport (= 5.2.2.1)
+    rack-test (1.1.0)
```
colinxfleming: check on this

colinxfleming: this is fine, just involves a new required argument for non-get requests that shouldn't affect us much per https://github.com/rack-test/rack-test/releases/tag/v1.0.0

colinxfleming:
new steps:

- [ ] `assert_routing` is no longer a thing, so search the codebase for it and remove it (replace if possible)
- [ ] run rails upgrade or rails update or whatever the command is
- [ ] get tests passing

config/application.rb
```diff
@@ -3,11 +3,23 @@
...
+require "rails"
+# Pick the frameworks you want:
+# require "active_model/railtie"
+# require "active_job/railtie"
+# require "active_record/railtie"
+# require "active_storage/engine"
+require "action_controller/railtie"
+require "action_mailer/railtie"
+# require "action_mailbox/engine"
+# require "action_text/engine"
```
colinxfleming: this might be something we try out? could be cool!

colinxfleming:
bad news:
* something in rails update broke builds
* something in commits after assert_routing blew up the size of the fontawesome icons

lots to revert here unfortunately. Might behoove us to roll out as is and incrementally make config changes. bleh.

colinxfleming:
It is an absolute mystery what in the Rails CSP handler changed to break this, but something did.

This is almost certainly a problem that is solveable if we can resolve these browser console errors: 
image

colinxfleming:
It looks like you have to drop a CSP flag thing in the headers; adding it above the webpacker snippet makes it work. woof.

colinxfleming:
I'm going to pull two changes out of this PR:
* webpacker upgrade to 4.2.0
* change to the gitignore (required for a puma setting in here)

I'll also be creating a few followup tickets:
* use 6.0 defaults and get zeitwerk working; confirm we aren't doing force ssl on a controller level
* move certain js packages to yarn

config/environments/production.rb
```diff
@@ -16,35 +16,27 @@
...
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
+
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
   config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
```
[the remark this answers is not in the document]
colinxfleming: took out uglifier - it's not in the default rails install anymore! wild.

colinxfleming:
Okay, I think this is in reasonably good shape. Here's what we should write followup tickets for:

* figure out what to do with the rails 6.0 defaults file - I think we're using new framework defaults since I just commented everything out, but we should check on that
* make sure we're using the new zeitwerk autoloading
* confirm that we're forcing SSL properly
* set up parallel testing

but I THINK we can roll this out as is.

PR #1590, opened 2019-01-27
bump google oauth gem to account for g+ deprecation

### LINKED ISSUE #1560: Update Oauth SSO

colinxfleming:

The google+ api is getting shut down. I thought this was going to be a big deal but it turns out it's as simple as bumping the library, which makes a few spare calls to the G+ api. Version 0.6.0 of the gem removes those calls and uses some always-on authentication apis, so in addition to this getting us out of the doghouse we can turn off the g+ api in our google project/org.

* omniauth-google-oauth2 0.5.3 -> 0.6.0
* `bundle install`

Note that after we ship this we can pop into our google cloud console and shut off the g+ api for that project.

No view changes.

* Fixes #1560 

@DarthHater or @rudietuesdays for review
@BintLopez for awareness


colinxfleming:
yep, looks like a flaky systemtest to me. I'll kick the build. thank you v much @rudietuesdays !

PR #1591, opened 2019-01-27
Make entries line up by being clearer about what nil is

### LINKED ISSUE #1338: Change Log entries do not line up properly when changed to/from nil
See screenshot in #1330 . Although this is not a severe issue (one can easily tell what was supposed to go where), it would be nice to make these line up properly.

I'll fix this myself, just wanted to put the issue out here.

colinxfleming:

a quick NITE PR -- I think this should resolve the problem described in #1338 . Screenshots in the morning. I'm getting old.

* `blank` changelog entries now show up as `(empty)`
* remix the view so breaks between individual field changes are always clear
* tests for same

image

* Fixes #1338

### DIFF
--- a/app/models/audit_trail.rb
+++ b/app/models/audit_trail.rb
@@ -5,7 +5,7 @@ class AuditTrail
   include Mongoid::Userstamp
   mongoid_userstamp user_model: 'User'
 
-  IRRELEVANT_FIELDS = %w[user_ids updated_by_id pledge_sent_by_id last_edited_by_id].freeze
+  IRRELEVANT_FIELDS = %w[user_ids updated_by_id pledge_sent_by_id last_edited_by_id identifier].freeze
   DATE_FIELDS = %w[appointment_date initial_call_date pledge_generated_at pledge_sent_at fund_pledged_at].freeze
 
   # convenience methods for clean view display
@@ -39,7 +39,7 @@ class AuditTrail
     all_fields = orig.keys | mod.keys
     changeset = {}
     all_fields.each do |key|
-      changeset[key.humanize] = { original: format_fieldchange(key, orig[key]),
+      changeset[key] = { original: format_fieldchange(key, orig[key]),
                                   modified: format_fieldchange(key, modified[key]) }
     end
     changeset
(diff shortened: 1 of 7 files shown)


colinxfleming:
hm, turns out this is really an issue with how we've written some methods on the audit trail model - this is what happens when I modified patient 0 from a fresh seed and added age:

```
irb(main):017:0> change.changed_fields
=> ["Age", "Special circumstances"]
irb(main):018:0> change.changed_from
=> ["not selected"]
irb(main):019:0> change.changed_to
=> [nil, 34, "not selected", nil]
```

I think the `empty` thing in this PR is helpful but that seems to be the source of this. marking wip.

app/models/audit_trail.rb
```diff
@@ -71,4 +32,33 @@ def changed_by_user
   def marked_urgent?
     modified.include?('urgent_flag') && modified['urgent_flag'] == true
   end
+
+  def shaped_changes
```
colinxfleming: This is the most significant change here -- rather than `changed_from` and `changed_to`, this rolls it all up into one big hash, because it's easier to keep them aligned if they're in the same column.

app/models/audit_trail.rb
```diff
@@ -71,4 +32,33 @@ def changed_by_user
...
+    all_fields.each do |key|
+      changeset[key.humanize] = { original: format_fieldchange(key, orig[key]),
+                                  modified: format_fieldchange(key, modified[key]) }
+    end
+    changeset
+  end
+
+  private
+
+  def format_fieldchange(key, value)
```
colinxfleming: This rolls up the shaping functions. I don't like that L51 and L61 do the same thing, but the early return on L51 prevents us from making a DB call on clinic that will fail. Is this sloppy on my part? Possibly! Does it make the app error? nah!

colinxfleming:
de-WIPing! tagging some folks in to approve the code and the appearance changes.

app/models/audit_trail.rb
```diff
@@ -5,63 +5,24 @@ class AuditTrail
...
-    if hash.key?('clinic_id')
-      @clinic = Clinic.where(:id => hash["clinic_id"]).first
-      hash["clinic_id"] = @clinic&.name # TODO not having a try here fails the update patient info test for some reason!
-    end
-    return hash
+  def has_changed_fields?
+    modified.reject { |x| IRRELEVANT_FIELDS.include? x }.present?
   end
 
   def changed_from
```
lomky: I think `changed_from` and `changed_to` are dead code now. Rip them out?

colinxfleming: great catch, thank you!

PR #1592, opened 2019-02-03
Translate string constants in dashboards

colinxfleming:

Tackles some of the translation on dashboards, specifically strings contained in partials. Strings that are inferred or produced from helper methods are tk.

It felt weird to pull single simple words into their own nested namespaces, so this PR also stands up the top level namespace `common`, which I stuck a bunch of one word things into like Add, Phone, etc.

This doesn't touch the slightly more complex stuff (in helpers, for example). But I am committing to doing that here.

* translate strings defined in the dashboard partials
* a little bit of view cleanup while I'm rooting around in there
image

cc @montanezp can you pop in here and review the translations?

* Bumps #1575 

ps s/o @lomky for the good directions and the reminder to run `il8n-tasks health`!

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1,5 +1,14 @@
 ---
 es:
+  common:
+    add: Añadir
+    appointment_date_short: Día de la cita
+    name: Nombre
+    notes: Notas
+    phone: Teléfono
+    search: Buscar
+    status: Estado
+    weeks_along_short: Semanas a lo largo
   configs:
     config:
       current_options: 'Las opciones actuales son:'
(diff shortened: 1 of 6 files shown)


config/locales/es.yml
```diff
@@ -1,12 +1,33 @@
 ---
 es:
+  common:
+    add: Añadir
+    appointment_date_short: Día de la cita
+    name: Nombre
+    notes: Notas
+    phone: Teléfono
+    search: Buscar
+    status: Estado
```
montanezp: "estatus"

colinxfleming:
thank you @montanezp !

PR #1593, opened 2019-02-04
Complete dashboard translation to Spanish

colinxfleming:

A companion PR to #1592 to translate the hard parts of the dashboard (the titles, the helpers). Note that it leaves a few components open

An important exception: It turns out that doing the localization as a URL param doesn't get passed down to `js.erb` files as far as I can tell, so the `Search Results` chokes. This is a surprising behavior of UJS but does feel consistent with how we've set it up, and I'm not sure how to resolve that, so I'm going to let it lay.

* fixes up a bunch of the tooltip metaprogramming
* some view cleanup while I'm rooting around
* translates the headers

Stuff it does NOT fix:
* the Search results partial header (because of the UJS bug)
* anything having to do with the budget bar
* new call partial
* tooltips
* contents of the Activity log

image

cc @montanezp for review!

* Bumps #1575 
* Contains #1592


app/helpers/dashboards_helper.rb
```diff
@@ -1,25 +1,25 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
-  def week_range(date: DateTime.now.in_time_zone, start_day: :monday)
+  def week_range(date: Time.zone.now, start_day: :monday)
     week_start = date.beginning_of_week start_day
     week_end = date.end_of_week start_day
-    week_start_string = week_start.strftime('%B %-d')
+    week_start_string = l week_start, format: '%B %-d'
```
colinxfleming: It rules that we get this for free from rails

app/helpers/tooltips_helper.rb
```diff
@@ -8,15 +8,15 @@ def tooltip_shell(help_text)
     end
   end
 
-  def dashboard_table_content_tooltip_shell(title)
-    return if title == 'Search results'
+  def dashboard_table_content_tooltip_shell(table_type)
```
colinxfleming: It makes more sense to use the non-localized table_type here, so we're rolling with that.

montanezp:
The only thing I see is this to change below. Good otherwise!

$0 sent (0 patients) = $0 mandado (0 pacientes)
$0 pledged (0 patients) = $0 prometido (0 pacientes)

lomky:
Tagging in the final pass ticket to investigate the exceptions mentioned! #1585

app/helpers/dashboards_helper.rb
```diff
@@ -1,25 +1,25 @@
...
-    vm_options.map! { |option| option.to_sym }
-
-    vm_options.map { |option| [enum_text[option], option.to_sym] }
+    enum_text = { not_specified: t('dashboard.helpers.voicemail_options.not_specified'),
+                  no: t('dashboard.helpers.voicemail_options.no'),
+                  yes: t('dashboard.helpers.voicemail_options.yes') }
+    Patient.voicemail_preference
+           .values
+           .map { |option| option.to_sym }
+           .map { |option| [enum_text[option], option.to_sym] }
```
lomky: Suggest maybe an explainer comment for why we're doing a series of maps? It's not super complicated, but it does take reading a couple times to parse. Might be worth having the intent spelled out to avoid 'fixes' later.

app/views/calls/create.js.erb
```diff
@@ -11,9 +11,30 @@ if (URL.indexOf(substring) !== 1) {
 }
 
 if (location.pathname == Routes.authenticated_root_path) {
-  $("#call_list").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your call list', table_type: 'call_list', patients: current_user.ordered_patients(current_line), sortable: true, autosortable: true }) %>")
-  $("#completed_calls").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Your completed calls', table_type: 'completed_calls', patients: current_user.recently_called_patients(current_line), autosortable: true }) %>")
-  $("#urgent_patients").html("<%= escape_javascript( render partial: 'dashboards/table_content', locals: { title: 'Urgent cases', table_type: 'urgent_patients', patients: Patient.urgent_patients(current_line), autosortable: true }) %>")
+  $('#call_list').html(
```
lomky: :green_heart: :green_heart: :green_heart:  the cleaned up js!

config/locales/es.yml
```diff
@@ -19,8 +19,18 @@ es:
   dashboard:
     activity_log:
       label: Actividad de línea de llamada %{current_line}
```
montanezp: Actividad de línea de _llamadas_

config/locales/es.yml
```diff
@@ -19,8 +19,18 @@ es:
...
+    helpers:
+      voicemail_options:
+        'no': No dejes un correo de voz
```
montanezp: No dejen mensajes de voz

config/locales/es.yml
```diff
@@ -19,8 +19,18 @@ es:
...
+        not_specified: Sin instrucciones; no ID VM
+        'yes': Correoo de voz bien, ID bien
```
montanezp: Si, se permiten mensajes de voz

app/helpers/dashboards_helper.rb
colinxfleming: good catch! I'll add a 'here's what we're doing' comment.

colinxfleming:
I'm gonna hold on the budget bar stuff because it's kinda complicated, and it's obvious it'll work. I think we're going to need to rewire how we surface the translations to deal with it (session-based rather than URL based), also.

PR #1594, opened 2019-02-09
Translate Spanish event strings (and move them out of the model) 

### LINKED ISSUE #1578: I18n Spanish: Events
## What are we trying to do?

This ticket is part of the DARIA translation effort. The goal is to implement all our text in both English and Spanish through the i18n* process. 

This ticket is for the Event view.  
Code link:  
app/models/event.rb - `EVENT_TYPES`  
may require more files updated, due to model string changes.

_This may not be a good intro i18n ticket_

The current English text should be moved to `config/locales/en.yml`,   
and new Spanish text should be added to `config/locales/es.yml`.

**Google translate is perfectly acceptable!**  
PRs will be reviewed by a fluent CM to correct any weirdness.

### Steps

For each view  
   `en.yml`:
   ```
   patient:
     dashboard:
       appt_date: Appointment date
    ```  
    `es.yml`:
    ```
    patient:
      dashboard:
        appt_date: Día de la cita
    ```

### Completed View Example

`i18n-tasks normalize`

Part of #1382

colinxfleming:

This translates the view-facing strings in the Event model and moves them over to the `events_helper`. This was the least intrusive way I could come up with to do this that didn't break the existing Event model.

* adds spanish translations for the event log
* moves the user-facing text out of the model

Note that we can't see the event view because this partial loads async, which can't set locale.

* Fixes #1578  ~(not done, because we need to move to session)~

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -42,17 +42,9 @@ class Event
     end
   end
 
-  def event_text
-    case event_type
-    when 'Pledged'
-      "sent a $#{pledge_amount} pledge for"
-    when 'Left voicemail'
-      'left a voicemail for'
-    when "Couldn't reach patient"
-      "called, but couldn't reach"
-    when 'Reached patient'
-      'reached'
-    end
+  # remove spaces and punctuation. A sin method because we did this as strs not syms.
+  def underscored_type
+    event_type.gsub(' ', '_').gsub(/\W/, '').downcase
   end
 
   # Clean events older than three weeks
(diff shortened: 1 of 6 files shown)


app/models/event.rb
```diff
@@ -42,17 +42,9 @@ def glyphicon
...
-    when 'Pledged'
-      "sent a $#{pledge_amount} pledge for"
-    when 'Left voicemail'
-      'left a voicemail for'
-    when "Couldn't reach patient"
-      "called, but couldn't reach"
-    when 'Reached patient'
-      'reached'
-    end
+  # remove spaces and punctuation. A sin method because we did this as strs not syms.
```
colinxfleming: this is to match the i18n key.

colinxfleming:
@montanezp for review too!

colinxfleming:
Bringing in the Spanish PR makes render_async work, so this will fix instead of bump the issue.

PR #1595, opened 2019-02-10
Add clinic finder gestation filter

### LINKED ISSUE #1365: Clinic finder: Implement gestation filter
TK

colinxfleming:

Adds gestation filter as a parameter you can filter down in clinic finder. If, for example, you are a CM working with someone who's 18w, clinics that don't go that far are unhelpful at best and misdirecting at worst. This pops those out early via some filtering in the controller.

Tests TK

* Adds gestation filter as a thing you can filter clinics by in clinic finder
* Adds gestation filter to the clinic finder results table
* includes #1596 

Getting a little cramped, but it works!

image

* Fixes #1365

### DIFF
--- a/app/controllers/clinicfinders_controller.rb
+++ b/app/controllers/clinicfinders_controller.rb
@@ -4,15 +4,15 @@ class ClinicfindersController < ApplicationController
   def search
     return head :bad_request if params[:zip].blank?
 
-    # Clinics intentionally excluded from ClinicFinder are assigned the zip 99999.
-    # e.g. so a fund can have an 'OTHER CLINIC' catchall.
-    excluded_zip = '99999'
+    gestation = calculate_gestation params
+    clinics = Clinic.with_zip_and_gestational_limit_above gestation
+                    .where(accepts_naf: )
+                    .where(:accepts_naf.in =>  true)
 
-    clinics = Clinic.where(:zip.nin => [nil, '', excluded_zip])
     # Get all clinics except those with invalid zip codes.
     clinic_finder = ClinicFinder::Locator.new(
       clinics,
-      gestational_age: (params[:gestation].to_i || 0),
+      gestational_age: gestation_limit,
       naf_only: params[:naf_only] == '1',
       medicaid_only: params[:medicaid_only] == '1'
     )
@@ -20,5 +20,13 @@ class ClinicfindersController < ApplicationController
     @nearest = clinic_finder.locate_nearest_clinics params[:zip]
     @cheapest = nil # clinic_finder.locate_cheapest_clinic
     respond_to { |format| format.js }
+  end
+
+  private
+
+  def calculate_gestation(params)
+    weeks = params[:gestation_weeks].present? ? params[:gestation_weeks].to_i * 7 : 0
+    days = params[:gestation_days].present? ? params[:gestation_days].to_i : 0
+    (weeks + days).to_i
   end
 end
--- a/app/controllers/clinicfinders_controller.rb
+++ b/app/controllers/clinicfinders_controller.rb
@@ -6,6 +6,7 @@ class ClinicfindersController < ApplicationController
 
     gestation = calculate_gestation params
     filtered_clinics = Clinic.where(:zip.nin => [nil, '', Clinic::EXCLUDED_ZIP])
+                             .gestational_limit_above(gestation)
                              .where(:accepts_naf.in => [true, params[:naf_only] == '1'])
                              .where(:accepts_medicaid.in => [true, params[:medicaid_only] == '1'])
 
--- a/app/controllers/clinicfinders_controller.rb
+++ b/app/controllers/clinicfinders_controller.rb
@@ -8,9 +8,10 @@ class ClinicfindersController < ApplicationController
     # e.g. so a fund can have an 'OTHER CLINIC' catchall.
     excluded_zip = '99999'
 
+    clinics = Clinic.where(:zip.nin => [nil, '', excluded_zip])
     # Get all clinics except those with invalid zip codes.
     clinic_finder = ClinicFinder::Locator.new(
-      Clinic.where(:zip.nin => [nil, '', excluded_zip]),
+      clinics,
       gestational_age: (params[:gestation].to_i || 0),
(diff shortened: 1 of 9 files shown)


app/controllers/clinicfinders_controller.rb
```diff
@@ -4,20 +4,23 @@ class ClinicfindersController < ApplicationController
...
-
-    # Get all clinics except those with invalid zip codes.
-    clinic_finder = ClinicFinder::Locator.new(
-      Clinic.where(:zip.nin => [nil, '', excluded_zip]),
-      gestational_age: (params[:gestation].to_i || 0),
-      naf_only: params[:naf_only] == '1',
-      medicaid_only: params[:medicaid_only] == '1'
-    )
+    gestation = calculate_gestation params
+    filtered_clinics = Clinic.where(:zip.nin => [nil, '', Clinic::EXCLUDED_ZIP])
```
colinxfleming: this is a lot of logic for a controller - could move this set of wheres/filters to the Clinic model, or let it be here. Not sure if I have an opinion.

app/views/clinicfinders/_search_form.html.erb
```diff
@@ -1,11 +1,15 @@
...
+    <small>This is an experimental feature.</small>
     <span class='glyphicon glyphicon-plus-sign pull-right'></span>
-  </h5>
+  </h4>
+
 
   <div class='col-sm-12 hide' id='clinic-finder-search-form'>
-    <h6>A tool for sorting clinics by cost and distance.</h6>
+    <h5>
+      Filters
```
colinxfleming: I don't think filters is quite the right verbiage here but I don't know what else to call these fields.

app/views/clinicfinders/_search_form.html.erb
```diff
@@ -15,9 +19,18 @@
...
-          <%#= f.select :gestation, weeks_options, label: "Patient's approx. gestation at appt" %>
+        <div class="col-sm-3 info-form-right">
+          <%= f.select :gestation_weeks,
+                       options_for_select(weeks_options, patient.last_menstrual_period_now_weeks),
+                       label: 'Gestation above',
+                       autocomplete: 'off' %>
+        </div>
+
+        <div class="col-sm-3 info-form-right hide-label">
+          <%= f.select :gestation_days,
```
colinxfleming: Is thinking in weeks AND days really useful here? Could we get most of the way there if we were just like `weeks.to_i * 7`

colinxfleming:
ready to go, assuming the build passes!

app/controllers/clinicfinders_controller.rb
lomky: Searches always flag like this for me. I don't have a strong opinion either.

app/controllers/clinicfinders_controller.rb
```diff
@@ -4,14 +4,23 @@ class ClinicfindersController < ApplicationController
...
     @cheapest = nil # clinic_finder.locate_cheapest_clinic
     respond_to { |format| format.js }
   end
+
+  private
+
+  def calculate_gestation(params)
+    weeks = params[:gestation_weeks].present? ? params[:gestation_weeks].to_i * 7 : 0
+    days = params[:gestation_days].present? ? params[:gestation_days].to_i : 0
+    (weeks + days).to_i
```
lomky: Gosh I love ruby.

app/helpers/clinicfinders_helper.rb
```diff
@@ -8,6 +8,8 @@ def parse_clinicfinder_field(clinic, field)
       "#{clinic[field].round(2)} miles"
     when :cost
       number_to_currency clinic[field], precision: 0
+    when :gestational_limit
+      clinic.gestational_limit.present? ? "Goes to #{(clinic.gestational_limit / 7).to_i}w" : 'Not specified'
```
lomky: Possible rounding weirdness. I can see this being a bit odd, if the clinic takes patients up to, say, 4.5 weeks, if this displays '4 weeks', and the patient we're looking to place is at 4.25 weeks, would this misinform our CMs?

app/views/clinicfinders/_search_form.html.erb
lomky: "Filter Clinics By: "?

app/views/clinicfinders/_search_form.html.erb
lomky: If it's going to make a large cost difference to our patients, I think it would be useful. Worse case the CMs will only use the week dropdown, best case we catch edge cases where days matter.

app/helpers/clinicfinders_helper.rb
colinxfleming: I think you're right about this. I'll change to XwYd (e.g. 5w6d, the format we use elsewhere in the patient view).

app/controllers/clinicfinders_controller.rb
colinxfleming: Let's leave it for now and extract if we decide we need to. My suspicion is that this won't get touched too often and we won't need to do this search elsewhere.

app/views/clinicfinders/_search_form.html.erb
colinxfleming: love it

app/views/clinicfinders/_search_form.html.erb
colinxfleming: yeah, I think clinics themselves (at least in DC) tend to work in days, so going with w/d instead of just weeks is the more portable option. Thank you very much for the gut check!

app/controllers/clinicfinders_controller.rb
colinxfleming: haaaaaard saaaaame

app/views/clinicfinders/_search_form.html.erb
colinxfleming: erred a little on the side of verbosity since this is sort of doing a sort-and-filter, but I think it works either way.

colinxfleming:
I think @lomky 's game plan of merging this and then refixing #1608 is a good call! Will proceed accordingly after getting approval.

PR #1596, opened 2019-02-10
Remove filters from ClinicFinder in favor of Mongoid/ActiveRecord#where

colinxfleming:

While working thru #1595 it kinda became apparent that we were just reimplementing activerecord filters in the clinic finder code. This strips out those filters in favor of activerecord. I separated this change out because it's sort of independent of the larger work of setting up gestation limit filtering.

* deletes filter methods from clinicfinder
* moves it all to the controller (for now)

no view changes

* Bumps #1365


BintLopez:
retro 👍 -- nice refactor!

PR #1527, opened 2018-11-21
Spanish switch link

DarthHater:
I rule and have completed some work on Case Manager that's ready for review!

This allows someone to switch from English to Spanish, and vice versa

This pull request makes the following changes:
* Addition of new logic to output a link in navbar template
* Small change to application controller to add a query string param of the locale (so you can keep it persisted)

screen shot 2018-11-20 at 5 11 50 pm

It relates to the following issue #s: 
* Bumps #1382

PR #1597, opened 2019-02-11
Switch to Spanish Link

DarthHater:

I borked #1527 so here's a new cleaner PR

This adds a link to switch from Spanish and English for the app, as well as enables default url options to include the locale.

* Tiny changes to the application controller
* A helper for outputting the spanish/english link
* Adding the helper into the navbar view

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,6 +22,11 @@ class ApplicationController < ActionController::Base
     I18n.locale = params[:locale] || I18n.default_locale
   end
 
+  def default_url_options
+    return { :locale => I18n.locale } if I18n.locale != I18n.default_locale
+    {}
+  end
+
   private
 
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -18,6 +18,10 @@ def configure_permitted_parameters
     end
   end
 
+  def default_url_options
+    { locale: I18n.locale }
+  end
+
   def set_locale
     I18n.locale = params[:locale] || I18n.default_locale
```
lomky: Do we set `default_locale`?

DarthHater: We don't need to unless we want to be explicit. It defaults to `en`.

From: "The I18n library will use English as a default locale, i.e. if a different locale is not set, :en will be used for looking up translations."

PR #1598, opened 2019-02-11
Don't tweak default url options if the locale is english

colinxfleming:
The special test stuff felt weird to me, so I tried something out controllerside and it seems to work. Has the benefit of keeping the URLs pretty!


colinxfleming:
resolved via the main branch, so closing

PR #1606, opened 2019-02-14
WIP Clinic views translations

jpkarvonen:
I added keys and values for views/clinics/index.html in en.yml and input `t'clinics.index.header'` into the `h1` tag on that page. Instead of getting 'Clinics', as expected, I got 'Header' as you can see below:

screen shot 2019-02-13 at 9 53 39 pm

I will continue to plug away at this when I can, but let me know if someone knows what's causing the unexpected result.


lomky:
We went a bit over the top and pushed through a bunch of Spanish before the next meetup. We'll get you on another one Tuesday!

jpkarvonen:
Sounds good! I'm looking forward to tomorrow :)

PR #1611, opened 2019-02-17
install webpack gem and run rails webpack:install

colinxfleming:

I wanted to see how hard it would be to set up webpack in daria, and it turns out, not very hard. I'm tagging in a bunch of people for review because I'd like a wide range of opinions on this, particularly from people who I know work in JS a lot.

This PR sets up the webpacker gem to compile rails assets. The next steps from this PR would be porting over our existing coffeescript stuff to ES6 and depreciating the sprockets asset pipeline in favor of webpack stuff. This would get us some longer term stability and keep us using what are considered Sensible, Well-Established Standards. (`webpack` is not the rails default, but way more people know ES6 than coffeescript, so we're making a technical decision here in favor of the lingua franca of frontend engineers.)

I am not really a js specialist and I don't know what I don't know, so please check these assumptions and make sure that I'm not off base on this:

* webpack is a reasonably stable and straightforward way of handling client side JS, which we should use instead of sprockets
* we should port over our SCSS to be part of the webpack compiled assets as well
* this would let us use stuff from npm pretty seamlessly
* we still want to have rails ERB views be the primary way we handle html generating (e.g. this is not an indicator that we want to split the frontend and backend and move to react or vue)
* porting our existing javascript and sass to be webpack friendly is not a pain in the behind
* we're not scared of the security stuff, because we can be covered by good security tools in CI. (A thing that does worry me about JS generally is the fact that npm is the wild west, but I'm okay with this.
* we can still do most of our testing of JS on the systemtests level, so we don't need to worry about setting up jest or anything like that. 

what do you all think? 

* install webpack

no issues


colinxfleming:
I'm tagging in a bunch of people for review because I'd like a wide range of opinions on this, particularly from people who I know work in JS a lot.

This PR sets up the webpacker gem to compile rails assets. The next steps from this PR would be porting over our existing coffeescript stuff to ES6 and depreciating the sprockets asset pipeline in favor of webpack stuff. This would get us some longer term stability and keep us using what are considered Sensible, Well-Established Standards. (`webpack` is not the rails default, but way more people know ES6 than coffeescript, so we're making a technical decision here in favor of the lingua franca of frontend engineers.)

I am not really a js specialist and I don't know what I don't know, so please check these assumptions and make sure that I'm not off base on this:

* webpack is a reasonably stable and straightforward way of handling client side JS, which we should use instead of sprockets
* we should port over our SCSS to be part of the webpack compiled assets as well
* this would let us use stuff from npm pretty seamlessly
* we still want to have rails ERB views be the primary way we handle html generating (e.g. this is not an indicator that we want to split the frontend and backend and move to react or vue)
* porting our existing javascript and sass to be webpack friendly is not a pain in the behind
* we're not scared of the security stuff, because we can be covered by good security tools in CI. (A thing that does worry me about JS generally is the fact that npm is the wild west, but I'm okay with this.
* we can still do most of our testing of JS on the systemtests level, so we don't need to worry about setting up jest or anything like that. 

what do you all think?

tingaloo:
Looks good. I can confirm we can starting using es6. Webpack definitely beats the alternatives.

colinxfleming:
thanks everyone for your insight and thoughtfulness here! I especially appreciate @geoffreyyip 's point about churn and @rudietuesdays 's experience with webpack. In the immediate term I'll be making some adjustments to our readme and build to reflect that this is a thing now, like adding some security tools and a static analyzer probably. The next steps after that are to gradually port over our existing JS to ES6 and get ourselves out of coffeescript.

PR #1616, opened 2019-02-21
Fix Some Flakey Tests

DarthHater:

This is an omnibus for fixing up some flakey tests on CI, hopefully

* TBD


colinxfleming:
we have a few flakey tests but this tends to be much less of a problem in the era of `minitest-optional_retry`, so closing it out.

Issue #1602, opened 2019-02-13
Add Practical Support view partial to patient view


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Add the Practical Support partial to the patient edit view for when we wire things together. We don't need it working entirely, but we do want to stub out the frontend for when we do want to wire it up. This is a ticket to set up the views.

* **What feature or behavior is this required for?**

The UI of practical support.

* **How could we solve this issue? (Not knowing is okay!)**

Basically implementing the form pictured in #1493 , which itself is the external fund thing on abortion information, pretty much. You're doing the following:

* Adding a practical support tab on the left sidebar in the patient edit view
* Making a new partial, `app/views/patients/_practical_support.html.erb` and giving it the header `Practical Support`
* Adding a stub inline form with the following fields: Type (select), Confirmed (checkbox), Provided by (select), Remove (button) 

See the documentation of rails_bootstrap_forms to see how to make a form not backed with a model. You can probably jumpstart by looking at something like `app/views/_abortion_information.html.erb` and cribbing some of that.

* **Anything else?**
Stems from #1493

colinxfleming:
saving this for @harumhelmy

PR #1620, opened 2019-02-24
Practical Support Config

### LINKED ISSUE #1619: Add practical support config option

We'll want funds to be able to set their own categories of practical support. We should do this with a config so someone can just add this in the admin panel and we aren't getting fancy.

Practical support implementation

* new config for practical support options
* should just be comma separated strings
* should have a practical_support_options helper method that works the same way the insurance one does

* **Anything else?**

stems from #1493

lomky:

Implementation & tests for the Practical Support Config options!

* Adds the practical support config
* Adds a helper for displaying them
* Includes i18n support
* Helper tests
* Config system test

Not in this ticket:
* Any tests relying on the patient view of practical support (probably should do in/after #1602)

screenshot from 2019-02-24 08-46-52

* Fixes #1619
* Bump #1493


lomky:
Ready for Spanish review! @montanezp 

Is there a good option for a gender neutral 'Companion' in Spanish?

config/locales/en.yml
```diff
@@ -177,6 +177,11 @@ en:
         unknown: Don't know
       language:
         English: English
+      practical_support:
+        companion: Companion
```
colinxfleming: @harumhelmy can you verify that these are good as starter, general categories of practical support?

test/test_helper.rb
```diff
@@ -45,6 +45,12 @@ def create_insurance_config
                     config_value: { options: insurance_options }
   end
 
+  def create_practical_support_config
+    practical_support_options = ['Metallica Tickets', 'Clothing']
```
colinxfleming: 🤘

config/locales/es.yml
```diff
@@ -168,6 +168,11 @@ es:
         unknown: No Se
       language:
         English: Inglés
+      practical_support:
+        companion: Compañera
+        lodging: Alojamiento
```
montanezp: Hospedaje

config/locales/es.yml
```diff
@@ -168,6 +168,11 @@ es:
         unknown: No Se
       language:
         English: Inglés
+      practical_support:
+        companion: Compañera
```
montanezp: Acompañante

config/locales/en.yml
lomky: Thanks for gathering the feedback!

PR #1626, opened 2019-02-25
Translate accountants page

### LINKED ISSUE #1583: I18n Spanish: Accounting
## What are we trying to do?

This ticket is part of the DARIA translation effort. The goal is to implement all our text in both English and Spanish through the i18n* process. 

This ticket is for the Accounting views.  
Code link:  
app/views/accountants/index.html.erb  
app/views/accountants/_search_form.html.erb  
app/views/accountants/_table_content.html.erb  

The current English text should be moved to `config/locales/en.yml`,   
and new Spanish text should be added to `config/locales/es.yml`.

**Google translate is perfectly acceptable!**  
PRs will be reviewed by a fluent CM to correct any weirdness.

### Steps

For each view  
   `en.yml`:
   ```
   patient:
     dashboard:
       appt_date: Appointment date
    ```  
    `es.yml`:
    ```
    patient:
      dashboard:
        appt_date: Día de la cita
    ```

### Completed View Example

`i18n-tasks normalize`

Part of #1382

colinxfleming:

Run the accountants workflow thru i18n.

* i18n-icize the accountants page

image

* Fixes #1583

### DIFF
--- a/app/views/accountants/_table_content.html.erb
+++ b/app/views/accountants/_table_content.html.erb
@@ -1,22 +1,22 @@
 <table class="table border-side">
-  <table class="table border-side">
-    <thead>
-      <tr>
-        <th>Clinic Name</th>
-        <th>PT initials</th>
-        <th><%= FUND %> payout</th>
-        <th>Weeks at Procedure</th>
-        <th>Pledge Sent Date</th>
-        <th>Check #</th>
-        <th>Date of Check</th>
-        <th></th>
+  <thead>
+    <tr>
+      <th><%= t 'accountants.table_content.clinic' %></th>
+      <th><%= t 'accountants.table_content.pt_initials' %></th>
+      <th><%= t 'accountants.table_content.fund_payout', fund: FUND %></th>
+      <th><%= t 'accountants.table_content.weeks_at_procedure' %></th>
+      <th><%= t 'accountants.table_content.pledge_sent' %></th>
+      <th><%= t 'accountants.table_content.check_number' %></th>
+      <th><%= t 'accountants.table_content.date_of_check' %></th>
+      <th></th>
+    </tr>
+  </thead>
+
+  <tbody id="patient_information_form">
+    <% patients.each do |patient| %>
+      <tr id='row-<%= patient.id %>'>
+        <%= render 'accountants/table_content_row', patient: patient %>
       </tr>
-    </thead>
-    <tbody id="patient_information_form">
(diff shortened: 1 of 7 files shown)


app/views/accountants/_table_content.html.erb
```diff
@@ -1,22 +1,22 @@
...
-        <th></th>
+  <thead>
+    <tr>
+      <th><%= t 'accountants.table_content.clinic' %></th>
+      <th><%= t 'accountants.table_content.pt_initials' %></th>
+      <th><%= t 'accountants.table_content.fund_payout', fund: FUND %></th>
+      <th><%= t 'accountants.table_content.weeks_at_procedure' %></th>
+      <th><%= t 'accountants.table_content.pledge_sent' %></th>
+      <th><%= t 'accountants.table_content.check_number' %></th>
+      <th><%= t 'accountants.table_content.date_of_check' %></th>
```
lomky: Suggest moving most of these to a shared - I just ran into most of them in the patient pledge PR! #1621 

If they don't feel like a good case for a 'shared' namespace, I think it's more appropriate to name them for the pledge tree, but I don't feel strongly.

config/locales/es.yml
```diff
@@ -280,3 +280,16 @@ es:
     resolved_without_fund: Esto se utiliza para indicar que un paciente ya no necesita o quiere nuestros servicios.
     status_definition: Definición de estado
     urgent_cases: Estos son pacientes que requieren un poco más de atención. Esta lista es compartido entre todos los administradores de casos que trabajan en una sola línea. Los pacientes son eliminados de esta lista automáticamente después de que estén marcados con “compromiso enviado”, o marcado como “resuelto sin ayuda”.
+  accountants:
+    title: Tablero contable
```
montanezp: Tablero de contador

app/views/accountants/_table_content.html.erb
colinxfleming: These might be specialized to accountants, but if they're not I'll move them to the mongoid attributes namespace. Great call.

colinxfleming: After taking a stab at this the language is like EVER so slightly different, in such a way that I don't think this is a free basket rn. My vote is that we go with the redundancy for now in the name of expedience.

colinxfleming: (I will happily commit to cleaning this mess up.)

PR #1628, opened 2019-02-27
Get yarn working in Alpine

### LINKED ISSUE #1623: Set up docker for yarn
Since we're on webpack we'll need to update our docker build. Let's make it so `docker-compose build` and `docker-compose up` work again. (Yarn needs to be installed as an executable.)

DarthHater:

This gets yarn working in docker-compose

* Adds npm
* Installs yarn via npm (this gets us a fresher version of yarn without writing a LOT of code to do it)
* Modifies the docker-compose command to run `yarn install` before `rails s` so that the integrity check works

* Fixes #1623

### DIFF
--- a/.docker/entrypoint.sh
+++ b/.docker/entrypoint.sh
@@ -4,7 +4,7 @@ start_server() {
 }
 
 if bundle check; then
-    start_server
+    yarn install && start_server
 else
-    bundle install && start_server
+    bundle install && yarn install && start_server
 fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: .docker/Dockerfile
-    command: rails s -p 3000 -b 0.0.0.0
+    command: sh -c "yarn install && rails s -p 3000 -b 0.0.0.0"
     image: colinxfleming/dcaf_case_management:latest
     volumes:
       - .:/usr/src/app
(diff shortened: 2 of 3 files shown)


.docker/entrypoint.sh
```diff
@@ -4,7 +4,7 @@ start_server() {
 }
 
 if bundle check; then
```
colinxfleming: I remember you mentioned in slack that we weren't using this that you could tell. Does that continue to be the case?

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: .docker/Dockerfile
-    command: rails s -p 3000 -b 0.0.0.0
+    command: sh -c "yarn install && rails s -p 3000 -b 0.0.0.0"
```
colinxfleming: I'd really like to keep it so the up command is just `rails s -p 3000 -b 0.0.0.0` to be consistent with more general rails behavior, and `yarn install` feels closer to a `bundle install` than it does booting up puma. How much of a pain would it be to move this command to the dockerfile?

.docker/entrypoint.sh
DarthHater: As far as I know! I had this a while ago to do fun things with `docker` but I think it got removed at some point. We can nix it if you'd like.

docker-compose.yml
DarthHater: To do that we either need an `entrypoint.sh`, or something akin. The reason we have to do this is because the `yarn.lock` is copied into the Dockerfile with `COPY .` and this ensures that the yarn.lock integrity is valid. We could in theory exclude the yarn.lock from being copied in but I don't know how much I like that either. Open to ideas!

colinxfleming: I did some cursory googling and this seems not totally wild: https://medium.com/soulmates-ai/running-a-rails-app-with-webpacker-and-docker-8d29153d3446 Specifically the part about:

```
# config/environments/development.rb
config.webpacker.check_yarn_integrity = false
```

Issue #1633, opened 2019-03-03
Final Pass - User Facing Spanish Review


lomky:
This ticket is to capture the last needed Spanish translations _for the Case Manager facing sections_. Admin & Data manager sections _are not required_ for this review.

* **What are we trying to do?**

Browse around on the [Spanish Review](https://github.com/DCAFEngineering/dcaf_case_management/tree/spanish_review) branch and look for English. Look at the regular text, any helper text (the "?"s), error messages, drop down selections. Try clicking and editing things. Sort things. Edit the call list. Try to break things!

Do not worry about the user entered text entered in fields. Values in notes, the patient name itself, other contact relationship, etc are not what we're looking for here. If you're uncertain though, just go ahead and add a comment of it! 

Screen shot, or carefully describe in a comment here, the English.

Also look out for any errors related to translations.

* **What feature or behavior is this required for?**

User facing Spanish release!

* **How could we solve this issue? (Not knowing is okay!)**

Lotsa eyes.

lomky:
"Forgot your password?". Would also be nice to have a language swap button on this screen.

![screenshot from 2019-03-03 10-55-23](https://user-images.githubusercontent.com/6129479/53697782-ddff4080-3da2-11e9-96b9-111062dbfccc.png)

lomky:
Alert upon login: `translation missing: es.devise.sessions.user.signed_in`

![screenshot from 2019-03-03 10-56-39](https://user-images.githubusercontent.com/6129479/53697797-1272fc80-3da3-11e9-96ed-d891052ac726.png)

lomky:
Confirmation box for deleting a call: "Are you sure you want to remove this call from the call log?", "Cancel", "OK".

![screenshot from 2019-03-03 11-03-15](https://user-images.githubusercontent.com/6129479/53697890-00de2480-3da4-11e9-8c1f-39e38417cd05.png)

(Unrelated, this is also an ugly box so we should theme it)

colinxfleming:
From a fresh seed, logging a call for Other Contact:

![image](https://user-images.githubusercontent.com/3866868/53697981-dd67a980-3da4-11e9-9daa-dd96c2006d85.png)

colinxfleming:
Most Recent Note in the note expander on the patient dashboard:

![image](https://user-images.githubusercontent.com/3866868/53697993-fa03e180-3da4-11e9-9434-0367d0099ff2.png)

colinxfleming:
[...]

PR #1647, opened 2019-03-06
Fixing Spanish with Paulina

lomky:
Working with @montanez at hacknight to translate some remaining sneaky English and improve the Spanish we had to be more consistent. (Pledge is Promesa!)

* Bumps #1633


colinxfleming:
lgtm. I think it needs an i18n-tasks normalize, after that let's merge

Issue #1648, opened 2019-03-06
Spanish Bugs!


lomky:
More tk

_sometimes_ the locale is lost on login and logout?

Cancel/Ok buttons
Cancel/Next/Complete on Pledge gen

Error on log out: `translation missing: es.devise.failure.user.unauthenticated` (fixed in #1652)

User Page needs translated! (fixed in #1650 )

Weird behavior:
Insurance options - No se reset to Don't Know, both options in list. ( fixed in #1651  )
Database options not routing back to Spanish "Seguro Médico del Paciente"
Does _not_ affect language pref.

colinxfleming:
What do you think about opening up longer term tickets for the cancel/ok buttons, and sometimes the locale being lost?

The cancel/ok button is a particularly hard one since we've rarely touched the multimodal since setting it up, and I don't think the text on a few buttons should hold up a more general launch. (though maybe @montanezp would disagree.) 

The language not sticking on logout is a weird bug but not one that hits us hard, since at that point they aren't really interacting with patients I think?

lomky:
I'll break this out into real tickets tonight at the hacknight!

lomky:
All issues addressed or broken out into distinct issues.

PR #1650, opened 2019-03-09
User page in Spanish!

lomky:

Not sure how I forgot the user page in the translation list, but shrug! Here it is.

* Translate the user page

Screenshot from 2019-03-09 12-10-14

* Bumps #1648 

Ready for spanish review @montanezp

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -199,10 +199,10 @@ es:
         phone: Teléfono
         state: Estado
         street_address: Dirección
-        zip: Código postal```
+        zip: Código postal
       user:
         current_password: Contraseña actual
-        email: Email
+        email: Correo Electrónico/Email
         name: Nombre y apellido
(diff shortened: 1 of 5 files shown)


config/locales/es.yml
```diff
@@ -201,7 +201,12 @@ es:
         street_address: Dirección
         zip: Código postal```
```
montanezp: Double checking the apostrophes after "postal" are intentional. If not, please remove.

config/locales/es.yml
```diff
@@ -201,7 +201,12 @@ es:
...
       user:
+        current_password: Contraseña actual
         email: Email
```
montanezp: Correo Electrónico/Email

config/locales/es.yml
lomky: they were not! Thanks for spotting!

PR #1661, opened 2019-03-24
Turn off check_yarn_integrity in development and add sassc to docker

### LINKED ISSUE #1623: Set up docker for yarn

colinxfleming:

Our docker build has been busted for a minute because of the sassc requirement and because of the yarn integrity check preventing rails s from starting. This jumps off @DarthHater 's work in #1628 . (I couldn't get the app to start in that branch, unfortunately.) As of 9bedaa0  I'm able to get things to start again.

I recognize that turning off the integrity check in dev is not a great solution to this but I'm willing to make this temporary sacrifice for the following reasons:

* we had a hacknight happen where people who weren't rails savvy tried to set up the project using docker, and weren't able to get it rigged up because this wasn't set up. This put @rudietuesdays who was doing the onboarding in a really tough spot and cost us some vols.
* the attack surface here seems super small to me. The libraries we're pulling in are webpacker only right now and we aren't pulling over other deps. To get hit by a miner or something worse one of those libraries would have to succumb to it, and it still wouldn't hit our production environment and expose patient data.

I think we should aim to turn this back on sooner rather than later, but I'm weighing this against DARIA's docker version continuing to be broken, and I think the current status is costing us volunteer energy. In the event that this gets merged I will make a ticket to explore how we turn on the integrity check in development.

@DarthHater can you weigh in on this plan?
@rudietuesdays can you weigh in on this plan and confirm that this branch gets the docker build working for you again?

* add `sassc`
* turn `yarn_integrity_check` off in development
* some other docker cleanup

no views

* Fixes #1623


colinxfleming:
we have a hacknight tomorrow, so sorry, gotta merge this. Made an issue to turn the integrity check back on in #1667

PR #1663, opened 2019-03-28
Get build green again: ruby 2.6.2, webpacker 4

colinxfleming:

there are some real and semi-real things making our build fail, which is preventing us from shipping spanish. this fixes that.

I'm justifying temporarily shutting off yarn audit because we don't currently have any actual js packages in webpack. I will file an issue for that. 

* start using ruby 2.6.2
* bump webpacker to v4, because I thought that would fix the yarn audit error, but turns out no
* temporarily comment out yarn audit

no issues


colinxfleming:
yarn issue popped in #1665 . thanks @lomky !

PR #1668, opened 2019-04-02
Remove secure_headers in favor of default CSP

colinxfleming:

Rails got a CSP system in 5.1 and then `nonce` got added in 5.2. The person who runs secure_headers appears to be thinking of archiving it. (See https://github.com/twitter/secure_headers/issues/375) 

Since rails CSPs have caught up, let's swap this to rails.

* Remove `secure_headers` gem
* Set up rails CSP
* use new rails CSP nonce syntax (`nonce: true`) in JS junk

no issues or views


config/initializers/content_security_policy.rb
```diff
@@ -1,25 +1,17 @@
-# Be sure to restart your server when you modify this file.
+Rails.application.config.content_security_policy do |policy|
+  policy.default_src :self
+  policy.font_src    :self, :https, :data
+  policy.img_src     :self, :https, :data
+  policy.object_src  :none
+  policy.font_src    :self, 'fonts.gstatic.com'
+  policy.connect_src :self
+  policy.script_src  :self, :unsafe_eval
```
DarthHater: SUGGESTION: You can likely just do :unsafe_eval if the env is `development`, pretty sure in a production setting it SHOULD work right :)

colinxfleming: I think we actually have to do this because jquery uses eval in a bunch of places! Unfortunately. I'm curious as to whether there's a way to get jquery out but that is another issue for another time probably.

PR #1672, opened 2019-04-03
Fix missing location and distance translations

### LINKED ISSUE #1657: Clinic finder: Location and distance headers do not translate

If you use the clinic finder utility in spanish mode, the result table does not translate the Distance and Location headers. They should be, respectively, `Distancia` and `Ubicación`.

Translation consistency!

I'm filing this as a ticket because it might require wrenching around in a helper. The view that builds this is in `app/views/clinicfinders/result_table`.

* **Anything else?**

not a blocker for spanish deployment.

jmhooper:
Ref https://github.com/DCAFEngineering/dcaf_case_management/issues/1657

I fixes some missing Spanish translations in the clinic finder.

* Fixes #1657

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -193,8 +193,10 @@ en:
         accepts_naf: Accepts NAF
         active: Active
         city: City
+        distance: Distance
         fax: Fax
         gestational_limit: Gestational limit (in days)
+        location: Location
         name: Name
         phone: Phone
         state: State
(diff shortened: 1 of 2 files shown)


lomky:
Thank you for the PR @jmhooper!

@montanezp could you doublecheck our Spanish here? Just a couple words.

lomky:
Ignoring the circle fail - its the fork bug again.

colinxfleming:
I feel safe merging this without a full spanish review, since it's pretty minor.

Thank you very much @jmhooper, we're super glad to have you!

Issue #884, opened 2017-02-27
Leave our case managers a nice note!


colinxfleming:
When our case managers log in, we show them a nice message of encouragement, randomly selected from the strings in the array `app/helpers/welcome_message_helper.rb`. 

I would encourage you to add something to this array during the weekend!

colinxfleming:
@cmacaulay

sofijaj:
I love this <3 thanks for doing it, CMs will appreciate it for sure.

Ryan-Koch:
https://github.com/DARIAEngineering/dcaf_case_management/pull/2821

PR #1674, opened 2019-04-03
Adds encouraging message in French

bravogolfgolf:

Adds encouraging message in French

1) Adds additional message to welcome_message_helper.rb welcome_messages

* Bumps #884


colinxfleming:
We're super glad to have you, @bravogolfgolf ! Thanks you for adding to the messages list, this is in need of some freshening up!

PR #1676, opened 2019-04-03
Fixing the Build

lomky:

Cleaning up build failures previously disguised by our friendly' circle-fork-bug.

* i18n-tasks normalize
* Update fields on the Data Entry tests to match standardized field names


colinxfleming:
thanks for this fixup @lomky !

PR #1679, opened 2019-04-07
Convert clinic_finders.coffee to ES6 as a proof of concept

colinxfleming:

Port over clinic-finders js as an es6 thing as a proof of concept.

We should continue loading turbolinks, jquery, ujs, and related stuff via sprockets for the time being. Unfortunately, we have to double-load jquery as part of the pack.

After this gets approved and merged, I'll split out the existing coffeescript files into their own 'port this file' issues.

* install some yarn deps
* port over junk
* other place fixes to get CI to pass

no views, no issues


colinxfleming:
verified that @rudietuesdays 's commits work, so this lgtm!

colinxfleming:
invoking godmode since @rudietuesdays signed off in her comment.

PR #1688, opened 2019-04-13
Abbreviate semanas in the same places as days

### LINKED ISSUE #1656: Abbreviate semanas and dias in a few places

If you look at a table on the dashboard in spanish, you'll see something like this:

image

Note that in the weeks along column, this shows up as X semanas, Y dias, but in english we abbreviate it:

image

We do this in two places that I know of -  dashboard table content, and under the Weeks Along at Intake input on the patient edit view. We can and should get these in agreement by changing the spanish to abbreviate `X semanas Y días` to `Xs Yd`.

Consistency!

Probably by wrenching in the localization. Possibly by wrenching in the LMP calculation code, which I think is in `last_menstrual_period_measurable.rb`.

* **Anything else?**

not a blocker for spanish release.

colinxfleming:

There was a slight difference in a few spots (the call list, current LMP) where we were not abbreviating `semanas` and `días`, but abbreviating weeks and days. This PR gets that to match across concerns.

This was a good exercise in refactoring, it turns out! our LMP calculators were doing a little too much in terms of spitting out strings, and the path of least resistance involved reducing those methods down to just 'spit out numbers and let templates handle the rest'. As a consequence, the changes involve a lot of model wrenching, but nbd.

I got preapproval from @montanezp on these text changes.

* make abbreviation of `semanas` and `días` match the abbreviation of weeks and days
* remix the LMP concern in the process

image
image

* Fixes #1656


app/models/concerns/last_menstrual_period_measureable.rb
```diff
@@ -2,42 +2,45 @@
 module LastMenstrualPeriodMeasureable
```
lomky: This refactor is :nail_care:!

PR #1693, opened 2019-04-23
Moves autosave functionality to separate file

### LINKED ISSUE #1686: JS: Port over autosave part of patients.coffee from coffeescript to ES6

L57-58 and L65-66 of patients.coffee into app/src/javascript/autosave.js

* **Anything else?**

bravogolfgolf:

Moves autosave functionality from patients.coffee to autosave.js

* Removes two function from patients.coffee
* Adds two functions to autosave.js

* Fixes #1686

### DIFF
--- a/app/javascript/src/autosave.js
+++ b/app/javascript/src/autosave.js
@@ -1,7 +1,9 @@
-$(document).on("change", ".edit_patient", () => {
-    $(this).submit();
-});
+$(document).on("turbolinks:load", () => {
+    $(document).on("change", ".edit_patient", () => {
+        $(this).submit();
+    });
 
-$(document).on("change", ".edit_external_pledge", () => {
-    $(this).submit();
+    $(document).on("change", ".edit_external_pledge", () => {
+        $(this).submit();
+    });
 });
--- a/app/javascript/src/autosave.js
+++ b/app/javascript/src/autosave.js
@@ -1,0 +1,7 @@
+$(document).on("change", ".edit_patient", () => {
+    $(this).submit();
+});
+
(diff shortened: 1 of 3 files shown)


colinxfleming:
the code here looks good, doubly so because this has unearthed an issue with our webpack junk. For whatever reason this isn't autosubmitting, so we should investigate that.

app/javascript/src/autosave.js
```diff
@@ -0,0 +1,7 @@
+$(document).on("change", ".edit_patient", () => {
```
colinxfleming: Can you wrap this whole thing in `$(document).on('turbolinks:load', => {...}`? This is basically the turbolinks version of `$(document).ready` or whatever it is -- we want these handlers to attach to things once we're sure they've loaded.

```suggestion
$(document).on("turbolinks:load", () => {
  $(document).on("change", ".edit_patient", () => {
```

bravogolfgolf: I think I captured what you are looking for in the latest commit.

nikobozi:
I tried this out locally and apparently the `this` variable is not available within the new arrow function, which is why it's not being executed. After triggering the submit I get the invalid authenticity token error. Maybe it could be something related to [this issue](https://github.com/turbolinks/turbolinks/issues/430#issuecomment-444767978).  
The browser console also keeps complaining about CSP, but I'm not sure what the best option is

colinxfleming:
this has turned out to be a small nightmare. I'm closing this PR for the time being and we should revisit this after we take the steps laid out in #1686 of not double-loading bootstrap.

PR #1695, opened 2019-04-24
table_sort? more like "table_port" if you ask me.

### LINKED ISSUE #1684: JS: Port table_sorting.coffee from coffeescript to ES6

table_sorting.coffee into app/javascript/src/table_sorting.js

* **Anything else?**

jmhooper:

Convert `table_sort.coffee` to `table_sort.js`

* Convert the `table_sort.coffee` coffeescript into JS
* Import the new JS in the application pack
* Import the stupidtable plugin from NPM instead of vendoring it
* Add `document` to the list of globals that ESLint is aware of

* Fixes #1684


colinxfleming:
CI failures are from our known circle issues with forks, so smashing that merge button. Thank you @jmhooper !

PR #1703, opened 2019-05-31
Fix system test that is sad about a change to static control

colinxfleming:

Bootstrap form seems to handle static elements a little differently, so we're changing them to disabled and readonly input fields.

* `static_control` => `text_field, disabled: true, readonly: true`

no view change

no issues


colinxfleming:
godmoding since minor changes to get test to pass again

PR #1707, opened 2019-06-11
Convert call list coffeescript to ES6

### LINKED ISSUE #1680: JS: Port calls.coffee from coffeescript to ES6

calls.coffee into app/javascript/src/call_list_drag_and_drop.js

* **Anything else?**

colinxfleming:

guess who knows a lot about handling CSRF tokens in ES6 now!

There was a routing problem in here that was one of those 'how on earth did this ever work' things. glad to fix that.

* runs the script thru decaffeinate
* installs jquery-ui
* takes advantage of some es6 nice to haves

no view changes

* Fixes #1680


colinxfleming:
got an approval, merging

Issue #1698, opened 2019-05-18
Add option to start budget week on non-Monday


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Per correspondence, BRAF is experimenting with weekly budgets and planning on having their budget start day be on something that isn't monday, which isn't as helpful since there's a day shift. To that effect, we should make it so that this is config-able.

* **What feature or behavior is this required for?**

proper budget bar functioning if you are not dcaf or otherwise do not have monday/sunday budgets.

* **How could we solve this issue? (Not knowing is okay!)**

my hope is that this is a pretty minor thing, and that we can continue to keep mondays as the defaults. To that end, I'd say let's try adding a config called like `budget_start_day` that should be a day name (e.g. monday, tuesday, wednesday). The budget bar functions should use that day if exists, otherwise default to monday.

I don't know how tangled a web this is, I don't think very, but you know.

* **Anything else?**

not a terrible thing if this doesn't get done for a bit but we probably owe BRAF a solid.

This is probably a fun fishing expedition thru the budget bar code.

valeriecodes:
I'll take a look at this.

valeriecodes:
@colinxfleming I'm looking around and not finding where the budgets are calculated based on date, just the individual patient pledges. Would you be able to point me to a file in the codebase or a page in the app where I could see?

colinxfleming:
@valeriecodes sorry, my bad for not adding more detail for this! Should have added more of a grand tour.

The method that generates the report is here: https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/models/patient.rb#L143 

The place this actually shows up in a view is the patient dashboard here: https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/views/dashboards/_budget_bar.html.erb and the helper that calls that actual method is here: https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/controllers/concerns/budget_bar_calculable.rb

My guess is that the move here is to add a start day optional argument to `Patient#pledged_status_summary` and alter `BudgetBarCalculable#budget_bar_calculations` so that it passes in a config. If that doesn't wind up being a viable option or not super straightforward let me know!

mdworken:
@valeriecodes are you still working on this one? Mind if I grab it?

valeriecodes:
Go ahead @mdworken !

[...]

PR #1709, opened 2019-06-12
Adds option to start budget week on non-Monday #1698

mdworken:

Implements #1698

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -2,7 +2,7 @@
 module DashboardsHelper
   def week_range(date: Time.zone.now)
     week_start = date.beginning_of_week Config.start_day
-    week_end = date.end_of_week start_day
+    week_end = date.end_of_week Config.start_day
     week_start_string = l week_start, format: '%B %-d'
     week_end_string = if week_start.month == week_end.month
                         l week_end, format: '%-d'
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -1,8 +1,9 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
   def week_range(date: Time.zone.now)
-    week_start = date.beginning_of_week Config.start_day
-    week_end = date.end_of_week Config.start_day
+    start_day = Config.start_day
+    week_start = date.beginning_of_week start_day
+    week_end = date.end_of_week start_day
     week_start_string = l week_start, format: '%B %-d'
     week_end_string = if week_start.month == week_end.month
                         l week_end, format: '%-d'
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -1,6 +1,6 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
-  def week_range(date: Time.zone.now, start_day: :monday)
+  def week_range(date: Time.zone.now)
     week_start = date.beginning_of_week start_day
     week_end = date.end_of_week start_day
     week_start_string = l week_start, format: '%B %-d'
@@ -36,5 +36,11 @@ module DashboardsHelper
                  aria: { hidden: true }),
       tag(:span, class: ['sr-only'], text: 'Call')
     ]
+  end
+
+  def start_day
+    start = Config.find_or_create_by(config_key: 'start_of_week').options.try :last
+    start ||= "monday"
+    start.downcase.to_sym
   end
 end
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -1,7 +1,7 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
   def week_range(date: Time.zone.now)
-    week_start = date.beginning_of_week start_day
+    week_start = date.beginning_of_week Config.start_day
     week_end = date.end_of_week start_day
     week_start_string = l week_start, format: '%B %-d'
     week_end_string = if week_start.month == week_end.month
@@ -36,11 +36,5 @@ module DashboardsHelper
                  aria: { hidden: true }),
       tag(:span, class: ['sr-only'], text: 'Call')
     ]
-  end
-
-  def start_day
-    start = Config.find_or_create_by(config_key: 'start_of_week').options.try :last
-    start ||= "monday"
-    start.downcase.to_sym
   end
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -59,4 +59,10 @@ class Config
       end
     end
   end
+
+  def self.start_day
+    start = Config.find_or_create_by(config_key: 'start_of_week').options.try :last
+    start ||= "monday"
+    start.downcase.to_sym
+  end
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -19,7 +19,7 @@ class Config
     resources_url: 'A link to a Google Drive folder with CM resources. ' \
                    'Ex: https://drive.google.com/drive/my-resource-dir',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
-    start_of_week: 'The start day of your budget week, if it doesn't start on Monday. ex: Sunday.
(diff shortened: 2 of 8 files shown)


app/helpers/dashboards_helper.rb
```diff
@@ -37,4 +37,10 @@ def call_glyphicon
       tag(:span, class: ['sr-only'], text: 'Call')
     ]
   end
+
+  def start_day
```
DarthHater: SUGGESTION: This is virtually identical to `def self.start_of_week`. I would see if you could abstract it out into a helper somewhere and just call that rather than have two methods that are largely the same floating around.

app/models/patient.rb
```diff
@@ -311,4 +310,10 @@ def self.fulfilled_on_or_before(datetime)
                   updated_at: { '$lte' => datetime })
   end
 
+  def self.start_of_week
```
DarthHater: See previous comment about being identical more or less to other method.

test/models/patient_test.rb
```diff
@@ -152,6 +152,18 @@ class PatientTest < ActiveSupport::TestCase
       assert_equal summary[:pledged].count, 1
       assert_equal summary[:sent].count, 1
     end
+
+    it "should support a different start of week" do
+      # if the week did NOT start on thursday then patients would not be counted
```
colinxfleming: I'm not sure I quite follow what this test is proving -- I think that this is creating a fund_pledged_at on a Saturday 

I would suggest having the config_value be `Saturday` and pledging to @patient_2 on Jan 19, 1973 (so we can prove that Saturday -> Friday gets picked up). A good practice I learned from @lomky is that when you're doing dates or times with edges, do one on the included side and one on the filtered side, and I thnk that's an appropriate rule of thumb here.

I would also suggest this comment which I think is a hair clearer (and which means whoever touches this next won't have to google what day Jan 20 was):

```suggestion
      # For a start day of Saturday, Saturday should be counted; the Friday before should not
      # Jan 19, 1973 is a Friday; Jan 20 is a Saturday
```

test/models/patient_test.rb
```diff
@@ -152,6 +152,18 @@ class PatientTest < ActiveSupport::TestCase
...
+      Config.create config_key: :start_of_week,
+                    config_value: {options: ['Wednesday']}
+      @patient.update fund_pledged_at: Date.parse("January 20, 1973"), fund_pledge: 300
+        shaped_patient = patient_to_hash @patient
+      Timecop.freeze("January 22, 1973") do
```
colinxfleming: 💅

test/models/patient_test.rb
```diff
@@ -152,6 +152,18 @@ class PatientTest < ActiveSupport::TestCase
       assert_equal summary[:pledged].count, 1
       assert_equal summary[:sent].count, 1
     end
+
+    it "should support a different start of week" do
+      # if the week did NOT start on thursday then patients would not be counted
+      Config.create config_key: :start_of_week,
+                    config_value: {options: ['Wednesday']}
+      @patient.update fund_pledged_at: Date.parse("January 20, 1973"), fund_pledge: 300
+        shaped_patient = patient_to_hash @patient
```
colinxfleming: please forgive me for this suggestion, I've been writing a lot of python lately and the extra two spaces makes me think this is a continuation of L160, haha. 

```suggestion
      shaped_patient = patient_to_hash @patient
```

app/models/config.rb
```diff
@@ -58,4 +59,10 @@ def self.autosetup
       end
     end
   end
+
+  def self.start_day
+    start = Config.find_or_create_by(config_key: 'start_of_week').options.try :last
```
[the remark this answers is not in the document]
mdworken: I included it as a guard against nil. We use the pattern elsewhere too.

app/helpers/dashboards_helper.rb
```diff
@@ -1,8 +1,8 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
-  def week_range(date: Time.zone.now, start_day: :monday)
-    week_start = date.beginning_of_week start_day
-    week_end = date.end_of_week start_day
+  def week_range(date: Time.zone.now)
```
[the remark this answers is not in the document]
mdworken: The default 2nd parameter/default argument wasn't causing a problem, it just became unnecessary since Config.start_day defaults to Monday on it's own.

If I were to worry about optimizing this I'd probably go for memoizing the Config.start_day itself, but more than happy to assign to a local variable since that's easy too.

test/models/patient_test.rb
mdworken: Whoops, that's a typo, good catch!

config/locales/en.yml
```diff
@@ -500,7 +500,7 @@ en:
         pledge_unfulfilled: Pledge Not Fulfilled
         resolved: Resolved Without %{fund}
   tooltips:
-    budget_bar: The budget bar lists money tentatively set aside for a patient (by entering a value in the %{fund} pledge field) and money sent to a clinic for a patient (by checking the I sent the pledge checkbox) over a given week. It resets on Mondays.
+    budget_bar: The budget bar lists money tentatively set aside for a patient (by entering a value in the %{fund} pledge field) and money sent to a clinic for a patient (by checking the I sent the pledge checkbox) over a given week. It resets on Mondays by default.
```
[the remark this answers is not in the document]
mdworken: My inclination was that since there's text showing the start/end days of the current week that would be sufficient, and that anything more might be over-engineering, but I'd be happy to add this. Your call.

test/models/patient_test.rb
mdworken: And I didn't even use shaped_patient after all, removed it completely.

test/models/patient_test.rb
mdworken: Addressed.

app/helpers/dashboards_helper.rb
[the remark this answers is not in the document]
mdworken: Fixed it to only call once.

config/locales/en.yml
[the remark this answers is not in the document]
colinxfleming: The point about showing the start/end days of the week is a good one. Let's merge this as is and I'll whip up a quick PR simplifying that i18n text to just take out what is (as you point out here) kinda redundant.

app/models/config.rb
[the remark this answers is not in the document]
colinxfleming: works for me!

app/models/config.rb
```diff
@@ -10,15 +10,16 @@ class Config
...
   ].freeze
 
   # Define overrides for particular config fields.
   # Useful if there is no `_options` method.
   HELP_TEXT_OVERRIDES = {
     resources_url: 'A link to a Google Drive folder with CM resources. ' \
                    'Ex: https://drive.google.com/drive/my-resource-dir',
-    fax_service: 'A link to your fax service. ex: https://www.efax.com'
+    fax_service: 'A link to your fax service. ex: https://www.efax.com',
+    start_of_week: 'Ex: Monday'
```
colinxfleming: Looking at the UI I think we should make this a little bit clearer:

image

I think this might be a little clearer to someone looking for clarity on what this changes:

```suggestion
    start_of_week: 'The start day of your budget week, if it doesn't start on Monday. ex: Sunday.
```

Please smash if it looks good to you!

app/models/config.rb
```diff
@@ -10,15 +10,16 @@ class Config
...
   ].freeze
 
   # Define overrides for particular config fields.
   # Useful if there is no `_options` method.
   HELP_TEXT_OVERRIDES = {
     resources_url: 'A link to a Google Drive folder with CM resources. ' \
                    'Ex: https://drive.google.com/drive/my-resource-dir',
-    fax_service: 'A link to your fax service. ex: https://www.efax.com'
+    fax_service: 'A link to your fax service. ex: https://www.efax.com',
+    start_of_week: 'The start day of your budget week, if it doesn't start on Monday. ex: Sunday.
```
colinxfleming: lol @ bad advice from me

```suggestion
    start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday."
```

PR #1712, opened 2019-06-15
See if CI passes on rails 6 rc1

colinxfleming:

CI test on rails 6.


colinxfleming:
tests pass on rc1!

PR #1713, opened 2019-06-15
Add ability to edit practical support forms

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

Not shipping because there's some additional UI to do, but this sets up update and destroy in the UI. Still not quite primetime ready but materially closer.

This pull request makes the following changes:
* Sets up entry update and destroy (no tests yet)
* some UI cleanup and breaking things into more rest-y routes
* turn on the autosave!

The state of the world:

image

Still todo are tests and localization and exports at least, and doing a quick UI check with @mlindberg61 .

It relates to the following issue #s: 
* Bumps #1493

PR #1714, opened 2019-06-16
System tests for practical support

colinxfleming:

Adds integration tests for creating, editing, and destroying practical supports using the UI.

* system tests for #1713 !

* Bumps #1493


test/system/practical_support_behaviors_test.rb
```diff
@@ -0,0 +1,122 @@
...
+    @user = create :user
+    @patient = create :patient, line: 'DC'
+  end
+
+  describe 'creating a new practical support entry' do
+    before { go_to_practical_support_tab }
+
+    it 'should start empty' do
+      within :css, '#existing-practical-supports' do
+        refute has_text? 'Category of support'
```
mdworken: has_text? only returns after the capybara default wait time, so using it when the expectation is negative adds unnecessary waits to test suite. Writing this as assert has_no_text? will still wait for page load but will avoid this issue

colinxfleming:
great catch, thank you!

PR #1716, opened 2019-06-19
Fixing seeds so the CM resources link works

rcackerman:
Super small fix :-). I changed what looks like a typo in the seed file, which was breaking the CM Resources link in the navbar.


colinxfleming:
CI failure is fake. Merging!

PR #1717, opened 2019-06-24
Add sentry error reporting

### LINKED ISSUE #1706: Set up sentry
I had a lovely conversation with the Sentry reps at Railsconf and I think the free plan of sentry might be a good investment. Let's investigate and see if we can rig this up.

colinxfleming:

We're getting into the business of error tracking, and we can easily squeeze into the sentry free plan, so I'm standing that up here. This way we can stop troubleshooting errors thru logstash, which is nearly impossible near as I can tell.

The other steps of this is to create the projects in the actual sentry UI and to add the heroku environment variables, which I'll be doing after I get approval on this.

* add sentry-raven gem and configure it

no view changes, but it does report based on setting the reporting endpoint locally!

* Fixes #1706


config/initializers/sentry.rb
```diff
@@ -0,0 +1,5 @@
+# Configure sentry
+Raven.configure do |config|
+  config.dsn = ENV['SENTRY_DSN']
+  config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
```
lomky: The scrubs all parameter values from being sent to Sentry, yes?

colinxfleming:
I'll fake an error on a patient save and post it in here to make sure.

lomky:
Buuuump

colinxfleming:
😬

colinxfleming:
I chucked up a sample error here to sentry (I threw a `raise` in the controller action) and here's what posted. We can see the filters here limiting just to the pt id in the URL, and the name and phone get filtered, because we have them in the list in `filter_parameter_logging.rb`.

image

look ok @lomky ?

colinxfleming:
(as a note, I bumped heroku envs so this is ready to ship)

PR #1718, opened 2019-06-26
Practical support: Correct spacing/alignment/refresh on create/update/destroy actions

colinxfleming:

I put this together kinda sloppy in terms of alignments/proper element id usage, so this adjusts them and empties out the right partials.

I think this is covered by existing system tests. Also want to note that after this, the partial will start showing up in sandbox for real.

* align everything along the rail
* properly empty out the entries partial if there are no entries

Now left aligned on both create and page load like so:

image

* Bumps #1493


app/views/practical_supports/_show.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="row" class="practical-support-item" id="practical-support-item-<%= practical_support.id %>">
+<div class="row practical-support-item" id="practical-support-item-<%= practical_support.id %>">
```
mdworken: 👍

PR #1724, opened 2019-06-30
Adds configuration option to turn off practical support

### LINKED ISSUE #1719: Add config to turn off practical support module

Not every fund does practical support and it might confuse CMs that do.  We should add an option to turn this shit off.

Practical support rollout!

I'd like to have a config option that's basically "Put yes in this box to make the practical support view stuff not show up". I'd prefer to hook it up in a separate PR, so this is just for the model work.

* **Anything else?**
stems from #1493

mdworken:
Wiring it up was OOS but I went ahead and did it anyways. Self-tested, mini-test specs TK

Fixes #1719

### DIFF
--- a/app/controllers/practical_supports_controller.rb
+++ b/app/controllers/practical_supports_controller.rb
@@ -1,6 +1,5 @@
 # Create a non-monetary assistance record for a patient.
 class PracticalSupportsController < ApplicationController
-  before_action :ensure_showing_practical_support
   before_action :find_patient, only: [:create]
   before_action :find_support, only: [:update, :destroy]
   rescue_from Mongoid::Errors::DocumentNotFound,
@@ -52,9 +51,5 @@ class PracticalSupportsController < ApplicationController
   def find_support
     find_patient
     @support = @patient.practical_supports.find params[:id]
-  end
-
-  def ensure_showing_practical_support
-    Config.show_practical_support?
   end
 end
--- a/app/controllers/practical_supports_controller.rb
+++ b/app/controllers/practical_supports_controller.rb
@@ -1,5 +1,6 @@
 # Create a non-monetary assistance record for a patient.
 class PracticalSupportsController < ApplicationController
+  before_action :ensure_showing_practical_support
   before_action :find_patient, only: [:create]
   before_action :find_support, only: [:update, :destroy]
   rescue_from Mongoid::Errors::DocumentNotFound,
@@ -51,5 +52,9 @@ class PracticalSupportsController < ApplicationController
   def find_support
     find_patient
     @support = @patient.practical_supports.find params[:id]
-  end    
+  end
+
+  def ensure_showing_practical_support
+    Config.show_practical_support?
+  end
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -20,7 +20,7 @@ class Config
                    'Ex: https://drive.google.com/drive/my-resource-dir',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
     start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday.",
-    show_practical_support: 'Toggle the visibility of practical support information. Enter "No" to hide this view. Note: This will not remove any existing data.'
+    show_practical_support: 'Toggle the visibility of practical support information. Enter "Yes" to show or "No" to hide. Note: This will not remove any existing data.'
   }.freeze
 
   # Fields
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -20,7 +20,7 @@ class Config
                    'Ex: https://drive.google.com/drive/my-resource-dir',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
     start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday.",
-    show_practical_support: 'Toggle the visibility of practical support information. Enter "Yes" to show or "No" to hide. Note: This will not remove any existing data.'
+    show_practical_support: 'Enter "no" to hide the Practical Support panel on patient pages. This will not remove any existing data.'
   }.freeze
 
   # Fields
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -10,7 +10,7 @@ class Config
   CONFIG_FIELDS = [
     :insurance, :external_pledge_source, :pledge_limit_help_text,
     :language, :resources_url, :fax_service, :referred_by,
-    :practical_support, :show_practical_support, :start_of_week
+    :practical_support, :hide_practical_support, :start_of_week
   ].freeze
 
   # Define overrides for particular config fields.
@@ -20,7 +20,7 @@ class Config
                    'Ex: https://drive.google.com/drive/my-resource-dir',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
     start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday.",
-    show_practical_support: 'Enter "no" to hide the Practical Support panel on patient pages. This will not remove any existing data.'
+    hide_practical_support: 'Enter "yes" to hide the Practical Support panel on patient pages. This will not remove any existing data.'
   }.freeze
 
   # Fields
@@ -67,9 +67,8 @@ class Config
     start.downcase.to_sym
   end
 
-  def self.show_practical_support?
-    show_value = Config.find_or_create_by(config_key: 'show_practical_support').options.try :last
-    show_value ||= "Yes"
-    show_value.downcase.strip.to_sym != :no
+  def self.hide_practical_support?
+    hide_value = Config.find_or_create_by(config_key: 'hide_practical_support').options.try(:last).to_s
+    hide_value =~ /yes/i
   end
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -10,7 +10,7 @@ class Config
   CONFIG_FIELDS = [
     :insurance, :external_pledge_source, :pledge_limit_help_text,
     :language, :resources_url, :fax_service, :referred_by,
-    :practical_support, :start_of_week
+    :practical_support, :show_practical_support, :start_of_week
   ].freeze
 
   # Define overrides for particular config fields.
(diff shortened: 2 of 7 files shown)


app/models/config.rb
```diff
@@ -19,7 +19,8 @@ class Config
     resources_url: 'A link to a Google Drive folder with CM resources. ' \
                    'Ex: https://drive.google.com/drive/my-resource-dir',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
-    start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday."
+    start_of_week: "The start day of your budget week, if it doesn't start on Monday. ex: Sunday.",
+    show_practical_support: 'Toggle the visibility of practical support information. Enter "Yes" to show or "No" to hide. Note: This will not remove any existing data.'
```
colinxfleming: This descriptive text isn't quite what's going on in the code. I would suggest:

```suggestion
    show_practical_support: 'Enter "no" to hide the Practical Support panel on patient pages. This will not remove any existing data.'
```

(or the equivalent of this being a negative per comment above)

app/models/config.rb
```diff
@@ -65,4 +66,10 @@ def self.start_day
     start ||= "monday"
     start.downcase.to_sym
   end
+
+  def self.show_practical_support?
+    show_value = Config.find_or_create_by(config_key: 'show_practical_support').options.try :last
+    show_value ||= "Yes"
```
colinxfleming: Since we're only checking for equivalence to no, I don't think it's necessary to use an or-equals operator here to set the value. I would suggest optimizing for readability and stripping this line in favor of `|| ''` on the line above, or packing this into one expression instead of three (e.g. `Config.find_or_create_by(...).options.try(...).to_s.downcase.strip.to_sym != :no`).

app/controllers/practical_supports_controller.rb
```diff
@@ -1,5 +1,6 @@
 # Create a non-monetary assistance record for a patient.
 class PracticalSupportsController < ApplicationController
+  before_action :ensure_showing_practical_support
```
colinxfleming: My instinct would be that we don't need a controller-level verification and that just working on the view layer is fine, but I'm curious -- what are we trying to get out of this by checking on the controller level?

app/models/config.rb
```diff
@@ -10,7 +10,7 @@ class Config
   CONFIG_FIELDS = [
     :insurance, :external_pledge_source, :pledge_limit_help_text,
     :language, :resources_url, :fax_service, :referred_by,
-    :practical_support, :start_of_week
+    :practical_support, :show_practical_support, :start_of_week
```
colinxfleming: This is kinda particular so please forgive me, but showing practical support is the default behavior here, right? Would this make more sense as `hide_practical_support`, which I think infers that most of the time it should show up? Compare it with, for example, `if current_user.allowed_data_access?` in menu.html.erb -- that indicates that generally, the thing that's wrapping isn't the default behavior. Would love your thoughts on this.

app/models/config.rb
```diff
@@ -65,4 +66,10 @@ def self.start_day
     start ||= "monday"
     start.downcase.to_sym
   end
+
+  def self.show_practical_support?
```
colinxfleming: ps I do really like the `Config.config_key` pattern and think it was a great introduction

app/controllers/practical_supports_controller.rb
mdworken: I was probably just being overly cautious, but I was a little concerned about the fact that if you were already in the view and then practical support was turned off, you could still do updates from there.

In retrospect, it feels like too much code for not enough edge case - want me to pull it out?

app/models/config.rb
mdworken: :D

app/models/config.rb
mdworken: I really only included this because nil.downcase is undefined so I'd have to do nil.to_s.downcase, and then turn the nil string into a symbol, which I'm not confident there aren't dragons lurking if I do it. I considered digging or using a ton of safe navs, but it felt sloppy.

I also liked keeping with the existing pattern from my previous PR when possible to reduce cognitive overhead.

app/models/config.rb
mdworken: :+1: wooh!

app/models/config.rb
mdworken: That's a reasonable objection - I'll do some more wordsmithing here.

test/system/practical_support_behaviors_test.rb
```diff
@@ -105,6 +105,29 @@ class PracticalSupportBehaviorsTest < ApplicationSystemTestCase
...
+    end
+
+    it 'toggles the showing of the practical support tab' do
+      assert has_text? /Practical Support/i
+      c = Config.find_or_create_by(config_key: 'hide_practical_support')
+      c.config_value = { options: ["Yes"] }
+      c.save!
+      page.refresh
+      assert has_no_text? /Practical Support/i
+      c.config_value = { options: ["No"] }
```
colinxfleming: I don't think we need to test this toggle back and forth, should be covered by the model tests. Can you delete this line thru 127?

test/system/practical_support_behaviors_test.rb
```diff
@@ -105,6 +105,29 @@ class PracticalSupportBehaviorsTest < ApplicationSystemTestCase
...
+    before do
+      @patient.practical_supports.create support_type: 'lodging',
+                                         source: 'Other funds (see notes)',
+                                         created_by: @user
+      go_to_practical_support_tab
+    end
+
+    it 'toggles the showing of the practical support tab' do
+      assert has_text? /Practical Support/i
+      c = Config.find_or_create_by(config_key: 'hide_practical_support')
```
colinxfleming: My experience is that creating objects like this in the test block (instead of the setup block) makes it kind of tough to infer exactly what we're trying to test, and doing stuff in the database after having the UI loaded is a pattern that can cause a lot of unintended side effects due to the database being out of sync with the UI. What do you think about either turning this into a one-liner and move it to the `before` block (see suggestion) or do it all using selenium (see, for example, `updating a config - insurance` in `update_configs_test`)?

In short, I think we should pare this test down to:

before block:
* given that I have a patient
* and a config of hide_practical_support: no (at this point, we're done changing the state of the database, and ready to load the UI)
* and I visit the `edit_patient_path(@patient)`

test block
* there shouldn't be a link for practical support

Thoughts?

```suggestion
      Config.create config_key: 'hide_practical_support', config_value: { options: ['No'] }
```

app/models/config.rb
```diff
@@ -65,4 +66,8 @@ def self.start_day
     start ||= "monday"
     start.downcase.to_sym
   end
+
+  def self.hide_practical_support?
```
colinxfleming: Can we add a model test for this?

test/system/practical_support_behaviors_test.rb
mdworken: Totally agree that the pattern can be problematic, which is why I threw `page.refresh` calls in. I think you're right though that this is overkill.

app/models/config.rb
mdworken: Done!

test/system/practical_support_behaviors_test.rb
mdworken: Broke it out into two separate it blocks - looks like it actually increased the run time of the test slightly*, but it should be easier for future peeps to reason with.


* this is to be expected since updating db and calling page.refresh is cheap relative to logging in a second time

test/system/practical_support_behaviors_test.rb
```diff
@@ -105,13 +105,36 @@ class PracticalSupportBehaviorsTest < ApplicationSystemTestCase
...
+                                         created_by: @user
+    end
+
+    it 'can hide the practical support tab' do
+      @config = Config.find_or_create_by(config_key: 'hide_practical_support', config_value: { options: ["Yes"] })
+      go_to_edit_page
+      assert has_no_text? /Practical Support/i
+    end
+
+    it 'shows the practical support tab by default' do
```
mdworken: This test doesn't give us any extra coverage in the present, but since the main test above is a negative test it's important to test the positive case too. 

Reasoning: 
Without checking the positive case, if we changed the text "Practical Support" to "New Text" and then introduced a regression where the text showed, the first test would fail silently. By checking the positive case, we ensure that a test failure occurs in that first step, allowing us to update assertions.

test/models/config_test.rb
```diff
@@ -94,5 +94,25 @@ class ConfigTest < ActiveSupport::TestCase
         assert(Config.start_day.equal? :tuesday)
       end
     end
+
+    describe '#hide_practical_support?' do
+      it 'can return true' do
+        c = Config.find_or_create_by(config_key: 'hide_practical_support')
```
colinxfleming: Any reason not to do all this on create like we do in the system test, esp since we're not really changing the object state? Would work either way, but would save us assigning the local variable and a database call (and I like the consistency). I'm good either way, let me know yr thoughts.

test/models/config_test.rb
```diff
@@ -94,5 +94,25 @@ class ConfigTest < ActiveSupport::TestCase
...
+        c.config_value = { options: ["Yes"] }
+        c.save!
+        assert(Config.hide_practical_support? == true)
```
colinxfleming: apologies for the ruby style nitpick, but `assert Config.hide_practical_support?` and (`assert_not Config.hide_practical_support?`) is a bit tidier than `assert boolean == true`. If it's not a ton of trouble I'd appreciate making that bump for consistency but am also happy to do it since this is my own weird neurosis, haha.

test/system/practical_support_behaviors_test.rb
colinxfleming: I'm convinced, thumbs up emoji

test/models/config_test.rb
mdworken: So the reason I did it this way was to enforce that it is returning a proper true/false instead of truthy/falsey, happy to change this if we don't want to enforce that strictly though!

colinxfleming: If we were very specifically looking for a `true` boolean value here I think we'd want to use `assert_equal`, but I don't think we need to strictly enforce this and checking for truthiness/falsiness is fine. That said I want this PR in, so going to merge.

Issue #1723, opened 2019-06-26
Practical support omnibus


colinxfleming:
placeholder ticket for review

UX portion of #1493

colinxfleming:
some UX feedback:

* possible to be exploring multiple practical support providers, so probably don't want to enforce uniqueness on that.
* not clear whether the organization is quarterbacking or providing. Probably should be quarterbacking, but need to make that clear.
* some ext pledge options aren't relevant and we should strip them from the list (Clinic Discount doesn't make sense, for example)
* concern about CMs missing that patients have practical support - ML suggested putting a flag in abortion options if anything's filled out
* change 'category of support' verbiage to 'support needed'
* thing I don't understand about backup requests. I think we might be able to solve by putting 'patient handling it themselves' here as an option, or not enforcing uniqueness. The exact text:

> To Harum's point, I think it's most likely that we request support from multiple funds/sources per section, but then only one actually covers each section. For example, the caller might be able to come up with an escort, but we also put in a backup request to DAPSN. I think having an option that says "Not sure yet (see notes)" or "Multiple requests out (see notes)" or something like that would cover this. I played through a few other ideas, but I don't think providing the option to select multiple options would resolve this.

colinxfleming:
particulars are done here, but keeping open until we get final sign off.

colinxfleming:
:shipit:

PR #1725, opened 2019-07-13
Add distinct practical support source options

colinxfleming:

Marie pointed out in #1723 that some of the practical support source options don't make sense in context, so we should massage them a little bit. This addresses that by setting the options in the dropdown to nil, whatever the fund is, and external pledge source options.

Along the way we move these functions to their own helper and remix the tests a little bit, and add the same config change protection that insurance_options has.

As a followup we should find a way to surface the fact that `practical_support_source_options` is consuming the external pledge source options config in the config panel, but that would probably complexify the ticket a lot, so let's hold for now.

* Adds a specialized helper for practical support source options
* some refactoring of tests
* adds value protection in the event of a config change

No real view changes

* Bumps #1723


app/helpers/practical_supports_helper.rb
```diff
@@ -1,2 +1,22 @@
...
+    if current_value.present? && full_set.map { |opt| opt.is_a?(Array) ? opt[-1] : opt }.exclude?(current_value)
+      full_set.push current_value
+    end
+
+    full_set.uniq
+  end
+
+  def practical_support_source_options(current_value = nil)
+    full_set = [nil, FUND_FULL] + Config.find_or_create_by(config_key: 'external_pledge_source').options + ['Clinic', 'Other (see notes)']
+    full_set.push current_value if full_set.exclude? current_value
```
mdworken: Do we need this check here? We're already calling uniq.

app/helpers/practical_supports_helper.rb
```diff
@@ -1,2 +1,22 @@
 module PracticalSupportsHelper
+  def practical_support_options(current_value = nil)
+    standard_options = [
+      [ t('patient.helper.practical_support.travel_to_area'), 'Travel to area' ],
+      [ t('patient.helper.practical_support.travel_inside_area'), 'Travel inside area' ],
+      [ t('patient.helper.practical_support.lodging'), 'Lodging' ],
+      [ t('patient.helper.practical_support.companion'), 'Companion' ],
+    ]
+    full_set = [nil] + standard_options + Config.find_or_create_by(config_key: 'practical_support').options
+    if current_value.present? && full_set.map { |opt| opt.is_a?(Array) ? opt[-1] : opt }.exclude?(current_value)
```
mdworken: Do we need this check? It seems to me we could remove it and just lean on the uniq call.

app/helpers/practical_supports_helper.rb
mdworken: Same comment about uniq as above.

test/helpers/practical_supports_helper_test.rb
```diff
@@ -0,0 +1,103 @@
...
+          'Clinic',
+          'Other (see notes)'
+        ]
+        assert_same_elements expected, @options
+        assert Config.find_by(config_key: 'external_pledge_source')
+      end
+    end
+
+    describe 'with an orphaned value' do
+      it 'should push orphaned value onto the end' do
```
mdworken: It looks like this will only support the orphaned value on the one pulldown it was selected on, not all of them. This seems okay-ish to me, but just wanted to make sure we're explicitly okay with that.

colinxfleming: I don't think I know what you mean by this, can you clarify?

app/helpers/practical_supports_helper.rb
colinxfleming: Probably true, I'll try just uniqing and see if tests still pass. Good catch.

test/helpers/practical_supports_helper_test.rb
mdworken: If there are multiple pulldowns and one has an orphaned value, that orphaned value will be an option on only one of the pulldowns. Anyways, not a big deal either way.

PR #1726, opened 2019-07-13
Change verbiage of practical support columns and add indicator

colinxfleming:

#1723 noted that the verbiage was kinda confusing, so we clarify it to `Support needed` and `Point of contact`.

We also add an indicator to abortion information to make a note when a patient has practical support entries. This seems so straightforward and noncritical that I don't think we need to ajax changes in or test for it. Open to other opinions.

* change verbiage
* add indicator to abortion information when a patient has practical supports

image

* Bumps #1723


colinxfleming:
Need these in, so invoking godmode and reving.

PR #1727, opened 2019-07-14
loosen uniqueness restrictions on practical support

colinxfleming:

Per feedback from @harumhelmy  it's possible to have multiple funds investigating practical support (though probably only one should confirmed? idk). So we're loosening that a bit.

* loosens uniqueness scope to source and support type
* bumps tests

* Bumps #1723


colinxfleming:
taking a different tack on this

PR #1728, opened 2019-07-14
add other/wildcard options to support source and type options

colinxfleming:

Had an open question about how to handle more than one active request out -- I think we're trying to actively discourage the possibility that it won't be clear who's quarterbacking a practical support supply effort, and that need supersedes what we lose on reporting here by not knowing 'we asked this fund but they went in a diff direction'.

* adds new wildcard options and patient as practical support source
* keeps current validation of only one source type the way it is now

* Bumps #1723


colinxfleming:
need these in, so invoking godmode

PR #1729, opened 2019-07-22
Do i18n translations for practical support

### LINKED ISSUE #1722: Translate Practical Support module
This ticket is part of the DARIA translation effort. We are trying to implement our text in both English and Spanish thru the i18n process.

This ticket is for the practical support module.
Code link: 
app/views/views/practical_supports/*
app/controllers/practical_supports_controller.rb (flash messages)

In addition, there's an additional menu item to translate, `practical_support`. Add that in en.yml and es.yml under `patient.menu`.
We also need to translate the dropdown options, which should be under `patient.helper.practical_support`. Those are located in app/helpers/patients_helper.rb.

### Steps

For each view  
   `en.yml`:
   ```
   patient:
     dashboard:
       appt_date: Appointment date
    ```  
    `es.yml`:
    ```
    patient:
      dashboard:
        appt_date: Día de la cita
    ```

### Completed View Example

`i18n-tasks normalize`

stems from #1493

colinxfleming:

This does the translation for practical support.

It involved a little bit of gut work in practical_support_helper; I remember this now, but the reason we had the complex checking system is because i18n keys-value pairs don't reduce to english values easily. So we do a little bit of of a dance, which I've rolled up into a private method here.

* adds i18n keys for practical support
* fixes up some helpers to account for different keys and values because of translation

Example of the view changes: 
image

@montanezp can you review the translations the same way we reviewed the old stuff? Take a look at the green stuff in es.yml and compare it to the corresponding language in en.yml to make sure I didn't screw up any translations. You can ignore the stuff outside of those two files. Let me know if I can be helpful.

* Fixes #1722

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -65,6 +65,7 @@ es:
     add: Añadir
     appointment_date_short: Día de la cita
     back: Atrás
+    clinic: Clínica
     days_along: Días de embarazo al inicio
     email: Email
     medicaid: Medicaid
@@ -72,6 +73,7 @@ es:
     name: Nombre
     'no': 'No'
     notes: Notas
+    patient: Patiente
     phone: Teléfono
     save: Guardar
     search: Buscar
@@ -354,8 +356,10 @@ es:
       practical_support:
         companion: Acompañante
         lodging: Hospedaje
-        travel_inside_area: Viajar dentro de la región
-        travel_to_area: Viajar a la región
+        travel_inside_region: Viajar dentro de la región
+        travel_to_region: Viajar a la región
+        other: Otro (ver notas)
+        not_sure_yet: No estamos seguro (ver notas)
       race:
         asian_south_asian: Asiática o Sur de Asia
         black_african_american: Negra/Afroamericana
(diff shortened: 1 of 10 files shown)


config/locales/es.yml
```diff
@@ -474,6 +479,14 @@ es:
         procedure_date: Fecha de procedimiento
         weeks_along_at_procedure: Semanas a lo largo del procedimiento
       title: Cumplimiento de la promesa
+    practical_support:
+      delete_confirm: "¿Está seguro de que desea eliminar esta entrada de soporte práctico?"
+      support_needed: Soporte Necesarios
+      point_of_contact: Punto de contacto
+      create: Crear una nueva entrada de soporte práctico
+      title: Apoyo practico
+      remove: Quita
```
montanezp: Replace "Quita" with "Elimina"

config/locales/es.yml
```diff
@@ -474,6 +483,14 @@ es:
         procedure_date: Fecha de procedimiento
         weeks_along_at_procedure: Semanas a lo largo del procedimiento
       title: Cumplimiento de la promesa
+    practical_support:
+      create: Crear una nueva entrada de soporte práctico
+      delete_confirm: "¿Está seguro de que desea eliminar esta entrada de soporte práctico?"
+      new: 'Grabar nueva información de soporte práctico:'
+      point_of_contact: Punto de contacto
+      remove: Quita
```
colinxfleming:
```suggestion
      remove: Elimina
```

PR #1735, opened 2019-08-07
Respect time zone in budget bar

colinxfleming:

BRAF found a bug with the budget bar that stems from our casting to date being in UTC instead of eastern. This enforces timezone on our budget bar filter.

Tests TK because it's a late code change

* filters on a time instead of a date in budget bar, to account for timezone

no view changes, no issue


colinxfleming:
going to try to write a test for this, but honestly it's pretty miserable going, so might say screw it

colinxfleming:
I've had an exceptionally hard time writing a unit test for this, so I'm going to just check it in the UI and merge if it works the way I'm pretty sure it does.

colinxfleming:
I'm satisfied this does what we think it does. Merging.

Issue #1721, opened 2019-06-26
Hook up practical support to exportable


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
We should represent practical support in exportable.

* **What feature or behavior is this required for?**
Making good data on practical support available!

* **How could we solve this issue? (Not knowing is okay!)**
Follow the patterns of external pledges in implementing this.

* **Anything else?**
stems from #1493

PR #1737, opened 2019-08-07
Adds error messages view to devise to squelch deprecation warnings #1721

mdworken:

* adds view to squelch devise error message


mdworken:
Before: image

After: 
image

mdworken:
Yeah, looks like it might help to replace the calls with the new view itself! Lolooops, - I'll get to this at the hacknight tonight hopefully

PR #1738, opened 2019-08-07
Adds practical support to exportable

mdworken:

* Adds practical support to exportable

Proof:

image

Fixes: #1721

### DIFF
--- a/app/models/concerns/exportable.rb
+++ b/app/models/concerns/exportable.rb
@@ -174,7 +174,7 @@ module Exportable
   end
 
   def all_practical_supports
-    practical_supports.map {|ps| "Type: #{ ps.support_type}, Source: #{ps.source}, Confirmed: #{ps.confirmed?}"}.join('; ')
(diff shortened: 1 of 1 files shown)


app/models/concerns/exportable.rb
```diff
@@ -172,6 +173,10 @@ def all_external_pledges
     external_pledges.map { |x| "#{x.source} - #{x.amount}" }.join('; ')
   end
 
+  def all_practical_supports
+    practical_supports.map {|ps| "Type: #{ ps.support_type}, Source: #{ps.source}, Confirmed: #{ps.confirmed?}"}.join('; ')
```
colinxfleming: let's have this follow the pattern in external pledges if that's cool?

```suggestion
    practical_supports.map { |ps| "#{ps.source} - #{ps.support_type} - #{ps.confirmed? ? 'Confirmed' : 'Unconfirmed'}" }.join('; ')
```

Not sure there's a need to define keys inline.

PR #1747, opened 2019-08-18
WIP Get autosave working in webpacker and port over patients.js

### LINKED ISSUE #1686: JS: Port over autosave part of patients.coffee from coffeescript to ES6

colinxfleming:

So it turns out webpacker requires doing UJS a little differently. https://github.com/rails/jquery-ujs/wiki/How-to-trigger-a-form-submit-from-code

In a nutshell, if we don't do arrow functions (to pass the context of `this`) and fire events with`$.rails.fire(element, 'submit')` it's smooth sailing.

However, this raises another associated weird thing where the CSP gets sad about executing what's in app/views/patients/update.js.erb. I don't know how to get that sorted out and don't want to shave the yak right now, but we have to resolve before merging.

* port over patients with working autosave
* place changes to get it to work with the CSP (eventually)

no view changes

* Fixes #1686


config/initializers/content_security_policy.rb
```diff
@@ -14,4 +14,4 @@
   end
 end
 
-Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
+# Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
```
colinxfleming: haxx to see if it works in general!

colinxfleming:
the issue at the root of this is that `$.rails` isn't set in webpack. I would have expected adding `jquery-ujs` to the package.json bundle but I think it's conflicting with stuff.

I'm nuking this for now - should try again when we aren't double-loading bootstrap.

PR #1756, opened 2019-08-22
Initial GitHub Actions workflow

DarthHater:

This is the first lil bit to test out a build with GitHub Actions

* Initial `.github/workflows/dcaf-build-workflow.yml`

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -8,5 +8,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   before { Capybara.reset_sessions! }
 
-  driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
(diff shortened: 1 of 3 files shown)


test/application_system_test_case.rb
```diff
@@ -8,5 +8,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   before { Capybara.reset_sessions! }
 
-  driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
+  # if in CI, run system tests headlessly.
+  browser = ENV['GITHUB_WORKFLOW'] ? :headless_chrome : :chrome
```
colinxfleming: this is basically the nut of why selenium tests were failing; either we need to be running headlessly (easier!) or we need to be running xvfb (hard!).

PR #1757, opened 2019-08-22
Requires roles for users. Adds default role of case manager to users#create

### LINKED ISSUE #1749: It's possible to create users without a role?

Here's a weird bug:

* go to user management page
* add a new user
* fill in email and name, don't change the role, click add
* for some reason, they don't have a role set?

The expected behavior should be that the user is created with a default role of `cm`.

I don't think this is preventing logins, but I do think it's sketchy, so we should fix it.

I THINK this is because role isn't a permitted parameter in devise, but I don't know that for sure. I think we can resolve this by mucking with `configure_permitted_parameters` to include whatever the user model field is, but maybe not.

* **Anything else?**

mdworken:
Previously users were created without roles by default. This makes a few changes to correct this:
- Newly created users have role CM
- Users are required to have a role
- Updated tests/factories
Fixes #1749

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,8 @@ class UsersController < ApplicationController
 
   def create
     raise Exceptions::UnauthorizedError unless current_user.admin?
-    @user = User.new(user_params)
+    binding.pry
+    @user = User.new(user_params.merge(role: :cm))
     hex = SecureRandom.urlsafe_base64
     @user.password, @user.password_confirmation = hex
     if @user.save
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,6 @@ class UsersController < ApplicationController
 
   def create
     raise Exceptions::UnauthorizedError unless current_user.admin?
-    binding.pry
     @user = User.new(user_params.merge(role: :cm))
     hex = SecureRandom.urlsafe_base64
     @user.password, @user.password_confirmation = hex
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,7 @@ class UsersController < ApplicationController
 
   def create
     raise Exceptions::UnauthorizedError unless current_user.admin?
-    @user = User.new(user_params)
+    @user = User.new(user_params.merge(role: :cm))
     hex = SecureRandom.urlsafe_base64
     @user.password, @user.password_confirmation = hex
     if @user.save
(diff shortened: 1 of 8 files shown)


mdworken:
Note: will also want to run something like `User.where(role: nil).map(&:update(role: :cm))` to fix the existing records

test/controllers/users_controller_test.rb
```diff
@@ -101,14 +101,15 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
...
 
   describe 'create method' do
-    it 'should create a user' do
+    it 'should create a user with the role cm' do
       attributes = attributes_for(:user)
-      attributes.delete :role
       assert_difference 'User.count', 1 do
         post users_path, params: { user: attributes }
       end
+      assert(User.last.role == :cm)
```
colinxfleming: Appreciate the attention to detail on our existing tests!

Minitest makes these nice semantic checks available, and we generally follow them since they do make it a little easier to read as plain english. Thoughts about this?

```suggestion
      assert_equal :cm, User.last.role
```

app/models/user.rb
```diff
@@ -82,6 +82,7 @@ class User
   # Validations
   # email presence validated through Devise
   validates :name, presence: true
+  validates :role, presence: true
```
colinxfleming: Would it be helpful to declare a `default: :cm` on this model's `role` field as well? Do you think that would be a useful clue on what role was supposed to be if you were looking at the user model for the first time?

app/controllers/users_controller.rb
```diff
@@ -51,7 +51,7 @@ def update
 
   def create
     raise Exceptions::UnauthorizedError unless current_user.admin?
-    @user = User.new(user_params)
+    @user = User.new(user_params.merge(role: :cm))
```
colinxfleming: Could we get a comment here on what we're accomplishing by setting the role? `cm` is a pretty ab-fund-specific parlance, so I think some clarification of what CM means would be useful to someone seeing this for the first time who's less familiar.

app/views/users/new.html.erb
```diff
@@ -3,7 +3,6 @@
     <%= bootstrap_form_for @user do |f| %>
         <%= f.email_field :email, label: t('common.email') %>
         <%= f.text_field :name, label: t('common.name') %>
-        <%= f.select :role, options_for_select(user_role_options), label: t('user.common.role') %>
```
colinxfleming: I think some help text here on how to escalate someone's role might be helpful. Maybe something like `New users by default have CM permissions, which let them use the case management workflow. To give someone data access or admin permissions, edit them after you create their accounts here.` I'm spitballing a bit, but I think it'd be helpful.

PR #1763, opened 2019-08-29
set ruby point release in github actions

colinxfleming:
trying to squash a config failure in github actions


colinxfleming:
this doesn't seem to be possible yet in github's ci, unfortunately. Sounds like they're working on it.

Issue #1632, opened 2019-03-03
Move to bootstrap 4 and bootstrap_form 4


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

We're a major version behind on bootstrap, which wouldn't be a terribly huge deal except our form builder is also moving on, and I want to make sure we're taking advantage of what it offers us in terms of security. So we should move to 

* **What feature or behavior is this required for?**

Long term health of the project, rails security, the ability to use form_with in code.

* **How could we solve this issue? (Not knowing is okay!)**

At a minimum, `gem 'bootstrap'` and `bundle update bootstrap_form`, then work to make all the tests pass again and make sure the views don't get wild.

* **Anything else?**

This is going to be a lot of work, as our CSS is probably the least touched part of this project, but it is absolutely worth doing so we can keep our form building up to snuff.

colinxfleming:
This is underway. We're doing it in such a way that at the end we'll still be in sprockets, but will be firmly on bootstrap 4 (with much cleaner HTML as a fringe benefit).

After that, we should ticket moving our BS4 install from sprockets to webpacker, and then write a gameplan for a JS framework implementation if that still feels like a good idea.

PR #1766, opened 2019-09-01
Port patient dashboard partial; un-doubleapply container; restore form styling

colinxfleming:

BS4's grid system uses flexboxes now, which are MUCH more forgiving about equal spacing and don't require us to define columns with numbers, so we port the top header there to that style.

Along the way we make some changes to not double-apply container (it was happening on both `main` and `div.container` blocks), and do some tweaking to restore form fields being bolded, and to compress the form line height a little bit so it doesn't run on.

Note that this goes into a feature branch rather than master.

* removes double-applied container
* adds some form css to bold form labels and compress forms
* refactors patient dashboard into columns

before:
image

after (only real diff is that status moves up to label level, I think):
image

* Bumps #1632

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -18,13 +18,11 @@
         <%= render 'layouts/navigation' %>
       </header>
     <% end %>
-    <main role="main">
-      <div class="container">
-        <%= render 'layouts/messages' %>
+    <main class="container" role="main">
+      <%= render 'layouts/messages' %>
 
-        <br>
-        <%= yield %>
-      </div>
+      <br>
+      <%= yield %>
     </main>
     <% if current_user.present? %>
       <footer>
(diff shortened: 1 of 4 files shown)


app/views/layouts/application.html.erb
```diff
@@ -18,13 +18,11 @@
         <%= render 'layouts/navigation' %>
       </header>
     <% end %>
-    <main role="main">
-      <div class="container">
-        <%= render 'layouts/messages' %>
+    <main class="container" role="main">
```
colinxfleming: `main` it turns out extended container, so we take it out of the css and just give it the class to make it explicit (and not double apply)

app/views/patients/_patient_dashboard.html.erb
```diff
@@ -1,39 +1,42 @@
 <div id="patient_dashboard_content">
```
colinxfleming: the diff on this looks big, but it's basically changing col values and putting things in two rows (and spacing out the form fields).

lomky: github's ignore whitespace to the rescue!

colinxfleming:
thanks @lomky ! I'll merge once we get a prelim thumbs up on the feature branch.

PR #1767, opened 2019-09-01
Convert patient edit menu to BS4

colinxfleming:

Not quite ready for review yet because it hinges on some changes in #1766 so the diffs aren't as clear as they could be, but here it is.

This converts the menu over through some strategic changes to the css and html. (Pill nav is now done thru `a` tags.

* remove some css styling we don't need anymore (they're overwritten by regular bootstrap styling and possibly weren't doing anything)
* convert html elements to be bootstrap 4-ey

old:

image

in here, with slightly better alignment of some elements but no glyphicon:

image

* Bumps #1632

### DIFF
--- a/app/views/patients/_menu.html.erb
+++ b/app/views/patients/_menu.html.erb
@@ -1,5 +1,5 @@
-<section id='menu-content' class="abortion-left-menu">
-  <ul class="nav nav-pills nav-stacked" id='menu-list'>
+<section id='menu-content'>
+  <div class="nav nav-pills flex-column abortion-left-menu" id='menu-list' role='tablist' aria-orientation="vertical">
     <%= render 'menu_item', link_for: 'patient_information', active: true %>
     <%= render 'menu_item', link_for: 'abortion_information', active: false %>
     <%= render 'menu_item', link_for: 'practical_support', active: false unless Config.hide_practical_support? %>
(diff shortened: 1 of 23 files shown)


app/views/patients/_menu.html.erb
```diff
@@ -1,5 +1,5 @@
-<section id='menu-content' class="abortion-left-menu">
-  <ul class="nav nav-pills nav-stacked" id='menu-list'>
+<section id='menu-content'>
+  <div class="nav nav-pills flex-column abortion-left-menu" id='menu-list' role='tablist' aria-orientation="vertical">
```
tingaloo: aria-orientation is new to me. Gonna use this one.

app/views/patients/edit.html.erb
```diff
@@ -5,13 +5,11 @@
...
-  <div>
-    <div id="menu" class="col-sm-3">
-      <%= render 'patients/menu', patient: @patient %>
-    </div>
+  <div id="menu" class="col-3">
+    <%= render 'patients/menu', patient: @patient %>
   </div>
 
-  <div id="sections" class="col-sm-9 tab-content">
+  <div id="sections" class="col tab-content">
```
tingaloo: Nice to see flexbox being used by bootstrap, what a world.

colinxfleming:
@tingaloo keeping me honest here! Thanks for those css flags -- swapped to variables and `1rem` for everything.

PR #1771, opened 2019-09-07
Replaces all instances of test.com with example.com in accordance with IANA RFCs 2606, 6761

mdworken:
See link here for more info. This will negate the risk of sending emails to test.com.

https://www.iana.org/domains/reserved

PROS: security, compliance with web specs

CONS: will need to inform anyone who is likely to use the sandbox (add a flag above?)


colinxfleming:
we should run comms around this before merging. I'll hit the DARIA listserv and smash go on this for the weekend to give a few days of lead, so as to not interfere with scheduled trainings or anything.

colinxfleming:
sorry for the delayon this this @mdworken ! I'll smash merge on this and set a reminder in slack to run the account setup rake task on Wednesday.

PR #1778, opened 2019-09-14
BS4ify Patient notes section

colinxfleming:

BS4ifys the patient notes partial.

Because of some form juggling it doesn't really snap to a neat document structure. I have left a note in there as penance. You can get away with just looking at `patient/_notes` and `notes_table`.

* BS4ify patient notes partial

new one is a little cleaner and snaps to grid better!

olde:
image

new:
image

* Bumps #1632


app/views/notes/_notes_table.html.erb
```diff
@@ -1,11 +1,13 @@
-<table class="table border-side" id='notes_log'>
+<table class="table border-side top-spacer" id="notes_log">
```
tingaloo: would use BS spacing utility

app/assets/stylesheets/application.css.scss
```diff
@@ -401,4 +401,4 @@ form {
 
 .hidden {
```
tingaloo: would use .d-none

PR #1780, opened 2019-09-14
Setup spacing utility in bootstrap

tingaloo:

Bootstrap has a nice inline spacing utility. Instead of creating our utility classes, we can leverage them from BS4. 

I just targeted unique ids and classes that were used once.

Anyways utility CSS is another coffee table discussion because it enforces a structure whilst being highly flexible. The current app has custom utility functions which can use a refactor. But it's definitely worth looking into.

https://getbootstrap.com/docs/4.0/utilities/spacing/

* Setups spacing 
```
    1: ($spacer * .25), //.125 rem
    2: ($spacer * .5), //.25 rem
    3: $spacer, //.5 rem
    4: ($spacer * 1.5), //.75rem
    5: ($spacer * 2), //1rem
    6: ($spacer * 3) //1.5rem
```

* Bumps #1776

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ db/*.sqlite3
 /db/*.sqlite3-journal
 /log/*
 /tmp/*
-/bin/*
 
 # cert file
 cacert.pem
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ db/*.sqlite3
 /db/*.sqlite3-journal
 /log/*
 /tmp/*
+/bin/*
 
 # cert file
 cacert.pem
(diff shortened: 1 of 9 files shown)


colinxfleming:
will review a little tighter when we're closer to adding the ab info partial into the working branch, but initial opinion is that I'm way into this!

.gitignore
```diff
@@ -8,6 +8,7 @@ db/*.sqlite3
 /db/*.sqlite3-journal
 /log/*
 /tmp/*
+/bin/*
```
colinxfleming: we do wanna keep the `/bin/` directory checked into version control -- these are some key rails-related binaries. are they doing something weird on your machine?

tingaloo: woops, i'll put em back.

tingaloo:
lol i named the branch ps4,

colinxfleming:
ps have been laughing every time I look at this because of `ps4`

PR #1789, opened 2019-09-23
BS4ify the new patient partial

### LINKED ISSUE #1632: Move to bootstrap 4 and bootstrap_form 4

colinxfleming:

BS4ify the new patient form. (Please note that the CI failures are intentional and we're merging this into a featurebranch.) 

The alignment goes out of sync, but I'm not extremely troubled by that. This form is a little clogged up regardless, but does space a bit nicer when you just give it over to the flexboxes.

No particular sins against programming in here.

* BS4 to the new patient form

olde:

image

new:

image

* Fixes #1632

### DIFF
--- a/app/views/patients/_new_patient.html.erb
+++ b/app/views/patients/_new_patient.html.erb
@@ -1,54 +1,66 @@
+<h3 class="mt-5 mb-5"><small><%= t('patient.new.title') %></small></h3>
 
-<% if @results.empty? %>
-  <h3><small><%= t('patient.new.title') %></small></h3>
-<%# else %>
-  <%# when this is ready to go, you can remix this partial according to instructions from team design %>
-  <!-- <h3><small>Add a new patient:</small></h3> -->
-<%# end %>
-
-
-  <%= bootstrap_form_for patient do |f| %>
-    <div class="row">
-      <div class="col-sm-2">
-        <%= f.text_field :primary_phone, label: t('common.phone'), value: phone, required: true, pattern: "[0-9]{10}|[0-9]{3}\.[0-9]{3}\.[0-9]{4}|[0-9]{3}[-][0-9]{3}[-][0-9]{4}" %>
-      </div>
-      <div class="col-md-4">
-        <%= f.text_field :name, label: t('common.name'), value: name, required: true %>
-      </div>
-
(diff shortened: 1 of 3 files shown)


app/views/patients/_new_patient.html.erb
```diff
@@ -1,54 +1,66 @@
...
 
-      <div class="col-sm-2">
-        <%= f.select :line, options_for_select(lines, current_line) %>
-      </div>
+    <div class="col">
+      <%= f.select :line, options_for_select(lines, current_line) %>
+    </div>
 
-      <div class="col-sm-2 btn-space">
-        <%= button_tag class: 'btn btn-primary', id: 'create_new_patient_button' do %>
```
colinxfleming: I have no idea how this ever worked, so this actually is a somewhat meaningful change.

app/views/patients/_new_patient.html.erb
```diff
@@ -1,54 +1,66 @@
...
-      <div class="col-sm-2">
-        <%= f.text_field :primary_phone, label: t('common.phone'), value: phone, required: true, pattern: "[0-9]{10}|[0-9]{3}\.[0-9]{3}\.[0-9]{4}|[0-9]{3}[-][0-9]{3}[-][0-9]{4}" %>
-      </div>
-      <div class="col-md-4">
-        <%= f.text_field :name, label: t('common.name'), value: name, required: true %>
-      </div>
+<%= bootstrap_form_with model: patient do |f| %>
+  <div class="row">
+    <div class="col">
+      <%= f.text_field :primary_phone,
```
lomky: All of these single-line to mutli-line are making me so happy :100:

PR #1794, opened 2019-09-28
BS4ify events log

colinxfleming:

Convert the events log on the main dashboard to BS4. Nothing terribly interesting.

Note that CI fails intentionally here because we're merging this into the feature branch.

* bs4ify event log
* use fontawesome icons instead of glyphicons (rip)

old:

image

new -- the font size is a little increased and the thumbs ups are white now to make them easier to see:

image

* Bumps #1632

### DIFF
--- a/app/assets/stylesheets/events.scss
+++ b/app/assets/stylesheets/events.scss
@@ -2,13 +2,21 @@
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: http://sass-lang.com/
 
-.event-item {
-  padding: 5px;
+svg.event-item {
   border-radius: 50%;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/events.scss
```diff
@@ -2,14 +2,22 @@
...
+  padding: 4px;
+  background-color: #337ab7;
+  font-size: 1.25em;
+  
+  &.fa-thumbs-up {
     background-color: green;
   }
+
+  &.fa-comment {
+    background-color: #f8f9fa; // TODO replace with $light;
```
tingaloo: noticed the white barely shows.

PR #1795, opened 2019-09-28
BS4ify submit button

colinxfleming:

BS4ify the pledge modal. Mostly this means turning cols into rows, form_for into form_with, and hacking in a little bit of css that our multimodal vendor library was expecting.

When this is over (possibly when we start adding js components) we should work hard to figure out if we can ditch that library. Lots of magic going on, and that magic assumes bs3 which makes it hard to work on.

CI failures are expected because feature branch.

* bs4ify the pledge flow

olde:

image

new: this isn't super pretty. You'll notice that cancel doesn't float correctly. I think that's probably okay as we'd have to get into the vendor library and rewire things a little bit to get that to work right. Also the modal is a little slimmer? I'm going to make a note of that, we have a lot of screen real estate to work with here.

image

* Bumps #1632

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -405,3 +405,8 @@ form {
 .bottom-spacer {
   margin-bottom: .75em;
 }
+
+// This is necessary because of a vendored multimodal plugin that we should aim to deprecate
+.hide {
+  @extend .d-none;
+}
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/application.css.scss
```diff
@@ -405,3 +405,8 @@ form {
 .bottom-spacer {
   margin-bottom: .75em;
 }
+
+// This is necessary because of a vendored multimodal plugin that we should aim to deprecate
+.hide {
```
tingaloo: interesting, i guess it sticks around.

PR #1797, opened 2019-10-01
BS4: Widen the default modal

colinxfleming:

The modal in BS4 slims down to a max width of 500px at the highest breakpoint, which is kinda tiny. I'm setting it to 60% instead so that it doesn't get super cramped up in, for example, the accountants view, new call modal, or submit pledge modal.

like other BS4 PRs, CI fails intentionally because we're smashing this into a feature branch `bs4-2`, so don't be put off by that when reviewing.

* `max-width: 60%` on `.dialog-modal`

olde:

image

new:

image

* Bumps #1632

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -410,3 +410,8 @@ form {
 .hide {
   @extend .d-none;
 }
+
+// widen default modal from bootstrap default
+.modal-dialog {
+  max-width: 60%;
+}


app/assets/stylesheets/application.css.scss
```diff
@@ -410,3 +410,8 @@ form {
 .hide {
   @extend .d-none;
 }
+
+// widen default modal from bootstrap default
+.modal-dialog {
+  max-width: 60%;
```
tingaloo: no utility class exists for this, we got em.

PR #1802, opened 2019-10-07
BS4 - Replace remaining glyphicons

colinxfleming:

Sub in the remaining fontawesome icons for the glyphicons. rip glyphicons.

Along the way, I have learned some valuable lessons about doing col-s in bootstrap 4, so subbed those in, and removed some extraneous comments.

* remove remaining glyphicons

* Bumps #1632

### DIFF
--- a/app/assets/stylesheets/pregnancies.scss
+++ b/app/assets/stylesheets/pregnancies.scss
@@ -36,8 +36,10 @@ $font-size: 17px;
     background-color: white;
 
     &:before {
-      font-family: 'Glyphicons Halflings';
-      content: "\e080";
+      display: none;
+      font-family: "Font Awesome 5 Free";
+      font-weight: 900;
+      content: "\f054";
       margin-left: -1.5em;
       padding-right: .85em;
       color: $daria-color;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/pregnancies.scss
```diff
@@ -49,6 +51,11 @@ $font-size: 17px;
   }
 }
 
+.svg-inline--fa {
```
colinxfleming: this is going to need to be nested

app/views/layouts/application.html.erb
```diff
@@ -5,7 +5,7 @@
     <title><%= content_for?(:title) ? yield(:title) : "DARIA" %></title>
     <meta name="description" content="<%= content_for?(:description) ? yield(:description) : "DCAF" %>">
     <meta name='ROBOTS' content='NOINDEX, NOFOLLOW'>
-    <%= javascript_pack_tag 'application' %>
+    <%= javascript_pack_tag 'application', 'data-search-pseudo-elements': true %>
```
colinxfleming: this is necessary because we're doing some `before` selector stuff; fontawesome can handle this but it requires some extra junk. I don't think it quite rises to the level of tech debt to figure out how to do the grid stuff without this but it was kind of a pain to figure out.

tingaloo: can you take a before and after screenshot? i'm trying to figure out how the before selector is used.

colinxfleming:
before selector screenshots!

old:

image


new: 

image

app/views/layouts/application.html.erb
colinxfleming: added a screener!

PR #1805, opened 2019-10-08
ruby version update

andrewrosario:
Updated ruby version in all files.


.docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.6.2-alpine
+FROM ruby:2.6.5
```
colinxfleming: @DarthHater this kind of raises an interesting question, in your opinion is alpine still serving us? 

@andrewrosario in lieu of @DarthHater weighing in, would you mind restoring the docker image to alpine? 

```suggestion
FROM ruby:2.6.5-alpine
```

package.json
```diff
@@ -2,6 +2,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.10.2",
     "@rails/webpacker": "4.0.7",
+    "docker": "^1.0.0",
```
colinxfleming: Can you help me understand where these changes from? I'm a little surprised to see JS differences in here given the nature of this update. If this is doing something in particular, could we maybe roll it into a different PR?

For now, I'd like this change removed, and for there to not be changes to package-lock.json / yarn.lock, if that's cool!

andrewrosario: Sorry about that... not sure what happened... This is my first contribution to an open source project. I'll fix it and open a new PR.

colinxfleming: no apologies necessary! I do this all the time too,  it's why we have peer review on pull requests.

andrewrosario:
Thanks so much... I believe I've made the changes you requested.

.docker/Dockerfile
DarthHater: It's buyers choice. All alpine gets the project is a smaller image to work with ultimately, but Alpine is as you know, tricky to work with. Since we don't use the image in production it's kinda whatever is best for developers. Slow download time vs lots of Alpine specific hacks.

PR #1810, opened 2019-10-13
BS4 system tests part one - the easy part

colinxfleming:

System tests are shaking out to be two buckets: One is slight changes to ids or one-liner fixes, and the others are a little more complex - probably form surgery or some things that don't translate as well in BS4. This PR is the easy stuff.

* easy PR fixes - id swaps, links becoming buttons, make some forms synchronous

no view changes

* Bumps #1632

### DIFF
--- a/test/system/navbar_links_test.rb
+++ b/test/system/navbar_links_test.rb
@@ -7,7 +7,7 @@ class NavbarLinksTest < ApplicationSystemTestCase
   end
 
   describe 'user dropdown' do
-    before { click_button @user.name }
+    before { click_on @user.name }
     it 'should display the profile link' do
       assert has_link? t('navigation.user_tools.profile'), href: edit_user_registration_path
     end
@@ -15,7 +15,7 @@ class NavbarLinksTest < ApplicationSystemTestCase
 
   describe 'admin dropdown' do
     describe 'admin view' do
-      before { click_button 'Admin' }
+      before { click_on 'Admin' }
 
       it 'should display User management link' do
         assert has_link? t('navigation.admin_tools.user_management'), href: users_path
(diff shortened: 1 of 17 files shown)


test/system/navbar_links_test.rb
```diff
@@ -48,7 +48,7 @@ class NavbarLinksTest < ApplicationSystemTestCase
         @user2 = create :user, role: :data_volunteer
         log_in_as @user2
         visit authenticated_root_path
-        click_link t('navigation.admin_tools.label')
+        click_button t('navigation.admin_tools.label')
       end
```
mdworken: If we want to future proof these a little, we could use 'click_on' which works with both links and buttons

test/system/navbar_links_test.rb
```diff
@@ -85,7 +85,7 @@ class NavbarLinksTest < ApplicationSystemTestCase
...
 
       it 'should not display an admin link' do
-        refute has_link? t('navigation.admin_tools.label')
+        refute has_button? t('navigation.admin_tools.label')
```
mdworken: Since matchers like 'has_button' return after either the matcher is satisfied OR the default wait time is reached, this anti-pattern may be adding 5 seconds to each test of this form.This isn't a blocker thought so I can spin off another PR for it if you'd like

colinxfleming: The matcher waits are intentional here - poltergeist used to be pretty spotty about this kind of thing, and I still don't completely trust the new system test regime completely. You are right that this could slow build times down, but I'm less convinced that long build times are a pain point we're experiencing at the moment. Do you feel differently, or is there a different issue worth raising?

test/system/navbar_links_test.rb
colinxfleming: this is a great call -- will sub!

PR #1813, opened 2019-10-14
added the webpacker service to dockerization

### LINKED ISSUE #1741: Docker Build Broken
image

ChrisBarthol:

On a fresh `docker-compose up --build` the app fails to load due to the unavailability of webpack and assets.

* add webpacker as a service
* moved the entry point commands to their own scripts for easier modification
* included the pid server removal in the script

* Fixes #1741

### DIFF
--- a/scripts/start_rails.sh
+++ b/scripts/start_rails.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b 0.0.0.0
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,17 @@
 version: '3.1'
 services:
+  webpacker:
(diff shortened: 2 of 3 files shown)


scripts/start_rails.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b 0.0.0.0
```
colinxfleming: I think currently we're handling the `rm tmp/pids/server.pid` by having it in the dockerignore file, so I'm curious if something bit you here?

colinxfleming:
(ps the CI failure is a CircleCI problem and totally fake; you're good)

scripts/start_rails.sh
ChrisBarthol: Yep, I was still getting this on some up's so I included it here.

Issue #1791, opened 2019-09-24
Add new fields to data entry form


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Our data entry view is a little dusty and a lot has happened to the data model since we built it, so we should spruce it up by adding the new fields.

* **What feature or behavior is this required for?**

Keeping data entry mode fresh, which means that people doing that will have to enter less data into the

* **How could we solve this issue? (Not knowing is okay!)**

Please add the following fields to the form in `app/views/patients/data_entry.html.erb`:

* Language
* Voicemail preference
* Resolved without fund
* Referred to clinic
* Completed ultrasound

You are welcome to use your best judgment about where to put them in the rest of the form - either put them on the end or mix them in, in a way that makes sense to you.

Further, please pop open `test/system/data_entry_test.rb` and add these new fields to the list of values we test. Be sure to test for values that aren't the default.

* **Anything else?**

Best to have in place before Nov 1!

colinxfleming:
resolved via #1815 ; tests in #1828

PR #1815, opened 2019-10-15
Added Requested Fields Issue #1791

andrewrosario:

Added requested fields to data entry form

- added Language field
- added Voicemail preference field
- added Resolved without fund field
- added Referred to clinic field
- added Completed ultrasound field

1791


app/views/patients/data_entry.html.erb
```diff
@@ -29,6 +29,9 @@
   <%= f.text_field :state, autocomplete: 'off' %>
   <%= f.text_field :county, autocomplete: 'off' %>
 
+   <%= f.select :language, options_for_select(language_options), autocomplete: 'off' %>
+   <%= f.select :voicemail_preference, options_for_select(voicemail_preference_options), autocomplete: 'off' %>
```
lomky:
```suggestion
   <%= f.select :voicemail_preference, options_for_select(voicemail_options), autocomplete: 'off' %>
```

Great thinking on the helper! We actually have a similar one implemented on the Dashboard helper`app/helpers/dashboards_helper.rb:voicemail_options` that we can take advantage of here! That will mean you won't need to new patient helper function, nor the config/locale/en changes, I believe.

andrewrosario:
I've made the requested changes to the voicemail_options and related files. I was unsure about the tests, because only two of the added fields already appear in the patient_information.html.erb file to check for. I thought adding them might be outside of the scope of the issue, so I didn't want to mess with it.

colinxfleming:
Per discussion, we're going to handle the tests in a separate issue. Thank you very much for grinding this out @andrewrosario , you nailed it!

PR #1819, opened 2019-10-18
Create successful Docker build locally

rudietuesdays:

The local Docker server was failing on Webpacker. I updated the Dockerfile to include a couple more necessary build tools, and generated a fresh yarn.lock file to squash the error.

* Adds deps to Dockerfile
* Updates yarn.lock


colinxfleming:
I think this might overtaken by events thanks to adding webpacker as a docker service in #1813 but let's keep this open in the event that something goes sideways.

colinxfleming:
closing - moving webpack to its own docker service resolves this for us, so we're cool I think?

PR #1820, opened 2019-10-18
Update Docker config to get passing build

rudietuesdays:

Very small tweak to the Docker `config.yml` file to get the build into the green


.docker/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Colin Fleming <c3flemin@gmail.com>
 # configure environment variable
 # note: move this to three ARG commands when CircleCI updates their docker
 ENV DCAF_DIR=/usr/src/app \
-    BUILD_DEPENDENCIES="build-base libxml2-dev libxslt-dev linux-headers bash openssh fontconfig fontconfig-dev curl" \
+    BUILD_DEPENDENCIES="build-base libxml2-dev libxslt-dev linux-headers bash openssh fontconfig fontconfig-dev curl ruby-dev alpine-sdk" \
```
colinxfleming: can you help me understand what's going on with these changes? I would have thought we'd have gotten ruby-dev and alpine-sdk from the base image, though I admittedly don't know that much about what alpine does/comes with!

.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-          docker_layer_caching: true
+          docker_layer_caching: false # true
```
colinxfleming: hm I think this is in circle's CI system, which works fine for builds for people who are contributors, but which consistently errors for reasons we can't fix for people working from forks. (For example, you are possibly experiencing this as someone working from a fork!) I think you're going to want to pull in changes from upstream, which should include a github action CI thing (I hope! if that's not the case then we can re-eval)

colinxfleming:
@rudietuesdays please do reopen if I have this wrong, but I think the meat of this is in #1819 so I'm going to close this. please do correct me if I'm off and reopen!

PR #1829, opened 2019-10-19
Has Special Circumstance Always Returns True

### LINKED ISSUE #1827: CSV Export shows all TRUE for special circumstances
Our CSV export provides a limited about of anonymized information, including a boolean field specifying whether the patient had any Special Circumstances checked under Patient Information. But right now it is exporting TRUE for everyone, regardless of whether there are any checked.

 - Fix the CSV export to properly export `has special circumstances`
 - Update our tests to check that!

To reproduce, click the admin dropdown in the upper right and click 'Export anonymized CSV'

BRAF bug request!

ChrisBarthol:

The `has_special_circumstance` method was always returning true to boolean safety checks against ints.

```
irb(main):001:0> !!0
=> true
irb(main):002:0> !!1
=> true
irb(main):003:0> !!false
=> false
irb(main):004:0> !!true
=> true
```

* switched from `0` and `1` to `true` and `false`
* added a test

* Fixes #1827

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -246,9 +246,9 @@ class Patient
   end
 
   def has_special_circumstances
-    has_circumstance = 0
+    has_circumstance = false
     special_circumstances.each do |cir|
-      has_circumstance = 1 if cir.present?
+      has_circumstance = true if cir.present?
       break
     end
     !!has_circumstance
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -246,12 +246,7 @@ class Patient
   end
 
   def has_special_circumstances
-    has_circumstance = false
-    special_circumstances.each do |cir|
-      has_circumstance = true if cir.present?
-      break
-    end
-    !!has_circumstance
+    special_circumstances.map { |circumstance| circumstance.present? }.any?
   end
 
   def archive_date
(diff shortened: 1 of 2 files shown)


colinxfleming:
Checks out! Thank you much @chrisbarthol, glad this was an easy fix. I will review in more depth later this wknd.

app/models/patient.rb
```diff
@@ -246,9 +246,9 @@ def notes_count
   end
 
   def has_special_circumstances
-    has_circumstance = 0
+    has_circumstance = false
```
colinxfleming: This is great! I think we got tripped up by the handling here and what ruby considers truthy.

@ChrisBarthol what do you think about simplifying this a little? The variable assignment and then boolean casting is kinda tricky. Maybe something like this would work?

`special_circumstances.map { |x| x.present? }.any?`

That would cast each of the special circumstances to true or false, and then return true if there are any trues or false otherwise. Ex:

```
irb(main):001:0> [false, true, true].any? 
=> true
irb(main):002:0> [false, false, false].any? 
=> false
```

what do you think?

ChrisBarthol: Great catch on the simplification!  Switched to map and added a few extra test cases.

colinxfleming:
thank you very much @ChrisBarthol !

PR #1840, opened 2019-10-23
Add system specs for new fields

### LINKED ISSUE #1828: Update system tests to account for changes to data_entry workflow

We recently merged https://github.com/DCAFEngineering/dcaf_case_management/pull/1815/files which updated our data entry workflow with some new fields. We want to now update the test in test/system/data_entry_test.rb to match.

Specifically, we added these form fields:

* language
* voicemail_preference
* referred_to_clinic
* completed_ultrasound
* resolved_without_fund

We should follow the pattern in data_entry_test of entering in values for these fields, saving the patient, and then checking that the values saved properly.

Keeping our data entry mode well tested, and ensuring that it doesn't eat data!

We do something pretty similar in test/system/update_patient_info.rb - there's probably stuff in there we can crib for this!

* **Anything else?**

Rails' documentation on system tests is here: https://guides.rubyonrails.org/testing.html#system-testing

These are tests that are basically issuing commands to a web browser and seeing what happens. You can run them with `rails test:system`.

You also have to install a lot of junk to run them to get them to run, so don't worry about seeing it work, just whip up something that seems reasonable and pull request it. Usually we run these in an external system that's set up to handle them so developers don't have to.

valeriecodes:

Adds system tests for new form fields

* Fixes #1828

### DIFF
--- a/test/system/data_entry_test.rb
+++ b/test/system/data_entry_test.rb
@@ -41,6 +41,10 @@ class DataEntryTest < ApplicationSystemTestCase
       fill_in 'Procedure Cost', with: '200'
       fill_in 'Patient contribution', with: '150'
       fill_in 'National Abortion Federation pledge', with: '50'
+      select 'English', from: 'patient_language'
+      select 'Do not leave a voicemail', from: 'patient_voicemail_preference'
+      check 'patient_referred_to_clinic'
+      check 'patient_completed_ultasound'
       check 'Pledge Sent'
       check 'Fetal_patient_special_circumstances'
       check 'Home_patient_special_circumstances'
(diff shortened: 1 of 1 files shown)


test/system/data_entry_test.rb
```diff
@@ -93,6 +98,9 @@ class DataEntryTest < ApplicationSystemTestCase
         assert has_field? 'Patient contribution', with: '150'
         assert has_field? 'National Abortion Federation pledge', with: '50'
         assert has_field? 'DCAF pledge', with: '100'
+        assert has_checked_field? 'Referred to clinic'
+        assert has_checked_field? 'Ultrasound completed?'
+        assert has_unchecked_field? 'Resolved without assistance from DCAF'
```
colinxfleming: in light of #1841 should this have a thing up near L47 like `check 'patient_resolved_without_fund'` and an associated `assert has_checked_field?` here?

valeriecodes: Done!

colinxfleming:
thank you very much @valeriecodes !

PR #1843, opened 2019-10-26
coerce URI to use HTTPS protocol

### LINKED ISSUE #1826: Make the Fax Service link less finicky
We have the fax URL set as a user configurable option. Right now if the user forgets to type `https://` at the front of the link in the Configuration Management entry, the link at the bottom of the page is broken. It works fine if they have the https, but let's make the app worry about this instead of putting it on our users! Let's make it work regardless of https or not.

#### Broken
Screenshot from 2019-10-18 21-00-56

#### Working!
Screenshot from 2019-10-18 21-01-27

Kevin-Wei:

Use new URIService Utility 

* Fixes #1826


lomky:
Closing in favor of #1845

Kevin-Wei:
Will be refactoring to use the utility in #1845

app/services/uri_service.rb
```diff
@@ -2,14 +2,17 @@ class URIService
   attr_accessor :uri
 
   def initialize(uri)
+    return nil if uri.nil?
```
Kevin-Wei: Not a huge fan of this PORO handling nil initializations

app/services/uri_service.rb
```diff
@@ -2,14 +2,17 @@ class URIService
...
       @uri = URI.parse(URI.encode(uri))
     rescue URI::InvalidURIError
       nil
     end
   end
 
-  def secure_scheme!
+  def secure_scheme_uri!
     @uri.scheme = "https"
+    @uri
```
Kevin-Wei: added this return here to and changed method name to make the interface to the PORO friendlier

colinxfleming:
looks good to me but want to get @lomky 's review as the issue reporter here.

Kevin-Wei:
if we need something more robust, its possible [this gem](https://github.com/sporkmonger/addressable) handles the edgiest of edge cases.

colinxfleming:
@lomky ready for another pass

PR #1845, opened 2019-11-01
Add URI https-ification service

Kevin-Wei:

Adds a resilient URI service utility that can force the scheme to https

adds utility and test

* Connects #1826 (fix coming in #1843)


Kevin-Wei:
placeholder idea for the moment

PR #1847, opened 2019-11-05
Status helper text updates when patient status updates (#1825)

### LINKED ISSUE #1825: Status Helper Text should update alongside the Status Text
Our Status field dynamically updates without having the whole page reload, which is great! But the helper text associated with it (hover on the "?") does not! So they can mismatch :frowning: Can we make it update dynamically too?

To recreate: After seeding your data, click a patient with status "Needs Appointment". Notice the hover text matches! Pick an appointment date. The Status updates to Fundraising, but the helper text is left behind!

** Another Example
Screenshot from 2019-10-18 19-29-06

Clarity for our Case Managers.

Probably similar to how we make the Status field update?

harumhelmy:

Hello! I added a little bit of code in update.js.erb to dynamically update the tooltip content rendered in _patient_dashboard.html.erb. 

* Without having to refresh the page, the helper text content in the tooltip (which shows up when you hover over the '?' next to a patient's status text) updates accordingly when the status itself changes (the change usually happens when a CM has entered the patient's  appointment date, or a pledge has been sent). 

dynamic-tooltip

* Fixes #1825

### DIFF
--- a/app/views/patients/update.js.erb
+++ b/app/views/patients/update.js.erb
@@ -5,7 +5,13 @@ $('#patient_appointment_date')
   .text("<%= t('patient.dashboard.approx_gestation', weeks: @patient.last_menstrual_period_at_appt_weeks, days: @patient.last_menstrual_period_at_appt_days) %>");
 
 // rerack status
-$('#patient_status').text('<%= @patient.status %>');
+$('#patient_status')
+  .text('<%= @patient.status %>');
+
+// rerack tooltip content 
+$('.daria-tooltip.tooltip-header-help')
+                  .attr('title', "<%= status_help_text(@patient)%>")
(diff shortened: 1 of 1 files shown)


app/views/patients/update.js.erb
```diff
@@ -5,7 +5,13 @@ $('#patient_appointment_date')
...
 
 // rerack status
-$('#patient_status').text('<%= @patient.status %>');
+$('#patient_status')
+  .text('<%= @patient.status %>');
+
+// rerack tooltip content 
+$('.daria-tooltip.tooltip-header-help')
+                  .attr('title', "<%= status_help_text(@patient)%>")
+                  .tooltip('fixTitle')
```
colinxfleming: Today I learned that this was a thing!

Issue #1816, opened 2019-10-16
Begin porting SCSS into webpacker


colinxfleming:
*WARNING: This one's tough and involves configuring webpacker, the rails implementation of webpack. Not for the faint of heart, or the new to rails!*

* **What are we trying to do?**

We want to start managing frontend assets in webpacker for easier maintenance, and for simpler package management. (Asset gems are a little clunky, unfortunately.) As part of that, let's set up webpacker 

Let's start by setting up webpacker to handle CSS, and port over `app/assets/lines.scss` to our webpacker build as a proof of concept.

I picked this file in particular because it's real straightforward and has a couple nested classes, so we can confirm any SCSS transpilation is working right.

* **What feature or behavior is this required for?**

Consolidation into webpack; installing a JS framework someday!

* **How could we solve this issue? (Not knowing is okay!)**

I'm pretty clueless about the mechanics of this but this looks promising: https://github.com/rails/webpacker/blob/master/docs/css.md

* **Anything else?**

colinxfleming:
(Please leave for tech rebalanced volunteers!)

rudietuesdays:
I'm on this!

PR #1849, opened 2019-11-06
Working on adding ability to compile styles in Webpack

rudietuesdays:

Webpack is compiling but not able to glob or deal with complicated extentions. WIP!

* #1816


colinxfleming:
@rudietuesdays can I bump you on this, or may I finish it out? I think this is standing between us and being able to release a lot of hacknight friendly stuff into the pool, and also some react stuff I think (since when we have webpacker compiling our scss, we'll be able to move over bootstrap).

colinxfleming:
I used this as a foundation for #1948 and it was super helpful in getting it to work! We're squared away here so closing.

PR #1858, opened 2019-11-11
Use a btn and not a link on add new clinics too

### LINKED ISSUE #1632: Move to bootstrap 4 and bootstrap_form 4

colinxfleming:

my reign of terror against a:visited by switching these to buttons continues

* Fixes #1632

### DIFF
--- a/app/views/clinics/index.html.erb
+++ b/app/views/clinics/index.html.erb
@@ -1,6 +1,6 @@
 <h1><%= t 'clinics.index.title' %></h1>
 
-<%= link_to t('clinics.new.title'), new_clinic_path, class: 'btn btn-primary' %>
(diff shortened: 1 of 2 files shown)


colinxfleming:
godmode for iteration

PR #1860, opened 2019-11-11
Remove section tag and bootstrap_form_tag

### LINKED ISSUE #1787: Remove section tags and bootstrap_form_tags from app
bootstrap_form_tag / bootstrap_form_for

section elements

colinxfleming:

Removes a few things we don't need to use anymore now that we're in full on BS4.

The new rails hotness is `bootstrap_form_with`, so we're converting some of these over. The bootstrap_form_for stuff is more complicated to port and there's more of it, so we can do that in its own ticket which I'll write up in a sec.

* remove section tag and styling
* convert bootstrap_form_tag to bootstrap_form_with url:

* Fixes #1787

### DIFF
--- a/app/views/lines/new.html.erb
+++ b/app/views/lines/new.html.erb
@@ -2,7 +2,7 @@
   <div class="col lines-form">
     <h1 class="border-bottom title"><%= t 'lines.new.welcome_to_daria', fund: FUND %></h1>
 
-    <%= bootstrap_form_tag url: lines_path, class: 'line-select-form' do |f| %>
+    <%= bootstrap_form_with url: lines_path, local: true, class: 'line-select-form' do |f| %>
       <%= f.form_group :line, label: { text: t('lines.new.what_line') } do %>
         <% lines.each do |line| %>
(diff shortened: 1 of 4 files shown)


app/views/lines/new.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="col lines-form">
     <h1 class="border-bottom title"><%= t 'lines.new.welcome_to_daria', fund: FUND %></h1>
 
-    <%= bootstrap_form_tag url: lines_path, class: 'line-select-form' do |f| %>
+    <%= bootstrap_form_with url: lines_path, local: true, class: 'line-select-form' do |f| %>
```
lomky: Is this `local: true` a change, or is it simply explicit on `with` when it was implied on `tag`?

colinxfleming: the latter -- `bootstrap_form_tag` was local by default, but `bootstrap_form_with` is remote by default. Not a behavior change, just a difference in what we have to declare.

colinxfleming:
thank you @harumhelmy and @lomky !

PR #1862, opened 2019-11-20
Add a welcome message

Elexc:
Issue #884

* Added an encouraging message for case managers :)


app/helpers/welcome_message_helper.rb
```diff
@@ -37,6 +37,7 @@ def welcome_messages
       "Thank you for taking the time to be the most thoughtful CM that you can be! <3",
       "Você é demais!",
       "Your work is so valuable and you're doing a wonderful job! <3",
+      "Your work empowers women to make decisions about their own bodies. Bravo!",
```
colinxfleming: Love this message, which gets to the very core of what abortion funds are about. I have very slight edit in keeping with DCAF's inclusivity policy re: gender identity from https://dcabortionfund.org/about/ -

```suggestion
      "Your work empowers people to make decisions about their own bodies. Bravo!",
```

@Elexc if that sounds good, please hit accept suggestion and I'll merge! Otherwise we can talk about it next time we run into each other.

Issue #1835, opened 2019-10-20
Build rake task to generate fake data


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

This would have been extremely helpful as we let some data scientists at DARA today! Let's build a rake task that will just generate like 5k rows of fake patients for export.

* **What feature or behavior is this required for?**

Safely giving people something they can try analysis scripts against.

* **How could we solve this issue? (Not knowing is okay!)**

Here's a starter kit. We need to add calls, external pledges, practical supports, and fulfillments at the least, and also account for the fact that a lot of this stuff is naturally null.

```
users = User.all
clinics = Clinic.all

10000.times do |i|
  flag = (rand(10) % 5 == 0)

  initial_call = Date.today - rand(1000)

  has_appt = rand(3) % 2 == 0

  has_pledge = rand(3) != 2

  patient = Patient.create!(
    name: 'Randomized Patient',
    primary_phone: "#{i}".rjust(10, '0'),
    initial_call_date: initial_call,
    urgent_flag: flag,
    line: i.even? ? 'DC' : 'MD',
    clinic: has_appt ? clinics.sample(1).first : nil,
    created_by: users.sample(1).first,
    appointment_date: has_appt ? initial_call + rand(30) : nil,
    last_menstrual_period_weeks: rand(15) + 7,
    last_menstrual_period_days: rand(7),
    procedure_cost: has_appt ? rand(600) : nil,
    pledge_sent: has_appt && has_pledge,
    patient_contribution: rand(400),
    fund_pledge: has_appt ? rand(300) : nil
  )
end
```

harumhelmy:
I've never built a rake task before but I'm interested in trying it out - maybe pairing with @mdworken this weekend? :)

marycodes2:
@harumhelmy I'd be happy to pair with you if you'd like!

harumhelmy:
Sorry this is taking me ten million years to get started on! But I am working on it, hoping to show my work sometime this week :)

PR #1873, opened 2019-12-10
A first attempt at creating a rake task that generates fake patient data

harumhelmy:

Hallo! Based off of Colin's request and starter code outlined in #1835, I built a rake task that generates fake randomized patient data, including external pledges, calls, practical supports and fulfillments -- accounting for the fact that some of them is null. You can run the task by running `rails db:seed:create_fake_data`. This is a first stab! Edits and suggestions super welcome.

* Created a file called create_fake_data.rake in lib/tasks 
* In that file, in the namespace db:seed, created a rake task that generates fake patient data 5000 times

Question/musing for @colinxfleming:
* The randomized logic wasn't at all based on reality (e.g. 75-80% fulfillment rate -- I just made fulfillments for fake patients whose index numbers are even and has a pledge sent). I wasn't sure whether they need to be!
* What else would be helpful? Tests? Logs?


lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+            last_menstrual_period_weeks: rand(15) + 7,
+            last_menstrual_period_days: rand(7),
+            procedure_cost: has_appt ? rand(600) : nil,
+            pledge_sent: has_appt && has_pledge,
+            patient_contribution: rand(400),
+            fund_pledge: has_appt ? rand(300) : nil
+          )
+
+          # create external_pledges
+          if idx.odd? && has_appt
```
colinxfleming: this is great idiomatic ruby

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
+namespace :db do
+  namespace :seed do
+    desc 'Generate fake patient entries for data wranglers'
+    task :create_fake_data => :environment do
+
+        users = User.all
+        clinics = Clinic.all
+
+        5000.times do |idx|
+          flag = (rand(10) % 5 == 0)
```
colinxfleming: A thing that we appreciate about the seeds file is that it spits out consistent data every time -- when we're faking it, that's kind of an asset, since you know the base layer is consistent when you're just screwing around. You've already got the modulo set up here, so what do you think about doing `idx % 5 == 0`?

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+
+          initial_call = Date.today - rand(1000)
+          
+          has_appt = rand(3) % 2 == 0
```
colinxfleming: Same deal for this and `has_pledge` - might behoove us to use the modulo of `idx` to be consistent. (I'm less worried about `initial_call` since there are so many possible values, but you could do it there too if you were interested in getting weird with it!)

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+            initial_call_date: initial_call,
+            created_by: users.sample(1).first,
+            urgent_flag: flag,
+            line: lines[idx%3], # thank you seeds.rb! 
+            clinic: has_appt ? clinics.sample(1).first : nil,
+            appointment_date: has_appt ? initial_call + rand(30) : nil,
+            last_menstrual_period_weeks: rand(15) + 7,
+            last_menstrual_period_days: rand(7),
+            procedure_cost: has_appt ? rand(600) : nil,
+            pledge_sent: has_appt && has_pledge,
```
colinxfleming: this is great ruby!

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+          
+          has_appt = rand(3) % 2 == 0
+
+          has_pledge = rand(3) != 2 
+
+          lines = ['DC', 'VA', 'MD'] # need to add Spanish maybe? 
+
+          patient = Patient.create!(
+            name: 'Randomized Patient',
+            primary_phone: SecureRandom.random_number(10**10).to_s.rjust(10, '0'),
```
colinxfleming: Since we require phone numbers to be unique, there's a slim but real chance that this will cause conflicts and error out. Your use of `rjust` is really inspired though. Can you think of a way that you can produce 5k unique valid phone numbers in a sequence? Something like `999-99X-XXXX` would work. (Let me know if that makes sense, I might notbe explaining it well.)

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+
+          lines = ['DC', 'VA', 'MD'] # need to add Spanish maybe? 
+
+          patient = Patient.create!(
+            name: 'Randomized Patient',
+            primary_phone: SecureRandom.random_number(10**10).to_s.rjust(10, '0'), 
+            initial_call_date: initial_call,
+            created_by: users.sample(1).first,
+            urgent_flag: flag,
+            line: lines[idx%3], # thank you seeds.rb!
```
colinxfleming: this is a cool solution to picking a line - part of programming is having a good eye for what ideas to borrow and this is great

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+            source: 'Metallica Abortion Fund',
+            amount: 100,
+            created_by: User.first
+            )
+          end
+
+          # create calls
+          call_status = ["Left voicemail", "Reached patient", "Couldn't reach patient"]
+
+          if idx.odd? && !has_appt
```
colinxfleming: I'm a little confused by the conditionals. Will the second one always happen, or will there be times when people sometimes have no calls?

It might be worth tweaking for clarity, or adding a comment about how this works.

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,89 @@
...
+          
+          # create practical_support
+          support_types = [
+            'Companion', 
+            'Lodging', 
+            'Travel to the region', 
+            'Travel inside the region', 
+            'Other (see notes)']
+          
+          if has_appt && patient.procedure_cost > 350 && has_pledge
```
colinxfleming: I don't know the answer to this: do patients without pledges ever have practical supports?

colinxfleming:
oh ps on your questions: if you know off the top of your head how to make the rates a little closer then go ahead and adjust the math, but don't put more time than that into it, as we can always tune it later. 

Don't worry about tests or logging - if you want to be thorough you can add a `puts Patient.count`  at the end similar to what the seeds file does but I've never really gotten much mileage out of testing rake tasks themselves like that, and what we're doing here isn't production-facing anyway.

lib/tasks/create_fake_data.rake
harumhelmy: probably not, tbh!

lib/tasks/create_fake_data.rake
harumhelmy: Yes, I was initially thinking that some people will have no calls and the conditionals was kind of a lazy way of randomizing that. Thinking about it further now, it's more likely that people will have at least one call once they're entered in DARIA.

lib/tasks/create_fake_data.rake
harumhelmy: Love it! Also, gosh, so much of this beautiful logic was laid out in your starter code, so I have you to thank for all this Ruby brilliance! SO MUCH LEARNIN HAPPENED

harumhelmy: Love it! Thank you.

lib/tasks/create_fake_data.rake
harumhelmy: So `'#{idx}'.rjust(10, '9')`, which would produce 999-999-XXXX, creates the same phone number for indeces 9, 99, and 999! -_- I'll keep looking around to see what else can create unique valid phone numbers in a sequence, but for now I kept the SecureRandom.random_number thing, but added the index as the second argument to the `rjust`, which I think should make it more unique?

harumhelmy:
Hiya! I incorporated some of your comments, but not sure if my modulo stuff makes sense. Would love to chat about it more. Didn't really have anything off the top of my head to make the rates closer to reality, but would love to be present when folks are fine-tuning that! :) Another thing that could be done better is making sure the `external_pledge` + `patient_contribution` + `fund_pledge` doesn't exceed/isn't under the `procedure_cost`.

harumhelmy:
(FYI still working on this — getting some help reducing the use of `rand`! :)

colinxfleming:
@harumhelmy what's next steps on this, anything I can help with to push it over the finish line?

harumhelmy:
Thanks for the patience and the bump, @colinxfleming! Got some help from @amackcrane and learned things about seeding Ruby's pseudo-random number generator. The task spits out the same thing every time now! It's still missing some demographic information, and I'm wondering if that's something we want, as well, and how realistic we'd want that information to look (i.e. how taking into account the gaps in our CMs data collection). I guess I'm still unclear how the data wranglers will use this.

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,93 @@
...
+  namespace :seed do
+    desc 'Generate fake patient entries for data wranglers'
+    task :create_fake_data => :environment do
+
+        Patient.destroy_all
+
+        users = User.all
+        clinics = Clinic.all
+
+        gen = Random.new(2020) # our random number generator: https://ruby-doc.org/core-2.7.0/Random.html
```
colinxfleming: love this seed technique

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,93 @@
+namespace :db do
+  namespace :seed do
+    desc 'Generate fake patient entries for data wranglers'
+    task :create_fake_data => :environment do
+
+        Patient.destroy_all
```
colinxfleming:
```suggestion
```

fine to not overwrite the current seeds, imo.

lib/tasks/create_fake_data.rake
```diff
@@ -0,0 +1,93 @@
...
+
+        users = User.all
+        clinics = Clinic.all
+
+        gen = Random.new(2020) # our random number generator: https://ruby-doc.org/core-2.7.0/Random.html
+
+        5000.times do |idx|
+          flag = gen.rand < 0.05 # gen.rand will always return a float between 0 and 1
```
colinxfleming: Let's put an additional logging statement in, since making 5k patients takes a little while:
```suggestion
          idx % 10 == 0 ? print("\nMaking fake pt #{idx} of 5000") : print('.')
          flag = gen.rand < 0.05 # gen.rand will always return a float between 0 and 1 
```

colinxfleming:
The goal of having this rake task around is that when an analyst is hacking on our data or trying to build a report, we'd like to provide them a kinda-close example of what it looks like, without actually having to give them a real pull. We have gaps in our real-life data, too, because we don't require demographic fields and CMs don't always drop them in, even for folks who get pledges.

I think this looks pretty good to my eyes - @harumhelmy please mark this ready when you feel done working on it and we can get it in. (I think it's mergeable for sure, but want to make sure you don't have any punchlist items left!)

harumhelmy:
Thanks Colin! Marked it as ready for now — I figured we can always add more demographic info later.

lib/tasks/create_fake_data.rake
amackcrane: maybe replace '5000' with a variable for to DRY?

PR #1877, opened 2019-12-15
Tell our ruby mongodb connection to not support rewrites

colinxfleming:

Not sure this will work but when I shipped rails 6 to prod it got mad about this:

```
Mongo::Error::OperationFailure (Mongo::Error::OperationFailure: Transaction numbers are only allowed on storage engines that support document-level locking (20) This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string or use the retry_writes: false Ruby client option):
```

This seems to be because mlab doesn't support this; hopefully this is as simple as changing the database connection and not indicative of a bigger, prod-only problem!

* adds a mongoid connection option: `retry_writes: false`

no view changes

* Bumps #1589

### DIFF
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -3,6 +3,7 @@ development:
     default:
       uri: mongodb://<%= ENV['mongohost'] || 'localhost' %>:<%= ENV['mongoport'] || 27017 %>/dcaf_case_management_development
       options:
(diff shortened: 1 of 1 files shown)


colinxfleming:
godmode-ing this since it's prod-only config garbage

PR #1879, opened 2019-12-16
set lockfile to use bundler 2 instead of bundler 1

colinxfleming:

CircleCI doesn't seem to support bundler v1 in their buildpacks anymore. Guess how I found this out.

* ran `gem install bundler && bundle update --bundler` to install new bundler gem and rebundle

no views, no issues


colinxfleming:
yeah, GHActions is regrettably a work in progress due to it being super hard to run headless browser tests in a docker container!

PR #1884, opened 2019-12-21
Configurable BudgetBar Max

### LINKED ISSUE #1883: Create Config for Budget Bar Maximim
Make the budget bar maximum a Configuration option.

This will allow the funds to set their own budgets, instead of having it hardset to a value.

Implement a new option in [the config model](https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/models/config.rb)

* **Anything else?**
Would be great to have before 2020 to support BAF's increased budget :tada:

lomky:
I rule and have completed some work on Case Manager that's ~ready for review!~

Made the max value of the BudgetBar configurable.

Old  (_number for demo purposes only_)
Screenshot from 2019-12-21 18-20-48

New (_number for demo purposes only_)
Screenshot from 2019-12-29 06-45-11

 - [x] Add Config option for budget bar max
 - [x] Budget Bar sizes proportional to that number

* Fixes #1883

### DIFF
--- a/app/helpers/budget_bar_helper.rb
+++ b/app/helpers/budget_bar_helper.rb
@@ -5,7 +5,7 @@ module BudgetBarHelper
     "bg-#{color}"
   end
 
-  def progress_bar_width(value, cash_ceiling = 1_000)
+  def progress_bar_width(value, cash_ceiling = Config.budget_bar_max)
     "width: #{to_pct(value, cash_ceiling)}%"
   end
 
@@ -27,10 +27,11 @@ module BudgetBarHelper
 
   private
 
-  def to_pct(value, cash_ceiling = 1_000)
+  def to_pct(value, cash_ceiling = Config.budget_bar_max)
     return '0' if cash_ceiling == 0
 
     pct = (value.to_f / cash_ceiling.to_f) * 100
     pct.round.to_s
   end
+
 end
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -1,7 +1,7 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
   def date_range(date: Time.zone.now)
-    if (Config.start_day.downcase == :monthly )
+    if (Config.start_day == :monthly )
       month_range(date: date)
     else
       week_range(date: date)
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -10,7 +10,7 @@ class Config
   CONFIG_FIELDS = [
     :insurance, :external_pledge_source, :pledge_limit_help_text,
     :language, :resources_url, :practical_support_guidance_url, :fax_service, :referred_by,
-    :practical_support, :hide_practical_support, :start_of_week
+    :practical_support, :hide_practical_support, :start_of_week, :budget_bar_max
   ].freeze
 
   # Define overrides for particular config fields.
@@ -22,6 +22,7 @@ class Config
                    'Ex: https://drive.google.com/drive/my-practical_support',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
     start_of_week: "How to render your budget bar. Default is weekly starting on Monday. Enter \"Sunday\" for weekly budget starting on Sunday, or \"Monthly\" for a calendar month based budget.",
+    budget_bar_max: "The maximum for the budget bar. Default $1,000",
     hide_practical_support: 'Enter "yes" to hide the Practical Support panel on patient pages. This will not remove any existing data.'
   }.freeze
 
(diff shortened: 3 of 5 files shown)


lomky:
@colinxfleming I added the new config, and called it where we had previously hard coded `1_000`, but it doesn't seem to have a visual effect on the bar display proportions. I'll dig in more later, but if you have a quick-insight what I'm missing, give a shout!

hardcoded the config value to display above:
Screenshot from 2019-12-21 18-21-20
Screenshot from 2019-12-21 18-20-48

app/helpers/budget_bar_helper.rb
```diff
@@ -5,7 +5,7 @@ def progress_bar_color(type)
     "bg-#{color}"
   end
 
-  def progress_bar_width(value, cash_ceiling = 1_000)
+  def progress_bar_width(value, cash_ceiling = Config.budget_bar_max)
```
colinxfleming: hm, I think you'll want to go up a level - take a look at `budget_bar_calculable.rb` which I think is what's invoking these methods. I think that's overruling the default arg. relatedly it's kinda unnecessary that we have a default arg here whatsoever, since this is being called from `BudgetBarCalculable#budget_bar_calculations`, so if we wanted to make these standard regular arguments instead that would probably be wise.

It does not surprise me in the least that I wrote this as some medium-spaghetti code 😬 😅

app/models/config.rb
```diff
@@ -22,6 +22,7 @@ class Config
                    'Ex: https://drive.google.com/drive/my-practical_support',
     fax_service: 'A link to your fax service. ex: https://www.efax.com',
     start_of_week: "How to render your budget bar. Default is weekly starting on Monday. Enter \"Sunday\" for weekly budget starting on Sunday, or \"Monthly\" for a calendar month based budget.",
+    budget_bar_max: "The maximum for the budget bar. Default $1,000",
```
colinxfleming: I think a hint about how to configure this var might be a good idea:

```suggestion
    budget_bar_max: "The maximum for the budget bar. Defaults to 1000 if not set. Enter as a number with no dollar sign or commas.",
```

what do you think?

app/helpers/dashboards_helper.rb
```diff
@@ -1,7 +1,7 @@
 # Convenience methods consumed in the dashboards controller index view
 module DashboardsHelper
   def date_range(date: Time.zone.now)
-    if (Config.start_day.downcase == :monthly )
+    if (Config.start_day == :monthly )
```
colinxfleming: just to check, this is because the downcasing is handled in the `Config.start_day` class method?

lomky: Correct!

test/models/config_test.rb
```diff
@@ -82,16 +82,16 @@ class ConfigTest < ActiveSupport::TestCase
       end
     end
 
-    describe '#start_day' do
-      it 'should return the day of week as a symbol' do
-        assert(Config.start_day.equal? :monday)
+    describe '#budget_bar_max' do
+      it 'should return an integer of 1_000 if unconfigured' do
+        assert(Config.budget_bar_max == 1_000)
```
colinxfleming: some small minitest-isms

```suggestion
        assert_equal 1_000, Config.budget_bar_max
```

test/models/config_test.rb
```diff
@@ -82,16 +82,16 @@ class ConfigTest < ActiveSupport::TestCase
...
 
-      it 'should return another day of the week if configured' do
-        c = Config.find_or_create_by(config_key: 'start_of_week')
-        c.config_value = { options: ["Tuesday"] }
+      it 'should return another amount if configured' do
+        c = Config.find_or_create_by(config_key: 'budget_bar_max')
+        c.config_value = { options: [2000] }
         c.save!
-        assert(Config.start_day.equal? :tuesday)
+        assert(Config.budget_bar_max == 2_000)
```
colinxfleming:
```suggestion
        assert_equal 2_000, Config.budget_bar_max
```

test/models/config_test.rb
```diff
@@ -114,5 +114,18 @@ class ConfigTest < ActiveSupport::TestCase
         assert(Config.hide_practical_support? == false)
       end
     end
+
+    describe '#start_day' do
+      it 'should return the day of week as a symbol' do
+        assert(Config.start_day.equal? :monday)
```
colinxfleming:
```suggestion
        assert_equal :monday, Config.start_day
```

test/models/config_test.rb
```diff
@@ -114,5 +114,18 @@ class ConfigTest < ActiveSupport::TestCase
...
+      end
+
+      it 'should return another day of the week if configured' do
+        c = Config.find_or_create_by(config_key: 'start_of_week')
+        c.config_value = { options: ["Tuesday"] }
+        c.save!
+        assert(Config.start_day.equal? :tuesday)
```
colinxfleming:
```suggestion
        assert_equal :tuesday, Config.start_day
```

PR #1886, opened 2019-12-24
Modernize container links in docker-compose

colinxfleming:

Turns out `links` in docker-compose are on their way out in favor of `network`. Removing explicit links puts everything in the same implied network, so we do that instead.

* deprecate links in docker-compose in favor of network

no views, no issues

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
-version: '3.1'
+version: '3.7'
+
 services:
   webpacker:
     image: colinxfleming/dcaf_case_management:latest
@@ -17,12 +18,13 @@ services:
       - .:/usr/src/app
     ports:
       - "3000:3000"
-    links:
-      - db
     environment:
       mongohost: db
     env_file:
       - .docker/web-variables.env
+    depends_on:
+      - db
+      - webpacker
   db:
     image: mongo
     ports:


docker-compose.yml
```diff
@@ -17,12 +18,13 @@ services:
       - .:/usr/src/app
     ports:
       - "3000:3000"
-    links:
```
DarthHater: `links` and `depends_on` are not 1:1. Links was a shortcut for `network` more or less, and KINDA did `depends_on` to some degree (which doesn't 100% work either, but that's a whole other issue). Generally when you remove `links`, you add `network`, because `network` does what `links` did but it also sets up other good stuff.

Pretty sure things will work as is, but you might take a look at the `network` command, and see if there is any benefit to using it. I've been teaching some friends golang and you can see a very basic use of it here: https://github.com/DarthHater/hello-world/blob/TestBranch/docker-compose.yml#L27

`depends_on` in short will only wait for the container to start, for the record. If you truly want to wait till the DB is up you gotta kinda hand roll a `bash` script or something akin that can ping something, and then return true once it's ready to go.

DarthHater: I shared this mostly because I can't see the `network` stuff, for the record, and your branch has it in the name :)

colinxfleming: the nice thing about docker-compose here is that it will automatically do the network stuff if you don't declare one! So it infers that everything is in a shared network and automagically sets it up as part of `docker-compose (run|up)` which is pretty handy.

colinxfleming:
it's been 11 months so pretty sure we canlive without this

PR #1890, opened 2019-12-29
Assorted bundler 2 dependency bumps

colinxfleming:

We need to bump some gems to make them compatible with bundler 2. This is an omnibus PR to make that happen.

* bump gems (in individual commits)

no views, no issues


colinxfleming:
Although the `render_async` stuff is prod facing, we need it knocked over to get ahead of some rails 6 stuff, so I'm invoking godmode here and merging.

PR #1892, opened 2019-12-31
render_async: move data-turbolinks-track to html options

colinxfleming:

really dunked on myself with this one.

* pass `data-turbolinks-track: reload` in the right spot, so it actually gets passed into `content_for` instead of just disappearing into the void

no views, no issues - this is a hotfix of something I broke in #1890


colinxfleming:
godmode so we can get out the config change in #1884

Issue #1876, opened 2019-12-15
Parallelize test suite


colinxfleming:
Please forgive not following the issue template. These aren't necessary to ship the gem upgrade, but are necessary to declare victory on this upgrade generally.

- [x] Move to rails 6 defaults by setting `config.load_defaults 6.0` in `application.rb`; ditch the two framework defaults migration things we have. (#1893)
- [x] Affirm that we autoload with zeitwerk properly - https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#project-structure the command is `bin/rails zeitwerk:check` (#1894)
- [x] confirm that we're forcing SSL properly in config - https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#force-ssl I think this is true, but just to be on the safe side. (yes, we do this)
- [ ] set up parallel testing - https://blog.bigbinary.com/2019/04/29/rails-6-adds-parallel-testing.html - requires https://github.com/metaskills/minitest-spec-rails/issues/94 to be resolved, which I found out after a day in hell

Separate PRs for each please, since they will affect a lot of stuff!

colinxfleming:
I'm not sure we can parallelize on mongoid - even with the minitest-spec-rails fix attached it's blowing up, unfortunately. I'm going to backlog this and change the title from `Rails 6 Followups` to `parallelize test suite`.

PR #1893, opened 2020-01-04
Move to rails 6.0 defaults

colinxfleming:

Moves us to Rails 6 framework defaults!

Had to rename URIService to `UriService` along the way, unfortunately, because activesupport acronyms couldn't find it.

* use rails 6 framework defaults
* small changes to get builds to pass as part of that 

no view changes

* Bumps #1876


colinxfleming:
gh build is not super great about managing ruby versions, which is a real bummer, because the builds are so nice! but yeah, as a consequence we're not really able to put it to work for us. I have an open PR to take it out for the time being.

PR #1894, opened 2020-01-05
squelch zeitwerk warning about undefined module

colinxfleming:

Paired with #1893 this eliminates an outstanding zeitwerk warning about how it's expecting a module to be defined.

Also add `rails zeitwerk:check` to CI, so we don't accidentally intro new problems, kinda similar to how we're enforcing a bunch of i18n stuff.

Note that this leaves an existing warning with ActionMailer previews, which I've opened an issue about at https://github.com/rails/rails/issues/38156 

* fix existing warning
* add zeitwerk:check to CI

no view changes

bumps #1876

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,7 @@ jobs:
       - run: brakeman --exit-on-warn .
       - run: bundle exec ruby-audit check
       - run: bundle-audit update; bundle-audit check --ignore CVE-2015-9284
+      - run: rails zeitwerk:check
       # - run: yarn audit
       - store_artifacts:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       - run: brakeman --exit-on-warn .
       - run: bundle exec ruby-audit check
       - run: bundle-audit update; bundle-audit check --ignore CVE-2015-9284
+      - run: rails zeitwerk:check
```
lomky:
```suggestion
      - run: bundle exec rails zeitwerk:check
```

colinxfleming:
merging because this fixes a deployment bug

Issue #1752, opened 2019-08-22
Let's talk about whether installing some limited js framework is a good idea


colinxfleming:
I am admittedly starting from a place where I don't want to do this, but I think that the patient edit view is pretty complicated and I think we might benefit from some one-way state management of the variety that stimulus or react provides. Now that we are a webpack operation I think this would be much easier to maintain, and the fact is that stuff like `patients/update.js.erb` is really, really ugly and unfun to work with and test.

To be clear, I am interested in augmenting our HTML, not switching to a whole-hog single-page app. But I also think that moderating our approach to this some 

Some benefits off the top of my head of going with more JS in the frontend:
* state management gets easier! patients/update.js.erb is doing a LOT of work.
* frankly, more javascript work; JS is a language that most of our new volunteers write and this is an opportunity to do work on that front

Some potential drawbacks:
* app becomes harder to troubleshoot
* javascript ecosystem is scary and this is throwing javascript at a problem
* people are starting to write component libraries for rails, so, like, nonzero chance this isn't even a problem in rails 6.1?

If you have strong opinions about this, I am extremely interested in hearing them!

tingaloo:
Since we have webpacker in place, these guys open up https://github.com/rails/webpacker#integrations.

[React](https://github.com/facebook/react): Safe choice since it is widely adopted all around. (was thinking of other funds adopting the app).

[Stimulus](https://github.com/stimulusjs/stimulus): It looks really stable but neglected (The copyright on the github page has not been updated.)

[vuejs](https://github.com/vuejs/vue): i heard is balanced and has good traction. Can build components and scale.

[angular](https://github.com/angular/angular): Potential overkill, 2 frameworks is a bit much.

[npm Trends
](https://www.npmtrends.com/@angular/core-vs-vue-vs-react-vs-@stimulus/core) to give insight on popularity.

On the context of DCAF itself, html erbs can be swapped out with Stimulus. It feels closer to rails adopting state lifecycles of modern frameworks.

React or vuejs are great for using the modern component methodology. We would need to gut most of our existing ```js.erb``` for consistency.

I am kind of biased on vuejs because it is a js framework i look forward to learning. I leave these options based on what we have. Webpacker is constantly updating, which follows the same cadence as react and vuejs. Definitely open for conversation on this one.

colinxfleming:
[...]

PR #1895, opened 2020-01-05
manage turbolinks in webpack

colinxfleming:

We're trying to get out of the business of sprockets so we can go full webpack. To that effect, let's use the npm package to manage turbolinks rather than the gem.

* `yarn add turbolinks` plus some config
* remove it from gemfile
* a couple forms are using xhr where they should be using synchronous requests; bump those

no view changes

* Bumps #1752 kinda

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ ruby '2.6.5'
 # Standard rails
 gem 'rails', '~> 6.0.2'
 gem 'puma', '~> 4.3' # roar
-gem 'turbolinks', '~> 5.2.0'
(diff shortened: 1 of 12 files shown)


Gemfile
```diff
@@ -4,7 +4,6 @@ ruby '2.6.5'
 # Standard rails
 gem 'rails', '~> 6.0.2'
 gem 'puma', '~> 4.3' # roar
-gem 'turbolinks', '~> 5.2.0'
```
tingaloo: # gotem

PR #1896, opened 2020-01-08
near year new readme

colinxfleming:

bumping the readme to remove some OBE sections and update the directions. s/o to the person from december's hacknight who found out the hard way our mongo install docs were out of date.


colinxfleming:
godmode because readme

PR #1901, opened 2020-01-09
Dictionary change

### LINKED ISSUE #1836: Add note to data dictionary about why archived patient data is null before 2019-10-21
This is a good ticket if you are learning Github's pull request flow!

## Context
https://codefordc.slack.com/archives/C0E6APB36/p1571703212151600
We had a bug in the archive & export code leading to all patients being marked as having special circumstances (hint: 1/0 are not truthy/falsey, they're both truthy :sob:).

Because of this we bug we had every archived patient flagged as having special circumstances. We decided to mark all those patients as not having, instead, since we have no way of discovering this after archiving.

## Work to do

Add this note to the [Data Dictionary](https://github.com/DCAFEngineering/dcaf_case_management/blob/master/docs/DATA_DICTIONARY.md):

> Archived Patients prior to 2019-10-21 will always have `had_special_circumstances` set to null due to a bug in the archiving processing not preserving the value. Archived patients after that date correctly reflect whether or not they had special circumstances. See [issue #1836](https://github.com/DCAFEngineering/dcaf_case_management/issues/1836) for further context.

villatrue:

added note to data dictionary about why archived patient data is null before 2019-10-21 #1836

* Fixes #1836 
* Bumps #1836


colinxfleming:
welcome to the party @villatrue ! glad to have you!

PR #1910, opened 2020-01-21
Adds random pin to patient model

mdworken:
We want to add this pin first without removing the previous identifier to avoid interrupting CM operations. We will eventually migrate to only using this, and will then remove the previous pin.

To support this, a future PR will add this new pin to search and will surface it on the patient views.

This new pin will be entered into Quickbooks records instead of existing data. This will allow us to automatically fill Daria Fulfillment from QB records. We will continue using the Data Entry team until we are ready for this switch.

In the future we will ALSO remove patient names from pledge forms and send them to clinics separately to improve security practices (discuss further offline).

Organizational ToDos/Implementation to be handled by @harumhelmy and @mdworken offline.

* Adds new random pin to be migrated to instead of phone-based pins


app/models/patient.rb
```diff
@@ -270,6 +273,15 @@ def confirm_appointment_after_initial_call
     end
   end
 
+  def create_random_pin
+    return if random_pin
+    25.times do
+      self.random_pin = rand(100000..999999)
```
mdworken: Gonna pull out magic numbers later.

app/models/patient.rb
```diff
@@ -55,6 +56,7 @@ class Patient
   field :other_phone, type: String
   field :other_contact_relationship, type: String
   field :identifier, type: String
```
colinxfleming: We're basically subbing out what's populating identifier, right? Should we just use that field rather than adding another one?

mdworken:
Closing - paired with @harumhelmy to finish this in a different PR

colinxfleming:
@mdworken can you write up a description etc for this PR when you get a sec? I have the brain of a hamster these days and can barely remember my own phone number, ha.

colinxfleming:
note also that there's discussion in #1912

app/models/patient.rb
colinxfleming: (If I remember right the reason we weren't doing this was that we want to use both and gradually transition.)

app/models/patient.rb
```diff
@@ -55,6 +56,7 @@ class Patient
...
+  field :random_pin, type: String
```
colinxfleming: not sure if I called it out, but the use of string is a very good choice here

app/models/patient.rb
```diff
@@ -270,6 +273,15 @@ def confirm_appointment_after_initial_call
     end
   end
 
+  def create_random_pin
+    return if random_pin
+    25.times do
```
colinxfleming: I think this is still a blocker for me as written. I've softened on the pseudo-RNG here, but I'm worried about these two things:

* `raise` as a part of control flow - this is code that gets executed for every new patient, so we have to be totally sure that if there's an error it's because the CM made a mistake, not because the rails code bit it for an independent reason.
* requiring the object be in a `valid` state to escape from this function. I think that might have some weird consequences, esp early in the callbacks, or creating some kinda weird infinite loop under some circumstances.

I think the common theme here is: is there a way to just Get a number that we know is unique without having to take and cycle? Generally people solve this problem with hashing algorithms, but sha and md5, but those return strings; FarmHash returns numbers but they're huge. People also use autoincrement to solve this problem, but you've expressed concern over that being guessable.

Absent a hashing algorithm we're unlikely to find (since hashing algorithms care about a scale way bigger than we do), I wonder if something like this might work? https://docs.mongodb.com/v3.0/tutorial/create-an-auto-incrementing-field/#optimistic-loop but this is really heavy and mongo specific.

a side question: If we were going to do this in postgres, how would we do it?

mdworken: Idea 1: (recommended)
1) Choose primes p, and q as secrets, where q has 6 digits 
2) Store p and q as ENV variables and compute the caller_pin (no longer random) id as n^p mod q

Advantages: 
1) this is a bijection of the integers 1...(q-1) onto itself, so it beats the problem of sequentiality.
2) This would mean we don't need to store another DB field at all, and therefore also wouldn't have to worry about existing records.

Disadvantages:
1) Math

-----
Idea 2 (meh)
One potential solution might be something like this:
1) pre-generate a list of random numbers and save them in an array in the DB
2) whenever we need a new random number, we use the first element in that array and delete it.

The good news: can just do something like `100_000..999_999.shuffle!` to generate them, and they're guaranteed unique since they're just a permutation.

The bad news: 
1) it's a lot more work
2) if a bug, server error, whatever, caused us to reuse a number by accident, it would have drastic business consequences, so we'd probably want to do a uniqueness validation anyways, in which case we'd need a way to "try again" if this was violated, and then we're at the same problem we started with.

colinxfleming:
I'm cycling off so won't be able to think about this much for the rest of the year, but I'm going to leave it open. If we do take action on it, please note that before we merge anything we still need:
* a writeup about what, exactly, we're trying to accomplish here in an associated github issue
* handle dupe ids in some way that won't raise an error if things choke (see comments around https://github.com/DCAFEngineering/dcaf_case_management/pull/1910#discussion_r386782163 )

If we don't think we're going to take action on this anytime soon but we still want to do it, let's close this and document these discussions in an issue so we don't lose them -- I think the premise of this PR (that the line + last 5 digits of phone IDs we make suck) is dead on, but this might be a problem that only really affects DCAF, so idk.

colinxfleming:
@mdworken still live? I couldn't find an issue for this and it's been hanging out in this PR purgatory state for awhile - can we close?

app/models/patient.rb
mdworken: Resolving conversation so we can look at this with fresh eyes

PR #1912, opened 2020-01-26
Implements random PIN for patient records 

### LINKED ISSUE #1900: Refix patient ids to be more random

our current ids follow the old pattern that DCAF had, which generates ids that aren't REALLY unique. @verbingthenoun reports seeing duplicates in the course of looking up people for data entry, which is not an ideal behavior (and a flaw with the method of generating IDs that we adopted from DCAF pre-DARIA). We should explore generating an actual unique identifier, or pressing a different identifier into service.

Actual unique ids for data entry people to search by when doing fulfillments.

I think the only places we use ID are the search and pledge generation forms. Clinics also use them in giving funds remits.

I think to decide what we need, we need answers to the following:
* do non-dcaf funds use generated ids for anything besides pledge generation? (probably not, right?)
* is it safe to surface these? (I think so but not sure)
* what can we give to clinics? Do they need something that looks like our current pattern but is totally scrambled, or can we do like a full on hash (e.g. `af5c32b`) or something?

* **Anything else?**
stems from a conversation with @verbingthenoun

harumhelmy:

Screen Shot 2020-01-26 at 14 50 20

* Fixes #1900

### DIFF
--- a/app/helpers/footer_helper.rb
+++ b/app/helpers/footer_helper.rb
@@ -2,6 +2,7 @@
 module FooterHelper
   def fax_service
     fax = Config.find_or_create_by(config_key: 'fax_service').options.try :first
-    link_to fax, fax, target: '_blank', rel: 'noopener nofollow' if fax
+    fax_uri = URIService.new(fax)
+    link_to fax_uri.uri, fax_uri.secure_scheme_uri!.to_s, target: '_blank', rel: 'noopener nofollow' if fax_uri.uri
   end
 end
--- a/app/services/uri_service.rb
+++ b/app/services/uri_service.rb
@@ -2,6 +2,8 @@ class URIService
   attr_accessor :uri
 
   def initialize(uri)
+    return nil if uri.nil?
+
     begin 
       @uri = URI.parse(URI.encode(uri))
     rescue URI::InvalidURIError
@@ -9,7 +11,16 @@ class URIService
     end
   end
 
-  def secure_scheme!
+  def secure_scheme_uri!
+    fix_leading_slashes
     @uri.scheme = "https"
+    @uri
+  end
+
+  def fix_leading_slashes
+    if @uri.scheme.nil? && @uri.host.nil? && @uri.path.present?
+      @uri.host = @uri.path
+      @uri.path = ""
+    end
   end
 end
--- a/test/helpers/footer_helper_test.rb
+++ b/test/helpers/footer_helper_test.rb
@@ -6,8 +6,8 @@ class FooterHelperTest < ActionView::TestCase
   describe 'fax service' do
     it 'should return the config' do
       fax_link_html = fax_service
-      assert_match 'href="http://www.yolofax.com"', fax_link_html
-      assert_match '>http://www.yolofax.com</a>', fax_link_html
+      assert_match 'href="https://www.yolofax.com"', fax_link_html
+      assert_match '>https://www.yolofax.com</a>', fax_link_html
     end
   end
 end
--- a/test/services/uri_service_test.rb
+++ b/test/services/uri_service_test.rb
@@ -12,11 +12,24 @@ class URIServiceTest < ActiveSupport::TestCase
       assert_nil uri
     end
 
+    it 'instantiates nil on nil' do
+      uri = URIService.new(nil).uri
+      assert_nil uri
+    end
+
     it 'can force the scheme to use https' do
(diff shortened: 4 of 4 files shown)


mdworken:
Overall plan of attack:

We want to add this pin first without removing the previous identifier to avoid interrupting CM operations. We will eventually migrate to only using this, and will then remove the previous pin.

To support this, a future PR will add this new pin to search and will surface it on the patient views.

This new pin will be entered into Quickbooks records instead of existing data. This will allow us to automatically fill Daria Fulfillment from QB records. We will continue using the Data Entry team until we are ready for this switch.

In the future we will ALSO remove patient names from pledge forms and send them to clinics separately to improve security practices (discuss further offline).

Organizational ToDos/Implementation to be handled by @harumhelmy and @mdworken offline.

mdworken:
Things to do before making live:
- Confirm approach with treasurer
- Alert CM team that this ID will start coming through and not to be alarmed

app/models/patient.rb
```diff
@@ -270,6 +273,15 @@ def confirm_appointment_after_initial_call
     end
   end
 
+  def create_random_pin
+    return if random_pin
+    25.times do
+      self.random_pin = rand(100000..999999)
+      return if self.valid?
+    end
+    raise "Had trouble generating a random pin - please contact tech support"
```
colinxfleming: Is there something we can do to be fairly certain that this will never hit a collision? Programmatically generating things, or even serial ids padded with zeroes? (e.g. 0001, 0002, 0003, ....) If we're fine with letters in pins, maybe part of a SHA with a salt?

My concern here is that the failure mode isn't really possible for CMs to resolve on their own, and I want to avoid a system where that's the case. So I'd love to explore another option to generate a random pin here where we're a little more confident that there won't be a collision, even if the possibilities are mega remote.

app/models/patient.rb
```diff
@@ -55,6 +56,7 @@ class Patient
   field :other_phone, type: String
   field :other_contact_relationship, type: String
   field :identifier, type: String
+  field :random_pin, type: Integer
```
colinxfleming: Can you help me understand why the new field rather than pressing identifier into service? Are we going to maintain both fields?

mdworken: Yes - the plan right now is to maintain both fields for several months to give clinics + CMs a chance to adjust and to avoid disruption. If we replaced it, then we'd risk issues with existing pledges

app/models/patient.rb
mdworken: I totally agree with your discomfort, and did some digging before doing this. Believe it or not, this turns out to actually be one of the methods recommended in the mongodb docs.

https://www.mongodb.com/blog/post/generating-globally-unique-identifiers-for-use-with-mongodb

In terms of here, we generally have no more than 5000 patients unarchived at any point in time, and this gives 900,000 possible IDs. The odds of this happening then are (5/900)^25 < 3*10^-56. To put that in comparison, the odds of this are about 1 billion times better than those of winning 6 mega millions jackpots back to back (ballpark).

app/services/pledge_form_generator.rb
```diff
@@ -33,6 +33,10 @@ def patient_identifier
     patient.identifier
   end
 
+  def patient_random_pin
+    "[#{patient.random_pin}]"
```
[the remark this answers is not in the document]
mdworken: Sounds good to me! Did it like this to avoid having to conditionally render the [ ] => maybe I'll move this to being done at the point of pledge generation then so it shows for all new ones after merging!

app/models/patient.rb
colinxfleming: I recognize that the odds are low to zero, but I would want to take the raise out of the control flow here regardless - if we had to move everyone to a single tenant instance, for example, this could potentially be a real pain tounwind.

What's the argument against a scoped serial id (DCAF-000001, DCAF-000002, BAF-000001) in favor of a random scrambler?

mdworken: Serial IDs I think would be too visually confusing for CMs/Clinics since they'll be close to each other. They also run the human risk of being used for authentication, in which case a "guessable" ID would be super dangerous.

I do like the scoping with org name though - I'll work on that right now.

colinxfleming: Can you expound on what 'used for authentication' here means? I'm a little on the sleep-deprived side so having a tough time the scenario - can you walk me thru it in here?

PR #1917, opened 2020-02-06
fixed budget bar

### LINKED ISSUE #1911: Budget bar not catching pledge sents if they're in a different week from soft pledge

BRAF reported an unexpected behavior on the budget bar:

For a patient with a pledge amount filled in on one week and with a sent pledge the following week, that pledge doesn't show up in the budget bar. That's because the `Patient#pledged_status_summary` method only cares about `fund_pledged_at`, which we set when the `fund_pledge` field changes. We should probably ALSO make it care about `pledge_sent_at`, which is our timestamp for a pledge actually going out the door.

A budget bar that accurately reflects money moving. I think.

change https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/models/patient.rb#L150 to be either `fund_pledged_at` or `pledge_sent_at`.

* **Anything else?**

harumhelmy:

Heyoo, I updated the `self.pledged_status_summary` method  in `patient.rb` by adding an or query so that the budget bar shows hard pledges sent since the start of the week, as well as soft pledges made that week. Working on writing the test for it! 

* Fixes #1911

### DIFF
--- a/test/models/patient_test.rb
+++ b/test/models/patient_test.rb
@@ -184,7 +184,7 @@ class PatientTest < ActiveSupport::TestCase
                     config_value: {options: ['Monday']}
       Timecop.freeze("January 20, 1973") { @patient.update! fund_pledge: 250 }                      
       Timecop.freeze("January 23, 1973") { @patient.update! pledge_sent: true, 
-                                                            appointment_date: @patient.initial_call_date + 1, 
+                                                            appointment_date: @patient.initial_call_date + 1.day, 
                                                             clinic: create(:clinic) }
 
       Timecop.freeze("January 25, 1973") do
--- a/test/models/patient_test.rb
+++ b/test/models/patient_test.rb
@@ -176,6 +176,24 @@ class PatientTest < ActiveSupport::TestCase
         assert_equal 0,summary[:pledged].count
       end
     end
+
+    it "should include pledges sent on the current week, even when the soft pledged was entered in a previous week" do
+      # The week starts on Monday Jan 22, there's a soft pledge entered on Jan 20, the week before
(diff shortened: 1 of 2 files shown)


harumhelmy:
Just committed the test! Wahoo

test/models/patient_test.rb
```diff
@@ -176,6 +176,24 @@ class PatientTest < ActiveSupport::TestCase
         assert_equal 0,summary[:pledged].count
       end
     end
+
+    it "should include pledges sent on the current week, even when the soft pledged was entered in a previous week" do
+      # The week starts on Monday Jan 22, there's a soft pledge entered on Jan 20, the week before
```
colinxfleming: enormously helpful comment!

test/models/patient_test.rb
```diff
@@ -176,6 +176,24 @@ class PatientTest < ActiveSupport::TestCase
...
+      # Update pledge_sent to true on Jan 23, the current week, so it should show up on the week's pledge_status_summary as sent
+      Config.create config_key: :start_of_week,
+                    config_value: {options: ['Monday']}
+      Timecop.freeze("January 20, 1973") { @patient.update! fund_pledge: 250 }                      
+      Timecop.freeze("January 23, 1973") { @patient.update! pledge_sent: true, 
+                                                            appointment_date: @patient.initial_call_date + 1,
```
colinxfleming: rails has a kind of cool thing for doing date math that we might be able to use here - I don't remember if it will work exactly like this, but worth a shot - try it out, and reject the suggestion if it does?

```suggestion
                                                            appointment_date: @patient.initial_call_date + 1.day, 
```

harumhelmy:
Thank you, @colinxfleming! I don't have merging privileges, though, so can't merge it :eyes:

colinxfleming:
nicely done!

PR #1918, opened 2020-02-08
comment out workflow since circleci is working again

colinxfleming:

Completely out of nowhere, CircleCI builds on branches started working again, which makes the importance of a working github build much, much smaller.

I would like to eventually do github builds if they're fast enough because it's a good workout for our docker build, but that's not urgent right now.

* comment out github workflow

no issue


colinxfleming:
godmode since this is entirely build-facing

PR #1919, opened 2020-02-08
add information about roles to user edit/creation pages

### LINKED ISSUE #1868: Surface what roles do in app

@NerdyGirl537 quite accurately pointed out that roles are really confusing - it's not clear to anyone in the UI who has access to what, or what permissions they're granted. We should surface some information about what these roles actually do in the user management tools!

Not requiring specialized knowledge to operate a DARIA instance safely!

Let's do this in two ways:

* define role information on some constants in the user model
* on the user create page, let's add information about each of the roles below our current 'New users automatically...' helper text.
* on a user edit page, let's add the explanation text for each of the roles under the role buttons. (the user edit page is kinda cramped so feel free to add some whitespace if you like!)

* **Anything else?**

Here's some text you can use:

* CM: 'The Case Manager role is for people doing patient intake and making calls. This gives access to the main call list dashboard, the individual patient record page (minus the ability to edit fulfillment info), and the data entry page.'
* Data volunteer: 'The Data Volunteer role is for people who require access to data or accounting tools. This gives access to what CMs have, plus the following: accounting tools for pledge reconcilling, reporting tools with summary data on patient intake, the fulfillment tab on the individual patient record page, and the ability to download an anonymized spreadsheet of data for analysis.'
* Admin: 'The admin role is for people who are managing their fund's DARIA instance or case manager corps. This gives access to what CMs and Data Volunteers have, plus the following: user management tools, clinic management tools, DARIA instance configuration tools, and the ability to delete patients from the individual patient record page (such as in the case of a duplicate entry).'

Please keep these assigned to different variables in case we want to use them in different places later.

evornov:

I added the suggested text about what each role does to the user edit and create pages. Corresponding text is missing for the Spanish localization file, though. 

edit
create

* Fixes #1868

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -21,6 +21,10 @@
       <%= role_toggle_button @user, 'data_volunteer' %>
       <%= role_toggle_button @user, 'cm' %>
     </div>
+    <br /><br />
+    <% User.role.values.each do |r| %>
+      <p><%= t("user.roles.#{r}") %></p>
+    <% end %>
   <% end %>
 
   <div id='user-details-update' class="top-spacer">
--- a/app/views/users/new.html.erb
+++ b/app/views/users/new.html.erb
@@ -7,5 +7,8 @@
     <% end %>
     <br>
     <p><%= I18n.t('user.new.edit_role_after_creating') %></p>
+    <% User.role.values.each do |r| %>
+      <p><%= t("user.roles.#{r}") %></p>
+    <% end %>
     <%= link_to t('common.back'), :back %>
 </div>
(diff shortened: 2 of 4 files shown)


app/views/users/edit.html.erb
```diff
@@ -21,6 +21,10 @@
       <%= role_toggle_button @user, 'data_volunteer' %>
       <%= role_toggle_button @user, 'cm' %>
     </div>
+    <br /><br />
+    <% User.role.values.each do |r| %>
```
colinxfleming: Excellent use of the class constant enum here - really like how clean this is.

evornov:
Google translating done!

colinxfleming:
thank you so much @evornov !

PR #1923, opened 2020-02-14
Add some yarn scanning into the mix

DarthHater:

This adds `auditjs` to DCAF, and starts to audit the javascript we know and love

* Adds `auditjs` as a dev dependency
* Adds a script to `package.json` to run a scan and generate a xml Junit style report
* Runs `auditjs` in CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,8 @@ jobs:
       - run: yarn run scan
       - store_artifacts:
           path: /home/circleci/abortioneering/tmp/capybara/
+      - store_artifacts:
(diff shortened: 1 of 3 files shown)


colinxfleming:
I like this a lot! Looking forward to having this green.

colinxfleming:
@DarthHater  want me to finish this off?

colinxfleming:
This is getting a smidge out of sync so I'm going to close it for now, as it's probably at the point now where we'd have to reroll it to get it squared away. @DarthHater happy to adopt this in CI if we get the build passing on a similar PR.

PR #1932, opened 2020-02-24
update nav_bar and corresponding tests

### LINKED ISSUE #1924: Remove links to the reports endpoint from the navbar

Our reports endpoint has never really worked right and it's starting to become a thing that I have to explain pretty often when we do demos, so let's take the link down until we actually get it set up. The endpoints can stay up, but the links to it should go for now, until we have something useful at that endpoint.

People not thinking we have busted up software when we're doing demos!

The only link I could find with `git grep` was in the top dropdown, so let's remove it from there. (You will also want to comment out these lines from the `navbar_links_test`; the lines in `reports_controller_test` can stick around.

```
$ git grep "reports_path" .
app/views/layouts/_navigation_links.html.erb:      <%= link_to t('navigation.admin_tools.reporting'), reports_path, class: 'dropdown-item' %>
test/controllers/reports_controller_test.rb:        get reports_path
test/controllers/reports_controller_test.rb:          get reports_path
test/controllers/reports_controller_test.rb:          get reports_path
test/system/navbar_links_test.rb:        assert has_link? t('navigation.admin_tools.reporting'), href: reports_path
test/system/navbar_links_test.rb:        assert has_link? t('navigation.admin_tools.reporting'), href: reports_path
```

* **Anything else?**

n

kathrynyetter:

I removed the link to the reports from the admin navbar, and the corresponding tests. I kept the reports controller and tests.

image

* Fixes #1924

### DIFF
--- a/app/views/layouts/_navigation_links.html.erb
+++ b/app/views/layouts/_navigation_links.html.erb
@@ -20,7 +20,6 @@
         <%= link_to t('navigation.admin_tools.config_management'), configs_path, class: 'dropdown-item' %>
       <% end %>
       <%= link_to t('navigation.admin_tools.accounting'), accountants_path, class: 'dropdown-item' %>
-      <%= link_to t('navigation.admin_tools.reporting'), reports_path, class: 'dropdown-item' %>
       <%= link_to t('navigation.admin_tools.export'), patients_path(format: :csv), class: 'dropdown-item' %>
     </div>
   </div>
--- a/test/system/navbar_links_test.rb
+++ b/test/system/navbar_links_test.rb
@@ -33,10 +33,6 @@ class NavbarLinksTest < ApplicationSystemTestCase
         assert has_link? t('navigation.admin_tools.accounting'), href: accountants_path
       end
 
-      it 'should display the Reporting link' do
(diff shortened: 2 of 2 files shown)


colinxfleming:
looking good @kathrynyetter !

I had `rails test` complain about something, but I'm going to resolve since it's quick and kind of specific:

```
F

Failure:
I18nTest#test_no_missing_keys [/Users/colin/repos/dcaf_case_management/test/i18n_test.rb:13]:
Missing 1 i18n keys, run `i18n-tasks missing' to show them.
Expected es
  navigation
    admin_tools
      reporting: Reporting {:path=>"config/locales/es.yml", :locale=>"es", :type=>:missing_diff, :missing_diff_locale=>"en"} to be empty.


rails test test/i18n_test.rb:12

F

Failure:
I18nTest#test_no_unused_keys [/Users/colin/repos/dcaf_case_management/test/i18n_test.rb:18]:
1 unused i18n keys, run `i18n-tasks unused' to show them.
Expected en
  navigation
    admin_tools
      reporting: Reporting {:path=>"config/locales/en.yml", :locale=>"en"} to be empty.


rails test test/i18n_test.rb:17
```

this is nbd and fixable by removing the (now unused) i18n key that corresponds to reporting. I'll rack it really quick.

PR #1933, opened 2020-02-24
remove fade

### LINKED ISSUE #1913: Remove hover fade on flash messages

When you save an attribute on the patient edit page, you get a flash message. Hovering over it makes it more transparent, but this is kinda awkward. I'd like to remove this opacity change on hover.

CMs are reporting that it's fading in and out weird; this is simple enough that we shouldn't bother diagnosing the issue and should just not change it on hover.

We're probably setting an opacity attribute on hover somewhere in `app/assets/css/application.css.scss` is my guess. We should eliminate that and then check that this fixes it.

* **Anything else?**
cc @verbingthenoun

kathrynyetter:

removed the fade on hover function

* deleted the fade section of the flash section of the stylesheet

I tried to screenshot, but it doesn't last long enough! Did confirm that it works locally though.

* Fixes #1913 
* Bumps #Y

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -142,9 +142,6 @@ body {
   width: 100%;
   z-index: 999;
   transition-duration: .5s;
-  &:hover {
-    opacity: .5;
-  }
 }
 
 // General styling artifacts


colinxfleming:
CI failure is fake; merging.

PR #1935, opened 2020-02-26
try uploading ci test results to see if that prevents errors

colinxfleming:
can't prove it won't work, I guess.


colinxfleming:
let's see if this helps.

PR #1946, opened 2020-03-20
upgrade rails to 6.0.2.2 + other security patches

colinxfleming:

Rails released a security patch. Should be pretty straightforward and safe, let's bump it.

This also bumps some other outstanding security concerns, so we can get the build green again.

* `bundle update rails`
* set serialize-javascript to a safe version
* upgrade sprockets (gem bump and config file)

no view changes

no issues


colinxfleming:
changes are def intentional - it's a necessary thing to get sprockets 4 going - https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs

PR #1949, opened 2020-03-25
adds exportable fields, and updates docs

### LINKED ISSUE #1920: Include patient#pledge_sent_at and patient#fund_pledged_at in exports

A fund is trying to recreate the budget bar from the CSV export, but can't because the export doesn't have the two key fields used to figure out whether something belongs in the budget bar: `pledge_sent_at` and `fund_pledged_at`. We should add these to the data that gets exported via CSV.

I think this should cover it:
* add the fields in `app/models/concerns/exportable`
* ensure the entries in `DATA_DICTIONARY.md` about these fields are accurate

* **Anything else?**

Since these are fund-level records and not really patient level records, I think they're safe to export, but please flag if there's a circumstance I'm not thinking of.

To pull an export, log in, click the admin dropdown, and select 'export to csv'.

kathrynyetter:

I added pledge_sent_at and fund_pledged_at fields to the CSV. and updated the docs appropriately (I think)

* Fixes #1920
* Bumps #Y


docs/DATA_DICTIONARY.md
```diff
@@ -42,6 +42,7 @@ Field | Description | Exportable? | Reported by | Input Type
 **Initial Call Date** | The date a patient's file was first created in DARIA. | Exportable | Autogenerated | Autogenerated
 **Urgent Flag** | This boolean flag is used to indicate an urgent case. This is used primarily for communication between case managers and has no impact on funding. | Exportable | CM Reported | Checkbox
 **When was pledge generated** | Date autogenerated from the DARIA system when a pledge is generated. | Exportable | Autogenerated | Autogenerated
+**When was pledge sent** | Date autogenerated from the DARIA system when a pledge is sent. | Exportable | Autogenerated | Autogenerated
```
colinxfleming: nailed it on the meaning of `pledge_sent_at` here!

let's add `fund pledged at` too, and just the titles so they match the strings in exportable:

```suggestion
**Pledge sent at** | Date autogenerated from the DARIA system when a pledge is sent. | Exportable | Autogenerated | Autogenerated
**Fund pledged at** | Date autogenerated from the DARIA system when a pledge amount changes. | Exportable | Autogenerated | Autogenerated
```

colinxfleming:
thank you @kathrynyetter !

PR #1950, opened 2020-03-25
export last_menstrual_period_at_appt_weeks

### LINKED ISSUE #1922: Add LMP on appointment date to CSV export

The app does a lot of LMP calculation in the patient view, but that doesn't show up in the CSV export, which means that if you want that information as an analyst you would have to individually grab things. That is unpleasant, so we should add that to the CSV export.

Less over the top excel magic in exports.

Add this to `app/models/concerns/exportable.rb`. The model field in question is `Patient#last_menstrual_period_at_appt_weeks` if memory serves.

* **Anything else?**

stems from a conversation with a fund

This seems like a relatively safe thing to make exportable, but please flag if there's a data security concern.

kathrynyetter:

Added Patient#last_menstrual_period_at_appt_weeks to be exported. Checked the data doc for the field

* Fixes #1922
* Bumps #Y


colinxfleming:
hey @kathrynyetter , we have a test failure here through no fault of yours -- 

```
docker-compose run --rm web rails test

....

Error:
PatientsControllerTest::index method#test_0007_should not contain personally-identifying information:
NoMethodError: undefined method `last_menstrual_period_at_appt_weeks' for #<ArchivedPatient:0x00007f9ebaec29e8>
Did you mean?  last_menstrual_period_weeks
               last_menstrual_period_weeks?
               last_menstrual_period_weeks=
               last_menstrual_period_days_was
    app/models/concerns/exportable.rb:152:in `public_send'
    app/models/concerns/exportable.rb:152:in `get_field_value_for_serialization'
    app/models/concerns/exportable.rb:193:in `block (3 levels) in to_csv'
    app/models/concerns/exportable.rb:193:in `map'
    app/models/concerns/exportable.rb:193:in `block (2 levels) in to_csv'
    app/models/concerns/exportable.rb:192:in `block in to_csv'
    app/controllers/patients_controller.rb:152:in `each'
    app/controllers/patients_controller.rb:152:in `each'
    app/controllers/patients_controller.rb:152:in `block in render_csv'
    test/controllers/patients_controller_test.rb:73:in `block (2 levels) in <class:PatientsControllerTest>'


rails test test/controllers/patients_controller_test.rb:71
```

From the message there, it looks like `last_menstrual_period_at_appt_weeks` is not a method that ArchivedPatient has available. Fortunately, we can fix that up by adding `include LastMenstrualPeriodMeasureable` in `app/models/archived_patient.rb`, easy peasy - go ahead and stick it under `# Concerns`. Then we should be good to go!

In other news, gotta spend some time trying to figure out what's up with circle again, bleh.

PR #1951, opened 2020-03-26
add fund_pledged_at and textable to archived patients

colinxfleming:

These fields aren't populating - I think it's because `archived patient` is missing a few fields. This adds two of them that I noticed, textable and fund_pledged_at.

This is blocking a deploy so I'm going to smash it.

* add some fields that patient has to archivedpatient

no view changes
no issues


colinxfleming:
smashing this in to get the build green

Issue #1854, opened 2019-11-11
Webpackerize and refactor our SCSS


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

We do a lot of work making sure this codebase is easy to dig into; a place where we have historically had a hard time with that is our stylesheets. At this point, the codebase has a couple years of hacked together stylesheets, and we are at a point with webpacker where we can thoughtfully recompose them into something much more modern.

To that end, when porting over our SCSS into webpack, we should do a full on extended refix.

* **What feature or behavior is this required for?**

Having a codebase that people can contribute to!

* **How could we solve this issue? (Not knowing is okay!)**

We'll be breaking out smaller tickets from this primary ticket. This will entail doing some refactoring, moving bootstrap into webpacker, and eliminating stuff we're just no longer using.

* **Anything else?**
should follow after #1816  is resolved
also see https://github.com/DCAFEngineering/dcaf_case_management/pull/1748#discussion_r319740636

colinxfleming:
this is solved as part of our nuking webpack

PR #1956, opened 2020-03-29
Remove unused CSS junk

colinxfleming:

In prep for moving SCSS to webpacker, clean up what we already have; delete some files with no entries and remove some style definitions that don't appear in our codebase any longer.

I suspect the search bar was just overtaken by a different style, and the navbar stuff is overtaken by bootstrap 4.

* delete empty sprockets scss files
* remove some styles that aren't used

no view changes, but here's the search bar on both a local and sandbox:

local:
image

sandbox:
image

* Bumps #1854


colinxfleming:
approved in https://codefordc.slack.com/archives/C0E6APB36/p1588829286066100 by @lomky  (thank you!)

PR #1962, opened 2020-04-19
Upgrade Ruby to 2.7.1

colinxfleming:

It's about that time. Upgrade ruby to 2.7.1 to squish a vulnerability and also to stay sharp.

* ruby 2.6.5 -> 2.7.1

no view changes

no issues


lomky:
They deprecated and removed `beginning_of_week` from `Time`?! :scream_cat: 

[Looks like we need to substitue out our `Time` for a `Date` or `DateAndTime` call.](https://apidock.com/rails/ActiveSupport/CoreExtensions/Date/Calculations/beginning_of_week)

colinxfleming:
I think this is going to wind up being pretty involved since ruby 2.7 puts so much in motion. CLosing for now.

PR #1967, opened 2020-05-06
bump jquery-rails to squish a vulnerability

colinxfleming:

jquery-rails has an id'd CVE that we need to deal with to get the build green again.

* pin `jquery-rails`
* `bundle i`

none


colinxfleming:
invoking godmode for this one

PR #1969, opened 2020-05-07
add sentry_dsn env var to app.json

colinxfleming:

We're now happy sentry customers, so rig this up on app init.

* add `SENTRY_DSN` key to app.json

no view changes
no issues


colinxfleming:
godmode since it doesn't touch app code

PR #1972, opened 2020-05-09
Chunks the csv export to hopefully fix our timeout issues

mdworken:
This is what @DarthHater and I settled on as a hail-mary approach. 

@colinxfleming @lomky any chance we could merge this in and get a deploy to see if it unblocks the export problems? Would definitely like to knock that out before the board meeting if at all possible

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -158,6 +158,7 @@ class PatientsController < ApplicationController
     headers["Content-disposition"] = "attachment; filename=\"#{filename}\""
     headers['X-Accel-Buffering'] = 'no'
     headers["Cache-Control"] = "no-cache"
+    headers["Transfer-Encoding"] = "chunked"
     headers.delete("Content-Length")
   end
 end


colinxfleming:
I'm not sure this will do what we're hoping it will, based on my pulling it down and testing it. I've tried to provide some context about why that is.

Here's how this functioned previously (until at least commit 045b15a2698829781697a2662075d9536e9cfa95 ) on chrome. Note that the download basically starts instantly, it doesn't know the filesize in advance (so no % done), and just keeps dumping data in there. This is the behavior we're trying to get back to that changed when we upgraded sprockets/rails. (Sorry for the large file upload.)

May-09-2020 18-29-41

The current behavior (which persists for me with this change) is more like 'run the export, and then when the enumerator finishes, mark the file complete'. This was true for me on firefox and chrome both, so I'm not sure it's going to work on heroku.

The reason that this is a problem is because heroku's ELB or whatever hangs up on requests if it doesn't get a tap every 30 seconds or so. That happens when we're doing it the old way (every iteration counts as a tap tokeep the request alive) and not in the current thing (we do the request, and the iterations don't tap back until all of them are done).

As a next step here, I would suggest comparing the headers between current master and 045b15a2698829781697a2662075d9536e9cfa95 , both request and response, and see if there are differences. My hunch is that there's something we can dig into around there.

colinxfleming:
Closing in favor of #1973

PR #1980, opened 2020-05-18
Swap accountants search to use activerecord calls because they handle nulls more gracefully

### LINKED ISSUE #1970: AccountantsController#search bug when pledge_sent_at is nil

We got an error report in Sentry about a minor bug in the Accountants controller search. The nut of it is: `comparison of NilClass with ActiveSupport::TimeWithZone failed` 

The failure was around here: 

https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/controllers/accountants_controller.rb#L18

Accountant search working reliably

I think we can fix this by doing some activerecord chaining. Let's see if we can do something like `Patient.search(...).where(...).order_by(... desc)` rather than using `select` and `sort_by`, which load objects into memory.

* **Anything else?**

Test would be helpful!

colinxfleming:

We got a sentry error from the accountants panel - there's an edge case where data can get into such a state where `pledge_sent` is true but there's no `pledge_sent_at` value.

The accountants controller fails in that case because the `sort_by` is trying to compare nils with a real timezone. Mongoid does that a lot better, so we move it to an activerecord call instead.

* handle cases where pledge_sent_at is null for some reason
* add a test

no view change

* Fixes #1970

### DIFF
--- a/app/controllers/accountants_controller.rb
+++ b/app/controllers/accountants_controller.rb
@@ -10,8 +10,6 @@ class AccountantsController < ApplicationController
   end
 
   def search
-    # We're doing it janky because we implemented search to return an array,
-    # not an activerecord object. some room for improvement.
     @results = if params[:search].present?
                  Patient.where(pledge_sent: true)
                         .order_by(pledge_sent_at: :desc)
--- a/app/controllers/accountants_controller.rb
+++ b/app/controllers/accountants_controller.rb
@@ -13,9 +13,9 @@ class AccountantsController < ApplicationController
     # We're doing it janky because we implemented search to return an array,
     # not an activerecord object. some room for improvement.
     @results = if params[:search].present?
-                 Patient.search(params[:search])
-                        .select(&:pledge_sent?)
-                        .sort_by(&:pledge_sent_at).reverse
+                 Patient.where(pledge_sent: true)
+                        .order_by(pledge_sent_at: :desc)
+                        .search(params[:search])
                else
                  pledged_patients
                end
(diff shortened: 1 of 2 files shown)


app/controllers/accountants_controller.rb
```diff
@@ -13,9 +13,9 @@ def search
     # We're doing it janky because we implemented search to return an array,
     # not an activerecord object. some room for improvement.
     @results = if params[:search].present?
-                 Patient.search(params[:search])
-                        .select(&:pledge_sent?)
-                        .sort_by(&:pledge_sent_at).reverse
+                 Patient.where(pledge_sent: true)
+                        .order_by(pledge_sent_at: :desc)
+                        .search(params[:search])
                else
```
lomky, line 19: This is basically undoing a change some clever dude made two years ago. He wrote the comment above on lines 13-14 suggesting we couldn't return active record objects. 

https://github.com/DCAFEngineering/dcaf_case_management/commit/0714914b6dc94beb2d7ba4c32f02346d3134d171#diff-4f7fab0f140af92109627f65edf210c3R15-R21

That said, it does seem to be functional now for me in dev, with 1 results, 0 results, and multiple results. So if we're happy with returning this object, can we delete the comment?

colinxfleming: definitely!

So my explanation now that I understand the finer points of how activerecord works: When we hit `search` we return a real array because of some concatenation we do, but `search` starts with activerecord, so we can feed activerecord into it, and it'll still do a proper query.

I will delete the comment!

PR #1981, opened 2020-05-18
delete sessions older than 30 days in nightly_cleanup

### LINKED ISSUE #1975: Delete unused session objects from db

For a security boost we use the `mongo_session_store` gem, which persists session data to db. This is highly useful, but we aren't currently deleting sessions, so we should get on that.

The gem documentation is at: https://github.com/mongoid/mongo_session_store

DB cleanliness, freeing up space.

Add a step to the `nightly_cleanup` that queries for sessions older than a month and destroys them.

Something like this, in the file `lib/tasks/nightly_cleanup.rake`:

```
MongoSession.where(:updated_at.lte => 2.days.ago)  # or whatever the model that mongo_session_store produces is called
puts "#{Time.now} - removed old sessions"
```

You can play with the MongoSession query to tune it a little bit to ensure it's working right.

* **Anything else?**

nope

saralaffin:

I added to the nightly_cleanup to delete sessions older than 30 days.

* I've only changed nightly_cleanup.rake

* Fixes #1975 

When testing the rake file, I get the following warning:
image

### DIFF
--- a/lib/tasks/nightly_cleanup.rake
+++ b/lib/tasks/nightly_cleanup.rake
@@ -16,6 +16,9 @@ task nightly_cleanup: :environment do
   ArchivedPatient.archive_eligible_patients!
   puts "#{Time.now} -- archived patients for today"
 
+  MongoidStore::Session.where(:updated_at.lte => 30.days.ago).delete_all  
+  puts "#{Time.now} - removed old sessions"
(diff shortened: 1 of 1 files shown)


colinxfleming:
thank you @saralaffin !

PR #1992, opened 2020-05-27
Set up doc directory for admin stuff; write instructions for new fund contracts

colinxfleming:

Adding here two things:
* stands up an administering directory, so we can keep that out of onboarding docs
* write contract steps down

Will be doing some other doc work in advance of my bouncing, but this gets us started I think. I'm going to merge first and ask questions later.

no view changes
no issues


colinxfleming:
godmode because docs only

PR #2001, opened 2020-06-18
Sqreen upgrades

xmunoz:
Upgrade actionpack and rack, both of which have CVEs that sqreen alerts on.
This requires upgrading Rails a minor version.


xmunoz:
Sorry, I don't have the expertise (or time) to figure out what's going with webpack. Any help would be much appreciated! @kanth @lomky

colinxfleming:
it looks like the circleci failure is coming from `js-routes`, a library that makes js routes available in js (that frankly I would like to take out since we only use it in `calls/create.js.erb`). I think we can get things back to green by bumping the `js-routes` gem - `bundle update js-routes` should do the trick.

PR #2002, opened 2020-06-19
WIP validations

### LINKED ISSUE #1888: Extend Config Model with Validations
Extend the Config model to provide validations for each field.

These validations should be optional, and defined in the code. When the fund admin then enters data into the config field these validations will inform them if the input is invalid.

I.e. `budget_bar_max` will want to enforce an int with no `$` on it. `start_of_week` should be either a day (`Monday`, `Tuesday`, etc) or `Monthly`.

The scope of  this ticket is the new field and it's application to the field, and tests therefore. 

Once implemented, file issues for each implemented Config to create the validations!

lomky:

WIP: Working on the validation for configs!

* (your change here)
* (another change here)
* (etc)

* Fixes #1888 

For reviewer:
* Tag this PR:


lomky:
Closing this dusty PR in favor of Eli's!

PR #2029, opened 2020-08-06
WIP Postgresize

colinxfleming:
# DO NOT MERGE UNDER ANY CIRCUMSTANCES

This PR is the WIP branch I'm using to port things to postgres. As you can see from the file count this is intense, though I think we'll be able to extract pieces of it.

I think roughly the strategy I'm aiming for is:
* maintain mongodb; keep existing collections but port 
* create core tables in pg andport the model methods as is
* add paper_trail as our substitute for the audit history gem
* get tests passing, gently refactoring along the way *(we are here)*
* get seedfile working
* write a migrator to port things from mongo to pg and confirm their veracity
* rebase this
* extract independent changes into separate PRs to get this in easier
* try this out on a separate sandbox / test environment

At time of opening this is working on getting model tests to pass. After that we'll do the rest of the tests.


colinxfleming:
can declare this dead, as we have a diff approach that works

colinxfleming:
closing this as we now have a viable one that isn't just me winging it and seeing how bad tests fail

Issue #1907, opened 2020-01-20
Docker on Windows fails to start application due to CRLF line endings


Jello:
Docker-compose up step fails on Windows due to git autoconversion to CRLF. This can be solved globally with git on Windows with:
`git config --global core.autocrlf false`
Which forces the Unix line endings expected by the Docker instance.

Will create a pull request that creates a .gitattributes file to specify this setting for the repo.

colinxfleming:
resolved via https://github.com/DCAFEngineering/dcaf_case_management/pull/1908

PR #2044, opened 2020-08-21
Remove misbehaving line ending config

lomky:

Specifying line ending `lf` caused unix systems to sporadically throw a fit and declare a bunch
of design docs to be changed and needed to be committed. Looks like we will need to do a bit more work to make Docker for Windows and Unix play nicely together.

* Removes misbehaving line ending config

* Reverts #1907 

For reviewer:
* Tag this PR:


lomky:
Doesn't solve the issue.

PR #2048, opened 2020-08-26
gitattr fix for LF behavior with binary files

elimbaum:

`.gitattributes` is trying to enforce CRLF on binaries, not just text files. There may be a better way of fixing this, but for now, we can simply add the extensions to ignore to the file.

There is still one txt file associated with a font (_design/Source_Sans_Pro/OFL.txt) that contains CRLF and should be updated, but I didn't touch that for now.

* Adds the file extensions that were causing problems (mostly images)

* Fixes #2044 

For reviewer:
* Tag this PR:


.gitattributes
```diff
@@ -1,2 +1,11 @@
 #Set default behavior to use Unix line endings so Docker on Windows doesn't fail
 * text eol=lf
+
+# Don't normalize images and other binary files
+# more info: https://www.git-scm.com/docs/gitattributes
```
lomky, line 5:
```suggestion
# more info: https://www.git-scm.com/docs/gitattributes#_text
```

.gitattributes
```diff
@@ -1,2 +1,11 @@
...
+*.ai -text
+*.png -text
+*.ttf -text
+*.pdf -text
+*.jpg -text
+*.ico -text
```
lomky, line 11: I'm perfectly happy to add .txt to the list here

```suggestion
*.ico -text
*.txt -text
```

lomky:
Thanks for much for this PR @elimbaum - it's been hard to pin this particular issue down and I really appreciate your proactive fix! I have a couple suggestions for you to consider and then we're gtg!

.gitattributes
elimbaum: oh, i was thinking we _shouldn't_ add txt, because we would want those line breaks to be converted. just that one file has to be manually committed (i can do that on a separate commit)

elimbaum: the reason these other files were showing up is because the magic numbers (file headers) for some files include CRLF for legacy reasons - they definitely should not be converted; anything plain text/code should be (is my understanding)

lomky (approved):
I pushed the last tweak we talked about. Thanks so much for running this issue down - it'll save new devs on the project a lot of heartburn! :fire_extinguisher: :shipit:

Issue #2072, opened 2020-11-20
Move to postgres and off of MongoDB


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Move to postgres! (in stages)

MongoDB is less and less supported as time goes on, and means we're dependent on a lot of extra stuff. Moving to pg has its drawbacks but makes most of our problems much more googleable, and the advantages of mongo are starting to slow down since our db schema is pretty consistent. See #1751 for further discussion.

* **What feature or behavior is this required for?**

Long term health of app; not being reliant on people to maintain the core mongoid libraries; squashing one of our few non-rails-standard dependencies.

* **How could we solve this issue? (Not knowing is okay!)**

I tried a spec approach in #2029 and found that this was possible but ugly. Here's my proposed strategy after that:

* set up postgres as an available db; continue running daria thru mongoid
* port the config model as a proof of concept; in the process, do the heavy lifting around configuring gems
* write a rake task to migrate existing configs over
* production usage of pg for config model/controller 
* port other models

* **Anything else?**

discussion in https://github.com/DCAFEngineering/dcaf_case_management/issues/1751

xmunoz:
Remaining work:
- Remove the mongo_id field
-- https://github.com/DCAFEngineering/dcaf_case_management/blob/main/db/schema.rb#L55
- Delete mongo instances in Atlas so that we aren't getting billed for those dbs anymore

colinxfleming:
~ It Is finished ~

PR #2073, opened 2020-11-20
Remove unnecessary files + other non-pg-dependent changes

colinxfleming:

Pulls a few changes out from #2029 that aren't specifically related to actually implementing postgres, but which pad the landing a bit or clean up some scruff.

* remove some scruff files
* some slight changes to tests to increase coverage and prep things for postgresization

no view changes, and nothing public-facing.

* Bumps #2072

For reviewer:
* Tag this PR:


config/initializers/date.rb
```diff
@@ -0,0 +1,6 @@
+# Extends Date class with convenience methods
+class Date
+  def display_date
```
lomky, line 3: How does this interact with the Time definition of the same? I'm assuming it doesn't

https://github.com/DCAFEngineering/dcaf_case_management/blob/e4eeca3aef354644d2d1a8b101aff3e484ac91a8/config/initializers/time.rb#L3-L5

lomky (approved):
:100: :shipit:

PR #2074, opened 2020-11-20
Add pg + convert Config model to postgres

colinxfleming:

TK:
* docker
* ci
* migrator rake task
* explanation
* proveit

* (your change here)
* (another change here)
* (etc)

* Fixes #X
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -153,7 +153,4 @@ class User
   def send_account_created_email
     UserMailer.account_created(id).deliver_now
   end
-
-  def will_save_change_to_email?
-  end
 end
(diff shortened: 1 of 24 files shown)


app/models/user.rb
```diff
@@ -153,4 +153,7 @@ def send_password_change_email
   def send_account_created_email
     UserMailer.account_created(id).deliver_now
   end
+
+  def will_save_change_to_email?
```
colinxfleming, line 157: https://github.com/heartcombo/devise/issues/4542 🤷

colinxfleming:
This is now in operational shape. That said I feel like we should still break this out some as follows:

* postgres setup, paper trail, CI stuff in its own PR; deploy that (which is what I'll turn this PR into)
* do Config model; port; deploy; remove mongo thing (should be its own separate deal)

colinxfleming:
closing in favor of #2075 (which takes out the config model changes and does not have theugly history)

PR #2078, opened 2020-11-24
Voicemail options

### LINKED ISSUE #2055: Add Config to Extend "Voicemail preference" options

Allow funds to extend Voicemail preference with per-fund options.

Chicago supports leaving messages under a codename, but right now stores that detail under notes. Lets allow them and other funds to add their preferences here.

Mimic how the Config for "Language options" currently works - our current three should stay hardcoded (a la English) and funds can optionally add more.

* **Anything else?**

elimbaum:

Added the ability to configure voicemail options beyond the defaults of yes, no, unspecified.

* Adds an option to the configuration screen, similar to other user-configurable options

This does _not_ handle the case where changes (edits, deletions) are made to the configuration, but patient information is not updated. Currently, only the insurance field has handling for this; I can address it with a more general fix if necessary, so that all user-configurable fields will exhibit the same behavior.

* Fixes #2055 

For reviewer:
* Tag this PR:


test/helpers/calls_helper_test.rb
```diff
@@ -1,32 +1,52 @@
...
+      @patient.voicemail_preference = 'Text Message Only'
+      display_content = display_voicemail_link_with_warning(@patient)
+      assert_match(/Text Message Only/, display_content)
+      assert_match(/I left a voicemail for the patient/, display_content)
+    end
+
+    # this scenario will arise if an option is removed from the config
+    # we want to keep the deleted config value with the patient (instead of
+    # defaulting to a remaining config), and allow them to be updated as needed.
+    it 'invalid voicemail preference is still displayed' do
```
colinxfleming, line 44: 👍

colinxfleming (commented):
nicely done! This all scans as right on to me - thanks in particular for your attention to tests etc. I'm going to pull it down and play around with it a bit before hitting approve but from eyeballing the code I don't anticipate any problems. Thanks a ton for your hard work on this.

app/helpers/patients_helper.rb
```diff
@@ -28,6 +28,17 @@ def language_options
     standard_options + Config.find_or_create_by(config_key: 'language').options
   end
 
+  # TODO: do we want to do something similar to insurance here, to add new
+  # options to list if not already?
+  def voicemail_options
```
colinxfleming, line 33: @elimbaum , how much of a hassle would it be to add the something-similar-to-insurance to this branch? If we don't work this out then the behavior on delete becomes 'the next time the patient's record gets saved, VM preferences becomes No ID VM' which, given the sensitivity of this field for callers, exposes people to a little bit of risk. I don't think the code needs to be dry at this juncture (since, as you've noted, there's a clear next step best handled as a separate matter) but I do think we should sort this out before we ship. What do you think?

colinxfleming (commented):
The code is rocksolid and works pretty much exactly as I'd hoped! I'm of the opinion that we should do the 'keep existing option for a patient in the dropdown' like we do for insurance as part of this, due to the sensitivity of the voicemail field, but I'm interested in @elimbaum 's opinion here. Holding on approval until we get some clarity around that. 

Thanks for this excellent work @elimbaum !

app/helpers/patients_helper.rb
elimbaum: I was thinking that (since this is a common issue across many fields) we should rewrite `options_for_select` to exhibit this behavior for all fields within the patient information page. I'm not sure it makes sense to only copy it over to voicemail, unless you think there's a reason we wouldn't want that behavior for other fields (language, insurance, etc.).

@lomky pointed out that we could run in to a situation where someone mispells a config. We risk losing all users' configs if a correction to the config is made, but users aren't updated immediately.

If you think that makes sense, I was going to open an issue and work on that next.

elimbaum: I think this applies to:
- external pledge source
- language
- insurance
- support
- limit help text
- (and now) voicemail

colinxfleming: I think the options_for_select approach is a good idea as a next up, and that list of fields looks right to me at a glance (excepting limit help text, which I believe is an FYI and not something that gets saved on patient).

in the interim, could we copy the way `insurance_options` handles this? e.g. something like (pseudocoding here):

```
def voicemail_options(current_value = nil)
  standard_options = [...]
  custom_options = Config.find_by(config_key: 'voicemail_options')
  full_set = standard_options + custom_options

  if current_value.present? && ...
    full_set.push(current_value)
  end
  full_set
end
```

I think that will hit the goal of a) not deleting potentially sensitive entries from voicemail and b) keeping the door open to implement the more well-rounded and practical/safe approach you're advocating for here.

I think in the 'misspells a config' case @lomky identified, we'd be covered by the (misspelled) contents staying around with the current_value thing above, so we're safe on that front.

Does that seem like it makes sense?

elimbaum:
alrighty! have a look at that. also added a new test similar to the insurance options.

colinxfleming (approved):
This is perfect - maintains sensitive voicemail info if the configs change without defaulting to a potentially problematic default. Exactly what I was hoping for here - great work @elimbaum !

PR #2080, opened 2020-11-27
update render_async to use nonces by default

colinxfleming:

the `render_async` library has a feature in 2.1.8 that sets nonces by default on `render_async` partials. This sets that up so we don't have to think too hard about when to use nonces.

* bump `render_async` to 2.1.8
* use nonces by default in render_async partials

no views

no issues

For reviewer:
* Tag this PR:


colinxfleming:
not a free basket so closing for now

PR #2081, opened 2020-11-28
Add login URL to account creation email

### LINKED ISSUE #1897: Add a next step to the welcome email

@verbingthenoun noted that one of the things data vols consistently ask for is 'where exactly do I go now that I have an account'. The template that gets sent is in https://github.com/DCAFEngineering/dcaf_case_management/blob/master/app/views/user_mailer/ .

We should make it clear what the next step is, whether that's 'reach out to your volunteer manager for the URL', or including the URL in the mailer. Let's figure out which of those we should do and then implement.

making sure there's a clear next step when someone gets the welcome email; taking some strain off of volunteer managers.

Let's check the state of the art for welcome emails in sensitive environments. If it's cool, we should include a URL; if not cool, we should go with the explanation route. but we should do SOMETHING.

* **Anything else?**
stems from a conversation with @verbingthenoun

colinxfleming:

@verbingthenoun asked for this awhile ago - the new account creation welcome email is basically the same one we launched with, and doesn't really have a next step, so people need to have an additional touch with their volunteers. This tries to get ahead of that by including a URL to hit to log in.

The first commit sets things up to include 'reach out to your CM directors or volunteer lead' language, but the second commit includes the actual URL to go to (set as `SITE_URL` as an env var everywhere). 

I'd like @lomky 's opinion on sending the URL out in new account emails and @verbingthenoun 's opinion on the language here.

* bumps the new account email to include a next step (log in, and to email the CM director if you don't know the URL)
* includes a site url to go to

no view changes

* Fixes #1897 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/user_mailer/account_created.html.erb
+++ b/app/views/user_mailer/account_created.html.erb
@@ -1,5 +1,5 @@
 <p>Hi <%= @user.name %> --</p>
-<p>A DARIA account has been created for you! You can head <a href="<%= ENV.fetch('SITE_URL', 'https://www.google.com/?q=cats' %>">here</a> to log in.</p>
+<p>A DARIA account has been created for you! You can head <a href="<%= ENV.fetch('SITE_URL', 'https://www.google.com/?q=cats') %>">here</a> to log in.</p>
 <p>We recommend logging in with Google. If you don't log in with Google, you will need to reset your password.</p>
 <p>If you experience any difficulties, please reach out to your case management director or volunteer contact.</p>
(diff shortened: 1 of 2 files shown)


lomky:
Is the concern on sending out the URL just on automated sharing? I think I'm fine with it here since is only kicked off when an Admin has actively created their account.

lomky (approved):
:shipit:

colinxfleming:
Got clearance from @verbingthenoun that this was good to go -- merging

Issue #1859, opened 2019-11-11
Move bootstrap 4 to npm/webpacker


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Bootstrap 4 is a frontend package and we should try our best to keep our frontend packages in npm. (This would also let us start REALLY modernizing our js.) So that end, I'd like to drop our bootstrap gem dependency and have it be an npm package instead, and load it through webpack.

* **What feature or behavior is this required for?**

Grouping frontend dependencies together; opening the door for a javascript framework (see #1768 )

* **How could we solve this issue? (Not knowing is okay!)**

The following:

* we've removed the bootstrap gem from our dependencies
* we've installed bootstrap and any dependencies (maybe `popper_js`?) as npm packages
* we're loading bootstrap in webpacker

the `bootstrap_form` gem can probably stay where it is, but please flag for me if we need to figure out something out on that front.

* **Anything else?**

may depend on #1816

colinxfleming:
taking this off hacknights because it does, in fact, dpeend on #1816

colinxfleming:
we're getting this through npm and what a relief it is. I'm declaring victory

PR #2087, opened 2020-12-07
Port dashboards/call list SCSS into webpacker

colinxfleming:

One of my goals for 2021 is to go full webpacker and only be in sprockets a minimal amount and move as much as possible over to webpacker. As part of that, we should start 

* move more variables into the variables scss file
* move over dashboards into call_list.scss

No actual view changes, but here's a before and after to prove things match:

local:
image

sandbox:
image

* Bumps #1859 

For reviewer:
* Tag this PR:


lomky (approved):
:art: :shipit:

PR #2089, opened 2020-12-08
Port Config model to postgres

colinxfleming:

This rigs up the Config model to be in postgres rather than mongodb. It also adds a migration script as a rake task.

To acid test:
- check out `main` and run `rails db:drop db:create db:migrate db:seed` to start over
- start the rails server, go to the Config page, and add `cats rule` to one of the options, so we have an arbitrary out of nowhere value to check for
- switch to the new branch, migrate `rails db:migrate`, and reload the Config page. See all the values disappear
- run `echo "Config.destroy_all" | rails c` to delete the values that the last step autosetup
- run `rails migrate_to_pg:config` to run the rake task that migrates over mongo things (which are still intact). You should get a notice that 13 configs were ported
- reload the Config page. See all the values reappear, including your feelings about cats

Deploy plan here is to get this situated and then ship by deploying, then running the rake tasks on each server.

* Port Config model to PG
* Add a rake task to copy over Config objects

no view changes

* Bumps #2072

For reviewer:
* Tag this PR:


colinxfleming:
Unlike last time codecov kinda has a point this time - there's some code we should comment out in a few places as it isn't really being used yet. MongoConfig will be dead weight though and isn't worth writing tests for.

test/factories/config.rb
```diff
@@ -2,6 +2,5 @@
   factory :config do
     config_key { :insurance }
     config_value { { options: ['DC Medicaid', 'No insurance', "Don't know"] } }
-    created_by { FactoryBot.create(:user) }
```
lomky, line 5: I get why we're untying this now, but its temporary until the User model migrates, right? If so, I'd rather we comment out & tag to update with when we move User. Thoughts?

test/models/paper_trail_version_test.rb
```diff
@@ -0,0 +1,81 @@
+require 'test_helper'
+
+class PaperTrailVersionTest < ActiveSupport::TestCase
+  # Most of this behavior is related to user/patient behavior, leaving it commented out for now
+  # before do
+  #   @user = create :user # Can't do this until
```
lomky, line 6: Until what?

app/lib/paper_trail_version.rb
```diff
@@ -0,0 +1,47 @@
+# Extensions to base class of PaperTrail.
+class PaperTrailVersion < PaperTrail::Version
+
+  IRRELEVANT_FIELDS = %w[user_ids updated_by_id pledge_sent_by_id last_edited_by_id identifier updated_at].freeze
```
lomky, line 4: Are these permanently irrelevant, irrelevant until their model is migrated, or a mix?

app/lib/paper_trail_version.rb
```diff
@@ -0,0 +1,47 @@
...
+  #                    value.reject(&:blank?).join(', ')
+  #                  elsif key == 'clinic_id'
+  #                    Clinic.find(value).name
+  #                  elsif key == 'pledge_generated_by_id'
+  #                    ::User.find(value).name # Use the User model instead of the Userstamp namespace
+  #                  else
+  #                    value
+  #                  end
+  #   shaped_value
+  # end
```
lomky, line 46: Is this all commented out pending other migrations?

app/models/concerns/paper_trailable.rb
```diff
@@ -0,0 +1,17 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    has_paper_trail versions: { class_name: 'PaperTrailVersion',
+                                scope: -> { order(id: :desc) } }
+  end
+
+  def created_by
+    versions.first&.actor
+  end
```
lomky, line 12: :star_struck:

app/models/mongo_config.rb
```diff
@@ -0,0 +1,86 @@
+# Class so that funds can set their own dropdown lists of things
+# Marked for deletion after migration
+class MongoConfig
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::Userstamp
+  include Mongoid::History::Trackable
+  extend Enumerize
+
+  store_in collection: 'configs'
```
lomky, line 10: This is the only change from the original file, right?

lomky:
I think I'd be most comfortable if, before we merge this in and run it, we:

 - write up the follow on PR that fully removes the Mongo config (base off of this branch instead of `main`)
 - run an acid test like above, but also including checking out that next PR and confirming all runs clean

especially since this is our first model we're migrating.

Thoughts?

app/models/mongo_config.rb
colinxfleming: Yep, that's right. Necessary to still get the model to point at the `configs` collection (rather than inventing one called mongo_configs`.

app/lib/paper_trail_version.rb
colinxfleming: These are only really relevant to the `patient` model, and more particularly the changelog section that gets populated. I think we could strip these if we wanted to.

app/lib/paper_trail_version.rb
colinxfleming: Yeah, the methods here are used by the patient edit view. We don't need those methods until then.

test/factories/config.rb
colinxfleming: This is a great flag - Config (and Event) are always machine-generated, and so the created_by is always going to be whatever user triggers the autosetup. In mongo world we were userstamping this mostly so we could get modified_by ids, but in paper trail world that gets delegated to the paper trail entry, and so created_by/modified_by fields on the Config model itself get dropped.

test/models/paper_trail_version_test.rb
colinxfleming: 😬 

```suggestion
  #   @user = create :user # Can't do this until we port over the user model, for example
```

colinxfleming:
I think the additional `nuke the Config model` is perfectly reasonable. I'll spin that up tmw.

app/models/config.rb
```diff
@@ -27,27 +16,28 @@ class Config
...
   }.freeze
 
-  # Fields
-  enumerize :config_key, in: CONFIG_FIELDS
-  field :config_value, type: Hash, default: { options: [] }
-
-  # Indices
-  index({ config_key: 1 }, unique: true)
+  enum config_key: {
+    insurance: 0,
```
mdworken, line 20: 💯 for doing the explicit key setting

app/models/config.rb
```diff
@@ -27,27 +16,28 @@ class Config
...
-                version_field: :version,
-                track_create: true,
-                track_update: true,
-                track_destroy: true
-  mongoid_userstamp user_model: 'User'
-
   # Methods
   def options
-    config_value[:options]
+    config_value['options']
```
mdworken, line 40: Why did this end up having to change?

app/models/config.rb
```diff
@@ -58,7 +48,7 @@ def help_text
   end
 
   def self.autosetup
-    CONFIG_FIELDS.map(&:to_s).each do |field|
+    config_keys.keys.each do |field|
```
mdworken, line 51: Nit: I'd consider doing `each_key` instead of `keys.each` here.

mdworken: Less nit: we should consider renaming "config_keys" to "config_fields" as that's less implementation detail and more "what is this thing to our users"

mdworken (changes requested):
Left some small nits on naming convention, but looks good so far!

lomky:
Ran through the tests! I added the step to run the normal migrations after checking out the new branch. I then checked out #2097 and confirmed the configs were still behaving even with the fully nuke'd mongo config. So this is GTG by me!

lomky:
@colinxfleming I'm hitting approved. Happy for you to either add the test coverage for `PaperTrailable` here or file an issue for our hacknight crew to pick up!

lomky (approved):
:leaves: :tada: :shipit:

app/models/config.rb
colinxfleming: pg kicks them out stringified rather than symbolized.

app/models/config.rb
colinxfleming: I'm _really_ reluctant to do field renaming in the scope of this PR given the parts in motion - renaming fields would mean we'd have to crack open controllers to get this to work and I want to keep those as stable as possible. It feels like the pain of renaming outweighs the benefit by quite a bit here but maybe I'm missing something @mdworken ?

colinxfleming:
I'm gonna put this in the holding pen until we get the YH instance square. Will still continue pushing on the far edge of it, but major db paradigm shift + our planned YH work is quite a bit, so going to slow this down until we have that square.

app/models/config.rb
lomky: I second the reluctance - let's file an issue on our backlog to do after we're six months stable on postgres, not try to be too cute now

PR #2093, opened 2020-12-11
Execute db:migrate on release phase in heroku + app.json bump

colinxfleming:

Heroku has a nice feature: in the procfile you can define a release action. We make that `rails db:migrate` here to exec any pending db migrations so the release phase doesn't fail.

We also bump app.json to be postgres instead of mongo. Appropriate since mlab is gone forever. This will only affect new instances.

* run `rails db:migrate` on deploy to heroku
* bump app.json to add pg

no view changes

* Bumps #2072 

For reviewer:
* Tag this PR:

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
+release: rails db:migrate
 web: bundle exec puma -p $PORT -C ./config/puma.rb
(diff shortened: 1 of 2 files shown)


mdworken (approved):
Looks reasonable to me.

PR #2095, opened 2020-12-13
Adds postgres configuration steps

lomky:

For folks installing without docker who are using Postgres for the first time, there are some steps we have to follow to get configured. 

The two possible paths are:
 - setting up a postgres role for your system user, setting those values in your ENV, and tweaking the `config/database.yml` to use them ([inexact reference](https://www.digitalocean.com/community/tutorials/how-to-use-postgresql-with-your-ruby-on-rails-application-on-ubuntu-18-04))
 - tweaking your postgres server to not require passwords by adding `trust` methods to `pg_hba.conf` 

I went with the first as it feels more user friendly to tweak your local env vs the database settings.

* adds user & pass to `config/database.yml`
* adds config instructions to `docs/SETUP.md`

For reviewer:
* Tag this PR:


docs/SETUP.md
```diff
@@ -70,6 +70,25 @@ We use Postgres around these parts. Installation will differ based on your OS.
...
+   ```
+   sudo -u postgres createuser -s `whoami`
+   ```
+ * Set your env variables:
+   * Mac:
+     ```
+     echo "export POSTGRES_USER=`whoami`" >>~/.bashrc
+     echo "export POSTGRES_PASSWORD=postgres" >>~/.bashrc
+     echo "export PGDATABASE=dcaf_case_management_development" >>~/.bashrc
+     ```
```
lomky, line 84: Feedback requested from a Mac user: do these steps result in the values appearing in your ENV on next terminal start up? If not, which file should they go in?

docs/SETUP.md
```diff
@@ -70,6 +70,25 @@ We use Postgres around these parts. Installation will differ based on your OS.
...
+   * Ubuntu:
+     ```
+     echo "export POSTGRES_USER=`whoami`" >>~/.bashrc
+     echo "export POSTGRES_PASSWORD=postgres" >>~/.bashrc
```
lomky, line 88: These two make the Rails server happy

docs/SETUP.md
```diff
@@ -70,6 +70,25 @@ We use Postgres around these parts. Installation will differ based on your OS.
...
+     echo "export PGDATABASE=dcaf_case_management_development" >>~/.bashrc
```
lomky, line 89: This one makes your `psql` command work without having to pass `-d dcaf_case_management_development`

docs/SETUP.md
colinxfleming: Macs default to zsh now, obnoxiously, which makes it tough to write directions like this. But the tldr is that I think if you are running a recent mac os your default shell rc file is going to be `~/.zshrc` so I think we prrrrrrrob want to point that in that direction.

docs/SETUP.md
colinxfleming: `rails db` also works for this fwiw.

This doesn't interact with the test env, right? The `database.yml` overrules it?

docs/SETUP.md
```diff
@@ -70,6 +70,25 @@ We use Postgres around these parts. Installation will differ based on your OS.
...
 
+* Configure Postgres
+ * Setup your user; the suggested password is `postgres` if you only use this machine for development purposes
+   ```
+   sudo -u postgres createuser -s `whoami`
+   ```
+ * Set your env variables:
+   * Mac:
+     ```
+     echo "export POSTGRES_USER=`whoami`" >>~/.bashrc
```
colinxfleming, line 81: I wonder if `dotenv` might save us some kinda particular/confusing shell mucking by letting us just stick this in an `.env` file in the root of the project? https://github.com/bkeepers/dotenv I will defer to you whether or not it's worth the dependency add.

colinxfleming (commented):
couple of notes and a question. I'm feelin' it - this is great work sanding the rough edges of our new setup. Interested in your thoughts on the comments. Thanks much @lomky !

docs/SETUP.md
mdworken: I personally like dotenv a lot.

config/database.yml
```diff
@@ -5,6 +5,8 @@ development:
   url: <%= ENV.fetch('pg_url', 'postgres://localhost') %>
   database: dcaf_case_management_development
   pool: 5
+  username: <%= ENV['POSTGRES_USER'] || `whoami` %>
```
mdworken, line 8: I'd default this to `postgres` unless there's a specific reason not to.

lomky: Could you say why?

lomky:
Finally bit the bullet and installed from scratch now that we have postgres - this now represents the functional steps and settings for a bare metal install without having to touch anything outside of the installed repo! ready for final look over

colinxfleming (approved):
thanks very much for seeing this thru!

PR #2098, opened 2020-12-18
No clobber selectboxes

### LINKED ISSUE #2084: Prevent changes to custom configs from clobbering existing values

When custom configurations change (insurance, voicemail, language, etc.), we don't want users with the old values to be broken. Currently, both insurance & voicemail implement special code to show the available options, _as well as_ the current value, if not already part of the database. The current system exhibits slightly weird behavior (i.e. if a config is changed or deleted, the user will keep their old value, until any other field is updated; this change occurs silently).

Reliability of custom configurations; maintenance of user data across config changes. 

I think the best way to do this is to change over calls to `options_for_select` to a helper function which adds the current value if not already present. Pretty much just move the custom insurance + voicemail code into a function that is accessible from all fields. Not all fields have to use the updated version, of course.

elimbaum:

Select boxes on the patient information screen no longer clobber the patient's current value if that value no longer exists in database. Previously this behavior only existed for voicemail and insurance options.

* consolidates current-value handling into a single function which all helpers call
* fields affected are language, insurance, practical support, referred by, and voicemail

* Fixes #2084

For reviewer:
* Tag this PR:


colinxfleming (commented):
From an app perspective this works pretty much exactly how I was hoping it would - very nicely done @elimbaum . Tested this on these fields:

language
voicemail
referred by
insurance
practical support options
practical support source options

A question before I hit approve - what do we want to do about testing here? I think, for example, `insurance_options` covered this this case before (and still covers this case!). Do you think it's worth applying that or something similar to the other *_options helpers, so we can prevent them from regressing?

elimbaum:
Oh, totally fair. I did forget to write tests. Voicemail and Insurance options have tests (voicemail from my last PR). I believe practical support is also tested - like insurance, it was previously being handled manually with duplicated code; I consolidated the logic. However, I can check language and referred by.

Is there a good _general_ way to implement this kind of test? That is, without duplicating the same `create option/select option/delete option/confirm still available` test?

And did you mean to _only_ test the helper itself, or should we also add new system tests?

colinxfleming:
My opinion is that I think we definitely need the helper tests. I think the system tests are a nice cherry on top (these tend to save us from weird surprises and subtle behavioral changes) but not strictly necessary.

elimbaum:
ok, added tests to `language` and `referred by`. Turns out `practical supports` was already testing this behavior, since it had duplicate code reimplementing what insurance options had already been doing. I change that to use my common function, but the test-level functionality hasn't changed.

elimbaum:
This is failing due to a Nokogiri security issue. I'll submit a separate PR with the fix, but I'm not sure how to get those changes into this commit.

elimbaum:
nevermind, looks like the security issue was already fixed. Just had to merge the changes. This is ready to go!

test/helpers/patients_helper_test.rb
```diff
@@ -137,4 +181,67 @@ class PatientsHelperTest < ActionView::TestCase
...
+    ]
+
+    it 'should return default referral options' do
+      assert_same_elements expected_referrals_base, referred_by_options()
+    end
+
+    it 'should append extra options' do
+      expected_referrals = expected_referrals_base + ['Social Worker']
+
+      assert_same_elements expected_referrals, referred_by_options('Social Worker')
```
colinxfleming, line 243: I didn't know this was an assertion in the minitest toolbox! neat

colinxfleming (approved):
looks good, thanks for staying on this!

test/helpers/patients_helper_test.rb
elimbaum: I picked it up from somewhere in the test cases! But it is super helpful.

PR #2100, opened 2020-12-23
Remove some dead clinics css

colinxfleming:

Our original design for the clinics index had a kind of 'expanded' view that involved an accordion expansion. That's no longer the case, and we can remove this css.

I did a quick git grep around and it doesn't look like the word 'accordion' shows up anywhere within our codebase, so I'm feeling pretty good about nuking this. Also confirmed that nothing gets weird on the clinics index/show pages if we take this out.

* Remove some css dealing with a previous iteration of the clinics index

No view diff

* Bumps #1854 

For reviewer:
* Tag this PR:


lomky (approved):
:shipit:

PR #2102, opened 2020-12-25
Remove unused reports css

colinxfleming:

We deprecated the reporting endpoint since it never really worked terribly well. The routing is still set up, but it doesn't really yield anything useful.

The CSS is similarly overtaken by events, so let's just strip it and start fresh when we _do_ overhaul the endpoint.

* remove unused reports css

no view changes

* Bumps #1854

For reviewer:
* Tag this PR:


lomky (approved):
:soap:

PR #2105, opened 2020-12-31
Convert call list into its own model

### LINKED ISSUE #2104: Extract call lists into their own model

Call lists and their ordering was never super elegant, but moving this over to postgres increases the ruby complexity quite a bit. This is a ticket to explore doing that in mongo to get the railsy parts right in isolation.

This is part of the larger postgres work.

postgresizing

swapping to a has_many :patients, through: :call_list thing

* **Anything else?**
stems from #2072

colinxfleming:

As part of the pg overhaul I discovered that mongoid does some fancy handling around the fake-ish relation we have between patient and user that powers the call list. As such it's necessary to bite the bullet and extract it into a bonafide model, and adjust the interactions with it accordingly.

To make that happen we unfortunately have to refactor a lot of how the call list works. So as part of this I've added a lot more line scoping and a few more convenience methods.

* extract call list functionality into its own model

no view changes

* Fixes #2104

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/call_lists_controller.rb
+++ b/app/controllers/call_lists_controller.rb
@@ -1,5 +1,7 @@
 # Controller for managing a user's call list.
 class CallListsController < ApplicationController
+  include LinesHelper
+
   before_action :retrieve_patients, only: [:add_patient, :remove_patient]
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :not_found }
 
@@ -18,14 +20,14 @@ class CallListsController < ApplicationController
   end
 
   def clear_current_user_call_list
-    current_user.clear_call_list
+    current_user.clear_call_list current_line
     respond_to do |format|
       format.js { render template: 'users/refresh_patients', layout: false }
     end
   end
 
   def reorder_call_list
-    current_user.reorder_call_list params[:order]
+    current_user.reorder_call_list params[:order], current_line
     head :ok
   end
 
@@ -33,6 +35,5 @@ class CallListsController < ApplicationController
 
   def retrieve_patients
     @patient = Patient.find params[:id]
-    @urgent_patient = Patient.where(urgent_flag: true)
   end
 end
--- a/app/models/call_list.rb
+++ b/app/models/call_list.rb
@@ -1,0 +1,20 @@
+# Object representing a patient's call list.
+class CallList
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  # Relationships
+  belongs_to :user
+  belongs_to :patient
+
+  # Fields
+  field :order_key, type: Integer
+  field :line, type: String
+
+  # Validations
+  validates :order_key, :line, presence: true
+  validates :patient, uniqueness: { scope: :user }
+
+  # Indices
+  index line: 1
+end
--- a/app/models/call_list_entry.rb
+++ b/app/models/call_list_entry.rb
@@ -1,5 +1,5 @@
 # Object representing a patient's call list.
-class CallList
+class CallListEntry
   include Mongoid::Document
   include Mongoid::Timestamps
 
--- a/app/models/concerns/call_listable.rb
+++ b/app/models/concerns/call_listable.rb
@@ -8,43 +8,57 @@ module CallListable
   # AND they would otherwise be in the call list
   # (e.g. assigned to current line and in user.patients)
 
-  def recently_called_patients(line)
-    patients.where(line: line)
-            .select { |patient| recently_called_by_user? patient }
+  def ordered_patients(line)
+    call_lists.includes(:patient)
+              .in(line: line)
+              .order_by(order_key: :asc)
+              .map(&:patient)
   end
 
   def call_list_patients(line)
-    patients.where(line: line)
-            .reject { |patient| recently_called_by_user? patient }
+    ordered_patients(line).reject { |x| recently_called_by_user? x }
+  end
+
+  def recently_called_patients(line)
+    ordered_patients(line).select { |x| recently_called_by_user? x }
+  end
+
+  def recently_reached_patients(line)
+    ordered_patients(line).select { |x| recently_reached_by_user? x }
   end
 
   def add_patient(patient)
-    patients << patient
-    if call_order
-      reorder_call_list(call_order.unshift(patient.id.to_s))
-    else
-      reorder_call_list([patient.id.to_s])
-    end
+    present_calls = call_lists.includes(:patient).where(line: patient.line).to_a
+    return if present_calls.map { |x| x.patient.id.to_s }.include? patient.id.to_s
+    call_lists.where(line: patient.line).destroy_all
     reload
+
+    # Rebuild the call list by putting the new pt in front,
+    # and then recreating the pieces after it
+    call_lists.create! patient: patient,
+                       line: patient.line,
+                       order_key: 0
+    present_calls.each do |entry|
+      call_lists.create! patient: entry.patient,
+                         line: entry.patient.line,
+                         order_key: entry.order_key + 1
+    end
   end
 
   def remove_patient(patient)
-    patients.delete patient
+    call_lists.find_by(patient_id: patient.id).destroy
     reload
+  rescue Mongoid::Errors::DocumentNotFound
   end
 
-  def reorder_call_list(order)
-    update call_order: order
-    save
-    reload
-  end
-
-  def ordered_patients(line)
-    return call_list_patients(line) unless call_order
-    ordered_patients = call_list_patients(line).sort_by do |patient|
-      call_order.index(patient.id.to_s) || call_order.length
+  def reorder_call_list(order, line)
+    call_lists.destroy_all
+    order.each_with_index do |pt, i|
+      call_lists.create patient_id: pt,
+                        line: line,
+                        order_key: i
     end
-    ordered_patients
+    reload
   end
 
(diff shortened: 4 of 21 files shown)


colinxfleming:
This is coming along reasonably nicely, despite being a complicated extraction. General state of the world is that there are a few additional spots to clean up, but hoping to have this ready for review soon.

colinxfleming:
Okay, this is ready for review finally. Leaving some notes and tagging in reviewers.

app/models/concerns/call_listable.rb
```diff
@@ -8,43 +8,57 @@ module CallListable
...
+    ordered_patients(line).select { |x| recently_called_by_user? x }
   end
 
-  def remove_patient(patient)
-    patients.delete patient
+  def recently_reached_patients(line)
+    ordered_patients(line).select { |x| recently_reached_by_user? x }
+  end
+
+  def add_patient(patient)
```
colinxfleming, line 30: This is a slightly different approach to what we've had before - call list objects are super ephemeral and basically meant to be thrown out/regenerated on the regular. I'm very amenable to another approach here, but this seems to be reasonably resilient.

app/models/patient.rb
```diff
@@ -189,6 +190,12 @@ def okay_to_destroy?
 
   def destroy_associated_events
     Event.where(patient_id: id.to_s).destroy_all
+    CallList.where(patient_id: id.to_s).destroy_all
+  end
+
+  def update_call_list_lines
+    CallList.where(patient: self)
+            .update(line: self.line, order_key: 999)
```
colinxfleming, line 198: the order key here is an indicator of 'shove it to the back'.

app/views/dashboards/_table_content.html.erb
```diff
@@ -85,7 +85,7 @@
     </tbody>
   </table>
 
-  <% if table_type == 'call_list' && current_user.patients.count.nonzero? %>
+  <% if table_type == 'call_list' && current_user.call_lists.count.nonzero? %>
```
colinxfleming, line 88: A surprising amount of this pr was some variation of `user.patients` => `user.call_lists`.

app/views/users/refresh_patients.js.erb
```diff
@@ -14,7 +14,7 @@ $('#completed_calls').append(
                                 locals: { title: t('dashboard.partial_titles.completed_calls'),
                                           table_type: 'completed_calls',
                                           patients: current_user.recently_called_patients(current_line), 
-                                          utosortable: true }) %>'
+                                          autosortable: true }) %>'
```
colinxfleming, line 17: oops

colinxfleming (commented):
notes

colinxfleming:
showtime!

test/factories/call_list.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :call_list do
+    user
+    patient
+    line { 'DC' }
+    sequence :order_key do |n|
+      n
+    end
```
lomky, line 8: You don't need the block in this case, [docs](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#without-a-block)
```suggestion
    sequence (:order_key)
```

app/controllers/call_lists_controller.rb
```diff
@@ -18,21 +20,20 @@ def remove_patient
...
-    current_user.reorder_call_list params[:order]
+    current_user.reorder_call_list params[:order], current_line
     head :ok
   end
 
   private
 
   def retrieve_patients
     @patient = Patient.find params[:id]
-    @urgent_patient = Patient.where(urgent_flag: true)
```
lomky, line 36: where'd this one go? :eyes:

app/models/call_list.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Fields
+  field :order_key, type: Integer
+  field :line, type: String
+
+  # Validations
+  validates :order_key, :line, presence: true
+  validates :patient, uniqueness: { scope: :user }
+
+  # Indices
+  index line: 1
```
lomky, line 19: What's the `1` do here?

app/models/concerns/call_listable.rb
lomky: my only worry with this approach, is without any sort of transaction lock we could theoretically end up in a space where we've deleted all of this users call list entries for this line and then error and lose the entire call list set. if we're going to loop every existing present call anyway, why not just loop and bump the order_key?

app/models/call_list.rb
```diff
@@ -0,0 +1,20 @@
+# Object representing a patient's call list.
+class CallList
```
lomky, line 2: Suggestion: `CallList` implies a has_many patients relationship to me. Maybe `CallListEntry` instead?

test/models/user_test.rb
```diff
@@ -116,24 +118,23 @@ class UserTest < ActiveSupport::TestCase
...
       end
     end
 
     describe 'reorder call list' do
       before do
-        set_of_patients = [@patient, @patient_2, @patient_3]
-        set_of_patients.each { |preg| @user.add_patient preg }
-        @new_order = [@patient_3.id.to_s, @patient.id.to_s, @patient_2.id.to_s]
-        @user.reorder_call_list @new_order
+        [@patient, @patient_2, @patient_3].each { |preg| @user.add_patient preg }
```
lomky, line 135:
```suggestion
        [@patient, @patient_2, @patient_3].each { |patient| @user.add_patient patient }
```

app/models/concerns/call_listable.rb
```diff
@@ -8,43 +8,57 @@ module CallListable
...
     reload
+  rescue Mongoid::Errors::DocumentNotFound
   end
 
-  def ordered_patients(line)
-    return call_list_patients(line) unless call_order
-    ordered_patients = call_list_patients(line).sort_by do |patient|
-      call_order.index(patient.id.to_s) || call_order.length
+  def reorder_call_list(order, line)
+    call_lists.destroy_all
```
lomky, line 55: I don't love that is implicitly deletes any call list entries that weren't provided in the new order. Could we make that effect more obvious from the method name, or write this to only drop the reordered entries and shove the rest to the end?

lomky (commented):
Some notes and questions!

app/controllers/call_lists_controller.rb
colinxfleming: believe it or not, unnecessary db call! Saves us some overhead.

app/models/concerns/call_listable.rb
colinxfleming: I thought about that and wound up not doing it, reasoning that it was easier to mass drop and mass recreate than to update in place. But that might be overkill and the collections are always going to be pretty small, so this might be overoptimization. I'll swap it back to use an update strategy.

(my kingdom for mongodb transactions)

app/models/call_list_entry.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Fields
+  field :order_key, type: Integer
+  field :line, type: String
+
+  # Validations
+  validates :order_key, :line, presence: true
+  validates :patient, uniqueness: { scope: :user }
+
+  # Indices
+  index line: 1
```
colinxfleming, line 19: (moving the file ate the comment on this)

the `1` seems to be the mongoid convention for whatever reason. See the docs: https://docs.mongodb.com/mongoid/current/tutorials/mongoid-indexes/

colinxfleming:
Okay, ready for another pass! The significant changes are:

* `CallList` => `CallListEntry`
* use an `update` based strategy for reordering call list + adding new patients to call list rather than a destroy-based one, since we don't have transactions, and the call lists are small enough that it shouldn't matter.

lomky:
@colinxfleming I'm seeing a bug in the frontend for the hot reload - the newly added call appears as first, the formerly first shifts down, the formerly second one just disappears. 
If you don't reload the page, additional searching and adding works properly and doesn't eat more entries, but that first eaten one is missing. Hard reloading corrects the list, but the next added call eats a row again until hard reload.
HotReloadBug

colinxfleming:
@lomky I think using the proper method fixed! I think this is ready for another pass

lomky (approved):
looking perfect! :medal_sports:

app/views/calls/create.js.erb
```diff
@@ -15,7 +15,7 @@ if (location.pathname == Routes.authenticated_root_path) {
     '<%= escape_javascript(render partial: "dashboards/table_content",
                                   locals: { title: t("dashboard.partial_titles.call_list"),
                                             table_type: "call_list",
-                                            patients: current_user.ordered_patients(current_line),
```
lomky, line 18: Should this be `ordered_patients` or `call_list_patients`?

colinxfleming: ordered! thank you for catching this. I'm going to add a note that this is basically a private func.

colinxfleming: actually I think I'm turned around on this? From a skim of the code kinda looks like we should be using `call_list_patients` for everything, and `ordered_patients` is basically a private func. Going to review tomorrow and add a documentation note.

colinxfleming:
Okay! Fixed the root cause here that @lomky identified - some mixing and matching of `ordered_patients` and `call_list_patients`. To square this away, we make `ordered_patients` a private method basically responsible for the db calls and mapping to patient, and get consistent with our use of `call_list_patients` and `recently_called_patients`.

PR #2106, opened 2021-01-01
Port calls scss into webpacker

colinxfleming:

Starting now to get into the tougher shots in our css - the call css was equal parts

I took a few forward steps here:

* px became rem
* some better organization of the css rather than a straight copy-paste
* ported some stuff that was really about call list/dashboard styling into a different spot

* moves the calls.scss file contents out of sprockets and into webpacker

PR:

There's a very slight difference between these two, primarily in the margin spacing between the modal edges and the content. I turned it down a little bit - still plenty of whitespace and now resizes a bit cleaner.

image

Drag and drop behavior / styling is preserved too (but harder to take a screencap of).

sandbox:

image

* Bumps #1854 

For reviewer:
* Tag this PR:


app/javascript/styles/call_list.scss
```diff
@@ -18,3 +18,25 @@
 .bootstrap_toggle_spacing .checkbox {
   margin-top: 0px;
 }
+
+// For autosortable columns, make the cursor a pointer
```
colinxfleming, line 22: Calls seemed to be half call-list/dashboard stuff and half call modal stuff. Have moved the call list stuff into here to reflect that.

app/helpers/calls_helper.rb
```diff
@@ -29,7 +29,7 @@ def display_reached_patient_link(patient)
               patient_calls_path(patient,
                                  call: { status: 'Reached patient' }),
               method: :post,
-              class: 'btn btn-primary', id: 'calls-btn'
+              class: 'btn btn-primary calls-btn'
```
colinxfleming, line 32: Started doing this by class and not id, as an excuse to do a css best practice.

app/javascript/styles/calls.scss
```diff
@@ -0,0 +1,27 @@
+// Styling related to log-a-call modals.
+
+@import './_variables';
+
+// Core new call modal
+.calls-layout {
```
colinxfleming, line 6: A soft refactor here to nest these appropriately, and take out the base html elements (a bunch of them were `h4.whatever`, and that's not necessary since it's overruling a bunch of those styles).

colinxfleming (commented):
couple of notes on where the significant changes are

app/assets/stylesheets/calls.scss
```diff
@@ -1,55 +0,0 @@
...
-#calls-btn {
-  font-size: 17px;
-  padding: 10px 40px;
-}
-
-.handle > span {
-  color: #ddd;
-  margin-left: 5px;
-  font-size: 1.25em;
-}
```
lomky, line 37: looking at the refactor this styling wasn't migrated anywhere - is this 

> the base html elements 

you mention [here?](https://github.com/DCAFEngineering/dcaf_case_management/pull/2106/files#r550838411)

lomky (approved):
One doublecheck question, otherwise all looks good on my machine!

:telephone_receiver: :cat2: :grey_exclamation:

app/assets/stylesheets/calls.scss
colinxfleming: oh, the `base html elements` is about the h4s on some of the call modal styling. Kinda overkill to specify the base element on those.

But you're keeping me honest here - the story of that handle class is that it's a holdover before we started using fontawesome icons (due to bootstrap 4 dropping their icon library). There isn't a handle element in a span now, they're all svgs, so that's a dead class.

colinxfleming:
thx much @lomky !

PR #2107, opened 2021-01-03
Port events scss into webpacker

colinxfleming:

Ports the events scss out of sprockets and into webpacker. Along the way, we split it into two different concerns, one to deal with pagination (which I believe the accountants controller also consumes) and one that deals with the events well.

Pretty clean, all things considered. (After this we get more into the complicated/heavier css.)

* Move events.scss from sprockets into webpacker
* Break it into two - events log and kaminari pagination

PR:

image

Sandbox:

image

* Bumps #1854 

For reviewer:
* Tag this PR:


lomky (approved):
looks good!

PR #2109, opened 2021-01-08
Allow YH pledge gen path

colinxfleming:

I'm not 100% sure they use it in practice but YH has pledge generation turned on on their instance, so we should adjust to their taste. (and eventually, but not now, tear this out of the rails app proper.)

I'm going to godmode merge this after tests pass.

* handle YH pledge generation path that got added on the fork

no view change
no issue, for now

For reviewer:
* Tag this PR:


colinxfleming:
godmode

PR #2112, opened 2021-01-16
Port pregancnies.scss into individual files in webpacker

colinxfleming:

This might be the last holdout of our original object design, when we were trying to force pregnancies to be a different object, with a different controller. What a time that was.

Anyway, moving over another file into webpacker - please gut check the way these files are split up and make sure I'm not overdoing it.

* extract pregnancies.scss into individual files in webpacker

Tables identical:

image

Patient edit page also still identical:

image

* Bumps #1854 

For reviewer:
* Tag this PR:


lomky (approved):
Looks great! :shipit:

PR #2119, opened 2021-02-01
Explicitly add MONGODB_URI as required config var

xmunoz:
The application won't be able to connect to the database without setting this variable as part of the dyno creation process. There might be some other code changes needed to create the db, but I'll defer to you for that.


colinxfleming (approved):
We _might_ need to do `rails db:create` but I think this is most likely automagic. Let's assume automagic until proven otherwise.

PR #2124, opened 2021-02-09
Remove an unused method + small unit test improvements

colinxfleming:

Two iso'd and severable changed here from larger postgres work.

* `Patient#contacted_since` is no longer in the codebase and the method hasn't been touched in four years, which I'm guessing is when we moved everything to concerns. I can't think of a place where this is used so I think we can nuke it.
* simplifies `ExternalPledgeTest` around uniqueness in fund.
* remove an OBE `PatientTest` around line scoping.

no view changes

* Bumps #2072  (loosely)

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/callable.rb
+++ b/app/models/concerns/callable.rb
@@ -13,22 +13,4 @@ module Callable
   def last_call
     calls.order(created_at: :desc).first
   end
-
-  class_methods do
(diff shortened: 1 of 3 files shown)


lomky (approved):
:+1: :shipit:

PR #2132, opened 2021-02-12
change zipcode to string + validate

### LINKED ISSUE #2131: Zipcode is a number but should be a string

A DCAF CM pointed out that leading zeroes, such as in New England area codes, are getting cut off from zip codes. 

This is workable, but not totally accurate, and since we have some funds potentially on deck we should rep their area codes properly.

The current DB field is a number. I think I would suggest making this a string and writing a validator for it.

* **Anything else?**
nope!

elimbaum:

Patient zipcode field is now a string; validation has been added to prevent issues.

* Changed zipcode datatype
* Added frontend + backend validation
* Added test to the patient model

* Fixes #2131 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -42,7 +42,8 @@
         </div>
 
         <%= f.text_field :county, autocomplete: 'off' %>
-        <%= f.text_field :zipcode, autocomplete: 'off', pattern: '\d{5}'%>
+        <%= f.text_field :zipcode, autocomplete: 'off',
+                          pattern: '^\d{5}(-?\d{4})?$'%>
 
         <h3><%= t 'patient.information.other_contact.title' %></h3>
         <%= f.text_field :other_contact,
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -43,7 +43,7 @@
 
         <%= f.text_field :county, autocomplete: 'off' %>
         <%= f.text_field :zipcode, autocomplete: 'off',
-                          pattern: '^\d{5}(-?\d{4})?$'%>
+                          pattern: '\d{5}(-?\d{4})?'%>
 
         <h3><%= t 'patient.information.other_contact.title' %></h3>
         <%= f.text_field :other_contact,
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -42,7 +42,7 @@
         </div>
 
         <%= f.text_field :county, autocomplete: 'off' %>
-        <%= f.number_field :zipcode, autocomplete: 'off' %>
+        <%= f.text_field :zipcode, autocomplete: 'off', pattern: '\d{5}'%>
 
         <h3><%= t 'patient.information.other_contact.title' %></h3>
         <%= f.text_field :other_contact,
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -134,7 +134,10 @@ class Patient
   validates_associated :fulfillment
 
   # validation for standard US zipcodes
-  validates :zipcode, format: /\d{5}/, length: {is: 5}, allow_blank: true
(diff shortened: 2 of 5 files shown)


elimbaum:
Leaving in draft while I try to fix webpacker funkiness -- currently can actually test out the new frontend behavior since my JS is broken.

elimbaum:
Forgot to change database seed code to use new type. Also, when making changes to the db model, do we need to worry about migration? How does that process work (i.e. when we give the funds a new release with a different database)?

lomky:
Running migrations is a part of the deploy! You can see the normal migration processes for rails [here](https://guides.rubyonrails.org/v3.2/migrations.html)

app/views/patients/_patient_information.html.erb
```diff
@@ -42,7 +42,7 @@
         </div>
 
         <%= f.text_field :county, autocomplete: 'off' %>
-        <%= f.number_field :zipcode, autocomplete: 'off' %>
+        <%= f.text_field :zipcode, autocomplete: 'off', pattern: '\d{5}'%>
```
lomky, line 45: I suggest we support five digit, nine digit, and 5-4 digits!
https://faq.usps.com/s/article/ZIP-Code-The-Basics

elimbaum: Now with more ZIP!

Supports normal (12345) and ZIP+4 (12345-6789). Nine digit (123456789) will be converted to ZIP+4 (dash added). Do we need to support any other separators? USPS says no, but I guess there's a question of whether we want to allow spaces, periods, etc.

lomky:
Looking great! I think the only concern left now is "what happens if the existing data is bad?"

Right now it can only be a number, so that saves us a lot of headache. Can you check what happens if you:

 - Checkout main
 - create a patient with a 6 digit zip
 - create a patient with a 9 digit zip (no dash)
 - checkout your branch
 - try to load the patient

Does it behave nicely or throw an error on load?

elimbaum:
You mean test out what happens if we have integer zip codes and then change the data type, right?

elimbaum:
Ok, just tested this out! Pretty interesting. Basically, _somebody_ (pg?) is autoconverting numbers to strings, so in the 80% case, this change will actually be transparent.

- For zero-prefix zipcodes (01111), our current code using integers improperly removes the zero. This PR would cause those zipcodes to become four-character strings, which are invalid; however, that validation would not be run until the next update. So if you had a patient with zipcode **01111** (stored as `1111` currently), and then you tried to update some other part of the patient record, after this change is in place, you would get a zipcode validation error.
- Full 5-digit zipcodes are converted and handled correctly. `90210` becomes "90210".
- 6-digit zipcode, like the 4-digit case above, is flagged on next patient update. (The current code places no limits on the zipcode format).
- (fun part) 9-digit zipcodes are converted to strings without a problem. Then, when you next update the patient, the `before_validation` callback auto converts to ZIP+4, all transparently! So if we _are_ using 9 digit ZIPs anywhere, they won't be converted to the proper format _until_ the fund makes some other changes to the patient record.

This seems like acceptable behavior, particularly without doing any database migrations manually!

lomky:
Sounds perfect! Thanks for being thorough and checking. This all looks GTG to me, so tagging it approved. I wanna get @colinxfleming a chance to look just as an FYI in case this implicates how he's converting our stuff to postgres, but once he's clear on that note I think this is ready to ship!

lomky (approved):

colinxfleming (approved):
no concerns from me on the pg transition front!

colinxfleming:
Just to check, I ran a couple queries to see what the prevalence of non-5 digit zipcodes was. here's what I got:

snippet: `Patient.where(:zipcode.nin => [nil]).pluck(:zipcode).select { |x| x.to_s.length != 5 }` 
* DCAF - two of em
* every other fund I spot checked: none of em

I might frankly just fix the DCAF ones by hand after this goes out, doesn't seem worth it. (Also seems like using zip9s is pretty minimal.)

PR #2133, opened 2021-02-21
Swap in `update` for `update_attributes`

colinxfleming:

Pulling this out of the larger mongo -> pg work - Rails is deprecating update_attributes in favor of update per https://github.com/rails/rails/pull/31998 and this takes these out of our codebase in favor of `update`.

* replace `update_attributes` with just `update`

no view changes

* stems from #2072 

For reviewer:
* Tag this PR:


lomky (approved):
Weird that we were using the validation-skipping update_attributes to begin with!

PR #2134, opened 2021-02-21
Load clinic medicaid/naf info on page load rather than through api call

colinxfleming:

Small quality of life improvement here getting extracted from mongo/pg - rather than making an api call and mapping results on checkbox click, do it in view, assign as data attributes, and then read those data attributes on the fly.

* take out the calls to clinics index endpoint
* instead, load up medicaid/naf as data attributes on selects
* for good measure, make it so you can't select the INACTIVE option

no real view changes

* stems from #2072 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -117,13 +117,17 @@ module PatientsHelper
                               { data: { naf: !!clinic.accepts_naf, medicaid: !!clinic.accepts_medicaid } }
                             ]}
                             .unshift nil
-    inactive_clinics = clinics.reject(&:active)
-                              .map { |clinic| [
-                                t('patient.abortion_information.clinic_section.not_currently_working_with_fund', fund: FUND, clinic_name: clinic.name),
-                                clinic.id,
-                                { data: { naf: !!clinic.accepts_naf, medicaid: !!clinic.accepts_medicaid } }
-                              ]}
-                              .unshift ["--- #{t('patient.abortion_information.clinic_section.inactive_clinics').upcase} ---", nil, { disabled: true }]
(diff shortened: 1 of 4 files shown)


app/helpers/patients_helper.rb
```diff
@@ -111,11 +111,23 @@ def household_size_options
...
+                              clinic.id,
+                              { data: { naf: !!clinic.accepts_naf, medicaid: !!clinic.accepts_medicaid } }
+                            ]}
                             .unshift nil
-    inactive_clinics = clinics.reject(&:active)
-                              .map { |clinic| [t('patient.abortion_information.clinic_section.not_currently_working_with_fund', fund: FUND, clinic_name: clinic.name), clinic.id] }
-                              .unshift ["--- #{t('patient.abortion_information.clinic_section.inactive_clinics').upcase} ---", nil]
+    inactive_clinics = if clinics.reject(&:active).count > 0
+                         clinics.reject(&:active)
+                           .map { |clinic| [
```
lomky, line 122: is the new `if` check to avoiding calling map on potential `nil`? How would we get a `nil` here, as opposed to just an empty array? If we do worry about `nil`, what do you think about using nil-check invocation `&map` and `&unshift` instead?

lomky:
just out of curiosity, is this techdebt from when our clinic finder was an external gem that we migrated in?

lomky (approved):
looks good and works fine locally!

app/helpers/patients_helper.rb
colinxfleming: The `if` check here is to skip this in case there aren't actually any clinics marked inactive. But I do see what you're saying here - let me see if I can simplify.

PR #2135, opened 2021-02-21
Minor, extractable changes/diffs from mongo -> pg

colinxfleming:

In the spirit of reducing the mental load from #2123, here are some of the diffs we can extract that don't feel

* spacing, comments, and other similar trivial changes from mongo -> pg work
* a few test improvements and hardenings that were necessary under pg

no view changes

* stems from #2072 

For reviewer:
* Tag this PR:


lomky (approved):
:broom: :soap: :tada:

Issue #2120, opened 2021-02-01
Go through Google App Verification


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
We're coming up on a domain limit for Oauth - google allows you 10 before you have to give them an explanation of who you are and what kinda ship you're running. We're there, so it's time.

* **What feature or behavior is this required for?**
Being able to continue to add new funds and keep oauth working.

* **How could we solve this issue? (Not knowing is okay!)**
Go through Google App Verification https://console.cloud.google.com/apis/credentials/domainverification and run thru the deal.

* **Anything else?**
admins! This is the rare issue that is behind a gated resource, so probably for @colinxfleming  / @lomky  to handle

colinxfleming:
I submitted us for app verification thru google on 3/19.

xmunoz:
Any updates?

colinxfleming:
This is on me to bump - looks like next steps are going thru the steps here https://stackoverflow.com/questions/54231660/i-cant-deploy-google-oauth-app-in-heroku

colinxfleming:
place to register individual domains which I think covers us: https://search.google.com/search-console/welcome?hl=en

colinxfleming:
domains are officially verified

xmunoz:
Cool! I'm going to try to add Sandbox back to verified domains list here, which we had previously removed to make space for a newly onboarded fund: https://console.cloud.google.com/apis/credentials?authuser=1&project=dcaf-single-sign-on-production

We need to add it back because:
- Sandbox should, as much as possible, closely mirror production 
- I want to test the omniauth upgrade before rolling it out to other sites (#2141)

xmunoz:
Good news! I was able to add back Sandbox to our Google Oauth config. I updated the config vars in Heroku, restarted the app, and add my user (xmunoz@opentechstrategies.com) from the admin login. I was then able to successfully login with Google. I think we can mark this done?

Bad news! The instructions here for creating a new user on the command line no longer work:
https://github.com/DCAFEngineering/dcaf_case_management/blob/8264dd3b411c769d4c2bbc25225207151de708c7/docs/administering/SETTING_UP_A_NEW_INSTANCE.md#smoke-test-to-confirm-everything-is-working-right

I can open a separate issue for that.

colinxfleming:
per conversation with @xmunoz I think we're all clear here!

[...]

PR #2145, opened 2021-03-07
Upgrade omniauth-google-oauth2 to 0.8.1, fixes #2141

### LINKED ISSUE #2141: upgrade omniauth
Sqreen currently recommends version 2.0.3.

xmunoz:

* upgrade omniauth-google-oauth2
* upgrade devise
* explicitly add new dependency: [omniauth-rails_csrf_protection](https://github.com/cookpad/omniauth-rails_csrf_protection) (reason: https://github.com/omniauth/omniauth/wiki/Upgrading-to-2.0#rails)

* Fixes #2141 

For reviewer:
* When this is ready, I will remove the `WIP` prefix from the PR title and assign reviewers.
* I will create user accounts for the reviewers on sandbox.
* I will deploy this branch to sandbox, temporarily hijacking the normal release flow for testing purposes.
* Reviewers will try to sign in with Google to confirm that the change works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
       - run: bundle exec rake knapsack:minitest
       - run: brakeman --exit-on-warn .
       - run: bundle exec ruby-audit check
-      - run: bundle-audit update; bundle-audit check --ignore CVE-2015-9284
+      - run: bundle-audit update; bundle-audit check
       - run: bundle exec rails zeitwerk:check
       - store_artifacts:
           path: /home/circleci/abortioneering/tmp/capybara/
--- a/Gemfile
+++ b/Gemfile
@@ -34,8 +34,8 @@ gem 'paper_trail', '~> 10.3'
 gem 'paper_trail-globalid'
 
 # Our authentication library is devise, with oauth2 for google signin
-gem 'devise', '~> 4.7'
-gem 'omniauth-google-oauth2', '~> 0.8.0'
+gem 'devise', '~> 4.7.3'
+gem 'omniauth-google-oauth2', '~> 0.8.1'
 
 # We report errors with sentry
 gem 'sentry-raven'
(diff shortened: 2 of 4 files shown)


xmunoz:
Current status: Login with Google test fails in this way:

imagen

Also, the current PR is still not great, because it requires pointing to the head of master for the `devise` package.

xmunoz:
Hmm, now that I'm remembering how we on-boarded the last fund, I don't think it will be possible to test this on sandbox until #2120 is closed. @colinxfleming @lomky can you confirm that this is a blocker?

colinxfleming:
hm, these are the endpoints that the sandbox creds are auth'd on:

image

so if you're grabbing the googly creds from sandbox, I would have expected that to not complain about `forbidden`. I'll try to recreate this tmw.

I'm not sure #2120 matters -- I thnk that's more for adding more authorized domains, and I don't _think_ we need that for this, right?

xmunoz:
My plan was to deploy to sandbox for testing the upgrade; I don't think telling Google to redirect to `localhost` will work.

Also, where did you get that screenshot from? AFAIK we removed the Sandbox Google OAuth creds when onboarding the latest fund because we ran out of domains (prompting you to go through app verification in #2120).

I think we are done with verification though, because I was able to add back Sandbox to the list of OAuth clients just now, and test that logging in with Google works.

xmunoz:
This is the failing test:
test_0001_can sign in with Google Auth Account - GoogleSSOTest::access top page

Why do we also need to upgrade devise? This is the thread: https://github.com/heartcombo/devise/pull/5327

In the mean time, there is a failing test, and it's unclear whether this is something devise still needs to fix for the new omniauth version, something broken in the test, or something broken in our app since the upgrade.

This can be tested by deploying this PR to Sandbox and seeing if "Login with Google" still works. Even if it does though, I would feel super uncomfortable merging a PR that essential points to HEAD of devise. That maintainer hasn't released a new version yet with the required fixes. In my opinion, the best thing to do is wait.

colinxfleming:
Definitely agree we should keep our powder dry until a new devise version comes out! I'll adjust issue title accordingly.

colinxfleming (approved):
lovely! I checked the devise release notes and this matches what I'd expect out of that. fire when ready!

PR #2146, opened 2021-03-07
Port Clinic model to postgres

colinxfleming:

(Depends on pg-papertrail branch)

This ports overt the Clinic model into postgres.

Ways you can try this out:
* run the tests with diffs (should be all green)
* run this command to try the porter rake task out: `git checkout main && rails db:drop db:create db:migrate db:seed && git checkout - && rails db:migrate && rails migrate_to_pg:clinic_user_patient`

* port Clinic model over
* start the rake migration cmd

no view changes

* Bumps #2072

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/clinics_controller.rb
+++ b/app/controllers/clinics_controller.rb
@@ -1,6 +1,6 @@
 # Clinic-related functionality.
 class ClinicsController < ApplicationController
-  before_action :confirm_admin_user, except: [:index]
+  before_action :confirm_admin_user
   before_action :find_clinic, only: [:update, :edit]
   rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
 
@@ -8,7 +8,6 @@ class ClinicsController < ApplicationController
     @clinics = Clinic.all.sort_by { |c| [c.name] }
     respond_to do |format|
       format.html
-      format.json { render json: @clinics }
     end
   end
 
--- a/app/controllers/clinics_controller.rb
+++ b/app/controllers/clinics_controller.rb
@@ -2,7 +2,7 @@
 class ClinicsController < ApplicationController
   before_action :confirm_admin_user
   before_action :find_clinic, only: [:update, :edit]
-  rescue_from Mongoid::Errors::DocumentNotFound, with: -> { head :bad_request }
+  rescue_from ActiveRecord::RecordNotFound, with: -> { head :bad_request }
 
   def index
     @clinics = Clinic.all.sort_by { |c| [c.name] }
@@ -23,9 +23,9 @@ class ClinicsController < ApplicationController
   end
 
   def new
-    # i18n-tasks-use t('mongoid.attributes.clinic.phone')
-    # i18n-tasks-use t('mongoid.attributes.clinic.fax')
-    # i18n-tasks-use t('mongoid.attributes.clinic.active')
+    # i18n-tasks-use t('activerecord.attributes.clinic.phone')
+    # i18n-tasks-use t('activerecord.attributes.clinic.fax')
+    # i18n-tasks-use t('activerecord.attributes.clinic.active')
     @clinic = Clinic.new
   end
 
--- a/app/models/mongo_clinic.rb
+++ b/app/models/mongo_clinic.rb
@@ -1,33 +1,60 @@
 # Object representing a clinic that a patient is going to.
-class Clinic < ApplicationRecord
-  # Concerns
-  include PaperTrailable
+class MongoClinic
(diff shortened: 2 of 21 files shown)


app/controllers/clinics_controller.rb
```diff
@@ -23,9 +23,9 @@ def create
   end
 
   def new
-    # i18n-tasks-use t('mongoid.attributes.clinic.phone')
-    # i18n-tasks-use t('mongoid.attributes.clinic.fax')
-    # i18n-tasks-use t('mongoid.attributes.clinic.active')
+    # i18n-tasks-use t('activerecord.attributes.clinic.phone')
```
colinxfleming, line 26: `activerecord` i18n works the same way `mongoid` i18n does, though under the activerecord namespace. We can basically square this away by copying them into this new namespace. Expect to see this for several models.

app/models/clinic.rb
```diff
@@ -1,62 +1,35 @@
...
   # e.g. so a fund can have an 'OTHER CLINIC' catchall.
   EXCLUDED_ZIP = '99999'
 
-  # Relationships
-  has_many :patients
-  has_many :archived_patients
-
   # Scopes
-  # Is gestatiional_limit either nil or above x?
+  # Is gestational_limit either nil or above x?
```
colinxfleming, line 11: spelling is hard.

app/models/clinic.rb
```diff
@@ -1,62 +1,35 @@
...
-                track_destroy: true
-  mongoid_userstamp user_model: 'User'
-
   # Methods
   def display_location
     return nil if city.blank? || state.blank?
     "#{city}, #{state}"
   end
 
+  def display_coordinates
```
colinxfleming, line 29: A reflection of pg storing numbers differently than mongo does.

app/models/mongo_clinic.rb
```diff
@@ -0,0 +1,82 @@
+# Object representing a clinic that a patient is going to.
+class MongoClinic
```
colinxfleming, line 2: Copying this over into its own deal so we can still access the collection in porting scripts, as we have previously.

lib/tasks/migrate_to_pg.rake
```diff
@@ -1,48 +1,63 @@
...
+        attrs['event_type'] = event_type_mapping[attrs['event_type']]
+        attrs
+      end
+
+      migrate_model(pg, mongo, extra_transform)
     end
+  end
 
-    migrate_model(pg, mongo, extra_transform)
+  task clinic_user_patient: :environment do
```
colinxfleming, line 32: I'm sticking everything in the same namespace - doesn't really make sense to me to have them in diff spots.

lib/tasks/migrate_to_pg.rake
```diff
@@ -1,48 +1,63 @@
...
+    ActiveRecord::Base.transaction do
+      # Start by moving clinics
+      pg = Clinic
+      mongo = MongoClinic
+      extra_transform = Proc.new do |attrs, obj|
+        attrs['mongo_id'] = obj['_id'].to_s
```
colinxfleming, line 38: So that we can preserve existing relationships we're adding a mongo_id field to stuff that relates to other stuff. This is part of the reason why we're doing Clinic first - it doesn't depend on much stuff compared to Patient (though we could also do User at this juncture).

test/models/clinic_test.rb
```diff
@@ -20,27 +17,16 @@ class ClinicTest < ActiveSupport::TestCase
 
     it 'should be unique on name' do
       clinic_name = @clinic.name
-      dupe_clinic = build :clinic, name: clinic_name, created_by: @user
+      dupe_clinic = build :clinic, name: clinic_name
```
colinxfleming, line 20: Another common pattern we'll see a lot - `created_by`s getting dropped. PaperTrail does these by storing everything on version objects rather than the model itself.

colinxfleming (commented):
notes

lomky:
I don't think this is up to day with `pg-papertrail` but I could be mistaken - I don't use the `force-push` yolo methodology! 😄

colinxfleming:
Okay, this is rebased and ready for a pass!

* rails db:drop db:create db:migrate db:seed works -- it builds a few patients, but doesn't get all the way through (makes sense, since Patient is still in Mongoland)
* rails test works (though there are a whole slew of failures as expected)
* rails s spins up a server
* the call list dashboard and patient edit view both work

lomky (approved):
:shipit: hype for postgres DARIA 📈📈📈

PR #2151, opened 2021-03-12
obsolete URI encode function

### LINKED ISSUE #2088: Fix deprecation warning on URI.escape

`rails test` is yielding the following warnings:

```
dcaf_case_management/app/services/uri_service.rb:8: warning: URI.escape is obsolete
```

The task here is track this down and switch it to something that keeps tests passing but doesn't yield that warning.

Keeping application up to date!

I would bet that there's some method with similar behavior - this is possibly an exercise in doing some googling to see what that method has been replaced by. If that's not the case and it winds up being some surprise hassle, please document in this ticket.

* **Anything else?**

nope!

elimbaum:

We were previously encoding URIs before parsing them. Instead, we should just parse them. There was a test that tried to encode a string (invalid URI) into a URI, which I have removed.

* removes obsolete `URI.encode` function
* removes test that we didn't need

* Fixes #2088 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/configs_controller.rb
+++ b/app/controllers/configs_controller.rb
@@ -9,12 +9,14 @@ class ConfigsController < ApplicationController
   def update
     @config = Config.find params[:id]
     @config.config_value = format_config_params(config_params)
+
     if @config.save
       flash[:notice] = t('flash.config_update_success')
       redirect_to configs_path
     else
-      flash[:danger] = t('flash.config_failed_update')
-      render 'index'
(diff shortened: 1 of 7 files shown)


elimbaum:
Leaving as draft until we decide if this is ok with the test removed.

elimbaum:
This may not actually work. Looking into it some more.

elimbaum:
There are actually bigger issues with the URI service which I am looking into. See #2153. The good news is that we shouldn't actually need `URI.encode`, in any form! 😄

PR #2152, opened 2021-03-12
turned list of options into a select!

### LINKED ISSUE #1990: On config page, dropdown configs should display as dropdowns

On the config management panel, there are several options that are for fleshing out dropdowns. In the UI, though, they show up as lists of options with some confusing array notation:

image

These are selects, and they should show up as selects when you're mucking around with them!

image

Not confusing people who are trying to configure a DARIA environment!

We do something similar in the `config.rb` model around help text overrides. I suspect we could do something similar to display the html over there. It may be tricky to map the specific config to the function using it though.

* **Anything else?**
As a bonus: if we can actually make them show are real dropdowns here, that's be _sweet_

elimbaum:

Rather than display config options as a list with weird arrays, now display them as selects.

* Displays options as a dropdown, not a text-list of items.
* For arrays (of the form `[display, internal]`) show only the first element. We don't care about internal representation here.
* Ignore `nil` elements (some fields allow empty values, but we don't need to show that here). 

For reference, this is what the new view code is:

```html
<div class="col" id="language_options_list">
      <label>Current options are:</label>
      <br>
      <select>
          <option>English</option>
          <option>Spanish</option>
          <option>French</option>
          <option>Korean</option>
      </select>
  </div>
```

(before it was just each option separated by `<br/>`)

Screenshot:

Screen Shot 2021-03-11 at 21 30 08

* Fixes #1990 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/configs/_config.html.erb
+++ b/app/views/configs/_config.html.erb
@@ -13,13 +13,9 @@
   <div class="col" id="<%= config.config_key.to_s %>_options_list">
     <% if Config::HELP_TEXT_OVERRIDES[config.config_key.to_sym] %>
     <% else %>
-      <label><%= t('configs.config.current_options') %></label>
-      <br />
-      <select>
-        <% public_send("#{config.config_key}_options").compact.each do |opt| %>
-          <option><%= opt.is_a?(Array) ? opt[0] : opt %></option>
-        <% end %>
-      </select>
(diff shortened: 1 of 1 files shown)


elimbaum:
oops, realized i'm not using rails forms here. changing.

colinxfleming (approved):
nicely done! Does what it says on the tin

Issue #2153, opened 2021-03-13
URI parsing (for fax service) is unstable


elimbaum:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Properly parse user-entered URLs, particularly ones that have slashes in them or are otherwise funky! Currently, we _only_ support URLs of the form `http://website.com`. Stumbled across this while working on PR #2151.

* **What feature or behavior is this required for?**
Fax service link in the footer.

* **How could we solve this issue? (Not knowing is okay!)**
Improved URL parsing! Currently, we are doing `@uri = URI.parse(URI.encode(uri))` which leads to all kinds of funky behavior!

* **Anything else?**
Here's the error:

> URI::InvalidComponentError in Configs#index
> Showing /usr/src/app/app/views/layouts/_footer.html.erb where line #15 raised:
> bad component(expected host component): a%20random%20string/testing
```ruby
  def fix_leading_slashes
    if @uri.scheme.nil? && @uri.host.nil? && @uri.path.present?
      @uri.host = @uri.path # this line!
      @uri.path = ""
    end
  end
```

Specifically, you can trigger this by entering a Fax service URL that has a slash in it, such as `https://yolofax.com/fax`. When the above code runs (while trying to render the footer), it will try to set the host (formerly `yolofax.com`) to the path (which is `/fax`), but hosts are not allowed to have slashes in them.

**Caution!!!** don't test this on the sandbox or any other environment where you can't easily rebuild the db! because the footer is rendered on _all_ pages, configuring an invalid url will make... all pages... crash. 😢 and you won't be able to fix it.

Another quirk of this bug is that you _can_ set the fax service url to something that would otherwise be invalid, such as `this is not a website`. Because of the call to `URI.encode`, Rails just renders the fax footer with an href to `https://this%20is%20%not%a%website` which is... not a website! my browser, at least, has no idea what to do with that.

colinxfleming:
thank you for spinning this up and tagging it @elimbaum !

elimbaum:
[...]

PR #2157, opened 2021-03-22
Updated URL handling in Configs

### LINKED ISSUE #2088: Fix deprecation warning on URI.escape

elimbaum:

Config URLs were previously not being handled in a super reliable way. Only `fax service` had any validation whatsoever, using the `UriService` class. I have revamped the URL validation (to do it at config-save time, rather than page-render time) and added it to all URLs.

* Removes usage of `UriService` (can probably delete this file)
* Adds a validator in the Config model for all fields that are tagged as URLs, using the built-in Ruby URI class.
* Converts all URLs into `https://`, and tries to add schemes to schemeless URLs (i.e. you can enter `www.example.com`, which is _technically_ not a valid URL, and makes it `https://www.example.com`.

* Fixes #2088  and #2153 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -82,7 +82,7 @@ class Config < ApplicationRecord
     
     url = UriService.new(maybe_url).uri
 
-    if not url
+    if !url
       errors.add :base, "\"#{maybe_url}\" is not a valid URL for #{config_key.humanize}."
     else
       config_value['options'] = [url]
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -81,6 +81,9 @@ class Config < ApplicationRecord
   def validate_urls
     url = options.try :last
 
+    # empty url is allowed
+    return if url.nil?
+
     if not url =~ /\A#{URI::regexp(['https'])}\z/
       errors.add :base, "\"#{url}\" is not a valid URL for #{config_key.humanize}."
     end
@@ -90,7 +93,10 @@ class Config < ApplicationRecord
   def clean_urls
     url = options.try :last
 
-    # don't have to do anything
+    # don't try to clean empty url
+    return if url.nil?
+
+    # don't have to do anything, already https
     return if url.start_with? 'https://'
 
     # convert http or // to https://
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -38,8 +38,7 @@ class Config < ApplicationRecord
   # Validations
   validates :config_key, uniqueness: true, presence: true
 
-  before_validation :clean_urls, if: -> { Config_URLs.include? config_key }
-  validate :validate_urls, if: -> { Config_URLs.include? config_key }
+  validate :validate_urls, if: -> { CONFIG_URLS.include? config_key }
 
   # Methods
   def options
@@ -76,38 +75,15 @@ class Config < ApplicationRecord
     start ||= "monday"
     start.downcase.to_sym
   end
+
   def validate_urls
-    url = options.try :last
+    maybe_url = options.last
+    url = UriService.new(maybe_url).uri
 
-    # empty url is allowed
-    return if url.blank?
-
-    if not url =~ /\A#{URI::regexp(['https'])}\z/
-      errors.add :base, "\"#{url}\" is not a valid URL for #{config_key.humanize}."
+    if not url
+      errors.add :base, "\"#{maybe_url}\" is not a valid URL for #{config_key.humanize}."
+    else
+      config_value['options'] = [url]
     end
   end
-
-
-  def clean_urls
-    url = options.try :last
-
-    # don't try to clean empty url
-    return if url.blank?
-
-    # don't have to do anything, already https
-    return if url.start_with? 'https://'
-
-    # convert http or // to https://
-    if url.start_with? /(http:)?\/\//
-        url = url.sub /^(http:)?\/\//, 'https://'
-
-    # convert no scheme to https:// (i.e. example.com -> https://example.com)
-    elsif not url.start_with? '/'
-        url = 'https://' + url
-    end
-
-    # set config back to what it was
-    config_value['options'] = [url]
-  end
-
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -32,8 +32,16 @@ class Config < ApplicationRecord
     voicemail: 12,
   }
 
+  # which fields are URLs
+  Config_URLs = %w[fax_service practical_support_guidance_url resources_url]
+
+
   # Validations
+  # before_validation :clean_urls
+
   validates :config_key, uniqueness: true, presence: true
+
+  validate :validate_urls, if: -> { Config_URLs.include? config_key }
 
   # Methods
   def options
@@ -70,4 +78,49 @@ class Config < ApplicationRecord
     start ||= "monday"
     start.downcase.to_sym
   end
+
+
+  def validate_urls
+    url = options.try :last
+    logger.info("==== RUNNING VALIDATE ===== ")
+    logger.info(url)
+
+    if not url =~ /\A#{URI::regexp(['https'])}\z/
+      errors.add :base, "\"#{url}\" is not a valid URL for #{config_key.humanize}."
+    end
+  end
+
+
+  def clean_urls
+    # only run this for URL configs, above
+    return unless Config_URLs.include? config_key
+
+    logger.info("===== RUNNING CLEAN URLS =======")
+
+    logger.info("#{config_key}: #{options.try :last}")
+    logger.info("full val #{config_value}")
+
+    url = options.try :last
+
+    # don't have to do anything
+    return if url.start_with? 'https://'
+
+
+    return false
+
+    # # convert http or // to https://
+    # if url.start_with? /(http:)?\/\//
+    #     url = url.sub /(http:)?\/\//, 'https://'
+
+    # # convert no scheme to https://
+    # elsif not url.start_with? '/'
+    #     url = 'https://' + s
+    # end
+
+    # # set config back to what it was
+    # config_value['options'] = [url]
+
+    logger.info("===== END CLEAN URLS ===========")
+  end
+
 end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -78,6 +78,8 @@ class Config < ApplicationRecord
 
   def validate_urls
     maybe_url = options.last
+    return if maybe_url.blank?
+    
     url = UriService.new(maybe_url).uri
 
     if not url
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -82,7 +82,7 @@ class Config < ApplicationRecord
     url = options.try :last
 
     # empty url is allowed
-    return if url.nil?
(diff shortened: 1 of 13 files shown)


app/models/config.rb
```diff
@@ -32,9 +32,17 @@ class Config < ApplicationRecord
     voicemail: 12,
   }
 
+  # which fields are URLs
+  Config_URLs = %w[fax_service practical_support_guidance_url resources_url]
```
elimbaum, line 36: This list is how the config model decides what is a URL.

app/models/config.rb
```diff
@@ -70,4 +76,40 @@ def self.start_day
     start ||= "monday"
     start.downcase.to_sym
   end
+
+
```
lomky, line 80:
```suggestion

```

app/models/config.rb
colinxfleming: Looks right! Let's constantize it per convention and have it in all caps:

```suggestion
  CONFIG_URLS = %w[fax_service practical_support_guidance_url resources_url]
```

app/models/config.rb
```diff
@@ -70,4 +76,40 @@ def self.start_day
...
+    url = options.try :last
+
+    # don't try to clean empty url
+    return if url.blank?
+
+    # don't have to do anything, already https
+    return if url.start_with? 'https://'
+
+    # convert http or // to https://
+    if url.start_with? /(http:)?\/\//
```
colinxfleming, line 103: Would this stuff make more sense in `UriService`? This is quite a bit of very particular uri handling, wonder if it makes sense to extract it into our service class.

app/services/uri_service.rb
```diff
@@ -5,7 +5,7 @@ def initialize(uri)
     return nil if uri.nil?
 
     begin 
-      @uri = URI.parse(URI.encode(uri))
+      @uri = URI.parse(uri)
```
colinxfleming, line 8: this is no longer necessary because the `clean` stuff in config is handling it, right?

test/models/config_test.rb
```diff
@@ -3,6 +3,31 @@
 class ConfigTest < ActiveSupport::TestCase
   before { @config = create :config }
 
+  describe 'callbacks' do
+    it 'should clean URLs before save' do
+      # doesn't touch https
+      c = Config.find_or_create_by(config_key: 'resources_url')
```
colinxfleming, line 9: I have a suggestion for how to simplify this test a little bit - why don't we:

* move the meat of the cleaning into `UriService#clean_uri`
* do the heavy duty testing of the cleaning logic there (L20 on down, but just testing various strings)
* change this test to something that does a spot check to make sure that urls are getting cleaned when necessary (maybe the `http` -> `https`), and things are getting skipped appropriately like on L12 and L17)
* loop through each of the url options to repeat the test, to make sure they're all covered

what do you think?

test/models/config_test.rb
```diff
@@ -19,6 +44,26 @@ class ConfigTest < ActiveSupport::TestCase
...
+
+    it 'should validate URLs' do
+      c = Config.find_or_create_by(config_key: 'fax_service')
+      
+      # good
+      c.config_value = { options: ["https://good.com"] }
+      assert c.valid?
+
+      # good but needs cleanup
+      c.config_value = { options: ["www.efax.com/path"] }
```
colinxfleming, line 56: Hm, can you say more about why this is allowed still? I would think that something that would get changed by cleanup should be invalid, but interested in your perspective here.

test/services/uri_service_test.rb
```diff
@@ -2,10 +2,12 @@
 
 class UriServiceTest < ActiveSupport::TestCase
   describe 'URI service utility' do
-    it 'instantiates a uri from string' do
-      uri = UriService.new("some yolo test uri").uri
-      assert uri.is_a?(URI)
-    end
+    ## this is an invalid URL
```
colinxfleming, line 5: What do you propose we do with this test? Is it no longer needed and deleteable, should we re-work it, or?

colinxfleming (commented):
porting over relevant comments from #2162 - same basic feedback applies. The approach is right on.

Let me know what you think, or if I need to explain what I'm looking for better, but again -- fundamental idea is spot on, just gotta do some work to make sure it stays stable long term. Nice job here.

app/services/uri_service.rb
elimbaum: Correct. We were actually misusing `URI.encode` before - that function is meant only for encoding form data or special characters in paths (it does the special-character-to-`%XX` conversion) - and we were using it to encode the entire URL, including the host. I believe it got put in there because `URI.parse` would crash if you gave it a bad URL; calling `encode` allows those bad URLs to fail silently...

test/models/config_test.rb
elimbaum: The cleanup is run by `before_validation` - perhaps I should make this more clear in the test. The cleanup is run automatically when we call `c.valid?` and adds `https://`, making it a valid URL. So, for example, after L57 is run, the URL is now `https://www.efax.com/path`.

test/services/uri_service_test.rb
elimbaum: I'll move all the URI testing over the URI service, and then that should be an _invalid_ URL. See above comment about `URI.encode` misuse - that's why this test claimed it should be valid.

test/services/uri_service_test.rb
```diff
@@ -20,15 +23,20 @@ class UriServiceTest < ActiveSupport::TestCase
...
       u = UriService.new("www.noleadingslashesurl.com")
      
-      assert_equal u.secure_scheme_uri!.to_s, "https://www.noleadingslashesurl.com"
+      assert_equal "https://www.noleadingslashesurl.com", u.uri.to_s
+      assert_equal "https", u.uri.scheme
+
+      # convert http to https
+      u = UriService.new("http://unsafe.biz/path")
+
+      assert_equal u.uri.to_s, "https://unsafe.biz/path"
```
elimbaum, line 39: Moved tests from `config_test` to `uri_service_test`

test/models/config_test.rb
```diff
@@ -3,6 +3,24 @@
...
+        
+        # confirm that scheme-less URLs are given one
+        c.config_value = { options: ["www.needs-scheme.net"] }
+        c.save
+        assert_equal "https://www.needs-scheme.net", c.options.last
+
+        # doesn't touch empty config
+        c.config_value = { options: [] }
+        c.save
+        assert_nil c.options.last
```
elimbaum, line 19: just doing basic testing of URL processing in the config test

app/services/uri_service.rb
```diff
@@ -2,25 +2,33 @@ class UriService
...
     rescue URI::InvalidURIError
-      nil
+      # Maybe we forgot the scheme... try adding
+      begin
+        @uri = URI.parse("https://" + uri)
+      rescue URI::InvalidURIError
+        # nope!
+        @uri = nil
+        return
+      end
```
elimbaum, line 23: Let me know if you don't like how this is structured now - I know nested exception handlers aren't 100% ideal. I realized that at the end of the day, all of my fancy regexes were just _approximating_  the `URI.parse` function, so I might as well use it. Basic overview of the new system:

1. Try to parse the URL as provided.
2. If it fails, it's either missing the scheme, or invalid. If the URL has class `URI::Generic` that also tells us there's no scheme, so raise the `InvalidURIError` to run the nested exception handler.
3. Prepend `https://` and try to reparse. If _this_ fails, it's definitely invalid; return.
4. Change the scheme to `https`. If users provide _valid_ HTTP URLs, steps 2 and 3 are skipped, and we just adjust the scheme.

app/models/config.rb
```diff
@@ -70,4 +75,17 @@ def self.start_day
...
     start.downcase.to_sym
   end
+
+  def validate_urls
+    maybe_url = options.last
+    return if maybe_url.blank?
+    
+    url = UriService.new(maybe_url).uri
+
+    if not url
```
lomky, line 85: Suggestion: use `!` preferrentially to `not` - you can get some nasty bugs from `not`, as its _much_ lower in the order of operations evaluation. I'm not sure I've ever used `not` in over a decade of programming! [See a full explainer here](https://stackoverflow.com/questions/27429711/difference-between-not-and-in-ruby)

lomky:
Thanks for the ping! a couple things of note, one managable and one I'm still thinking about.

First - we have to consider existing data, since validations won't run until they save the field again. Right now if they have, for example `www.yolo.com` as a URL, that link will now break:
broken fax

I think we can handle that one of two ways - 
1) undo the change to the footer that removes the explicit scheme
2) have me or Colin manually fix any existing URLs, which is ~3 URLs per instance, so ~20 items to check

Second, the URL parser is a bit too generous right now - if you type any single word string in it works:
lenient URL parsing
I'm a little puzzled how we check that, but worst case here is the URL won't work - it won't crash the app like our previous behavior, so this is still an improvement!

app/models/config.rb
elimbaum: totally! i went down this rabbit hole with `and` and `&&` and `or` and `||` (lol) but forgot about `not`!

elimbaum:
For your first concern, I'll think about that as well. We could retain the original functionality to some degree, and add `https://`, on page load, if the URL doesn't start with it already. This should take care of that edge case and be stable until funds update their URLs.

For #2, I agree that it's weird, but I think it is the intended behavior. We usually only see this (non-TLD-domains) on LANs, but my browser at least attempts to navigate to `https://com`. We could mandate all hostnames have at least two components, although that doesn't fix it, for example, in the UK (`co.uk` would be allowable). 

I think the *proper* way of handling this is by checking against the [public suffix list](https://publicsuffix.org/) but this quickly becomes a non-trivial problem.

lomky:
Adding in the fallback `https://` sounds like a good catch to me!

elimbaum:
Sounds good. I'll put the fallback in before Thursday!

elimbaum:
okie dokie, have a look! I've added (slightly redundant for now, but oh well) calls to `UriService` in each of the three views we are concerned about (footer/fax, nav bar/resources, practical support). This is _in addition to_ the calls to `UriService` in the config validations; ideally, we can eventually remove the redundancy.

I confirmed this worked by

1. checked out main
2. set all three urls to something of the form of `abc.com`
3. confirmed they didn't work (link pointed to `localhost/abc.com`)
4. checked out `uri_escape_obsolete`
5. Confirmed the urls work! and are `https://`

lomky (approved):
Ready for prime time! Thank you so much for the great PR! 🎉 :shipit:

PR #2158, opened 2021-03-25
update mime gem to avoid a rails left-pad style mess

### LINKED ISSUE #2154: docker setup issue - webpack out of sync

Setup instructions, on docker-compose up, are returning an error because webpack is being weird and not compiled yet. Maybe there's a depends_on we need to do or something?

non-frustrating setup instructions

idk

* **Anything else?**

cc @patriciaDAscientist  who discovered this

colinxfleming:

This happened: https://dev.to/cseeman/what-s-up-with-mimemagic-breaking-everything-he1 
and as a result the mimemagic version we were on got yanked. This moves to a version that exists. The version that exists unfortunately requires pulling in an external dependency, so we adjust the setup instructions too.

* `bundle update mimemagic`
* adjusts setup instructions / dockerfile to add an extra dependency

no view changes

fixes #2154 along the way 

For reviewer:
* Tag this PR:

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Colin Fleming <c3flemin@gmail.com>
 # configure environment variable
 # note: move this to three ARG commands when CircleCI updates their docker
 ENV DCAF_DIR=/usr/src/app \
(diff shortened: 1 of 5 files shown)


colinxfleming:
godmode on this since it fixes some stuff we gotta sort out for hacknight

PR #2162, opened 2021-04-03
Config Validations

### LINKED ISSUE #1888: Extend Config Model with Validations

elimbaum:

Add validations for config fields. This finishes, and relies heavily on, the work @lomky did in #2002.

* Validates config URLs, start of week, support hiding, and budget bar.
* Canonicalizes those fields where appropriate (URL format, capitalization)
* Tests the above.

* Fixes #1888

For reviewer:
* Tag this PR:


elimbaum:
uh oh, I'm in git hell

elimbaum:
ok, I hope that fixed things a little bit... squashed the mistaken commits all into one (fix typos). should be all ready!

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+
+      clean_f = CLEAN_PRE_VALIDATION[config_key.to_sym]
+
+      # no clean function, return
+      return if clean_f.nil?
+
+      method(clean_f).()
+    end
+
+    def clean_url
```
colinxfleming, line 114: Would this func make more sense in `UriService`?

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+      config_value['options'] = [options.last.capitalize]
+    end
+
+    ### Validation
+
+    # parent function. will handle errors; child validators should return true
+    # if value is valid for key, and false otherwise.
+    def validate_config
+      # don't try to validate if no key or no value
+      return if config_key.nil? || options.last.nil?
```
colinxfleming, line 146: Two things:
* when is `config_key` null? Is that something we have to look out for or are you playing it safe? If the latter, would love to take it out and just do `options.last.blank?`
* let's use `.blank?` for consistency - it's a bit more aggro about whitespace which is good here

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
   end
+
+  private
+    ### Pre-validation cleanup
+
+    def clean_config_value
+      # do nothing if empty
+      return if config_key.nil? || options.last.nil?
+
+      clean_f = CLEAN_PRE_VALIDATION[config_key.to_sym]
```
colinxfleming, line 106: The way you have this structured is really elegant - the function passing is really good here, nicely done

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+
+      # no clean function, return
+      return if clean_f.nil?
+
+      method(clean_f).()
```
colinxfleming, line 111: today I learned! Usually I use `public_send` for stuff like this, but this works well too.

`call` might be a bit more readable than `.()`, but that's a style choice so choice is yours.

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+      url = options.last
+
+      # don't try to clean empty url
+      return if url.blank?
+
+      # don't have to do anything, already https
+      return if url.start_with? 'https://'
+
+      # convert http or // to https://
+      if url.start_with? /(http:)?\/\//
```
colinxfleming, line 124: Would this make more sense in `UriService`? This is quite a bit of very particular url handling, wonder if it makes sense to extract it into our service class.

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+
+    # parent function. will handle errors; child validators should return true
+    # if value is valid for key, and false otherwise.
+    def validate_config
+      # don't try to validate if no key or no value
+      return if config_key.nil? || options.last.nil?
+
+      val_f = VALIDATIONS[config_key.to_sym]
+
+      logger.info("VALIDATE_CONFIG #{config_key}: #{config_value} func #{val_f}")
```
colinxfleming, line 150: Let's remove the logger here

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
...
+      # no validation for this field, ignore
+      return if val_f.nil?
+
+      # run the validator and get a boolean, exit if true
+      return if method(val_f).()
+
+      errors.add(:invalid_value_for, "#{config_key.humanize(capitalize: false)}: '#{options.last}'")
+    end
+
+    START_OF_WEEK = [
```
colinxfleming, line 161: Can we put all the constants up together? Much easier to find/read that way, and follows other model patterns we have.

app/views/layouts/_footer.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="row">
 	<div class="col text-center">
-		<small><%= t 'navigation.footer.please_send_feedback' %> <%= link_to 'DCAF.CMapp@gmail.com', mailto: 'DCAF.CMapp@gmail.com' %>.</small>
+		<small><%= t 'navigation.footer.please_send_feedback' %> <%= link_to 'DCAF.CMapp@gmail.com', 'mailto:DCAF.CMapp@gmail.com' %>.</small>
```
colinxfleming, line 5: lol

test/services/uri_service_test.rb
```diff
@@ -2,10 +2,12 @@
 
 class UriServiceTest < ActiveSupport::TestCase
   describe 'URI service utility' do
-    it 'instantiates a uri from string' do
-      uri = UriService.new("some yolo test uri").uri
-      assert uri.is_a?(URI)
-    end
+    ## this is an invalid URL
+    # it 'instantiates a uri from string' do
```
colinxfleming, line 6: What do you propose we do with this test? Is it no longer needed and deleteable, should we re-work it, or?

test/models/config_test.rb
```diff
@@ -19,6 +44,27 @@ class ConfigTest < ActiveSupport::TestCase
       refute @bad_config.valid?
       assert @bad_config.errors.messages[:config_key].include? "can't be blank"
     end
+
+    it 'should validate URLs' do
+      c = Config.find_or_create_by(config_key: 'fax_service')
```
colinxfleming, line 49: Can we do this for `practical_support_guidance_url` as well? Looks like `resource_url` and `fax_service` are both covered.

A fun experiment if you're feeling wild - test all three using a loop.

test/models/config_test.rb
```diff
@@ -3,6 +3,31 @@
 class ConfigTest < ActiveSupport::TestCase
   before { @config = create :config }
 
+  describe 'callbacks' do
+    it 'should clean URLs before save' do
```
colinxfleming, line 7: can we check the `fix_capitalization` cleans too?

test/system/updating_configs_test.rb
```diff
@@ -169,6 +169,31 @@ class UpdatingConfigsTest < ApplicationSystemTestCase
           assert has_content? 'https://metallicarules.com'
         end
       end
+
+      it 'should reject invalid values' do
```
colinxfleming, line 173: great work adding the system test here -- really appreciate this

colinxfleming (commented):
This is a very thoughtful approach to a complicated code structuring problem -- very good work! I have a couple of suggestions and notes I'd like to take care of before merging, but the core idea really nails it.

colinxfleming:
hmm, there might be some crosstalk here with #2157  - a bunch of the same stuff is in both, and also a mailto change that got merged. I'm going to extract a couple comments - let's make sure not to merge this until we merge #2157 and rebase. I'll port over my relevant comments now.

app/models/config.rb
elimbaum: IIRC I added this because some of the more pathological tests attempt to pass in a `nil` key. From what I understand of the system this would never happen in real life unless you, like, were messing around with developer tools and manually sending POST requests.

app/models/config.rb
elimbaum: this was actually @lomky 's idea from #2002 :)

app/views/layouts/_footer.html.erb
elimbaum: I think this is an artifact of my brief foray into git hell... some rebase + fast forwarding issues; should go away when I remerge main.

lomky (commented):
Oh no I forgot to hit submit on this yesterday! Here's some thoughts in addition to Colin's

app/models/config.rb
```diff
@@ -70,4 +95,100 @@ def self.start_day
     start ||= "monday"
     start.downcase.to_sym
   end
+
+  private
```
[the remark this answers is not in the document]
elimbaum: My only concern here is that there isn't a 1:1 mapping of cleaners/validators to fields. For example, `start of week` and `hide practical support` both use `fix_capitalization` as their cleaner, and all the URLs share cleaners and validators. If, in the future, we need more granular control, we can always create field-specific methods and then this may make more sense? Or are you thinking: move all the generic functions up top, and then the field-specific?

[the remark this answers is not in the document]
lomky: yeah, the latter!

elimbaum:
Ready for review. One note - I'm readding the `.keep` files in `tmp` because it looks like we accidentally deleted those last week! Offending commit is ea3a40283 - IIRC this is a fat-finger goof on my part from trying to help Kory fix webpack...

colinxfleming (commented):
The engine here is good and works as expected! I have a little tiny bit of hardening but not a ton of feedback otherwise. Great job here.

I have a slightly question about URL handling - I tried plugging `100` into a URL and it looked like it prepended https in front. Is this the expected behavior here? I think I'm good with it, since I think generally people will be good about this and config-set broken urls aren't really bad, but wanted to make sure.

Input:
image

After save:
image

lomky:
@colinxfleming we [spotted & chatted about the overly-generous URL parsing in the previous PR](https://github.com/DCAFEngineering/dcaf_case_management/pull/2157#issuecomment-822893518) and decided to let it through for now since its strictly better than the existing situation (crashing the app). But I think we then forgot to file a follow up issue!

app/controllers/configs_controller.rb
```diff
@@ -14,7 +14,8 @@ def update
       flash[:notice] = t('flash.config_update_success')
       redirect_to configs_path
     else
-      flash[:danger] = t('flash.config_failed_update', error: @config.errors.full_messages.to_sentence)
+      error_content = @config.errors.full_messages.to_sentence
+      flash[:alert] = t('flash.config_failed_update', error: error_content) unless flash[:alert]
```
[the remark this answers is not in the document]
elimbaum: oooh very cool!

app/models/config.rb
```diff
@@ -33,12 +33,29 @@ class Config < ApplicationRecord
...
   # which fields are URLs (run special validation only on those)
-  CONFIG_URLS = %w[fax_service practical_support_guidance_url resources_url]
 
-  # Validations
-  validates :config_key, uniqueness: true, presence: true
+  # symbols are required here because functions are not objects in rails :)
+  CLEAN_PRE_VALIDATION = {
+    start_of_week: :fix_capitalization,
+    hide_practical_support: :fix_capitalization,
+  }
```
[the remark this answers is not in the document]
elimbaum: awesome, didn't realize maps could be frozen too!

elimbaum:
Thanks for the review! Changes committed.

On the "weird URL" front... that's actually expected behavior. As @lomky and I discussed earlier, URL edge cases are _super_ weird, and if you're following the spec to the letter (as the `URI` class does, so therefore as we are doing here), you run into some unexpected results.

@colinxfleming, in this case, specifically, [https://100](https://100) is valid, because `100` is an alias for the IP address `0.0.0.100` (zeros are actually optional in IPv4...).

I agree this isn't ideal behavior, so if we want to only allow a subset of URLs down the line we could definitely do that (i.e. no IPs; domain names with TLD only, etc).

test/models/config_test.rb
```diff
@@ -5,20 +5,34 @@ class ConfigTest < ActiveSupport::TestCase
...
-        c.config_value = { options: ["www.needs-scheme.net"] }
-        c.save
-        assert_equal "https://www.needs-scheme.net", c.options.last
-
-        # doesn't touch empty config
-        c.config_value = { options: [] }
-        c.save
-        assert_nil c.options.last
+      Config::CLEAN_PRE_VALIDATION
+        .select{ |field, cleaner| cleaner == :clean_url }
```
lomky, line 9: Since you refactored out the clean step for URLs there are no cases where the cleaner is `:clean_url` anymore. I think this test is delete-able!

test/models/config_test.rb
```diff
@@ -39,20 +53,22 @@ class ConfigTest < ActiveSupport::TestCase
...
-        # bad
-        c.config_value = { options: ["bad url"] }
-        refute c.valid?
-
-        # allow no URL
-        c.config_value = { options: [] }
-        assert c.valid?
+      Config::CLEAN_PRE_VALIDATION
+        .select{ |field, cleaner| cleaner == :clean_url }
+        .each do |url_field, cleaner|
```
lomky, line 58: Also not running! I think you actually want:

```suggestion
      Config::VALIDATIONS
        .select{ |field, validator| validator == :validate_url }
        .each do |url_field, _validator|
```

(the `_` on `_validator` is us acknowleding we aren't using this variable!)

lomky (commented):
Okay _one_ last thing and then I think we can merge this in and celebrate!

lomky (approved):
😤 🎉 :shipit:

PR #2163, opened 2021-04-03
PG: Migrate user model

colinxfleming:

Ports most of the moving pieces around `User` model to pg. (There's some stuff around call lists that needs to wait, yet.)

* PG-ifies `User`
* re-sets up `devise`

The following acid tests work, more or less:
* `rails db:drop db:create db:migrate db:seed` works (at least until patient creation, which is okay)
* server spins up
* this breaks the call list dashboard, since call list/patient are still mongo, so part doesn't work yet. Going to free myself from being too invested in that
* users views at http://localhost:3000/users work though! http://localhost:3000/users/3/edit also works

* Bumps #2072 

For reviewer:
* Tag this PR:


colinxfleming:
ps if you want to run the migration so far, here's a snippet:

git checkout main && rails db:drop db:create db:migrate db:seed && git checkout pg-user && rails db:migrate migrate_to_pg:clinic_user_patient

That will set up the app according to main including a seed, run the new migrations, and then try to migrate existing pg objects over. At that point you can rails s and check localhost:3000/clinics and localhost:3000/users to make sure things ported over accurately.

config/initializers/devise.rb
```diff
@@ -142,17 +178,17 @@
 
   # ==> Configuration for :validatable
   # Range for password length.
-  config.password_length = 8..72
+  config.password_length = 6..128
```
lomky, line 181: 1) how about we bump this to 8, or better 12?
2) maaaybe this is a good excuse to add that strong passwords into this process, to take advantage of asking folks to change it?

lomky (approved):
🎉

colinxfleming (commented):
woof definitely made some comments and then didn't hit submit! here they are for posterity

config/initializers/devise.rb
colinxfleming: per discussion, using this as an opportunity to add strong password changes is a wonderful idea, and we should absolutely add that to the feature branch. I'lll bump back to 8 for now.

```suggestion
  config.password_length = 8..128
```

PR #2171, opened 2021-04-15
PG: Port fulfillment model

colinxfleming:

Port Fulfillment to pg. This should let us cleanly init patient objects again.

Try out the migration script:
```
git checkout pg-fulfillment && rails db:migrate
rails migrate_to_pg:clinic_user_patient
rails c
Fulfillment.where(fulfilled: true).count
```

* `Patient` model to pg

no view change

* Bumps #2072

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/mongo_fulfillment.rb
+++ b/app/models/mongo_fulfillment.rb
@@ -1,6 +1,6 @@
 # Indicator that a pledge by the primary fund was cashed in,
 # which in turn indicates that the patient used our pledged money.
-class Fulfillment
+class MongoFulfillment
   include Mongoid::Document
   include Mongoid::Timestamps
   include Mongoid::History::Trackable
--- a/app/models/mongo_fulfillment.rb
+++ b/app/models/mongo_fulfillment.rb
@@ -1,0 +1,36 @@
+# Indicator that a pledge by the primary fund was cashed in,
+# which in turn indicates that the patient used our pledged money.
+class Fulfillment
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::History::Trackable
+  include Mongoid::Userstamp
+
+  # Relationships
+  embedded_in :can_fulfill, polymorphic: true
+
+  field :fulfilled, type: Boolean
+  field :procedure_date, type: Date
+  field :gestation_at_procedure, type: String
+  field :fund_payout, type: Integer
+  field :check_number, type: String
+  field :date_of_check, type: Date
+  field :audited, type: Boolean
+
+  # Validations
+  validates :created_by_id,
+            presence: true
+
+  # History and auditing
+  track_history on: fields.keys + [:updated_by_id],
+                version_field: :version,
+                track_create: true,
+                track_update: true,
+                track_destroy: true
+  mongoid_userstamp user_model: 'User'
+
+  # Methods
+  def gestation_at_procedure_display
+    I18n.t('accountants.table_content.weeks_at_procedure_display', gestation: gestation_at_procedure) if gestation_at_procedure.present?
+  end
+end
--- a/app/views/patients/_fulfillment.html.erb
+++ b/app/views/patients/_fulfillment.html.erb
@@ -16,7 +16,7 @@
                           remote: true do |f| %>
     <div class="row">
       <div class="col">
-        <%= f.fields_for patient.fulfillment do |pt| %>
+        <%= f.fields_for :fulfillment do |pt| %>
           <%= pt.form_group :fulfilled do %>
             <%= pt.check_box :fulfilled, label: t('patient.pledge_fulfillment.form.pledge_fulfilled_box'), autocomplete: 'off' %>
           <% end %>
@@ -26,7 +26,7 @@
 
     <div class="row">
       <div class="info-form-left col-6">
-        <%= f.fields_for patient.fulfillment do |pt| %>
+        <%= f.fields_for :fulfillment do |pt| %>
           <%= pt.number_field :fund_payout,
                               label: t('patient.pledge_fulfillment.form.fund_payout', fund: "#{FUND}"),
                               autocomplete: 'off',
@@ -42,7 +42,7 @@
       </div>
 
       <div class="info-form-right col-6">
-        <%= f.fields_for patient.fulfillment do |pt| %>
+        <%= f.fields_for :fulfillment do |pt| %>
           <%= pt.text_field :check_number, label: t('patient.pledge_fulfillment.form.check_num'), autocomplete: 'off'%>
           <%= pt.date_field :date_of_check, label: t('patient.pledge_fulfillment.form.date_of_check'), autocomplete: 'off' %>
         <% end %>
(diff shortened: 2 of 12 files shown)


app/models/mongo_fulfillment.rb
```diff
@@ -0,0 +1,36 @@
+# Indicator that a pledge by the primary fund was cashed in,
+# which in turn indicates that the patient used our pledged money.
+class Fulfillment
+  include Mongoid::Document
+  include Mongoid::Timestamps
+  include Mongoid::History::Trackable
+  include Mongoid::Userstamp
+
+  # Relationships
+  embedded_in :can_fulfill, polymorphic: true
```
lomky, line 10: All of our previous ports added a line like:

```
  store_in collection: 'patients'
```

I'm assuming this preexisting relationship is why we don't need this here?

app/views/patients/_fulfillment.html.erb
```diff
@@ -16,7 +16,7 @@
                           remote: true do |f| %>
     <div class="row">
       <div class="col">
-        <%= f.fields_for patient.fulfillment do |pt| %>
```
lomky, line 19: do we expect these `patient.fulfillment` to be replaced everywhere after this PR? If so there are a few hanger-ons in the code.

db/migrate/20210415015227_create_fulfillment.rb
```diff
@@ -0,0 +1,20 @@
+class CreateFulfillment < ActiveRecord::Migration[6.0]
+  def change
+    create_table :fulfillments do |t|
+      t.boolean :fulfilled, null: false, default: false
```
lomky, line 4: So this was nullable before, but the default takes care of it now... yeah, that makes sense!

lib/tasks/migrate_to_pg.rake
```diff
@@ -92,3 +107,24 @@ def migrate_model(pg_model, mongo_model, transform = nil)
 
   puts "#{pg_model.count} #{pg_model} migrated to pg"
 end
+
+def migrate_fulfillment(pt_model, mongo_pt_model, pg_model, mongo_model, relation, parent_relation, transform)
```
lomky, line 111: is this in prep for reuse on archive patient?

lib/tasks/migrate_to_pg.rake
```diff
@@ -92,3 +107,24 @@ def migrate_model(pg_model, mongo_model, transform = nil)
...
+
+    pg_attrs = transform.call(mongo_obj.slice(*attributes), mongo_obj, doc)
+    pg_obj = pg_model.create! pg_attrs
+    if mongo_obj['created_by_id'].present?
+      pg_obj.versions.first.update whodunnit: User.find_by(mongo_id: mongo_obj['created_by_id'].to_s)&.id
+    end
+
+    # Check
+    if pg_model.count != mongo_model.count
+      'PG and mongo counts are in disagreement, aborting'
```
lomky, line 124: Is this rake task magic?

app/models/mongo_fulfillment.rb
```diff
@@ -0,0 +1,36 @@
+# Indicator that a pledge by the primary fund was cashed in,
+# which in turn indicates that the patient used our pledged money.
+class Fulfillment
```
lomky, line 3: Throwing an error on `rails migrate_to_pg:clinic_user_patient` because of this!

```suggestion
class MongoFulfillment
```

lomky (commented):
One MustFix (mongo model name) and a few questions!

Also what should the output of `rails> Fulfillment.where(fulfilled: true).count` be? I get `6`

lib/tasks/migrate_to_pg.rake
colinxfleming: yep! Because fulfillment is 1-1 rather than 1-many, we know we're always porting over one object, so we write a specialized handler here that doesn't iterate.

Side note,  when we're all done with this we should consider just slamming the fields onto the patient object rather than dealing with them separately. The separate fulfillment object feels increasingly silly. Not something we

lib/tasks/migrate_to_pg.rake
colinxfleming: nah, `pg_model` and `mongo_model` should be the classes in question. `pg_model = Fulfillment` and `mongo_model = MongoFulfillment` for ex.

db/migrate/20210415015227_create_fulfillment.rb
colinxfleming: exactly right! Nice and minor, but a quality of life improvement.

app/models/mongo_fulfillment.rb
colinxfleming: exactly right - the document store knows to save this as a subobj within patient objects in the appropriate collection, so no need to declare.

colinxfleming:
`6` is right, yeah! Showing my work a bit more:

```
$ git checkout main && rails db:drop db:create db:migrate db:seed
$ echo "Patient.where('fulfillment.fulfilled' => true).count" | rails c # mongo

6

$ git checkout pg-fulfillment && rails db:migrate migrate_to_pg:clinic_user_patient
$ echo "Fulfillment.where(fulfilled: true).count" | rails c # pg

Fulfillment.where(fulfilled: true).count
   (1.6ms)  SELECT COUNT(*) FROM "fulfillments" WHERE "fulfillments"."fulfilled" = $1  [["fulfilled", true]]
6
```

lib/tasks/migrate_to_pg.rake
lomky: I meant line 124 just being a string!

colinxfleming: lol my mistake, sorry! I assume you're asking whether or not this should `puts` and it def should. thank you!

colinxfleming:
```suggestion
      raise 'PG and mongo counts are in disagreement, aborting'
```

colinxfleming: (puts, raise, whatever.)

lib/tasks/migrate_to_pg.rake
```diff
@@ -92,3 +107,24 @@ def migrate_model(pg_model, mongo_model, transform = nil)
...
+    if mongo_obj['created_by_id'].present?
+      pg_obj.versions.first.update whodunnit: User.find_by(mongo_id: mongo_obj['created_by_id'].to_s)&.id
+    end
+
+    # Check
+    if pg_model.count != mongo_model.count
+      raise 'PG and mongo counts are in disagreement, aborting'
+    end
+    if pg_model.count != Patient.count
+      'Every patient not associated with only one fulfillment'
```
lomky, line 127: Same here then!
```suggestion
      raise 'Every patient not associated with only one fulfillment'
```

lib/tasks/migrate_to_pg.rake
lomky: not something we?

I don't have strong feelings on subobj vs embedded!

colinxfleming: Pull request experimental fiction 😂 

Not something we have to resolve in the scope of this of course. But when I was working on it I was like 'this reminds me of the original model mistake we made, trying to shoehorn pregnancy data into a separate object'. (though fulfillment is a much cleaner separation and a lot less conceptually hard.)

app/views/patients/_fulfillment.html.erb
colinxfleming: nah - it stops working in the form helpers, but actual data is still accessed via `patient.fulfillment.whatev`.

lomky (approved):
🎉

colinxfleming:
:)

PR #2176, opened 2021-04-22
Upgrade webpack-dev-server, which also upgrades sockjs

xmunoz:

* Title says it all

* Dependabot alert


colinxfleming (approved):
sockjs eh

PR #2177, opened 2021-04-23
Kory Welcome Message

KoryJCampbell:
issue #884

Added welcome message from Kory


app/helpers/welcome_message_helper.rb
```diff
@@ -63,6 +63,7 @@ def welcome_messages
       "You are appreciated, you are loved, your work makes a differnces! Thank you for all you do! 👏👏👏",
       "Thank you for the critical, wonderful work you do. You're awesome!",
       "Keep on crushing it! You are awesome."
```
colinxfleming, line 65:
```suggestion
      "Keep on crushing it! You are awesome.",
```

colinxfleming (approved):
welcome aboard @KoryJCampbell ! Thanks much for your contribution and very glad to have you.

Elements in arrays need to be separated by commas, but NBD - easy fix that I implemented.

Not quite sure why there are diffs in the .keep files (which are empty) but not worrying about it.

Approving and merging.

PR #2199, opened 2021-05-11
Migrate external pledge to pg

colinxfleming:

Migrate the ExternalPledge model to postgres.

The patient edit page isn't _quite_ up to speed with this - looks like there are a few small changes we'll need to make in addition. I'll leave those for a separate PR to keep this one reasonably scoped.

To test:

```
git checkout pg-extp && rails db:migrate migrate_to_pg:clinic_user_patient
echo "ExternalPledge.count" | rails c # 2
```

* port ExternalPledge to postgres

no view changes

* Bumps #2072 

For reviewer:
* Tag this PR:


lomky (approved):
No questions!

PR #2201, opened 2021-05-15
Make pg-related edits to get views to identically render

colinxfleming:

This PR aims to get the views a) all loading and b) in agreement with a fresh seed. (spoiler: it does)

After this, we'll be doing archived patient and tests, and then hopefully we're in the clear.

Note that generally we are relying on system tests to confirm that behavior is generally the same, which aren't ported yet, so I think you can generally take these code changes at more or less face value; we'll do a more involved click-around review when the feature branch is really ready to go I think.

* mostly small place changes to get stuff to properly render
* a slightly more involved change to changelog to accommodate timestamps coming out of the database differently

* Bumps #2072 

For reviewer:
* Tag this PR:


colinxfleming:
Views _mostly_ identically render now, with three exceptions (one of which doesn't really count yet):
* changelog - mongoid and created_at are showing up
* changelog - fund_pledged_at is showing up with full 
* export - errors, but I think it's because archived patient isn't ported, so not dealing with that yet

I'll fix the first two and then punt this up.

colinxfleming:
(There are a couple minor differences like ordering differences but I think that's more because of migrate_to_pg than it is actual logic changes.)

app/controllers/accountants_controller.rb
```diff
@@ -37,8 +37,8 @@ def paginate_results(results)
 
   def pledged_patients
     Patient.where(pledge_sent: true,
-                  :initial_call_date.gte => 6.months.ago)
```
lomky, line 40: I will never miss trying to wrap my head around gte/lte date math 🎉

lomky (approved):
image

PR #2207, opened 2021-05-20
Adds a clinic's city & state to the clinic dropdown

### LINKED ISSUE #2188: Add city/state to the clinics dropdown in patient edit view

If you go to a patient edit page, here's how the clinic dropdown renders:

image

Those are just the names of the clinics, and they're a little tough to read on the fly - where is `Sample Clinic with NAF`? Since geography matters quite a bit, I think it would be really nice to add city/state to the dropdown. So, that screenshot should change to something like this:

```
Sample Clinic 1 - DC (Washington, DC)
Sample Clinic 2 - VA (Arlington, VA)
Sample Clinic with NAF (Washington, DC)
```
etc. 

A nice quality of life improvement for CMs.

Take a look at `clinic_options` in `patients_helper.rb`. You might have to muck with i18n a little bit, but changing around the clinic name a bit is probably the way to go.

* **Anything else?**
Note that `Sample Clinic 1 - DC` and `Sample Clinic 2 - VA` are the names of the clinic in the seedfile - we definitely are not doing this right now!

nsiwnf:

* Adds clinic city, state to the drop-down in "Patient/edit" > "Abortion Information"

* Resolves #2188

### DIFF
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -61,7 +61,7 @@ module PatientsHelper
       [ t('patient.helper.referred_by.sexual_assault_crisis_org'),    'Sexual assault crisis org' ],
       [ t('patient.helper.referred_by.youth'),                        'Youth outreach' ], ]
     full_set = standard_options + Config.find_or_create_by(config_key: 'referred_by').options
-    
+
     options_plus_current(full_set, current_value)
   end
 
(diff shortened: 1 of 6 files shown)


lomky:
before and after of clinic dropdown

lomky (approved):
Wonderful work! Thank you! 🎉 🥇

app/helpers/patients_helper.rb
```diff
@@ -112,7 +112,7 @@ def clinic_options
     clinics = Clinic.all.sort_by(&:name)
     active_clinics = clinics.select(&:active)
                             .map { |clinic| [
-                              clinic.name,
+                              t('patient.abortion_information.clinic_section.clinic_display', clinic_name: clinic.name, city: clinic.city, state: clinic.state),
```
colinxfleming, line 115: love your inclination to i18n this 💯

lomky:
@all-contributors please add @nsiwnf  for code!

PR #2216, opened 2021-05-28
Get all unit tests working on pg + improve with_versioning

colinxfleming:

A couple of app changes, but mostly unit test fixes to get that back to green.

Sorry for the large file diff; I know anything over 15 files tends to make eyes glaze over. I've tried to leave comments where possible.

* some minor fixes to the app code - mistakes I made in translating things over from the original omnibus, or mistakes in rebasing
* improve the with_versioning helper to reduce its overhead
* `rails t` (but not the system tests) should return green now

Just for safety's sake you can verify stuff works by running the standard migration commands. I wouldn't fret about the exact numbers here:

```
git checkout pg-units && rails db:migrate migrate_to_pg:clinic_user_patient
```

* Bumps #2072 

For reviewer:
* Tag this PR:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,6 @@ gem 'enumerize' # Mongoid doesn't have enum out of the box, so we get it here
 # ...but hopefully soon it will be postgres
 gem 'pg', '~> 1.2'
 gem 'paper_trail', '~> 10.3'
-gem 'paper_trail-globalid'
 
 # Our authentication library is devise, with oauth2 for google signin
 gem 'devise', '~> 4.8'
--- a/app/lib/paper_trail_version.rb
+++ b/app/lib/paper_trail_version.rb
@@ -46,11 +46,11 @@ class PaperTrailVersion < PaperTrail::Version
              end
   end
 
-  private
-
   def marked_urgent?
-    object_changes['marked_urgent']&.last == true
+    object_changes['urgent_flag']&.last == true
   end
+
+  private
 
   def format_fieldchange(key, value)
     shaped_value = if value.blank?
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -94,7 +94,7 @@ class ArchivedPatient < ApplicationRecord
 
     )
 
-    archived_patient.build_fulfillment(dup_with_context(patient.fulfillment)).save
+    archived_patient.build_fulfillment(patient.fulfillment.attributes.except('id')).save
     archived_patient.clinic_id = patient.clinic_id if patient.clinic_id
 
     PaperTrail.request(whodunnit: patient.created_by_id) do
@@ -102,7 +102,7 @@ class ArchivedPatient < ApplicationRecord
     end
 
     patient.calls.each do |call|
-      calls.update can_call: archived_patient
+      call.update can_call: archived_patient
     end
     patient.external_pledges.each do |ext_pledge|
       ext_pledge.update can_pledge: archived_patient
--- a/app/models/concerns/paper_trailable.rb
+++ b/app/models/concerns/paper_trailable.rb
@@ -8,7 +8,7 @@ module PaperTrailable
   end
 
   def created_by
-    versions.first&.user
+    versions.last&.user
   end
 
   def created_by_id
@@ -16,7 +16,7 @@ module PaperTrailable
   end
 
   def updated_by
-    versions.last&.user
+    versions.first&.user
   end
 
   def updated_by_id
--- a/app/models/concerns/urgency.rb
+++ b/app/models/concerns/urgency.rb
@@ -4,11 +4,9 @@ module Urgency
 
   def still_urgent?
     # Verify that a pregnancy has not been marked urgent in the past six days
-    return false if recent_history_tracks.count.zero?
     return false if pledge_sent || resolved_without_fund
-    recent_history_tracks.sort.reverse.each do |history|
-      return true if history.marked_urgent?
-    end
+    recent = versions.where('created_at > ?', 6.days.ago)
+    return true if recent.any?(&:marked_urgent?)
     false
   end
 
@@ -28,11 +26,5 @@ module Urgency
       Patient.where(urgent_flag: true)
              .each(&:confirm_still_urgent)
     end
-  end
-
-  private
-
-  def recent_history_tracks
-    history_tracks.select { |ht| ht.updated_at > 6.days.ago }
   end
 end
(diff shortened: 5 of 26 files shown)


Gemfile
```diff
@@ -31,7 +31,6 @@ gem 'enumerize' # Mongoid doesn't have enum out of the box, so we get it here
 # ...but hopefully soon it will be postgres
 gem 'pg', '~> 1.2'
 gem 'paper_trail', '~> 10.3'
-gem 'paper_trail-globalid'
```
colinxfleming, line 34: We don't actually need this; we rolled our own created_by stuff and set up a relation of `version.user`, so we can remove this from the gemfile.

app/lib/paper_trail_version.rb
```diff
@@ -46,12 +46,12 @@ def shaped_changes
              end
   end
 
-  private
-
   def marked_urgent?
-    object_changes['marked_urgent']&.last == true
+    object_changes['urgent_flag']&.last == true
```
colinxfleming, line 50: oops

app/models/archived_patient.rb
```diff
@@ -94,15 +94,15 @@ def self.convert_patient(patient)
 
     )
 
-    archived_patient.build_fulfillment(dup_with_context(patient.fulfillment)).save
+    archived_patient.build_fulfillment(patient.fulfillment.attributes.except('id')).save
```
colinxfleming, line 97: I wrote a custom `dup_with_context` thing in the original draft of this (and then didn't port it over into the archived patient pr 🤦 ) but this basically does the same thing, so might as well not delegate it to its own method.

app/models/concerns/paper_trailable.rb
```diff
@@ -8,15 +8,15 @@ module PaperTrailable
   end
 
   def created_by
-    versions.first&.user
+    versions.last&.user
```
colinxfleming, line 11: I cannot tell the difference between asc and desc, which is embarrassing. The scope orders these by id desc, so the most recent id is the most recent change; created by would thereby be the first.

app/models/concerns/urgency.rb
```diff
@@ -4,11 +4,9 @@ module Urgency
 
   def still_urgent?
     # Verify that a pregnancy has not been marked urgent in the past six days
-    return false if recent_history_tracks.count.zero?
     return false if pledge_sent || resolved_without_fund
-    recent_history_tracks.sort.reverse.each do |history|
-      return true if history.marked_urgent?
-    end
+    recent = versions.where('created_at > ?', 6.days.ago)
```
colinxfleming, line 8: Can simplify this a bit now that we're in pg!

test/controllers/calls_controller_test.rb
```diff
@@ -58,10 +58,8 @@ class CallsControllerTest < ActionDispatch::IntegrationTest
 
   describe 'destroy method' do
     it 'should destroy a call' do
-      with_versioning do
-        PaperTrail.request(whodunnit: @user.id) do
-          @patient.calls.create attributes_for(:call)
-        end
+      with_versioning(@user) do
```
colinxfleming, line 61: A bunch of the diffs in here are changing `with_versioning { PaperTrail.request(whodunnit: ...) }` to just `with_versioning(@user)`. I made a minor change to the `with_versioning` helper to decrease lines of code to write, and to delegate thinking about how to handle the user object to the test helper rather than the programmer. (I ran into all sorts of problems and weird behavior around when to close blocks, use @user vs @user.id, etc, and hopefully this cuts down on the headache for us.)

test/models/audit_trail_test.rb
```diff
@@ -1,66 +1,66 @@
-require 'test_helper'
+# require 'test_helper'
```
colinxfleming, line 1: mongo-related, so I feel comfortable just commenting out until we remove these files.

colinxfleming (commented):
couple of notes

app/lib/paper_trail_version.rb
```diff
@@ -46,12 +46,12 @@ def shaped_changes
              end
   end
 
-  private
-
   def marked_urgent?
```
lomky, line 49: so we define this function twice - once here and once in the audit_trail model.  IIRC the audit_trail is mongodb stuff? I see you commented out its tests below - should we comment out the file as well, or in some other way check that instance isn't being invoked?

audit_trail
https://github.com/DCAFEngineering/dcaf_case_management/blob/c5811041293e634bbdcb560311a9c464e3b749d1/app/models/audit_trail.rb#L24-L26

Existing Invocation
https://github.com/DCAFEngineering/dcaf_case_management/blob/c5811041293e634bbdcb560311a9c464e3b749d1/app/models/concerns/urgency.rb#L10

app/models/archived_patient.rb
lomky: wow I should have caught that 😅

app/models/user.rb
```diff
@@ -35,7 +35,7 @@ class User < ApplicationRecord
   # email presence validated through Devise
   validates :name, :role, presence: true
   validates_format_of :email, with: Devise::email_regexp
-  validate :secure_password
```
lomky, line 38: This is already the case on `main`, so I'm a bit sus of this diff? do you need to pull in `main`?

 https://github.com/DCAFEngineering/dcaf_case_management/blob/c5811041293e634bbdcb560311a9c464e3b749d1/app/models/user.rb#L85

test/controllers/calls_controller_test.rb
lomky: nice!

app/lib/paper_trail_version.rb
colinxfleming: yeah, the audit trail model is dead code. We have to keep it because builds get wonky without it, but it shouldn't actually contain anything that gets invoked anymore since we've switched all the models over as of the archivedpatient branch. Commenting out the contents of the class is a good idea though, so I'll do that!

app/models/user.rb
colinxfleming: super plausible there's some rebasing out of sync. I'll check on this.

app/models/archived_patient.rb
colinxfleming: no harm no foul, usually that's the job of tests!

app/models/user.rb
colinxfleming: This is definitely a case of me accidentally screwing up a rebase conflict. See https://github.com/DCAFEngineering/dcaf_case_management/pull/2143/commits/58d30ecb15404c74603964054d6c6007b4e66881 for my exact mistake on the pg-shell branch. I think it's probably fine to leave this diff as atonement for my prior error but interested in your opinion.

lomky: 👍🏼 I don't spy any others.

lomky (approved):
I get some wonky behavior on this branch, but the system test PR resolves them so GTG!

PR #2218, opened 2021-05-31
Accountant dashboard - clinic filtering

### LINKED ISSUE #2214: Add clinic filter system to accountants view

I would like to add a clinic filter to the accounting workflow form, in addition to the existing patients filter.

When funds are cutting checks and reconciling, they're generally doing that by clinic; individual initials lookup is helpful but they're generally looking for like 10 patients who all went to a given clinic.

Here's what I think:

* add a dropdown to the search form that has the options 'All clinics' (default) or all individual clinics.
* If an individual clinic is selected, limit search results to just that clinic. (I don't _think_ this is super painful in either pg or mongo - hopefully just an additional `where` condition? if the field is set?) 

* **Anything else?**

Leaving the implementation purposefully a little vague for purposes of good exercise; please do ask me any questions if I've left anything out, as that is also part of the exercise!

The origin ticket here made reference to changing the sort. Let's not do that; a single-clinic filter will accomplish what they want more cleanly.

elimbaum:

Adds a clinic filter dropdown to the accountants dashboard. The default option, `[All Clinics]`, does not filter.

* Updates the view with a dropdown
* Changes the search logic to filter by name and/or clinic.
* Adds unit and system tests

*Searching for both phone number and clinic:*
image

*Dropdown options with test data:*
image

* Fixes #2214
* Bumps #Y

For reviewer:
* Tag this PR:


app/controllers/accountants_controller.rb
```diff
@@ -10,10 +10,22 @@ def index
...
+                  partial_result = Patient.where(pledge_sent: true)
+                                          .order_by(pledge_sent_at: :desc)
+
+                  if have_clinic
+                    partial_result = partial_result.where(clinic_id: params[:clinic_id])
+                  end
+
+                  if have_search
+                    partial_result = partial_result.search(params[:search])
+                  end
```
elimbaum, line 26: This is tickling my *"there must be a more idiomatic way"* spidey sense. Is there a method-call assignment operator? As in `x .= f()` to do `x = x.f()`. If this looks ok I'm fine keeping it though.

app/views/accountants/_search_form.html.erb
```diff
@@ -1,12 +1,21 @@
...
     <%= f.text_field :search,
                      placeholder: t('accountants.search_placeholder'),
                      hide_label: true,
-                     class: 'search_form_input accountant' %>
-    <%= f.submit t('common.search'), class: 'btn btn-primary ml-2' %>
+                     class: 'search_form_input accountant mr-4' %>
+
+    <%= f.select :clinic_id,
+                 options_for_select(clinic_options.reject{|x| x.blank?}),
+                 prompt: "[All Clinics]",
```
elimbaum, line 11: Adding a `prompt` adds an item to the dropdown with empty value. The above search code, then, sees a blank clinic search and ignores it for filtering purposes.

app/controllers/accountants_controller.rb
colinxfleming: If I were writing this (ymmv!) I'd probably do something like this, but it's basically the same approach you have going on:

```
result = Patient
result = result.where(clinic..) if params[:clinic_id].present?
result = result.search(...) if params[:search].present?
```
I'm not sure I've ever used a `.=` operator (I lead a somewhat charmed life), so not sure I can offer any advice on that!

app/views/accountants/_search_form.html.erb
colinxfleming: I had no idea rails had this! amazing

app/controllers/accountants_controller.rb
elimbaum: ok, that's cleaner. thanks!

app/views/accountants/_search_form.html.erb
elimbaum: I think is kinda an HTML thing!

elimbaum: Let me know if you're fine with that formatting etc. In particular, do I need to handle internationalization? Or does that happen in batches down the line?

lomky:
I think this one is just database-adjacent enough to hold until we get the PG epic released on Sat and we can compare - 🧷

elimbaum:
ok! are you concerned about possible controller changes?

colinxfleming:
@elimbaum do you want me to rebase this, or do you have it handled?

elimbaum:
I'll take care of it! Good git experience...

elimbaum:
pg merged & ready!

app/controllers/accountants_controller.rb
```diff
@@ -10,15 +10,22 @@ def index
   end
 
   def search
```
colinxfleming, line 12: Works as advertised! Very well done - the ActiveRecord chaining is perfect

app/views/accountants/_search_form.html.erb
```diff
@@ -1,12 +1,21 @@
...
     <%= f.text_field :search,
                      placeholder: t('accountants.search_placeholder'),
                      hide_label: true,
-                     class: 'search_form_input accountant' %>
-    <%= f.submit t('common.search'), class: 'btn btn-primary ml-2' %>
+                     class: 'search_form_input accountant mr-4' %>
+
+    <%= f.select :clinic_id,
+                 options_for_select(clinic_options.reject{|x| x.blank?}),
+                 prompt: t('accountants.search_all_clinics'),
```
colinxfleming, line 11: I think I said this elsewhere but I didn't know about prompt and I'm probably going to use it all the time now

test/controllers/accountants_controller_test.rb
```diff
@@ -117,6 +117,39 @@ class AccountantsControllerTest < ActionDispatch::IntegrationTest
...
+
+          create :patient, name: 'sally clinic2',
+                           appointment_date: 4.days.from_now,
+                           fund_pledge: 500,
+                           pledge_sent: true,
+                           pledge_sent_at: nil,
+                           clinic: @clinic2
+        end
+
+        it 'can filter by clinic' do
```
colinxfleming, line 141: Good choice here by delegating the heavy lifting to the system tests.

colinxfleming (approved):
Very well done on multiple fronts - the AR is good, the tests are good. Great work here.

PR #2219, opened 2021-06-01
Allow rubocop to run locally

elimbaum:

Minor changes to configs to allow `rubocop` to run locally.

After `bundle install`ing, run `rubocop [files...]` (or `docker-compose run --rm web rubocop [files...]`). With no arguments, checks everything, and there are a lot of suggestions! I think this will mostly be useful to look at *new* changes: you could use `git diff --name-only main` to get a list of ruby files changed on this branch, so

```
docker-compose run --rm web rubocop `git diff --name-only main *.rb`
```

I know we're not planning on doing a full re-lint of the codebase, but this might help, moving forward, to be able to keep things up to spec!

Open to comments/config changes/doc suggestions.

* Installs `rubocop-rails`.

For reviewer:
* Tag this PR:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,5 @@
+require: rubocop-rails
+
 Style/ClassAndModuleChildren:
   Enabled: false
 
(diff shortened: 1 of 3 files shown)


elimbaum:
Thanks @mdworken!!

lomky (approved):
Nice clean fix!

PR #2220, opened 2021-06-01
Update accountant information and UX

### LINKED ISSUE #2215: Change information displayed on accountants workflow

A fund has requested that we add some more information to the accountants workflow, to help ease their (currently very unpleasant through no fault of anyone's) accounting process. This is a ticket to adjust the UX a little and add a bit new information to the accounting patient roll.

In part, this is because a particular fund has a uniquely difficult situation around accounting - they aren't consistently getting remit lists, so they're having to draft them on their own.

I'd like to propose we do the following to accommodate:

* Make the display shell scrollable, to hold more information. (I believe there's a bootstrap class that does this. `scroll-overflow` maybe?)
* Ditch patient initials in favor of regular patient name.
* Add pledge amount and line.

* **Anything else?**

I would explicitly like @lomky 's thoughts here, since this is a slight change in our security posture - names would be exposed here in a spot it wasn't before, but still not bulk exportable (and click-thru-able). Not ideal, but this feels to me like a minor enough shift as to not really increase the risk level. I'd appreciate your gut check on this.

@elimbaum please also weigh in if you feel strongly; please don't start until @lomky 's had a chance to weigh in, in case we mix this up.

Screenshot of the current state of the world for ref:
image

elimbaum:

Added fund-requested fields to the table, and made some styling updates.

* Replace initials with name.
* Add pledge amount and line.
* Make table responsive and scrollable.
* Fix the results/pagination counter to be responsive with search results.

**todo**:
- confirm the SCSS updates look okay
- decide on behavior of pagination counter

*it scrolls! + updated table spacing -->*

Screen Shot 2021-06-01 at 16 41 14

* Fixes #2215

For reviewer:
* Tag this PR:


elimbaum:
IIRC, this is my first substantive change to the frontend, so feedback appreciated on JS/SCSS.

One thing I want to decide on is if we're ok with the behavior of the pagination counter. Currently, on page load, kaminari renders `Displaying all X patients`. Then, when you search, that will change to `Displaying all Y results` (assuming this all fits on one page).

I'm wondering if we want to do something more like `Displaying Y results (out of X patients total)`, just to clarify that `... all Y results` doesn't mean all patients. Also, if, after searching once, you un-search, it will still say `Displaying all Y results`. I can probably fix that, if you think it'd be good. [smth like `if # results == # patients, render regular`]

However, this is may not actually be that confusing, so it may be fine to just leave it up to kaminari to handle natively.

lomky:
before | after
--|--
main branch accountants page | PR branch accountants page

lomky:
On the counts - I think we're dealing with a high trust situation. I think its worth a decent amount of work for the accountants to feel comfortable and trust it. Do you think that would be better done in this PR or as a follow up?

lomky (commented):
Questions!

app/views/accountants/_table_content_row.html.erb
```diff
@@ -1,9 +1,11 @@
 <td><!-- spacer --></td>
```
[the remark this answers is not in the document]
elimbaum: Good idea!

app/views/accountants/_table_content.html.erb
```diff
@@ -1,12 +1,14 @@
-<table class="table border-side">
+<table class="table border-side table-responsive text-nowrap">
   <thead>
     <tr class="accountants-table-header">
       <th><!-- spacer --></th>
+      <th><%= t 'accountants.table_content.line' %></th>
```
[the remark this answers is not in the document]
elimbaum: Ha! I didn't know that was an edge case. Do you mean the actual row (vs. the header)?

elimbaum:
@counts: I feel like that's still UX, so I can include that in this PR! Basically, we have three cases:

- user has not searched; show all patients. Current language is `Displaying all X patients` using kaminari defaults.
- user has filtered; however, all patients are returned. Proposed language is (still) `Displaying all X patients`
- user has filtered. Current language is `Displaying all Y results`. Propose adding ` (out of X patients total)`

app/views/accountants/_table_content.html.erb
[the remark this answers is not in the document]
elimbaum: https://github.com/DCAFEngineering/dcaf_case_management/blob/3f9aba6e8e27bbdd2d37435a0f0eed95aa92eabc/config/initializers/_env_var_constants.rb#L2-L9

This suggests to me that all funds must have at least one line?

lomky:
(I think you mis-tagged? 😆)

 - Yup!
 - I like that proposed langauge
 - I prefer `(out of X patients total)` 
  
 @colinxfleming whatcha think?

app/views/accountants/_table_content.html.erb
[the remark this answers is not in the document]
lomky: So they do! Looks like we've configured folks who didn't need a line to just be `Main`.

elimbaum:
Also, question: the `index` and `search` methods use different queries. Is this intentional? Should they be rectified? It seems like they *should* match, so I'm wondering if we do or do not want the initial call date filter.

https://github.com/DCAFEngineering/dcaf_case_management/blob/6e0bae049b40baabe9e6c644a4dafdf1f51ea6fe/app/controllers/accountants_controller.rb#L14-L16

vs

https://github.com/DCAFEngineering/dcaf_case_management/blob/6e0bae049b40baabe9e6c644a4dafdf1f51ea6fe/app/controllers/accountants_controller.rb#L41-L43

elimbaum:
looks like this is all good! question above can be ignored - consolidated everything into using the same method.

- When searching, language looks something like `Displaying all 2 results (out of 17 patients total)`
- Edit button moved:

Screen Shot 2021-06-13 at 17 54 03

lomky (commented):
Something isn't quite working right in the search & the numbers showing. I populated a bunch more pledged patients, and I can only get it to return the result of 15 or none, and its out of 25 now, instead of the 47

image
image


And on the command line:
```ruby
irb(main):021:0> Patient.where("name LIKE '%submit%'").where(pledge_sent: true).where(initial_call_date: 6.months.ago..).count
=> 31
irb(main):022:1* Patient.where(pledge_sent: true,
irb(main):023:0> initial_call_date: 6.months.ago..).order(pledge_sent_at: :desc).search("submit").count
=> 15
```


<details><summary>quick ref, how to add more pledged patients, adapted from seedfile:</summary>

```ruby
> 10.times do |i|                                                                                                         
   patient = Patient.create! name: "Patient #{i}",                                                                    
     primary_phone: "723-123-123#{i}",                                                           
     initial_call_date: 3.days.ago,                                                              
     urgent_flag: i.even?,                                                                          
     last_menstrual_period_weeks: (i + 1 * 2),                                                      
     last_menstrual_period_days: 3,                                                                 
     line: 'DC'
  patient.update! clinic: Clinic.first,                                                                                 
   appointment_date: 3.days.from_now,                                                                                    
   naf_pledge: 200,                                                                                                      
   procedure_cost: 400,                                                                                                                              
   fund_pledge: 100,                                                                                                     
   pledge_sent: true,                                                                                                    
   patient_contribution: 100,                                                                          
   zipcode: "06222",                                                                                   
   pronouns: 'ze/zir',                                                                                 
   name: 'Pledge submitted - 3'
end
```
</details>

app/controllers/accountants_controller.rb
```diff
@@ -7,19 +7,28 @@ class AccountantsController < ApplicationController
...
   end
 
   def search
+    index
+
     @results = if params[:search].present?
-                 Patient.where(pledge_sent: true)
-                        .order_by(pledge_sent_at: :desc)
-                        .search(params[:search])
+                 pledged_patients.search(params[:search])
```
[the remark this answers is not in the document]
elimbaum: ah i see, we intended for the accountants to be able to browse only the last 6 months. edge case: if you search for nothing, this will return everything. do we actually want that to only return the last 6 months, or is it fine to treat that as the initial case?

elimbaum:
yeah, weird search behavior. I was seeing some of that before but I had assumed that was the intended behavior haha... i'll take a crack @ this tomorrow!

elimbaum:
Ok, well there's *definitely* something funky going on! Compare SQL for `where name` and `search`

```
irb(main):017:0> (Patient.where("name LIKE '%Patient%'") - Patient.search("Patient")).count
  Patient Load (1.5ms)  SELECT "patients".* FROM "patients" WHERE (name LIKE '%Patient%')
  Patient Load (1.4ms)  SELECT "patients".* FROM "patients" WHERE "patients"."line" IN ($1, $2, $3) AND (((name ilike '%Patient%') OR (other_contact ilike '%Patient%')) OR (identifier ilike '%Patient%')) ORDER BY "patients"."updated_at" DESC LIMIT $4  [["line", "DC"], ["line", "MD"], ["line", "VA"], ["LIMIT", 15]]
```

Search has a hardcoded limit of 15, for some reason. Trying to figure out if this is Kaminari or something we did. Also note that `search` method searches on more than name.

elimbaum:
@lomky, after some digging through git blame, I discovered that we added search limiting about four years ago (Issue #845, PR #883), and eventually this got rolled into the `PatientSearchable` concern.

Should we just move the search limiting to the dashboard controller, so that we can return all results on the accountants page?

colinxfleming:
Unexpected surprises of yesteryear! But this makes sense to me.

I think the general idea of having the limit be only on the call list dashboard and not having it here, like you suggested, is a good idea. It's also consistent with what this particular fund is trying to do.

elimbaum:
Search limiting now moved to `dashboards_controller`, and the associated test moved to `record_lookup_test`. (I couldn't quite get it to work as a unit test in `dashboards_controller_test`, but if you have ideas and would rather it be there than a system test, let me know!) Other review changes incorporated.

app/controllers/dashboards_controller.rb
```diff
@@ -5,6 +5,8 @@ class DashboardsController < ApplicationController
 
   before_action :pick_line_if_not_set, only: [:index, :search]
 
+  SEARCH_LIMIT = 15
```
colinxfleming, line 8: hmm, I see where your head's at, but I think I'd do this slightly differently - the LIMIT 15 is almost a safety feature (makes it harder to get mass records out) and it would be good to preserve as the default behavior. So to that end, what if:
* we open up patient_searchable and change the search method to take an argument, `limit`, with a default value of `SEARCH_LIMIT`
* in the accountants controller, we turn `Patient.search(...)` into `Patient.search(..., limit=nil)` which I think will return all results like we intend
* we revert the changes here in dashboards_controller, which will still keep the current behavior in place

my bad for not being clearer here about what I was thinking, I apologize! Also open to an argument about this as the preferred approach - let me know what you think.

elimbaum: oh, very fair, that seems safer for future search use-cases. makes sense!

elimbaum: turns out passing `.limit(nil)` does nothing, so we get that functionality for free!

app/controllers/accountants_controller.rb
```diff
@@ -7,24 +7,34 @@ class AccountantsController < ApplicationController
   def index
     patients = pledged_patients.includes(:clinic)
     @patients = paginate_results(patients)
+
+    @entry_name = t('common.patient')
   end
 
   def search
+    index
```
colinxfleming, line 15: Rails controllers (and this is true for stuff like express.js, django, etc too I think) do a lot of behind the scenes stuff, so it's not advisable to call endpoint methods from other endpoint methods. What are we trying to accomplish here? If it's complicated let's roll it up into a private method that both index and search endpoints can share.

app/models/concerns/patient_searchable.rb
```diff
@@ -2,11 +2,12 @@
...
   extend ActiveSupport::Concern
 
-  SEARCH_LIMIT = 15
+  DEFAULT_SEARCH_LIMIT = 15
 
   class_methods do
     # Case insensitive and phone number format agnostic!
-    def search(name_or_phone_str, lines = LINES)
+    # pass `search_limit = nil` for no limit.
+    def search(name_or_phone_str, lines: LINES, search_limit: DEFAULT_SEARCH_LIMIT)
```
colinxfleming, line 10: This swaps these to keyword args rather than positional args, right? I'm good with that, though reading the comment does it mean we want to do `search_limit: nil`?

colinxfleming (commented):
Thanks, appreciate this approach switch up! I'll take it for a test drive later tonight.

app/models/concerns/patient_searchable.rb
elimbaum: oops, yup!

app/controllers/accountants_controller.rb
elimbaum: ah, that's fair. we need access to `@patients` in `search` to know how many total patients there are, for Kaminari interface display. Maybe I'll make a new method for that setup stuff.

app/javascript/styles/tables.scss
```diff
@@ -9,13 +9,13 @@ table.border-side {
     text-transform: uppercase;
     letter-spacing: 1.5px;
     background: $light-gray;
-    padding-left: 0px !important;
+    padding: 0.8rem 0.5rem !important;
```
colinxfleming, line 12: The styling changes here affect a few other tables, including the call list dashboard, so I think we'll need to scope this a smidge tighter. (Def not your fault or anything; our CSS is the toughest part of our codebase to work in.) I would suggest slapping a new class on the accountants table, and adding a new CSS rule that applies these padding rules to it.

app/controllers/accountants_controller.rb
```diff
@@ -5,26 +5,33 @@ class AccountantsController < ApplicationController
   before_action :find_patient, only: [:edit]
 
   def index
-    patients = pledged_patients.includes(:clinic)
-    @patients = paginate_results(patients)
+    generate_base_patients
```
colinxfleming, line 8: _much_ better, thank you!

app/controllers/accountants_controller.rb
```diff
@@ -38,15 +45,21 @@ def edit
...
+
   def paginate_results(results)
     Kaminari.paginate_array(results)
             .page(params[:page])
             .per(25)
   end
 
   def pledged_patients
-    Patient.where(pledge_sent: true,
-                  initial_call_date: 6.months.ago..)
```
colinxfleming, line 49: Just to telegraph, I think I'm good taking out this filter; Kaminari prevents us from loading a million billion patients all at once, which is what we were going for with this original rule.

app/views/accountants/_table_content.html.erb
```diff
@@ -1,12 +1,14 @@
-<table class="table border-side">
+<table class="table border-side table-responsive text-nowrap">
```
colinxfleming, line 1: 👍 nicely done

colinxfleming (commented):
one CSS issue to address, so we don't alter styling outside of the accountants view, but then I'm all good on this. thanks for sticking with this @elimbaum , this was a tougher one and you've done very well with it. Would like @lomky to weigh in too, if possible!

app/controllers/accountants_controller.rb
elimbaum: to clarify, do you want to remove it entirely? and are you referring to both `pledge sent` and `initial call date` or just the latter?

app/javascript/styles/tables.scss
elimbaum: All your `table.border-side th` are belong to us

app/controllers/accountants_controller.rb
colinxfleming: The way you have it is right - we def want the `pledge_sent` (since that's what the accountants/reconcilers care about) but don't need the `initial_call_date` because of kaminari.

lomky:
I'm still seeing some funky numebsr when we actually invoke a search term, but not when we do the filter! I'll comment more inline 
Peek 2021-06-23 20-19

app/controllers/accountants_controller.rb
```diff
@@ -5,26 +5,33 @@ class AccountantsController < ApplicationController
...
+    end
+  
+    @results = paginate_results @results
+
+    if @results.length != @patients.length
+      extra_display = t('accountants.extra_count', count: @patients.length,
+                                                   entry: t('common.patient').pluralize(@patients.length))
+      @result_count_extra = " (#{extra_display})"
+      @entry_name = t('accountants.results')
     end
```
lomky, line 35: I think I'm coming to the conclusion that while I like the (out of X patients) theoretically, it just isn't a reliable number we can pull as we adjust the results through searching At this point I'm thinking we just nix the Extra Display rather than keep spinning on this otherwise awesome PR. Thoughts? @elimbaum @colinxfleming?

app/controllers/accountants_controller.rb
```diff
@@ -38,15 +45,21 @@ def edit
 
   private
 
+  def generate_base_patients
+    patients = pledged_patients.includes(:clinic)
```
elimbaum, line 49: note to self this includes is redundant

app/controllers/accountants_controller.rb
elimbaum: this is so weird! let me take another look at this - i can have an answer to you this weekend. what we're searching for shouldn't have a bearing on the total number of patients...!

also I think I should change the language. My `(out of X patients)` is dangerously similar to kaminari's `of X in total`. I'm realizing now you could end up in a situation that read something like `Displaying patients 1 - 25 of 37 in total (out of 47 patients total)` which is rather confusing!

also thank you for your diligence on this! i definitely need to do a leeeeeetle bit more manual testing and not rely entirely on unit tests! or, i need better unit tests ;)

colinxfleming: Enormously in favor of nuking the Extra display - low-to-nil value add to an accountant + high code complexity = great target for removal

elimbaum: Totally fair. do not like the below... in the future if we needed we could possibly do language like (`...out of X patients in database`) 

weird display

elimbaum: @lomky, did some digging, and the bug you observed is caused by Kaminari objects' `.count` method returning the number of elements on the current page, rather than the number of elements in the entire set. For that, you have to use `.total_count`; with that change I was able to replicate and then fix the issue you observed. However, agree with your instinct; I'll remove it.

elimbaum:
Another thing I want to check:

Screen Shot 2021-06-26 at 13 06 46

When you search, the Kaminari navigation links go away, because of line 6 below: https://github.com/DCAFEngineering https://github.com/DCAFEngineering/dcaf_case_management/blob/044bd8a5737a5dc1b5b4ec5d48d3198c0ab0348d/app/views/accountants/search.js.erb#L3-L10

This partially makes sense, because if you don't empty the `nav` links, they don't update with the search (i.e. clicking `Next` brings you to the second page of results for the unfiltered query). Since were in the neighborhood, I wanted to ask, but we can totally ignore this (or happy to make a new issue and look into it later).

elimbaum:
WAHOO time to go outside

app/controllers/accountants_controller.rb
```diff
@@ -5,28 +5,29 @@ class AccountantsController < ApplicationController
...
   def search
     have_search = params[:search].present?
     have_clinic = params[:clinic_id].present?
 
-    if have_search || have_clinic
-      partial = Patient.where(pledge_sent: true)
-                       .includes(:clinic)
-                       .includes(:fulfillment)
-                       .order(pledge_sent_at: :desc)
+    @results = if have_search || have_clinic
```
mdworken, line 16: Nitpick: this is way slicker + totally the right way to do things, but may make the codebase slightly less beginner friendly. No strong opinions,

app/controllers/accountants_controller.rb
```diff
@@ -5,28 +5,29 @@ class AccountantsController < ApplicationController
...
-      @results = partial
+      partial
     else
-      @results = pledged_patients
+      pledged_patients
     end
 
+    # when we search, counter says 'results' instead of 'patients'
+    @entry_name = t('accountants.results')
+    @results = paginate_results @results
```
mdworken, line 29: Rails convention nit: Usually we want to avoid using instance variables that aren't directly returned by the controller, so it would be slightly better to have it be `results` until it's the version that we'll return

app/controllers/dashboards_controller.rb
```diff
@@ -14,7 +14,7 @@ def search
     # n+1 join here
     if params[:search].present?
       @results = Patient.search params[:search],
-                                [current_line.try(:to_sym) || lines]
+                                lines: [current_line.try(:to_sym) || lines]
```
mdworken, line 17: I *think* this might lead to a behavior change, but it seems more correct

app/javascript/styles/tables.scss
```diff
@@ -12,10 +12,20 @@ table.border-side {
...
+    td {
+      padding: 0.5rem 0.5rem !important;
+    }
+  }
+
   tr {
     td {
       font-size: 17px;
-      padding: 10px 0 !important;
+      padding: 10px 0px !important;
```
mdworken, line 28: 🔥

test/system/accountant_workflow_test.rb
```diff
@@ -56,10 +55,37 @@ class AccountantWorkflowTest < ApplicationSystemTestCase
...
+      assert has_content? @fulfilled_patient.name
+      assert has_content? @pledged_patient.name
+      refute has_content? @nonpledged_patient.name
+    end
+
+    it 'should properly display counts' do
+      assert has_content? 'Displaying all 3 patients'
+    end
+
+    it 'should properly display large numbers of patients' do
```
mdworken, line 67: Love the new tests!

test/system/accountant_workflow_test.rb
```diff
@@ -56,10 +55,37 @@ class AccountantWorkflowTest < ApplicationSystemTestCase
...
+                        procedure_cost: 400
+      end
+
+      # gotta reload!
+      visit accountants_path
+
+      assert has_content? 'Displaying patients 1 - 25 of 33 in total'
+
+      # the table should have 25 rows
+      assert has_selector?('tbody#accountants-table-content tr', count: 25)
```
mdworken, line 88: Nit: I think the `tbody` makes this a little more confusing - would this work with just the ID and the `tr`?

mdworken (approved):
Looking great! Left some nits, none of which are blocking imo

app/controllers/dashboards_controller.rb
lomky: It could! but I did a check through the codebase and Eli caught all the instances 🎉

app/controllers/accountants_controller.rb
elimbaum: fair enough! i only did this because the (rubo)cops told me 👮‍♀️

app/controllers/dashboards_controller.rb
elimbaum: slash it took me a couple of Big Red Failing Tests to figure out lol

app/javascript/styles/tables.scss
elimbaum: consistency!!!!!

test/system/accountant_workflow_test.rb
elimbaum: duh yup it's an id, no need to further clarify!

lomky:
So there's still one bug, but I think we can merge this PR and file it as a follow on for our shared sanity!

Upon searching the next/last/>/>> buttons disappear, making it impossible to get to the second+ page.
image
image

elimbaum:
I believe this is existing behavior that was unexposed due to the search limit. I mentioned it [above, here](https://github.com/DCAFEngineering/dcaf_case_management/pull/2220#issuecomment-869031348.) I think this will require a deep dive into Kaminari -- my instinct is that we're not _quite_ using it as intended w.r.t. searching.

colinxfleming (approved):
approved with the fast follow, thankyou @elimbaum

PR #2224, opened 2021-06-05
Migrate events as part of pg; log places where n+1 queries will crop up for future work

colinxfleming:

* Mark a few places where n+1 joins are taking place - we used to get this for free from the mongodb document model and now we extremely don't. This is fine for now since we don't do that much traffic, but we should try and resolve after we get this into production. as noted this will be a good @elimbaum shaped ticket
* fix a few n+1 joins on the accountants page as a proof of concept
* fix patient ids on events - since they get numeric ids now, we need to bump them from the BSON IDs

* Bumps #2072 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/configs_controller.rb
+++ b/app/controllers/configs_controller.rb
@@ -2,6 +2,7 @@ class ConfigsController < ApplicationController
   before_action :confirm_admin_user
 
   def index
+    # n+1 join here
     Config.autosetup
     @configs = Config.all.sort_by(&:config_key)
   end
--- a/lib/tasks/migrate_to_pg.rake
+++ b/lib/tasks/migrate_to_pg.rake
@@ -139,6 +139,16 @@ namespace :migrate_to_pg do
         attrs
       end
       migrate_submodel(pt, mongo_pt, pg, mongo, 'notes', 'patient', extra_transform)
+
+      # Finally, clean up events - creating new objects fires events callbacks;
+      # we want to just change IDs on the old ones
+      # Start by nuking old events
+      Event.where('length(patient_id) != 24').destroy_all # anything with a letter is a BSON ID
+      # Then change ids of what's left
+      puts "Updating patient ids on #{Event.count} events"
+      Event.all.each do |event|
+        event.update patient_id: Patient.find_by(mongo_id: event.patient_id).id
+      end
     end
   end
 end
--- a/app/controllers/accountants_controller.rb
+++ b/app/controllers/accountants_controller.rb
@@ -12,6 +12,8 @@ class AccountantsController < ApplicationController
   def search
     @results = if params[:search].present?
                  Patient.where(pledge_sent: true)
+                        .includes(:clinic)
+                        .includes(:fulfillment)
                         .order(pledge_sent_at: :desc)
                         .search(params[:search])
                else
@@ -38,6 +40,8 @@ class AccountantsController < ApplicationController
   def pledged_patients
     Patient.where(pledge_sent: true,
                   initial_call_date: 6.months.ago..)
+           .includes(:clinic)
+           .includes(:fulfillment)
            .order(pledge_sent_at: :desc)
   end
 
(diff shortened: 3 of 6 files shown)


colinxfleming:
testing script to verify events:

```
echo "Event.first.patient_id" | rails c # should be a bson id
git checkout move-events && rails db:migrate migrate_to_pg:clinic_user_patient
echo "Event.first.patient_id" | rails c # should be a numeric id
```

elimbaum:
What's an `n+1` join?

lomky:
@elimbaum this one'll be good to look at after we release PG for reference if/when you need to resolve any conflicts with #2218

lomky:
Big O & N+1: https://stackoverflow.com/questions/3693185/how-does-big-o-relate-to-n1

elimbaum:
aha, the 1 isn't the issue, it's the N. I think I'll need to do some reading up on this but sounds good!

app/controllers/configs_controller.rb
```diff
@@ -2,6 +2,7 @@ class ConfigsController < ApplicationController
   before_action :confirm_admin_user
 
   def index
+    # n+1 join here
```
lomky, line 5: Can we throw an issue in the backlog like
> `grep "n+1 join here"` and fix em 

while we're at it?

lib/tasks/migrate_to_pg.rake
```diff
@@ -139,6 +139,16 @@ namespace :migrate_to_pg do
...
+
+      # Finally, clean up events - creating new objects fires events callbacks;
+      # we want to just change IDs on the old ones
+      # Start by nuking old events
+      Event.where('length(patient_id) != 24').destroy_all # anything with a letter is a BSON ID
+      # Then change ids of what's left
+      puts "Updating patient ids on #{Event.count} events"
+      Event.all.each do |event|
+        event.update patient_id: Patient.find_by(mongo_id: event.patient_id).id
+      end
```
lomky, line 151: does this make all the events look new? 🤔 

Before
```
irb(main):001:0> Event.first
=> #<Event id: 1, cm_name: [FILTERED], event_type: "reached_patient", line: "DC", patient_name: [FILTERED], patient_id: "60c
0184545cb72b055df1fb2", pledge_amount: nil, created_at: "2021-06-09 01:24:21", updated_at: "2021-06-09 01:24:21">
irb(main):002:0> Event.last
=> #<Event id: 158, cm_name: [FILTERED], event_type: "reached_patient", line: "DC", patient_name: [FILTERED], patient_id: "6
0c0184845cb72b055df21a0", pledge_amount: nil, created_at: "2021-06-09 01:24:24", updated_at: "2021-06-09 01:24:24">
```

After:
```
irb(main):001:0> Event.first
=> #<Event id: 1, cm_name: [FILTERED], event_type: "reached_patient", line: "DC", patient_name: [FILTERED], patient_id: "1",
 pledge_amount: nil, created_at: "2021-06-09 01:24:21", updated_at: "2021-06-09 01:28:09">
irb(main):002:0> Event.last
=> #<Event id: 158, cm_name: [FILTERED], event_type: "reached_patient", line: "DC", patient_name: [FILTERED], patient_id: "3
9", pledge_amount: nil, created_at: "2021-06-09 01:24:24", updated_at: "2021-06-09 01:28:09">
```

Looks like the `created_at` is retained and the `updated_at` is updated (natch). Does that matter?

  - ✅ Events Controller - cares about `created_at`  
  - ✅ Events Helper - cares about `created_at`
  - ✅ Events Model - cares about `created_at`
  - ✅ No other objects care about the Event's timestamps

all clear. 🎉 
and since these are ephemeral objects that delete after three weeks I'm not worried about them having distinctive updated timestamps

lomky (approved):
One ask for getting the n+1 ticket onto the backlog, otherwise clear! :shipit:

app/controllers/configs_controller.rb
colinxfleming: yeah! doing so now

colinxfleming: #2225

PR #2228, opened 2021-06-11
Final postgres fixes before showtime

colinxfleming:

Final pg fixes before showtime, to address issues that got ID'd by user tests (thank you @lomky  for leading that effort!). A couple things got punted to post-postgres.

* fix the js that updates the 'This patient will be audited on' text
* check presence on patient search to disallow returning random results. this was current behavior and would just return 15 random records, but I agree that this isn't great and now's a good time to nip it in the bud

no view changes

* Bumps #2072 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -12,8 +12,12 @@ class DashboardsController < ApplicationController
 
   def search
     # n+1 join here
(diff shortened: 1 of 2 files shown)


lomky (approved):
Get it! :shipit:

PR #2143, opened 2021-03-07
Postgres port omnibus

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

Here's the plan: 
* We start with a single trivial-ish change to `application_controller` to start up a shell PR
* Further PRs will port individual models or pieces of functionality. This is to make the actual code review as bite-sized as possible, similar to how we did bootstrap 3
* When we have a working, green build, we'll write up a plan to test it to confirm that the migration script does what we want

This pull request makes the following changes:
* eventually, converts to postgres and removes the need for mongo
* adds a migration rake task for this cluster of models

no view changes

It relates to the following issue #s: 
* Bumps #2072 
* pretty much individual components of #2123 

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #2233, opened 2021-06-12
comment out user email callbacks for a second

colinxfleming:

can't beat sendgrid

no issues


lomky:
Temp addition to #2143 to make sure we don't spam users

xmunoz (approved):

PR #2242, opened 2021-06-13
actually remove mongo session store

### LINKED ISSUE #2236: enable activerecord-session_store
We had this in mongo and we should do it in pg too -- https://github.com/rails/activerecord-session_store

colinxfleming:

SORRY @lomky 

I forgot to re-bundle install after removing mongo-session-store.

* actually remove mongo session store

* Fixes #2236  (continued)

For reviewer:
* Tag this PR:


lomky (approved):
lol np

PR #2257, opened 2021-06-23
Fix - Updating update_pledge_sent_by_sent_at  save for consistency

### LINKED ISSUE #2019: Pledged data entry patients should have pledge fields set on save

Right now the hook for update_pledge_sent_by_sent_at is only invoked on update, instead of on save. 
https://github.com/DCAFEngineering/dcaf_case_management/blob/f600471bb898df8cff1510c7857ee727f04c5a40/app/models/patient.rb#L29
This works smoothly in our regular flow, but can leave these fields unset when we create via the `/data_entry` route.

Make behavior on similar fields consistent, capture fallback pledging info for data entry. 

Update the call above to `before_save` instead of update, and update the function to ensure we are updating on change, similar to `update_fund_pledge_at`.
https://github.com/DCAFEngineering/dcaf_case_management/blob/f600471bb898df8cff1510c7857ee727f04c5a40/app/models/patient.rb#L285

* **Anything else?**

nsiwnf:

* Make behavior consistent with `update_fund_pledge_at`

* Fixes #2019 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -20,7 +20,7 @@ class Patient < ApplicationRecord
   # Callbacks
   before_validation :clean_fields
   before_save :save_identifier
-  before_update :update_pledge_sent_by_sent_at
+  before_save :update_pledge_sent_by_sent_at
   before_save :update_fund_pledged_at
   after_create :initialize_fulfillment
   after_update :confirm_still_urgent, if: :urgent_flag?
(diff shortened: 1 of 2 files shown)


colinxfleming (approved):
pulled down and affirmed it follows the behavior laid out in that issue - thank you @nsiwnf !

PR #2259, opened 2021-06-24
update sqreen gem

colinxfleming:

Bump sqreen gem since it's a smidge out of prod.

* `bundle update sqreen`

no view changes

no issues

For reviewer:
* Tag this PR:


lomky (approved):
Not too worried since this is a minor bump, but how would we test these sorts of deps?

colinxfleming:
honestly, this is a production group gem, so we're kinda at the mercy of the sandbox environment, sad but true. that said, so few of them and they're mostly for services like this so it doesn't stress me out much.

PR #2260, opened 2021-06-27
Allow system tests to run in local docker

### LINKED ISSUE #1874: Install Chrome in Docker to run system tests

I'd like to be able to run systemtests in github actions, but when you run this command it errors out with something like "chrome not found in path": `docker-compose run --rm web rails test test/system/routing_test.rb`

If we COULD run that, we'd be able to get builds working in github actions! which would be good because it would hopefully fix the problem of 'all builds on forks are failing'. So let's set up chrome in the docker container.

We can declare victory when `docker-compose run --rm web rails test test/system/routing_test.rb` returns a green result instead of erroring out.

running system tests in both local environments and github actions; swapping to a CI system that works.

In CircleCI we use an image they have with browsers installed: https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/ruby/images/2.6.5-buster/node-browsers/Dockerfile

We might be able to adapt this to our use.

* **Anything else?**

elimbaum:

This PR installs firefox inside the docker container, allowing system tests to be run locally.

I attempted to install Chrome, to match CircleCI, but ran into a bunch of issues with running it headlessly. FIrefox works out of the box.

⚠️ **Warning!** I was able to run individual system tests with no issues, but running the entire suite hung my computer terribly (adjusted load averages ~4). This was on a 3.3GHz 8-core machine with 8GB RAM that I was only using for DARIA work. YMMV.

* installs firefox
* uses firefox for system tests if running in docker.
  * this should not break existing functionality, specifically, running system tests on bare metal

* Fixes #1874 

For reviewer:
* Tag this PR:

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -45,6 +45,9 @@ RUN curl -sL https://deb.nodesource.com/setup_12.x | sudo bash - && \
 # remove unnecessary build dependencies
 RUN apt-get purge -y ${BUILD_DEPENDENCIES} && apt-get autoremove -y 
 
+# Install firefox for system tests
+RUN apt-get install -y firefox-esr
+
 # Check docker base image for vulnerable packages, ignore non zero exit code (just informative)
 RUN mkdir /tmp/ahab && \
     cd /tmp/ahab && \
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -19,5 +19,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   # if in CI, run system tests headlessly.
   browser = ENV['GITHUB_WORKFLOW'] ? :headless_chrome : :chrome
+
+  # if in docker, run headless firefox
+  browser = ENV['DOCKER'] ? :headless_firefox : browser
+
   driven_by :selenium, using: browser
 end


colinxfleming (approved):

test/application_system_test_case.rb
```diff
@@ -19,5 +19,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   # if in CI, run system tests headlessly.
   browser = ENV['GITHUB_WORKFLOW'] ? :headless_chrome : :chrome
+
+  # if in docker, run headless firefox
+  browser = ENV['DOCKER'] ? :headless_firefox : browser
```
colinxfleming, line 24: another way to do this (that involves a little less chaos) is sticking this in a conditional, a la:

```
browser = if ENV['GITHUB_WORKFLOW']
                   :headless_chrome
                 elif ENV['DOCKER']
                   ...
```

that said I'm opposite of worried about this, and merging as is!

.docker/Dockerfile
```diff
@@ -45,6 +45,9 @@ RUN curl -sL https://deb.nodesource.com/setup_12.x | sudo bash - && \
 # remove unnecessary build dependencies
 RUN apt-get purge -y ${BUILD_DEPENDENCIES} && apt-get autoremove -y 
 
+# Install firefox for system tests
+RUN apt-get install -y firefox-esr
```
colinxfleming, line 49: using firefox over chrome is a really sharp move. good call!

PR #2264, opened 2021-07-10
Allow patients with sent pledges to be marked urgent again

### LINKED ISSUE #2262: Restore ability to tag callers as "Urgent" after pledge is sent
* Flag cases as urgent after the pledge has been sent so they appear on the DARIA home page and allow CMs who are picking up a shift to see what needs to be prioritized/followed up on.

*BAF CMs use this feature frequently to flag cases during CM handoffs that may potentially fall through the cracks after a pledge has been sent. It also helps CMs during their shifts prioritize their callers on the DARIA landing page.

If you want a flag a PT (patient) for the next CM or yourself to keep track of, you can't (the box unchecks automatically & won't let you re-check it), whereas before you'd be allowed to mark as urgent.

Restore the ability to flag a PT as urgent after a pledge is sent.

* **Anything else?**

colinxfleming:

A few funds have reported that postgres caused a slight change in how urgency works, and that they can no longer mark patients with sent pledges urgent. That was original intent, but funds are used to the not-quite-right version, so this PR implements the not-quite-right version. (I think the source of this is callbacks working slightly differently in activerecord than in mongoid, but exact details aren't that important, I don't think.)

* remove pledge_sent from unmark-urgency criteria

no view changes

* Fixes #2262

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/urgency.rb
+++ b/app/models/concerns/urgency.rb
@@ -4,7 +4,7 @@ module Urgency
 
   def still_urgent?
     # Verify that a pregnancy has not been marked urgent in the past six days
-    return false if pledge_sent || resolved_without_fund
+    return false if resolved_without_fund
     recent = versions.where('created_at > ?', 6.days.ago)
     return true if recent.any?(&:marked_urgent?)
     false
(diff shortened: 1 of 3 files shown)


lomky (approved):
✅ :shipit:

PR #2266, opened 2021-07-11
Fix accounting pagination bug

### LINKED ISSUE #2261: Accountant Search & Filter hides the pagination navigation links

Searching and filtering on the Accountants page should retain the navigation links when there are paginated results.

Fully functional paginated accounting page

Not sure!

* **Anything else?**
Context!

> 
> When you use the accountants search, the Kaminari navigation links go away, because of line 6 below: https://github.com/DCAFEngineering https://github.com/DCAFEngineering/dcaf_case_management/blob/044bd8a5737a5dc1b5b4ec5d48d3198c0ab0348d/app/views/accountants/search.js.erb#L3-L10
> 
> This partially makes sense, because if you don't empty the `nav` links, they don't update with the search (i.e. clicking `Next` brings you to the second page of results for the unfiltered query). Since were in the neighborhood, I wanted to ask, but we can totally ignore this (or happy to make a new issue and look into it later).
> 
> Screen Shot 2021-06-26 at 13 06 46

_Originally posted by @elimbaum in https://github.com/DCAFEngineering/dcaf_case_management/issues/2220#issuecomment-869031348_

> I believe this is existing behavior that was unexposed due to the search limit. I think this will require a deep dive into Kaminari -- my instinct is that we're not _quite_ using it as intended w.r.t. searching.

_Originally posted by @elimbaum in https://github.com/DCAFEngineering/dcaf_case_management/pull/2220#issuecomment-872562005_

> 
> Upon searching the next/last/>/>> buttons disappear, making it impossible to get to the second+ page.
> image
> image
> 

_Originally posted by @lomky in https://github.com/DCAFEngineering/dcaf_case_management/issues/2220#issuecomment-872557387_

colinxfleming:

Kobayashi Maru-s the accounting workflow a little bit to make it properly paginate -- this is a pretty aggro approach but I think it gets us to where we want to be. It eliminates the search endpoint in favor of making search URL-params based on the index, and having the form redirect instead of making remote requests.

This is a little clunkier than our nice ajax, but it has the perk of not doing a confusing dance with kaminari's library magic.

* Fix pagination bug on accounting
* Eliminate accountants/search endpoint
* bump tests to match.

eyyy: 
image

(You can too with http://localhost:3000/accountants?button=&clinic_id=&locale=&search=reporting+patient)

* Fixes #2261 

For reviewer:
* Tag this PR:


elimbaum:
Are we OK with the additional fields showing up in the URL? (`locale=&button=`)

Alternatively, should the search just be a POST request?

app/controllers/accountants_controller.rb
```diff
@@ -5,30 +5,15 @@ class AccountantsController < ApplicationController
...
-    respond_to { |format| format.js }
+    @patients = Patient.where(pledge_sent: true)
+                       .includes(:clinic)
+                       .includes(:fulfillment)
+                       .order(pledge_sent_at: :desc)
+                       .page(params[:page] || 1)
+
+    @patients = @patients.where(clinic_id: params[:clinic_id]) if params[:clinic_id].present?
+    @patients = @patients.search(params[:search]) if params[:search].present?
+    @patients
```
elimbaum, line 16: ha this is much cleaner! and I guess this really is the way you're supposed to do it, huh?

config/routes.rb
```diff
@@ -53,7 +53,6 @@
     post 'data_entry', to: 'patients#data_entry_create', as: 'data_entry_create' # temporary
 
     resources :accountants, only: [:index, :edit]
-    post 'accountant/search', to: 'accountants#search', defaults: { format: :js }
```
elimbaum, line 56: oh ok, so what was your rationale for removing the post?

elimbaum:
one super nitpicky thing that we may totally not care about: for some reason the first and prev don't have the same margin that next and last do. i don't think this was something you did though, colin. i'm just noticing it now...

Screen Shot 2021-07-14 at 19 26 00

elimbaum:
oops, sorry! thought i was suggesting a change but it was a commit! well i think that'll do it.

elimbaum (changes requested):
this is great! this feels much better than our earlier behind-the-scenes javascript!

did stumble across one bug: if you search for something, page 1 will only have 15 results. (others have 25) This confuses kaminari, and it seems to be making the wrong number of pages - you can then get stuck on the phantom last page, where the buttons stop working.

I think this is just caused by https://github.com/DCAFEngineering/dcaf_case_management/blob/58326b742c9b83ec124e28abb396acbc8c3e49ee/app/models/concerns/patient_searchable.rb#L5

changing it to 25 seems to work! (this way it matches kaminari's expectations)

alternatively, adjust kaminari in accountants controller to expect 15 per page.

app/controllers/accountants_controller.rb
```diff
@@ -5,30 +5,15 @@ class AccountantsController < ApplicationController
...
-    @entry_name = t('accountants.results')
-    @results = paginate_results results
-
-    respond_to { |format| format.js }
+    @patients = Patient.where(pledge_sent: true)
+                       .includes(:clinic)
+                       .includes(:fulfillment)
+                       .order(pledge_sent_at: :desc)
+                       .page(params[:page] || 1)
+
```
elimbaum, line 13: I think if you add `.per(15)` here, that should fix it.

config/routes.rb
colinxfleming: This was actually not something I intended to do at the start, but it felt like it made a lot of sense as I progressed! Essentially went as follows:
* kaminari async causes a LOT of weird problems
* we COULD, in theory, do this all with URL parameters and redirects?
* If that's the case, we don't need a search endpoint, and can just filter using URL params on the base endpoint, I think?
* omg it works

It has the side effect of making searches copy-pasteable to others, since you can set them via URL param rather than post payload.

app/controllers/accountants_controller.rb
colinxfleming: 🤷  no idea honestly! I spent a lot of time flipping thru kaminari docs and I'm still not totally sure I got it right.

colinxfleming:
incredible catch on the pagination thing! I'll punt up a commit that fixes it in a sec.

lomky:
> Are we OK with the additional fields showing up in the URL? (`locale=&button=`)
> 
> Alternatively, should the search just be a POST request?

I'm not bothered by them here, personally

colinxfleming:
I took a slightly different tack and removed the search limit from the controller, which I think gets us to where we need to be. I _think_ this gets us to a good place; I did a little testing with paginations of 5 and seems to work pretty cleanly. Setting the limits is now kaminari's problem.

But I also accidentally dropped your commit fixing the css @elimbaum ! Sorry about that, if you remember what it was it would be clutch if you'd make it again.

colinxfleming:
I think this is ready for review again! At your leisure, @lomky  / @elimbaum

elimbaum (approved):

Issue #2285, opened 2021-07-21
Upgrade to heroku-20 stack


xmunoz:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
Upgrade our infrastructure. Currently we are using the heroku-18 stack, which has some slightly out-of-date software, and is not the currentl default anymore.
https://devcenter.heroku.com/articles/heroku-20-stack
> Heroku-20 is currently the default stack for newly created apps.

* **What feature or behavior is this required for?**
We need to upgrade a bunch of the Dockerfile stuff along with this change, since we are relying on older versions of node and probably other stuff, for example.

* **How could we solve this issue? (Not knowing is okay!)**
Upgrading our dev env dependencies, and running the appropriate heroku admin commands.

* **Anything else?**

xmunoz:
Turns out we also need to upgrade yarn:

> remote:        yarn install v1.22.4
remote:        [1/5] Validating package.json...
remote:        error @: The engine "node" is incompatible with this module. Expected version ">=14.16.0". Got "12.16.2"
remote:        error Found incompatible module.
remote:        info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
remote:        I, [2021-07-28T15:57:17.242330 #1459]  INFO -- : Writing /tmp/build_633179b0/public/assets/application-708d2201f25caac7dc1595243aea9fb9d245de2620434a7959bb3c3d6f5ca387.js
remote:        I, [2021-07-28T15:57:17.242673 #1459]  INFO -- : Writing /tmp/build_633179b0/public/assets/application-708d2201f25caac7dc1595243aea9fb9d245de2620434a7959bb3c3d6f5ca387.js.gz
remote:        Compiling...
remote:        Compilation failed:
remote:        yarn run v1.22.4
remote:        info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
remote:        
remote:        
remote:        error Command "webpack" not found.
remote:        
remote: 
remote:  !
remote:  !     Precompiling assets failed.
remote:  !
remote:  !     Push rejected, failed to compile Ruby app.
remote: 
remote:  !     Push failed
remote:  !
remote:  ! ## Warning - The same version of this code has already been built: b28cd9d2d195df653eae2b9ffa6e4def38468d67
remote:  !
[...]

PR #2299, opened 2021-07-24
Upgrade node and ruby

xmunoz:

* Makes explicit the require node version, and upgrades that version to 14
* Upgrades the ruby version docker container
* Upgrades some dependencies

* Tests the fix for #2285 

For reviewer:
* Bring this up locally and click around to make sure everything is gravy

Followup:
* After this is merged, ensure that the sandbox deploy works and looks good.
* Then upgrade the Heroku stack for sandbox
* Verify again that everything looks good
* Promote to prod
* Upgrade the stack for each prod deployment


xmunoz:
A couple of failures during the execution of `bundle exec rake knapsack:minitest`. @colinxfleming @lomky want to take a look?

xmunoz:
Nice, only down to one failing test now:
```
Failure:
CallListInteractionTest::completed calls section#test_0001_should add a call to completed when a call was made within 8 hrs [/home/circleci/abortioneering/test/system/call_list_interaction_test.rb:74]:
Expected false to be truthy
```

.nvmrc
```diff
@@ -0,0 +1 @@
+v14.16.0
```
colinxfleming, line 1: Putting on my 'community volunteer leader' hat: Should we add nvm / properly setting node version stuff to the readme? I think most people are using docker these days but some vols do still do the individual steps, either as a learning exercise or because of the intensity of docker.

I'll also node that we have `.ruby-version` in gitignore, so we'll probably want to follow a similar pattern there?

colinxfleming (commented):
couple of notes - pair of questions about nvm and an alt strategy for the bundle update that I think might fix the test problem. Excited to lock down node for real though!

.nvmrc
lomky: Hello itme one of (the only?) not-docker-runner 😄  

I definitely am onboard with both committing our `.nvmrc` this and getting this into the setup. If anything I'd be a fan of removing `.ruby-version` from the ignore file so we can play nicer with rbenv for the bare metal setup.

xmunoz:
I would prefer to just fix the failing test instead of downgrading everything. Dependabot is going to be hitting us up for these upgrades eventually anyway. Let me see if it's an easy fix.

.nvmrc
xmunoz: Please feel free to push to this branch. I don't have strong opinions about this, and also don't know what would be best for the community in this case.

Gemfile.lock
```diff
@@ -1,67 +1,67 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    Ascii85 (1.0.3)
-    actioncable (6.1.3.2)
-      actionpack (= 6.1.3.2)
-      activesupport (= 6.1.3.2)
+    Ascii85 (1.1.0)
```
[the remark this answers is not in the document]
xmunoz: Hmm yeah I don't want to just keep kicking the can down the road :/

.nvmrc
colinxfleming: I'll kick up a PR for this in a sec.

colinxfleming: https://github.com/DCAFEngineering/dcaf_case_management/pull/2309 is my attempt at setup instructions for nvm / keeping our bare metal instructions up to date. if that looks chill to you both I think we're good.

Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'bootsnap', '>= 1.4.2', require: false
 # Asset pipeline
 gem 'webpacker', '~> 5.4'
 gem 'sass-rails', '>= 6'
-gem 'bootstrap', '~> 4.5.0'
+gem 'bootstrap', '~> 4.5.0', '< 5' # we're on bs4 for now
```
colinxfleming, line 15: probably a minor rebase in your future, sorry!

colinxfleming (approved):
prob needs a rebase if theduplicative bootstrap 4.5.0 gemfile change is any indication, but I think fire when ready! we'll still need to figure out the heroku stack thing but I'm confident we'll resolve that before the drift becomes a problem.

PR #2300, opened 2021-07-24
[DO NOT MERGE] Try out acts_as_tenant gem

colinxfleming:
*Not intended for merge*

getting weird with acts_as_tenant. This tries to set up and configure the gem, and make appropriate model and other changes to get the seedfile and test suite running again.


config/application.rb
```diff
@@ -56,5 +56,8 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+
+    # Allow lvh.me for subdomain garbage
```
colinxfleming, line 60: The tutorial recommended using `lvh.me` (for ex, `dcaf.lvh.me`) to try out subdomain stuff in practice. Not sure if I like it as an approach or not, and esp not in `application.rb`, but can figure it out down the line.

db/migrate/20210724231608_allow_multi_email_on_user.rb
```diff
@@ -0,0 +1,6 @@
+class AllowMultiEmailOnUser < ActiveRecord::Migration[6.1]
```
colinxfleming, line 1: Gotta redo a couple of indexes.

test/fixtures/funds.yml
```diff
@@ -0,0 +1,11 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
colinxfleming, line 1: rm

test/models/fund_test.rb
```diff
@@ -0,0 +1,7 @@
+require "test_helper"
+
+class FundTest < ActiveSupport::TestCase
+  # test "the truth" do
```
colinxfleming, line 4: I do want to add some actual scoping tests here to prevent regression; one of those 'fix-it-in-post' kinda things. I'll get to it when we've agreed on the general broad outline here.

test/models/user_test.rb
```diff
@@ -14,8 +14,7 @@ class UserTest < ActiveSupport::TestCase
       it "should require content in #{attribute}" do
         @user[attribute.to_sym] = nil
         assert_not @user.valid?
-        assert_equal "can't be blank",
-                     @user.errors.messages[attribute.to_sym].first
+        assert @user.errors.messages[attribute.to_sym].include? "can't be blank"
```
colinxfleming, line 17: side effect of not being able to use validatable :(, or possibly I copypasta'd wrong in `user.rb`.

colinxfleming (commented):
couple notes

colinxfleming:
build is green! Left notes for tmw.

colinxfleming:
Here's my takeaway from this run at setup:
* this feels eminently doable! The gem configuration was surprisingly straightforward and I only ran into snags on properly setting things because the documentation is rspec-oriented.
* actually implementing this would involve something a little more devopsy - we'd have to turn 
* if we wanted to make this shovel-ready, we'd need to write some tests and do some additional work to turn `FUND` / a lot of the env var handling into something handled with models. We could probably hack that into place and do it in the scope of a separate PR, but we assume in a lot of places that fund is an environment variable, and there's a decent chunk to unravel there.
* We would also have to actually run comms on this with other funds, and let them know that our security posture is changing, and that they'd be accessing things through a different URL. Would probably be some comms we'd want to manage here, so we'd want to start running through
* The benefits of this include way stronger pg and ec2-s, a vastly simplified onboarding process ('pop into the rails console and create a fund object'), fewer servers to manage, smaller heroku bill, etc. so lots of major cascading benefits here. For ex, we could go to two standard-1x instance types and load balance them, and a pg standard instance, and we'd still be saving dollars in addition to saving headache.

Issue #2227, opened 2021-06-11
For Config, enforce uniqueness on config_key at the database level


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
I don't know how on earth we managed to do this but we somehow created two patient fulfillment archive config entries in staging. We only enforce their uniqueness in the rails model; let's do that in postgres too.

* **What feature or behavior is this required for?**
preventing extremely bizarre behavior

* **How could we solve this issue? (Not knowing is okay!)**
make a rails database migration that adds this index/uniqueness check

* **Anything else?**
nah. fairly low priority though, this probably won't bite us again regardless.

PR #2301, opened 2021-07-25
Enforce config key uniqueness to avoid weird errors when changing fund configs in the Config panel

mdworken:
1) Requires uniqueness of configkey at the db level
2) adds spec

#2227


colinxfleming (approved):

colinxfleming:
thanks much!

PR #2308, opened 2021-07-27
pin bootstrap gem to < 5

colinxfleming:

Dependabot tried to bump bootstrap; upgrading bootstrap is very very hard; we'll pin it for now to make it clear that we don't want to do this.

* pin bootstrap gem below 5

no views
no issues

For reviewer:
* Tag this PR:


colinxfleming:
godmoding

PR #2316, opened 2021-07-28
Revert "Upgrade node and ruby"

xmunoz:
Reverts DCAFEngineering/dcaf_case_management#2299

We need to upgrade yarn before this ready to deploy.


xmunoz:
Ref: #2285

colinxfleming (approved):
f

PR #2343, opened 2021-08-17
Configurable Urgent Reset Window Config

### LINKED ISSUE #2322: Change urgent patients tooltip to note six day period

A fund got gotcha-d by checking this helpful tooltip, which does not contain a note about the six day diff period.
image

Let's add a note about the six day check to the tooltip.

no surprises when funds depend on the urgents list

We'll need to change this in `en.yml` and `es.yml` under the appropriate key! Google translate is fine for getting the spanish translation.

* **Anything else?**

nah

### LINKED ISSUE #2338: Add configuration option for Patient urgency reset

Right now patient urgency is hardset to one week. Let's make this fund-configurable!

For some funds that use a more collective pool would like urgency to last a little but longer so they get more overlap.

 - add a config for urgency
 - update the code to use this
 - set the default to the current hardcode
 - update any docs!

* **Anything else?**
s/o to New River folks for raising!

See #2150 for a recent similar effort

lomky:

* Adds the Urgent Reset config, default to 6
* Uses the config instead of hard-coded 6
* Adds tests
* docs
* config alt text
* Urgent `?` tool tip

* Fixes #2338 
* Fixes #2322 

For reviewer:
* Tag this PR:

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -535,7 +535,7 @@ en:
     referred_to_clinic: Check this box if you as a case manager referred the patient to a particular clinic.
     resolved_without_fund: This is used to indicate that a patient does not require or want our services any longer.
     status_definition: Status definition
-    urgent_cases: These are patients who require a little more attention. This list is shared across all case managers working on a single line. Patients are removed from this list automatically after they are marked as having a sent pledge, or marked as resolved without assistance.
+    urgent_cases: 'These are patients who require a little more attention. This list is shared across all case managers working on a single line. Patients are removed from this list automatically after they are marked as having a sent pledge, are marked as resolved without assistance, or after %{urgent_reset} days of inactivity.'
   user:
     common:
       last_login: Last Login
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -535,7 +535,7 @@ en:
     referred_to_clinic: Check this box if you as a case manager referred the patient to a particular clinic.
     resolved_without_fund: This is used to indicate that a patient does not require or want our services any longer.
     status_definition: Status definition
-    urgent_cases: 'These are patients who require a little more attention. This list is shared across all case managers working on a single line. Patients are removed from this list automatically after they are marked as having a sent pledge, are marked as resolved without assistance, or after %{urgent_reset} days of inactivity.'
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -535,7 +535,7 @@ en:
     referred_to_clinic: Check this box if you as a case manager referred the patient to a particular clinic.
     resolved_without_fund: This is used to indicate that a patient does not require or want our services any longer.
     status_definition: Status definition
-    urgent_cases: These are patients who require a little more attention. This list is shared across all case managers working on a single line. Patients are removed from this list automatically after they are marked as having a sent pledge, or marked as resolved without assistance.
+    urgent_cases: 'These are patients who require a little more attention. This list is shared across all case managers working on a single line. Patients are removed from this list automatically after they are marked as having a sent pledge, are marked as resolved without assistance, or after %{urgent_reset} days of inactivity.'
```
lomky, line 538: So, this isn't true. Specifically:
> Patients are removed from this list automatically after they are marked as having a sent pledge.

Nope! Nothing removes pledged patients from the list!

@colinxfleming Is this a bug in the behavior, or the documentation? I would suggest we leave the existing behavior alone (since no one has complained in the past half decade) and update this tooltip & the docs

colinxfleming (approved):
world class catch on the docs/tooltip text being out of date, and this looks marvelous. Thanks for knocking it out so quickly!

test/models/patient_test.rb
```diff
@@ -321,18 +321,90 @@ class PatientTest < ActiveSupport::TestCase
...
+          assert_equal 2, Patient.urgent_patients('DC').count
+          assert_equal 1, Patient.urgent_patients('MD').count
         end
       end
 
-      it 'should return urgent patients by line' do
-        assert_equal 2, Patient.urgent_patients('DC').count
-        assert_equal 1, Patient.urgent_patients('MD').count
+      describe 'trim_urgent_patients method' do
+        it 'should trim pregnancies after they have been urgent for five days' do
```
lomky, line 341:
```suggestion
        it 'should trim patients from urgent after they have been inactive or resolved' do
```

PR #2344, opened 2021-08-17
add bullet gem, fix some of our identified n+1 joins

### LINKED ISSUE #2225: Fix postgres n+1 joins

For expedience, we're adding some n+1 joins to some controllers. We shouldn't let that sit for too long though!

fast app

`git grep "n+1 join here"` to find these trouble spots; then adjust the controller code to use appropriate `includes` or `joins` to eliminate the n+1 joins.

* **Anything else?**

colinxfleming:

This cropped up again in an inconvenient way, so going to make this my 'chip away at it' project this week.

* install bullet gem
* make bullet fail tests if it hits a problem
* fix the problems

* bumps #2225 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/call_listable.rb
+++ b/app/models/concerns/call_listable.rb
@@ -39,7 +39,7 @@ module CallListable
   end
 
   def reorder_call_list(order, line)
-    current_entries = call_list_entries.where(line: line).to_a
+    current_entries = call_list_entries.includes(:patient).where(line: line).to_a
     order.each_with_index do |pt, i|
       current = current_entries.find { |x| x.patient_id.to_s == pt }
       current.update order_key: i
@@ -69,7 +69,7 @@ module CallListable
 
   def ordered_patients(line)
     # n+1 join here
-    call_list_entries.includes(:patient)
+    call_list_entries.includes(patient: [:calls])
                      .where(line: line)
                      .order(order_key: :asc)
                      .map(&:patient)
--- a/app/models/concerns/call_listable.rb
+++ b/app/models/concerns/call_listable.rb
@@ -69,7 +69,7 @@ module CallListable
 
   def ordered_patients(line)
     # n+1 join here
-    call_list_entries.includes(patient: [:calls])
+    call_list_entries.includes(patient: [:calls, :fulfillment])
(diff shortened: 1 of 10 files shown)


app/models/concerns/exportable.rb
```diff
@@ -127,11 +127,11 @@ def last_call_timestamp
   end
 
   def call_count
-    calls.count
+    calls.size
```
colinxfleming, line 130: Fun fact: `size` will use the cached query, and `count` will always run a `count(*)` query. So this winds up saving us a lot of queries.

app/models/concerns/exportable.rb
```diff
@@ -158,17 +158,11 @@ def get_field_value_for_serialization(field)
...
 
   def external_pledge_sum
-    sum = 0
-    all_pledges = external_pledges.all
-
-    all_pledges.each do |pledge|
-      sum += pledge.try :amount
-    end
-    sum
+    external_pledges.inject { |x, acc = 0| acc = x.try(:amount) || 0}
```
colinxfleming, line 165: `external_pledges.all` was running an additional select * from external pledges, despite eager loading. So we turn this into a snappy one line reducer.

colinxfleming (commented):
notes

app/models/concerns/call_listable.rb
```diff
@@ -69,7 +69,7 @@ def clear_call_list(line)
 
   def ordered_patients(line)
     # n+1 join here
```
colinxfleming, line 71: There's more to do here, but we do fix some of the particular egregious cases (and get bullet to hush, so leaving this in.

lomky: since we've silence the alert, should we throw an issue on the backlog so we don't lose site of it?

lomky (approved):
_so freaking cool_. I love rails

app/models/concerns/call_listable.rb
colinxfleming: I think let's keep #2225 open? I'll add a note about what remaining problems there are.

colinxfleming:
thanks @lomky !

PR #2360, opened 2021-09-04
Revert "squash a security vuln in actionpack via bundle update rails"

colinxfleming:
Reverts DCAFEngineering/dcaf_case_management#2349 (experimental, to see if build passes)


colinxfleming:
Was hoping there was something about the rails point bump that was making puma not start in staging but that does not seem to be the case

PR #2348, opened 2021-08-21
Parallelize our test suite

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

Rails 6 or so introduced a test parallelization feature that would split tests between processors. We haven't been able to adopt it because of some complexity around mongo + circle + some other stuff, but this gets us to a pretty good shape.

This drops our test suite in github to around 7m, which is great, and a pretty substantial improvement for pretty low effort (not counting having to get on github actions in the first place).

This pull request makes the following changes:
* turn on rails parallelization
* do some other odds and ends cleanup

no view changes

It relates to the following issue #s: 
* Fixes #1876 

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #2361, opened 2021-09-05
Port CI to github actions

colinxfleming:

Port our existing circle ci tests/checks into github actions, and depreciates a CI deploy in favor of a heroku deploy.

~I'm very interested in a second opinion on this, but it DOES run capybara/system tests cleanly. We lose parallel splitting but might be able to make that up with rails' native test suite parallelization (which we haven't rolled out yet but which I've been experimenting with in general).~ (this is still true but it's less of a problem, since rails has a built-in test parallelizer and I'm pretty close to implementing it in #2348 

~note that this also does not get the app into a deployable state, as we deploy to staging via circle and this doesn't fix that.~ (it does this now)

This splits the CI config into two different subjobs - one that runs static scanners like brakeman / ruby-audit / bundle-audit etc, and one that sets up the application and runs the test suite including system tests.

(In this current commit circle is still hooked up - when this gets approved I will unhook circleci and merge on a full green build.)

The reason we're doing this is because circleci made some under the hood changes that caused system tests to stop working almost entirely due to puma not starting. Troubleshooting this was a pretty frustrating experience, was above my level of expertise to fix, and knocked out our ability to follow our deploy process for two weeks or so. We had been meaning to take another pass at github actions for a little while, and now seemed like a good time.

* lots of github ci config
* zero out circleci config

no views
no issues

For reviewer:
* Tag this PR:


colinxfleming:
okay, I think this gets us to a place where:
* builds fire
* tests run
* circle build will autodeploy to staging on merge (which sucks but does let us deploy again, until we can port over the deploy stuff to github actions too)

.github/workflows/test_dev_env.yml
```diff
@@ -18,4 +19,85 @@ jobs:
...
+        uses: ruby/setup-ruby@v1 # automagically reads from .ruby-version
+        with:
+          bundler-cache: true
+
+      # npm/yarn setup
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.2
+      - name: Find yarn cache location
```
DarthHater, line 55: For caching go ahead and try this:

```
- uses: actions/cache@v2
  with:
    path: '**/node_modules'
    key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
```

.github/workflows/test_dev_env.yml
```diff
@@ -18,4 +19,85 @@ jobs:
...
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Cache yarn
+        uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: yarn packages
+        run: |
```
DarthHater, line 66: You can just do `run: yarn install`, the `|` is only if you need to do multiple commands

colinxfleming: good flag, thank you!

.github/workflows/test_dev_env.yml
colinxfleming: https://github.com/DCAFEngineering/dcaf_case_management/pull/2361/commits/5080a9e0d02c43a9508e42050f216a7e42de439a addresses this, I think? I popped out the other two cache commands and am using your snippet here instead.

colinxfleming:
Here's where this landed:
* Build CI completely swapped to github actions
* Deploy step removed and ported over to heroku - via a heroku-side config, we now listen for changes on main and push them
* Ditch CircleCI

This increases our build time by about 5m per run, but VASTLY simplifies our CI configuration and removes a service we use entirely. I'm also convinced we can get some of the time back with rails' test suite parallelization feature, which we haven't turned on yet (because of mongodb).

lomky (approved):
All glory to GitHub, may all our eggs live in this basket 😂😭

colinxfleming:
whew

PR #2364, opened 2021-09-07
do builds on main

colinxfleming:

missed a spot! We want to trigger builds on main too, since that's what triggers deploys to heroku.


colinxfleming:
godmode because I want to get our deploy pipeline working again

PR #2368, opened 2021-09-09
DRAFT query improvements

elimbaum:

Continuation of #2225. However it looks like some of them weren't necessarily n+1 issues, just queries that were in need of some optimizing.

\~*placeholder while we work on this*~

still to do:
- configs helpers are doubling config page queries
- other @colinxfleming notes:
  - call_listable
  - patients_controller

* Removes duplicate config queries by leveraging `find_or_create_by` and `map`
* (another change here)
* (etc)

* Fixes #X
* Bumps #Y

For reviewer:
* Tag this PR:


app/controllers/configs_controller.rb
```diff
@@ -2,9 +2,9 @@ class ConfigsController < ApplicationController
   before_action :confirm_admin_user
 
   def index
-    # n+1 join here
-    Config.autosetup
-    @configs = Config.all.sort_by(&:config_key)
+    @configs = Config.config_keys.keys.map do |field|
+      Config.find_or_create_by(config_key: field)
```
colinxfleming, line 6: I'm sorta shocked this is the only spot we do autosetup in!

What do you think about the 'make sure all configs are there on application start' route? I think if we add this to development.rb and production.rb we can get away with it, and do your config map here:

```
  config.after_initialize do
    Config.autosetup
  end
```

colinxfleming (commented):
one down two to go! left a thought and I'll poke at this a bit more tmw

app/controllers/configs_controller.rb
elimbaum: sounds like a plan! so are those two files the place the do "start-of-app" things?

elimbaum: Have this working (will push commit in a few) but realizing one side effect of doing this is that any changes to the configs class require a server reboot - this could make development tricky. we could always turn this ON for prod and OFF for dev.

elimbaum:
Stumbled across another one we might want to address. Dashboard calls `most_recent_note_display_text` and `most_recent_note`:

https://github.com/DCAFEngineering/dcaf_case_management/blob/4ccbb48874aef6f77130b1d7a5399e769b0c1f77/app/views/dashboards/_table_content.html.erb#L46

but `most_recent_note_display_text` also calls `most_recent_note`!

https://github.com/DCAFEngineering/dcaf_case_management/blob/4ccbb48874aef6f77130b1d7a5399e769b0c1f77/app/models/concerns/notetakeable.rb#L5-L15

So we get two identical queries per patient per dashboard load. We can set an instance variable in the first function and just return it in the second, but that introduces a messy call-order dependency. Or, maybe we can set an instance variable in the template itself, and pass that along to a helper function?

elimbaum:
@colinxfleming should we close this? any other work you wanted to do?

colinxfleming:
this is a minor scope mess at this point (sorry, that's on me). In the spirit of 'if it ain't broke don't fix it' let's close this.

PR #2369, opened 2021-09-18
Ruby 3!

### LINKED ISSUE #2255: Upgrade to Ruby 3!

Now that the first bugfix on Ruby 3 is out, let's stay modern.

Keep up with Matz, having a Christmas in July.

2.7 should have started giving us warnings for anything incompatible with 3.0 - So muck around in the app and ID any warnings, address them. Once it looks clean? BUMP IT!

* **Anything else?**
Let's let the Rails bump complete first - #2251

lomky:
Upped some numbers! nbd

* Upgrades Ruby

* Fixes #2255 

For reviewer:
* Tag this PR:


colinxfleming (commented):
Incredibly, I went through a decent sample of core endpoints (login, line selection, patient search, logging a call, patient update, pledge sending, config update, clinic update, user lock, some accounting stuff) and not only did they all work great, they didn't so much as complain about a warning! I think your work earlier this year prepping for this made this pretty straightforward, huh? A++

I think you're right that we should have some else fire it up in docker and check, and I want to do this as its own deploy preferably, but imo we can fire it off when ready.

colinxfleming (approved):

elimbaum:
docker builds perfectly! (both clean repo and existing repo on the branch) will spend some time going through the workflows but nothing pops out yet!

PR #2384, opened 2021-09-27
Set up acts_as_tenant

### LINKED ISSUE #1817: Is having DARIA be a single tenant application still serving us?
TK

colinxfleming:

This sets up the acts_as_tenant gem, to enable us to do multitenancy.

The general approach here is to move a lot of the stuff that's in env vars now to a `Fund` object, and read from that rather than env vars. The rest of it is, frankly, mostly grepping around for `FUND` and removing that in favor of some form of `current_tenant.name`. (Exceptions to this are lines - more on lines later - and fund-mailer-domain, which has been daria-services for a little while now.) The diff is large, but most of it turned out to be find and replace.

Deployment-wise, what this will mean in the immediate term is that every fund's instance will get turned into multitenant apps with only one tenant. We'll need to declare a brief downtime to reconfigure, but the two step plan is basically:

* Create a Fund object with the proper subdomain set
* Run some sql to set every row in every table's fund_id to 1

This approach means we can punt on lines for now, but most likely we'll need to retool to make Line a proper object. (Having line be set on the `Fund` level doesn't work super well because of the enums and validation on other objects right now.) But we can roll this out as is without a problem.

To test:
* try the migration strategy: 

```ruby
git checkout main && rails db:drop db:create db:migrate db:seed && git checkout tenants-2 && rails db:migrate
echo "Fund.create name: 'CatFund', domain: 'supercat', subdomain: 'fluffy', full_name: 'Catte Fund" | rails c
sql="
begin;
update archived_patients set fund_id = 1;
update call_list_entries set fund_id = 1;
update calls set fund_id = 1;
update clinics set fund_id = 1;
update configs set fund_id = 1;
update events set fund_id = 1;
update external_pledges set fund_id = 1;
update fulfillments set fund_id = 1;
update notes set fund_id = 1;
update patients set fund_id = 1;
update practical_supports set fund_id = 1;
update users set fund_id = 1;
update versions set fund_id = 1;
commit;
"
echo $sql | rails db
```

* spin up the app as normal - `git checkout tenants-2 && rails db:drop db:create db:migrate db:seed` then `rails s` and try logging in as normal - localhost:3000 - to confirm you can hit the first fund
* log in to the alt development tenant - http://catbox.lvh.me:3000 to confirm you can hit the second fund

* sets up acts as tenant
* adjusts a lot of stuff to be tenant

no significant view changes

* bumps #1817 

For reviewer:
* Tag this PR:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,9 @@ gem 'devise', '~> 4.8'
 gem 'omniauth-google-oauth2', '~> 1.0.0'
 gem "omniauth-rails_csrf_protection", '~> 1.0'
 
+# Run multiple funds on one server
+gem 'acts_as_tenant', '~> 0.5.0'
+
 # Strong Password for user password validation for folks not on oauth
 gem 'strong_password', '~> 0.0.10'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -67,6 +67,9 @@ GEM
       minitest (>= 5.1)
       tzinfo (~> 2.0)
       zeitwerk (~> 2.3)
+    acts_as_tenant (0.5.1)
+      rails (>= 5.2)
+      request_store (>= 1.0.5)
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
     afm (0.2.2)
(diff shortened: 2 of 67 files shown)


colinxfleming:
per discussion/self-realization, breaking this into two so it's much easier to review

colinxfleming:
splitting this across two PRs instead

PR #2403, opened 2021-10-03
ActsAsTenant pt 2: Use fund objects instead of env vars for full name, phone, etc

colinxfleming:

Since everything being on one box will mean that env var config is no longer feasible without things getting messy, this ports the env var attributes onto the fund object.

In one instance, we squash the lone remaining fax env var - this should be user settable now, so feels clean to remove. (I spot checked a fund's config/env var and made sure they were the same, to be safe - looked clean!)

There's an alt world where we make these user settable configs, but they'll change infrequently enough that it felt live overkill to me.

I don't have great suggestions to review this - maybe spot check a couple of the core views? Shouldn't be any real diffs.

* Move full_name, site_domain, phone to Fund model and remove the env vars/constants
* remove all instances of these constants in the codebase, in favor of looking at the Fund object directly

no viewchanges

* builds on #2402
* Bumps #1817 

For reviewer:
* Tag this PR:


app/models/concerns/statusable.rb
```diff
@@ -19,7 +19,7 @@ module Statusable
                  help_text: I18n.t('patient.status.help.fulfilled')},
     dropoff: { key: I18n.t('patient.status.key.dropoff'),
                help_text: I18n.t('patient.status.help.dropoff')},
-    resolved: { key: I18n.t('patient.status.key.resolved', fund: FUND),
+    resolved: { key: I18n.t('patient.status.key.resolved', fund: ActsAsTenant.current_tenant&.name),
```
colinxfleming, line 22: Using the lonely operator here because tenant isn't resolved on startup, just on request. Need to check that it works properly before bringing out of draft.

app/services/pledge_form_generator.rb
```diff
@@ -1,12 +1,13 @@
 class PledgeFormGenerator
   include ActionView::Helpers::NumberHelper # for currency formatting
 
-  attr_reader :patient, :user, :case_manager_name
+  attr_reader :patient, :user, :case_manager_name, :fund
```
colinxfleming, line 4: Adjusting this is one of the few heavier things here. Test suite should keep us mostly safe but want to generate a pledge tobe sure.

app/views/patients/_pledge.html.erb
```diff
@@ -33,9 +33,9 @@
 
     <div class="hide" data-step="3" data-title="<%= t('patient.pledge.step_three.title') %>">
       <% unless ENV['DARIA_PLEDGE_GENERATOR_DISABLED'] %>
-        <p class="row"><%= t('patient.pledge.step_three.generated', fund: "#{FUND}") %></p>
+        <p class="row"><%= t('patient.pledge.step_three.generated', fund: current_tenant.name) %></p>
         <p class="row">
-          <%= t('patient.pledge.step_three.downloads') %> <%= link_to FUND_FAX_SERVICE, "http://#{FUND_FAX_SERVICE}", target: '_blank', rel: 'noopener nofollow' %>
+          <%= t('patient.pledge.step_three.downloads') %> <%= fax_service %>
```
colinxfleming, line 38: This was the only place we were using this env var constant, andit probably should be the config var regardless.

colinxfleming (commented):
a few notes

app/models/concerns/statusable.rb
colinxfleming: sure does! 
image

app/services/pledge_form_generator.rb
colinxfleming: Generated a pledge, confirmed it worked!

app/views/patients/_pledge.html.erb
colinxfleming: confirmed this was okay

db/schema.rb
```diff
@@ -195,6 +195,9 @@
     t.string "domain"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.string "full_name"
+    t.string "site_domain"
```
lomky, line 199: i'm gunna forget what this is - can we add [column comments](https://www.postgresql.org/docs/current/sql-comment.html)?

db/seeds.rb
```diff
@@ -27,13 +27,19 @@
 password = 'AbortionsAreAHumanRight1'
 
 # Create a few test funds
-fund1 = Fund.create! name: 'CatFund',
+fund1 = Fund.create! name: 'DCAF',
+                     domain: 'dcabortionfund.org',
+                     subdomain: 'sandbox',
+                     full_name: 'DC Abortion Fund',
+                     site_domain: '.dcabortionfund.org',
```
lomky, line 34: why the leading `.`?

test/integration_helper.rb
```diff
@@ -71,7 +71,7 @@ def _finished_all_ajax_requests?
   end
 
   def go_to_dashboard
-    click_on "DARIA - #{FUND_FULL}"
+    click_on "DARIA - #{ActsAsTenant.current_tenant.name}"
```
lomky, line 74:
```suggestion
    click_on "DARIA - #{ActsAsTenant.current_tenant.full_name}"
```

lomky (commented):
Let's snag the docs that need updated here! I catch `app.json` & `SETTING_UP_A_NEW_INSTANCE.md` with a grep

db/seeds.rb
colinxfleming: probably a mistake! fixing

app/services/pledge_form_generator.rb
```diff
@@ -92,34 +93,34 @@ def generate_pledge_pdf
   def build_header(pdf)
     y_position = pdf.cursor
     pdf.bounding_box([0, y_position], width: 200, height: 100) do
-      pdf.image "#{Rails.root.join("public", "#{FUND.downcase}_logo.png")}",
+      pdf.image "#{Rails.root.join("public", "#{@fund.name.downcase}_logo.png")}",
                 position: :left,
-                width: CONFIG_DATA[FUND][:width],
-                height: CONFIG_DATA[FUND][:height]
+                width: CONFIG_DATA[@fund.name][:width],
+                height: CONFIG_DATA[@fund.name][:height]
```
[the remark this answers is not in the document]
colinxfleming: okay, here's where this wound up:
* add a config setting to the Fund object
* adjust pledge generator + the view to use that setting
* change the seeds so that pledge generation is turned off by default

You can try it out by running `Fund.update_all pledge_generation_config: 'DCAF'`. It'll be a little weird since some of the config is in the PDF class and some of it is in tenant, but it'll come out clean in production.

I'll also note that pledge generation is kind of overdue for an overhaul - I think this strikes a balance between 'getting this working now' and also 'still keeps things open for further development'.

PR #2409, opened 2021-10-21
nuke all reports related code

colinxfleming:

We haven't used this endpoint for awhile, and it's only gotten more obsolete over time. Now it's especially obsolete - still hooks into mongo. I think we can safely just blow this all away.

* nuke reporting routes
* nuke all reporting code

no view changes
no issues, but came up during porting Lines out of an env var and into a model

For reviewer:
* Tag this PR:


lomky (approved):
🧹🧹🧹

PR #2417, opened 2021-10-28
multiple validations and singletons

### LINKED ISSUE #2194: Singleton Configs act funny with commas

Parse singleton values (integers, URLS, etc) in configs.

Budget bar, archive days (upcoming), URL handling, few other single-word configs.

Even though the help text says to enter numbers without commas, if you do enter a comma (i.e. budget bar max = `10,000`), it seems like that's being interpreted as a list (so budget bar max = `[10, 0]`). This *is* the expected behavior given the current code, so it's not really a bug, but it may make sense to add validation for singleton values versus things that _could be_ a list.

This also applies to URLs, start of week, hide practical support, etc... if you enter `https://cats,dogs.com` as a URL, it will _successfully validate_ and turn into `https://dogs.com` because the URL handling code calls `options.last`. I would argue that, given that input, we should fail to validate, rather than succeed in the way we are. (I could see this happening in practice with a minor typo: `sub.domain.com => sub,domain.com => domain.com`)
 
* **Anything else?**

I know this could make things more complicated because it will bifurcate our config handling. So may not be a priority issue; just came across it while adding new config fields.

elimbaum:

This PR extends the config validation system to allow multiple validators per field. It also validates that singleton values aren't lists.

* Adds a new `validate_singleton` validator
* Turns validators from single functions to lists of functions
* Adds a few tests (let me know if we should test every field, rather than just a few)

* Fixes #2194 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -44,24 +44,34 @@ class Config < ApplicationRecord
 
   # symbols are required here because functions are not objects in rails :)
   CLEAN_PRE_VALIDATION = {
-    start_of_week: :fix_capitalization,
-    hide_practical_support: :fix_capitalization,
+    start_of_week: [:fix_capitalization],
+    hide_practical_support: [:fix_capitalization],
+    language: [:fix_capitalization]
   }.freeze
 
   VALIDATIONS = {
-    start_of_week: :validate_start_of_week,
+    start_of_week:
+      [:validate_singleton, :validate_start_of_week],
 
-    hide_practical_support: :validate_hide_practical_support,
+    hide_practical_support:
+      [:validate_singleton, :validate_hide_practical_support],
 
-    budget_bar_max: :validate_number,
+    budget_bar_max:
+      [:validate_singleton, :validate_number],
     
-    resources_url: :validate_url,
-    fax_service: :validate_url,
-    practical_support_guidance_url: :validate_url,
+    resources_url:
+      [:validate_singleton, :validate_url],
+    fax_service:
+      [:validate_singleton, :validate_url],
+    practical_support_guidance_url:
+      [:validate_singleton, :validate_url],
 
-    days_to_keep_fulfilled_patients: :validate_patient_archive,
-    days_to_keep_all_patients: :validate_patient_archive,
-    urgent_reset: :validate_urgent_reset,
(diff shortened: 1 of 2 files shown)


app/models/config.rb
```diff
@@ -134,15 +144,15 @@ def clean_config_value
...
 
       # no clean function, return
-      return if cleaner.nil?
+      return if cleaners.blank?
 
       # we need to use `method` because `cleaner` is a symbol... this converts
       # the symbol into a Method object, which we can then `call`...
       # See https://ruby-doc.org/core/Object.html#method-i-method
-      method(cleaner).call
+      cleaners.map { |cleaner| method(cleaner).call }
```
colinxfleming, line 155: Pretty minor thing here - generally when I see `map` a reader (for example, me!) is expecting to see some values manipulated and returned. I think in this case the methods are all doing mutation rather than returning values, right? So we can use `each`, which will signal to the reader we don't really care what the return value is, and the interesting stuff is happening via mutation.

test/models/config_test.rb
```diff
@@ -41,6 +41,25 @@ class ConfigTest < ActiveSupport::TestCase
       assert @bad_config.errors.messages[:config_key].include? "can't be blank"
     end
 
+    it 'should reject lists for singleton values' do
```
colinxfleming, line 44: 👨‍🍳

colinxfleming (approved):
This PR is a lot of your thoughtful work paying off really beautiflully. Love this approach. and how smooth it is! One minor signal-to-other-writers suggestion in here around use of map over each, and then I think we can get this out. Outstanding work.

test/models/config_test.rb
elimbaum: are you ok with how this is structured (i.e. test of the functionality itself but not actually checking that all the fields validate in that way)? or flip it around?

app/models/config.rb
elimbaum: oops derp yes

test/models/config_test.rb
colinxfleming: nah I think this is a good test - taking a pair of representative examples is I think a good enough proveit here.

PR #2435, opened 2021-11-21
Don't parallelize system tests in docker

elimbaum:

In #2348 we parallelized our system tests, but this broke things in Docker. :'( There may be a more clever to get the parallel tests working in docker, but since we'll probably only be running system tests locally for individual debugging, the time is less of a concern, and I think this quick fix is appropriate for the time being. This should _not_ affect non-docker (bare metal, CI) execution.

* adds a check for docker environment
* (another change here)
* (etc)

* Fixes #X
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -32,7 +32,7 @@ class ActiveSupport::TestCase
     teardown_tenant
   end
 
-  parallelize(workers: :number_of_processors)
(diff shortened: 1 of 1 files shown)


colinxfleming (approved):

colinxfleming:
thanks for digging into this!

PR #2436, opened 2021-11-22
Save test artifacts

xmunoz:
I rule and have completed some work on Case Manager that's ready for review!

This pull request makes the following changes:
* Stores test coverage artifact
* Stores screenshot artifacts

It relates to the following issue #s: 
* Fixes #2430 

For reviewer:
* Check CI and make sure you can download artifacts (e.g. https://github.com/DARIAEngineering/dcaf_case_management/actions/runs/1492613843#artifacts)

PR #2437, opened 2021-11-23
Conditionally run artifact storage steps in test job

xmunoz:
We should always preserve the coverage artifacts. The screenshot artifacts, however, should only get stored when a job fails. Follow up from #2436


colinxfleming (approved):
lgtm, merge when ready!

PR #2442, opened 2021-11-26
declare asset site url specifically via env var

colinxfleming:

for multitenant, we need to be able to point the asset loader at a particular url; this lets us do that via environment variable, with a fallback to current behavior.

(This is because otherwise in a multitenant environment it'll try to point it at a domain that isn't set up.)

* lets us dynamically set up asset urls for multitenant envs

no view changes

* Bumps #1817 

For reviewer:
* Tag this PR:

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -53,7 +53,7 @@ Rails.application.configure do
   # Prevent host header injection
   config.action_controller.default_url_options = { host: ENV['SITE_URL'] }
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.action_controller.asset_host = ENV['SITE_URL']
+  config.action_controller.asset_host = ENV['ASSET_SITE_URL'] || ENV['SITE_URL']
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   # config.active_job.queue_adapter     = :resque
--- a/app.json
+++ b/app.json
@@ -54,6 +54,10 @@
       "description": "The app name you just selected plus .herokuapp.com. e.g. daria-xaf.herokuapp.com",
       "value": ""
     },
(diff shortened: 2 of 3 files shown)


config/environments/production.rb
```diff
@@ -53,7 +53,7 @@
   # Prevent host header injection
   config.action_controller.default_url_options = { host: ENV['SITE_URL'] }
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.action_controller.asset_host = ENV['SITE_URL']
```
lomky, line 56: is this a two-part roll out where eventually we can drop the `|| ENV['SITE_URL']`, or do we always want the fall back for local dev?

lomky (approved):

lomky:
Do we have a go-live checklist for multitenant? I'm a little worried about losing track of the smaller things to change, like this ENV variable setup tweak

lomky (commented):
Oop - also #2443 reminded me that `app.json` exists - do you want to add this new var there?

colinxfleming:
(I made a CSP adjustment and am currently fighting with some redirect logic)

config/environments/production.rb
colinxfleming: I think keep the fallback for the time being - this is pretty specifically for multitenant which otherwise tries to grab assets from weird places. Maybe we'll revisit after we've consolidated.

colinxfleming:
I'll post more about multitenant as we go, but the short of it is that I'm messing around with a separate production-shaped environment (no real data loaded in). There aren't that many real structural changes to make from other environments, thankfully! but there are a couple odds and ends like this. I'll ensure that we have a working staging and prod environment before we load anybody into it.

PR #2443, opened 2021-11-26
Use model attribute rather than env var for pledge generation

### LINKED ISSUE #2439: Remove remaining refs to DARIA_PLEDGE_GENERATOR_DISABLED

another multitenancy fix

* **Anything else?**

colinxfleming:

We store this on the model now, so we should use it, eh?

(Note that 'default' in the test is a deliberate choice, coming soon!)

* use model attribute instead of env var for displaying pledge generator

no view changes

* Fixes #2439

For reviewer:
* Tag this PR:


lomky (approved):
🎉 :shipit:

PR #2444, opened 2021-11-27
Add script to safely merge daria databases

colinxfleming:

We are ready to go here! The idea here is:
* we make a new fund in the instance
* we set that fund's id and that fund's current db creds as env vars
* we run something like `rails migrate_db_multitenant`

This then copies db rows over and offsets everything to preserve relations/refs, including in papertrail. It's pretty fast (I've been running it against a prod db for some stress testing and it takes roughly a minute). 

How to test drive:
* run `rails db:seed`, then run this sql to clean out your db entirely except for funds: `delete from call_list_entries; delete from events; delete from patients; delete from archived_patients; delete from configs; delete from clinics; delete from users; delete from lines; delete from calls; delete from versions; delete from notes;  delete from practical_supports; delete from external_pledges; delete from fulfillments; `
* get the first fund id with `echo "Fund.first.id" | rails c`
* `export MIGRATION_FUND_ID=(that id)`
* `export MIGRATION_DB_URL=(sandbox db url, assuming sandbox is set up to have one tenant currently)`
* `rails multitenant_db_merge`
* reset the password of test@example.com
* log in to your local and ensure it has parity with the remote db you just loaded in. it should!

* postgres migration script

no view changes

* Bumps #1817 

For reviewer:
* Tag this PR:


colinxfleming:
one remaining problem: created_at / updated_at are -5h as they're getting converted to the same time in eastern. an easy fix: just add 5 hours to every timestamp?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+
+  # Port patients
+  patient_map = -> (x) {
+    x.except('id', 'fund_id', 'created_at', 'updated_at', 'line_id', 'clinic_id', 'pledge_generated_by_id', 'pledge_sent_by_id', 'last_edited_by_id')
+     .merge({
+       'fund_id' => @fund_id,
+       'created_at' => x['created_at'].asctime.in_time_zone("America/New_York"),
+       'updated_at' => x['updated_at'].asctime.in_time_zone("America/New_York"),
+       'clinic_id' => @clinic_mappings[x['clinic_id']],
+       'line_id' => @line_mappings[x['line_id']],
```
lomky, line 133: nit to make eyeballing to the except easier
```suggestion
       'line_id' => @line_mappings[x['line_id']],
       'clinic_id' => @clinic_mappings[x['clinic_id']],
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+
+  # Port archived patients
+  archived_patient_map = -> (x) {
+    x.except('id', 'fund_id', 'created_at', 'updated_at', 'line_id', 'clinic_id', 'pledge_generated_by_id', 'pledge_sent_by_id', 'last_edited_by_id')
+     .merge({
+       'fund_id' => @fund_id,
+       'created_at' => x['created_at'].asctime.in_time_zone("America/New_York"),
+       'updated_at' => x['updated_at'].asctime.in_time_zone("America/New_York"),
+       'clinic_id' => @clinic_mappings[x['clinic_id']],
+       'line_id' => @line_mappings[x['line_id']],
```
lomky, line 149: nit
```suggestion
       'line_id' => @line_mappings[x['line_id']],
       'clinic_id' => @clinic_mappings[x['clinic_id']],
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+       'pledge_generated_by_id' => @user_mappings[x['pledge_generated_by_id']],
+       'pledge_sent_by_id' => @user_mappings[x['pledge_sent_by_id']],
+       'last_edited_by_id' => @user_mappings[x['last_edited_by_id']]
+     })
+  }
+  @patient_mappings = easy_mass_insert Patient, 'patients', patient_map
+
+  # Port archived patients
+  archived_patient_map = -> (x) {
+    x.except('id', 'fund_id', 'created_at', 'updated_at', 'line_id', 'clinic_id', 'pledge_generated_by_id', 'pledge_sent_by_id', 'last_edited_by_id')
```
lomky, line 143: This has `last_edited_by_id` excepted, but not included back in in the merge below?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+        'updated_at' => x['updated_at'].asctime.in_time_zone("America/New_York"),
+        'can_call_id' => if x['can_call_type'] == 'Patient'
+                           @patient_mappings[x['can_call_id']]
+                         elsif x['can_call_type'] == 'ArchivedPatient'
+                           @archived_patient_mappings[x['can_call_id']]
+                         else
+                           raise "unexpected type - row #{x}"
+                         end
+      })
+    res['can_call_id'].nil? ? nil : res
```
lomky, line 173: What's this line doing? Is it dropping abandoned subobjects by having the lamda return nil?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+    'User' => @user_mappings,
+    'PracticalSupport' => @practical_support_mappings,
+  }
+
+  def transform_obj(obj, item_type, value_will_be_array)
+    obj.each_pair do |k, v|
+      obj['id'] = @item_type_mappings[item_type][v] if k == 'id'
+      if @fkey_mappings.keys.include? k || (k.start_with?('can_') && k.end_with?('_id'))
+        keyset = if k == 'can_call_id'
+                   obj['can_call_type'] == 'Patient' ? @patient_mappings : @archived_patient_mappings
```
lomky, line 312: Why do we trust this to be either/or now, unlike up in the subobjects? are we more certain we won't find any abandoned ones here?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+  @call_mappings = {}
+  @external_pledge_mappings = {}
+  @fulfillment_mappings = {}
+  @practical_support_mappings = {}
+  @note_mappings = {}
+  @patient_mappings = {}
+  @archived_patient_mappings = {}
+  @call_list_entry_mappings = {}
+  @event_mappings = {}
+
```
lomky, line 52:
```suggestion

  # object agnostic function for handling each insert 
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+             'created_at' => x['created_at'].asctime.in_time_zone("America/New_York"),
+             'updated_at' => x['updated_at'].asctime.in_time_zone("America/New_York"),
+             'line_id' => @line_mappings[x['line_id']],
+             'patient_id' => @patient_mappings[x['patient_id'].to_i],
+           })
+    res['patient_id'].nil? ? nil : res
+  }
+  @event_mappings = easy_mass_insert Event, 'events', event_map, false
+
+  # Versions, maybe a little more complicated...
```
lomky, line 276: I'm gunna need a bit more verbosity from you here friend 😂 Can you give me a bit more on how you intend the mappings and the helper function feed into the versions port lambda?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+
+  # Transfer versions, porting keys along the way
+  version_map = -> (x) {
+    res = x.except('id', 'created_at', 'item_id', 'fund_id', 'object', 'object_changes', 'whodunnit')
+           .merge({
+             'item_id' => @item_type_mappings[x['item_type']][x['item_id']],
+             'created_at' => x['created_at'].asctime.in_time_zone("America/New_York"),
+             'whodunnit' => x['whodunnit'].nil? ? nil : @user_mappings[x['whodunnit'].to_i],
+             'object_changes' => x['object_changes'].nil? ? nil : transform_obj(JSON.parse(x['object_changes']), x['item_type'], true),
+             'object' => x['object'].nil? ? nil : transform_obj(JSON.parse(x['object']), x['item_type'], false)
```
lomky, line 337: nit reorder
```suggestion
             'object' => x['object'].nil? ? nil : transform_obj(JSON.parse(x['object']), x['item_type'], false)
             'object_changes' => x['object_changes'].nil? ? nil : transform_obj(JSON.parse(x['object_changes']), x['item_type'], true),
             'whodunnit' => x['whodunnit'].nil? ? nil : @user_mappings[x['whodunnit'].to_i],
```

lomky (commented):
Some thoughts on the code!

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
+desc 'Merge an ext daria db into main db'
+task multitenant_db_merge: :environment do
+  raise 'SET MIGRATION_FUND_ID ENV VAR' if not ENV['MIGRATION_FUND_ID']
+  raise 'SET MIGRATION_DB_URL ENV VAR' if not ENV['MIGRATION_DB_URL']
+
+  # Make absolutely sure the new fund has been created!
+  fund = Fund.find ENV['MIGRATION_FUND_ID']
+  @fund_id = fund.id
```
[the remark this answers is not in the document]
colinxfleming: `find` does this naturally!

```
irb(main):003:0> Fund.find 999
  Fund Load (1.7ms)  SELECT "funds".* FROM "funds" WHERE "funds"."id" = $1 LIMIT $2  [["id", 999], ["LIMIT", 1]]
/Users/colin/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/activerecord-6.1.4.4/lib/active_record/core.rb:338:in `find': Couldn't find Fund with 'id'=999 (ActiveRecord::RecordNotFound)
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
+desc 'Merge an ext daria db into main db'
+task multitenant_db_merge: :environment do
+  raise 'SET MIGRATION_FUND_ID ENV VAR' if not ENV['MIGRATION_FUND_ID']
+  raise 'SET MIGRATION_DB_URL ENV VAR' if not ENV['MIGRATION_DB_URL']
```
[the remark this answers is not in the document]
colinxfleming: My plan is to pretty much run this as a rake on heroku dynos, a la `heroku run rails multitenant_db_merge -a ...` after setting particular env vars. Not worried about spillover - to avoid this we'll write the playbook to explicitly unset the env vars, I think.

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+  # Database we're migrating things out of
+  def connect_to_migration_db
+    ActiveRecord::Base.establish_connection(
+      adapter: 'postgresql',
+      encoding: 'unicode',
+      pool: 10,
+      url: ENV['MIGRATION_DB_URL']
+    )
+    ActiveRecord::Base.connection.execute(
+      "set session characteristics as transaction read only;"
```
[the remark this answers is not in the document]
colinxfleming: postgres: it's good

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,344 @@
...
+    )
+  end
+
+  puts "Migrating into fund_id #{@fund_id}..."
+
+  if Patient.count() > 0
+    raise 'Possibly already ported? patients belonging to that fund_id are found'
```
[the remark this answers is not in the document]
colinxfleming: It's frankly not pretty! But we'd probably set tenant and `Patient.destroy_all`, `Clinic.destroy_all`, etc to clean up. To avoid this we can dry run it in a local environment.  I agree that it's less than ideal but since this code will outlive its usefulness quickly, felt fine to me.

That said, we COULD stick this in a `begin/rescue` block and do the cleanout if we wanted to, though I think if this DID choke I'd want to open up a rails console and clean it up by hand, rather than letting therobots do it. idk, what do you think?

lib/tasks/multitenant_db_merge.rake
colinxfleming: yeah, I encountered a few cases of exactly this - abandoned subobjects. This guards against that case.

lib/tasks/multitenant_db_merge.rake
colinxfleming: nah, def possible that some of these will be abandoned, but the versions table is storing everything as json rather than references like on rails models, so it won't raise errors if it runs into an abandoned obj. In other words, here we're just manipulating numbers in json rather than ensuring foreign key relations are valid, so we can care a lot less.

lib/tasks/multitenant_db_merge.rake
colinxfleming: winds up being a noop -- `last_edited_by_id` isn't on archived patient, turns out. I'll take it out of here regardless.

lib/tasks/multitenant_db_merge.rake
colinxfleming: lol I took a whack, hopefully that's helpful

lib/tasks/multitenant_db_merge.rake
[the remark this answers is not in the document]
lomky: I want to defer to @xmunoz's comfort here, since the plan is to have y'all running it!

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+      url: ENV['MIGRATION_DB_URL']
+    )
+    ActiveRecord::Base.connection.execute(
+      "SET SESSION CHARACTERISTICS AS TRANSACTION READ ONLY;"
+    )
+  end
+
+  puts "Migrating into fund_id #{@fund_id}..."
+
+  if Patient.count() > 0
```
lomky, line 34: Let's check the first model we port, as well as the most important?

```suggestion
  if Patient.count() > 0 || Config.count() > 0
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+    'Line' => @line_mappings,
+    'Note' => @note_mappings,
+    'Patient' => @patient_mappings,
+    'User' => @user_mappings,
+    'PracticalSupport' => @practical_support_mappings,
+  }
+
+  def transform_obj(obj, item_type, value_will_be_array)
+    obj.each_pair do |k, v|
+      obj['id'] = @item_type_mappings[item_type][v] if k == 'id'
```
lomky, line 315:
```suggestion
      # Handle `id` fields
      obj['id'] = @item_type_mappings[item_type][v] if k == 'id'
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+      if @fkey_mappings.keys.include? k || (k.start_with?('can_') && k.end_with?('_id'))
```
lomky, line 316: Additionally, for clarity
```suggestion

      # Process the object mappings to handle polymorphic properly
      if @fkey_mappings.keys.include? k || (k.start_with?('can_') && k.end_with?('_id'))
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+                   obj[k] = obj['can_pledge_type'] == 'Patient' ? @patient_mappings : @archived_patient_mappings
+                 elsif k == 'can_fulfill_id'
+                   obj[k] = obj['can_fulfill_type'] == 'Patient' ? @patient_mappings : @archived_patient_mappings
+                 elsif k == 'can_support_id'
+                   obj[k] = obj['can_support_type'] == 'Patient' ? @patient_mappings : @archived_patient_mappings
+                 else
+                   @fkey_mappings[k]
+                 end
+
+        obj[k] = value_will_be_array ? v.map { |x| keyset[v] } : keyset[v]
```
lomky, line 329:
```suggestion
        # handle non-`id` fields with polymorphic-aware mappings
        obj[k] = value_will_be_array ? v.map { |x| polymorphic_aware_mappings[v] } : polymorphic_aware_mappings[v]
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+    'Patient' => @patient_mappings,
+    'User' => @user_mappings,
+    'PracticalSupport' => @practical_support_mappings,
+  }
+
+  def transform_obj(obj, item_type, value_will_be_array)
+    obj.each_pair do |k, v|
+      obj['id'] = @item_type_mappings[item_type][v] if k == 'id'
+      if @fkey_mappings.keys.include? k || (k.start_with?('can_') && k.end_with?('_id'))
+        keyset = if k == 'can_call_id'
```
lomky, line 317:
```suggestion
        polymorphic_aware_mappings = if k == 'can_call_id'
```

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+    'pledge_generated_by_id' => @user_mappings,
+    'pledge_sent_by_id' => @user_mappings,
+    'last_edited_by_id' => @user_mappings,
+    'user_id' => @user_mappings,
+    'line_id' => @line_mappings,
+    'patient_id' => @patient_mappings,
+    'can_call_id' => @patient_mappings,
+    'can_pledge_id' => @patient_mappings,
+    'can_fulfill_id' => @patient_mappings,
+    'can_support_id' => @patient_mappings
```
lomky, line 295: Do these actually get polymorphic mappings ever get referenced, or are they always preemted by the `if` at 317?

lib/tasks/multitenant_db_merge.rake
```diff
@@ -0,0 +1,350 @@
...
+    res['patient_id'].nil? ? nil : res
+  }
+  @event_mappings = easy_mass_insert Event, 'events', event_map, false
+
+  # Versions, maybe a little more complicated...
+  # Essentially, versions have a lot of fields we have to kick over, both at the top level (item_id, whodunnit)
+  # and nested (object, object_changes).
+  # This is essentially a more complicated version of the lambdas above, except it changes anything that's known to be an id
+  # to the new value that's just been inserted into db. End effect is that all the ids stowed in versions change to the new appropriate value.
+  @fkey_mappings = {
```
lomky, line 283: Assuming its true that this mapping shouldn't handle the polymorphic directly, I suggest a rename:

```suggestion
  @monomorphic_mappings = {
```

lomky (commented):
More Comments!

lib/tasks/multitenant_db_merge.rake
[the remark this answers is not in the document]
colinxfleming: In the spirit of expedience, let's assume that this is okay, and if we have concerns after the first few we can retcon that into the script?

lib/tasks/multitenant_db_merge.rake
colinxfleming: hm, this variable doesn't seem to actually get used (probably an orphan of debugging) - I'll just take it out.

lib/tasks/multitenant_db_merge.rake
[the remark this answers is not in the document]
colinxfleming: reasonable enough!

lib/tasks/multitenant_db_merge.rake
colinxfleming: they get used in two spots - the `if` on 317, and in the version mapping func, where we rewrite the version with the object's new `id` field.

lib/tasks/multitenant_db_merge.rake
colinxfleming: (am holding on changing this for now, don't want to thrash too much on variable naming)

lib/tasks/multitenant_db_merge.rake
lomky: It was used! You accepted a suggested change from me that did the sibling rename on line 317 that did the rename as well as add a comment

colinxfleming:
per offline discussion, I feel good about where this is at, so merging!

PR #2451, opened 2021-12-05
Adjust routing to pass subdomain on redirect + fix url in new user emails

colinxfleming:

This PR stems from whack-a-mole trying to get the staging environment to work. Here's the sitch:
* adjusts the rerouting logic to respect domain. previously, going to sandbox.dariaservices.org/configs would redirect you to just dariaservices.org/users/sign_in without the domain, which would 404, which makes urls less shareable. For some reason adding a shim into the devise logic corrects this. (I briefly considered using `authenticated`, which just 404s everything within if not logged in, but that again makes stuff not really shearable. what a world.) This current posture, I think, both respects the current state of the world and.
* after that, I invited my gmail to test out oauth and noted that the url was going to the wrong spot, so after some back and forth I figured out a way that will be respectful of both dev and prod environments. It does continue to lean on site url which isn't great but that is a real Later Problem. (Anyway, oauth works.) 

* correct subdomain redirect logic for unauthenticated hits
* fix invitation url

no view changes

* Bumps #1817 

For reviewer:
* Tag this PR:


colinxfleming:
(you can take this for a spin @ https://sandbox.dariaservices.com/ or https://catbox.dariaservices.com/ )

mdworken (approved):
Would prefer a test, but lg2m!

PR #2453, opened 2021-12-05
Explicitly declare form methods on patient edit page because for some reason they're trying to POST sometimes

colinxfleming:

In certain cases, when I load a patient edit page I'm getting forms that POST to the patient route rather than PATCH (which causes failures). This is because method is set to a CSRF token, for some reason I cannot fathom - maybe a rails_bootstrap_form behavior or something? See:

image

Here's what it should look like (note the proper hidden field `_method` value):

image

I have no idea why this is doing this, but explicitly declaring method seems to resolve for me in firefox.

* explicitly declare method=patch on patient edit page forms

no view changes
no issues

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/patients/_abortion_information.html.erb
+++ b/app/views/patients/_abortion_information.html.erb
@@ -12,6 +12,7 @@
   <%= bootstrap_form_with model: patient,
                           html: { id: 'abortion-information-form-1' },
                           remote: true,
+                          method: 'patch',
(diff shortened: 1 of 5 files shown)


DarthHater (approved):
These changes are innocuous, and just set the helper method explicitly. Curious why this is happening, but 🤷 , would probably file an issue on the project to see what they say!

PR #2454, opened 2021-12-06
Adjust nightly_cleanup to accommodate multitenant

colinxfleming:

as @lomky pointed out, some of the stuff in nightly_cleanup depends on fund-level configs, so we gotta scope some of these to fund. In general, if it's got a fund_id on the table, we do it in the fund loop now.

Others of these we can do all at once, like clinic coordinates updates, clearing out old events, or trimming db sessions, though, so those can stay out o the loop.

* do fund-specific cleanup subtasks in the context of a fund

no views

* Bumps #1817 

For reviewer:
* Tag this PR:


lib/tasks/nightly_cleanup.rake
```diff
@@ -1,27 +1,31 @@
...
-  puts "#{Time.now} -- cleared all recently reached patients from call lists"
+  Rake::Task['db:sessions:trim'].invoke
+  puts "#{Time.now} -- removed old sessions"
 
-  User.disable_inactive_users
-  puts "#{Time.now} -- locked accounts of users who have not logged in since #{User::TIME_BEFORE_DISABLED_BY_FUND.ago}"
+  Fund.all.each do |fund|
+    ActsAsTenant.with_tenant(fund) do
+      User.all.each { |user| user.clean_call_list_between_shifts }
+      puts "#{Time.now} -- cleared all recently reached patients from call lists"
```
lomky, line 10: Whatcha think about adding the fund to the logging msg?

```suggestion
      puts "#{Time.now} -- cleared all recently reached patients from call lists for fund #{fund}"
```

repeat for each?

lib/tasks/nightly_cleanup.rake
```diff
@@ -1,27 +1,31 @@
...
+      Event.destroy_old_events
+      puts "#{Time.now} -- destroyed old events"
 
-  Rake::Task['db:sessions:trim'].invoke
-  puts "#{Time.now} -- removed old sessions"
+      if Time.zone.now.monday?
+        # Run these events weekly
+        Clinic.update_all_coordinates
+        puts "#{Time.now} -- refreshed coordinates on all clinics"
+      end
```
lomky, line 25: Description says this one can be fund agnostic - but its per fund here?

colinxfleming:
wow was my brain paste when I wrote this one. bumping now, thank you @lomky 😅

lomky (approved):
🎉

PR #2456, opened 2021-12-08
redirect an env var is set, to point legacy systems to new urls

colinxfleming:

Hopefully one of the last bits of multitenant - this redirects to new urls if an env var is set. Plan is to temporarily keep the urls up for a month or two while people get used to the new one.

To test, `export REDIRECT_DESTINATION='https://www.google.com' and then try loading the app, and see yourself redirected.

* add REDIRECT_DESTINATION env var that will send people to new domains when they're set up

no view changes

* Bumps #1817

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -12,12 +12,20 @@ class ApplicationController < ActionController::Base
   if Rails.env.development?
     before_action :confirm_tenant_set_development
   end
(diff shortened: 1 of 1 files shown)


colinxfleming:
ps I'm doing this because I don't want to deal with DNS in heroku

xmunoz (approved):
Makes sense to me!

PR #2463, opened 2021-12-19
allow practical supports to be decimals

### LINKED ISSUE #2452: Change practical support amount to a decimal

I guessed wrong on this one. Let's change this int to a decimal.

accurate expense tracking. oops.

* migrate the db field to a two digit decimal
* bump the input validations

* **Anything else?**
nah

colinxfleming:

A fund noted to me that practical support amounts weren't as clean as pledge amounts, and cents mattered, so let's make this a decimal.

@elimbaum as our local expert on the practical support code, would you mind pulling this down and reviewing? 

* allows practical support amount to be a decimal in db
* ...and in form

image

* Fixes #2452 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/practical_supports/_edit.html.erb
+++ b/app/views/practical_supports/_edit.html.erb
@@ -19,6 +19,7 @@
 
       <div class="col-4">
         <%= f.number_field :amount,
+                           value: number_to_currency(f.object.amount.to_f, unit: '', delimiter: ''),
                            hide_label: true,
                            autocomplete: 'off',
                            prepend: '$',
--- a/app/views/practical_supports/_edit.html.erb
+++ b/app/views/practical_supports/_edit.html.erb
@@ -21,7 +21,8 @@
         <%= f.number_field :amount,
                            hide_label: true,
                            autocomplete: 'off',
-                           prepend: '$' %>
+                           prepend: '$',
+                           step: '0.01' %>
       </div>
 
       <div class="col-2">
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_18_014015) do
+ActiveRecord::Schema.define(version: 2021_12_19_044530) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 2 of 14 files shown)


db/schema.rb
```diff
@@ -299,7 +299,7 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "fund_id"
-    t.integer "amount"
+    t.decimal "amount", precision: 8, scale: 2
```
elimbaum, line 302: when i was a wee youth i was raised to never represent currency as a float, lest you get floating point errors in your math (instead using integer cents). are we now a sufficiently advanced society that float currency is ok? is that the idea of the `precision`/`scale`?

app/views/practical_supports/_edit.html.erb
```diff
@@ -21,7 +21,8 @@
         <%= f.number_field :amount,
                            hide_label: true,
```
elimbaum, line 22:
```suggestion
                           value: number_to_currency(f.object.amount.to_f, unit: '', delimiter: ''),
                           hide_label: true,
```

This is a bit gross, but works. Rails's `number_to_currency` is meant to do all the work for you, but for number fields, we need to drop the unit and delimiter (`,`). This displays 2 decimal units in the edit view.

This of course doesn't work while making changes (and therefore we can't use it on the `new` partial) but I think that's a reasonable compromise. Then we can also make our tests handle currency normally. :)

elimbaum: voila
Screen Shot 2021-12-19 at 18 35 18

test/system/practical_support_behaviors_test.rb
```diff
@@ -21,15 +21,15 @@ class PracticalSupportBehaviorsTest < ApplicationSystemTestCase
...
+        fill_in 'Amount', with: 500.10
         check 'Confirmed'
         click_button 'Create new practical support'
       end
 
       within :css, '#practical-support-entries' do
         assert_equal 'Companion', find('#practical_support_support_type').text
         assert_equal 'Other (see notes)', find('#practical_support_source').value
-        assert_equal '500', find('#practical_support_amount').value
+        assert_equal '500.1', find('#practical_support_amount').value
```
elimbaum, line 32: if you like the `number_to_currency` change then we can fix this!

elimbaum (approved):

app/views/practical_supports/_edit.html.erb
elimbaum: also i found it here: https://stackoverflow.com/questions/18898947/rails-number-field-alternative-for-decimal-values

db/schema.rb
colinxfleming: yeah! There is mercifully a difference between floats and decimals (also called numerics) in postgres. Decimals do not do any of the unholy stuff that floats are notorious for.  https://www.postgresql.org/docs/9.5/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

(Though honestly, there's a `money` datatype too, I'll screw around with that a bit in a commit.)

Anyway, it's not that we're an advanced enough society that float currency is okay, just that we're an advanced enough society that we no longer use float most of the time (I think).

app/views/practical_supports/_edit.html.erb
colinxfleming: minor holiday miracle in here. Thank you for doing this digging!

PR #2490, opened 2022-01-23
Show practical support, ext pledges, and fulfillments in changelog

### LINKED ISSUE #826: Fields in fulfillment, ext pledge not showing up in patient changelog
Issue description TK -- affects $$ fields, resolved without DCAF.

Should also do this for external pledges

colinxfleming:

Since we're off mongo this is a bit more straightforward (ed note: me and my big mouth)! Required some adjustments to i18n though. Made some quality-of-life improvements to the changelog while I was in there.

* Display changes for external pledges, practical supports, and fulfillments in the patient changelog
* clean up some long-running display issues (missing keys + null changesets)
* add keys to i18n registry

The display isn't perfect but they're likely deducible from context clues to a CM. An example with an ext pledge and a few psupport entries:

image

* Fixes #826 

For reviewer:
* Tag this PR:


app/models/patient.rb
```diff
@@ -204,6 +204,14 @@ def recent_history_tracks
     versions.where(updated_at: 6.days.ago..)
   end
 
+  def all_versions(include_fulfillment)
+    all_versions = versions || []
+    all_versions += (external_pledges.includes(versions: [:item, :user]).map(&:versions).reduce(&:+) || [])
+    all_versions += (practical_supports.includes(versions: [:item, :user]).map(&:versions).reduce(&:+) || [])
+    # all_versions += fulfillment.includes(versions: [:item, :user]).versions || [] if include_fulfillment # next up
```
colinxfleming, line 211: I had a little trouble getting this to play nice, so I'd prefer to do it in isolation in a diff pr.

colinxfleming (commented):
notes. it wasn't pretty, but we got there!

app/models/patient.rb
lomky: cool cool, then this shouldnt be marked as fixing the issue, right?

app/views/patients/_change_log.html.erb
```diff
@@ -11,7 +11,7 @@
     </thead>
 
     <tbody>
-      <% patient.versions.each do |revision| %>
+      <% patient.all_versions(current_user.allowed_data_access?).each do |revision| %>
```
lomky, line 14: why only admins & data users getting fulfillment info?

test/models/patient_test.rb
```diff
@@ -646,6 +646,41 @@ class PatientTest < ActiveSupport::TestCase
     end
   end
 
+  describe 'all_versions' do
+    before do
+      # For some reason bullet doesn't play nicely with these unit tests,
+      # but does with the system tests. Given the choice I prefer the system
+      # tests, so we assume this is a false negative and turn off bullet for these.
```
lomky, line 653: Should there be system tests in this PR?

app/views/patients/_change_log.html.erb
colinxfleming: we generally only show fulfillment data (in the UI via the partial, for example) if someone's a data vol or above - if you aren't a data vol or above you can't do stuff like see the accounting pane, the fulfillment partial, or run a data export. This gates it according to our existing standard.

colinxfleming:
given the time in drafting, going to bump to include fulfillments too

colinxfleming:
this is now in good shape, except for bullet thinking we're eager loading when we don't need to. Going to iron that out soon I hope, and then this PR should be in mergeable shape again

app/models/patient.rb
```diff
@@ -204,6 +204,16 @@ def recent_history_tracks
     versions.where(updated_at: 6.days.ago..)
   end
 
+  def all_versions(include_fulfillment)
+    all_versions = versions || []
+    all_versions += (external_pledges.includes(versions: [:item, :user]).map(&:versions).reduce(&:+) || [])
+    all_versions += (practical_supports.includes(versions: [:item, :user]).map(&:versions).reduce(&:+) || [])
+    if include_fulfillment
+      all_versions += (fulfillment.versions.includes(fulfillment.versions.count > 1 ? [:item, :user] : []) || [])
+    end
```
colinxfleming, line 213: This is an awful hack, and in fact so awful that it makes me wonder if we should take out bullet from tests. The story here is that for most tests there's only one version associated with fulfillment, and so bullet just sees one result and assumes there's never a risk of an n+1 query. We get around that here by conditionally running the include if there _are_ multiple which gets bullet to chill, but potentially a 'cure worse than the disease' kinda thing here.

test/models/patient_test.rb
colinxfleming: nah, by this I meant that system tests would trigger any problems in bullet (though if there's a behavior you want to see locked in, definitely _can_ add system tests)

app/lib/paper_trail_version.rb
```diff
@@ -47,7 +47,7 @@ def changed_by_user
 
   def shaped_changes
     changeset.reject { |field| IRRELEVANT_FIELDS.include? field }
-             .reject { |field| field[1][0].blank? && field[1][1].blank? }
+             .reject { |_, values| values[0].blank? && values[1].blank? }
```
lomky, line 50: 🤩

test/models/patient_test.rb
```diff
@@ -669,15 +669,15 @@ class PatientTest < ActiveSupport::TestCase
       assert_includes version_types, 'ExternalPledge'
       assert_includes version_types, 'PracticalSupport'
       assert_includes version_types, 'Patient'
-      # assert_not_includes version_types, 'Fulfillment'
+      assert_not_includes version_types, 'Fulfillment'
```
lomky, line 672: is this not backwards? 
> `it 'should show fulfillment if include_fulfillment'`
and
> `assert_not_includes version_types, 'Fulfillment'`

app/models/patient.rb
lomky: I don't fully follow - what's happening if we just run it flat? Bullet is throwing a false positive?

colinxfleming: yeah - bullet can't seem to tell that versions is a has_many relation, and throws false positives if there's only one version. pretty annoying!

test/models/patient_test.rb
colinxfleming: ah, yep, you're right. fixing

lomky:
I think we have some confusing behavior introduced here, unfortunately. When you change only one part of an External Pledge or a Practical Support, the Change Log is ambiguous as to what happened. Can we make them show more context in the logs?

image

colinxfleming:
@lomky I think this might be a bit better of an approach - stapling the exact obj type and id in the changelog. So we wind up with something like this:

image

thoughts?

lomky (approved):
I like it! let's ship it!

Issue #2462, opened 2021-12-19
Upgrade to Rails 7


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

It's Time

* **What feature or behavior is this required for?**

There are a host of features that I'd like to try:
* the better frontend strats ('webpack makes everything asset pipelineable' has a LOT of appeal for me)
* the encrypted db fields stuff looks frankly incredible
* probably some other stuff

* **How could we solve this issue? (Not knowing is okay!)**

* `bundle update rails` to 7
* `rails app:upgrade` and accept any relevant diffs
* `rails new` and see if any ~ Cool New Gems ~ are in there
* get the test suite to pass

* **Anything else?**

PR #2491, opened 2022-01-24
Road to rails 7: change form_with to local by default

colinxfleming:

In rails 6.1 they changed the standard of form_with defaulting to `remote: true` to `local: false` to generate an ajax form. but only for forms, not for `link_to` or `button_to`, which still use `remote: true` (a fact that was so frustrating that I might submit a patch to rails about it, haha). This makes changes around form_with, and explicitly declares in button_to and link_to functions. (Also one minor test hack to make some ajax more reliable.)

* turn on the default config in the framework default
* change a bunch of `remote: true` to `local: false`-es on forms
* explicitly declare `remote: true/false` on button_to and link_tos

no view changes

* Bumps #2462 

For reviewer:
* Tag this PR:


app/views/patients/_fulfillment.html.erb
```diff
@@ -13,8 +13,7 @@
 
   <%= bootstrap_form_with model: patient,
                           html: { id: 'pledge_fulfillment_form' },
-                          method: 'patch',
-                          remote: true do |f| %>
+                          local: false do |f| %>
```
lomky, line 16: Why'd we lose the `patch` here?

app/views/dashboards/_search_form.html.erb
```diff
@@ -3,7 +3,7 @@
     <%= t 'dashboard.search.header' %>
   </h1>
 
-  <%= bootstrap_form_with url: '/search', layout: :inline do |p| %>
+  <%= bootstrap_form_with url: '/search', layout: :inline, local: false do |p| %>
```
lomky, line 6: is this one a change, or just an implicit made explicit?

colinxfleming: yep, this is an implicit made explicit!

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -45,7 +45,7 @@
 Rails.application.config.active_record.legacy_connection_handling = false
 
 # Make `form_with` generate non-remote forms by default.
-# Rails.application.config.action_view.form_with_generates_remote_forms = false  # HARD ONE
```
lomky, line 48: Nit: kill the comment?

lomky (approved):
🎉

PR #2447, opened 2021-12-02
add practical supports to db seed data

cameron-hicks:
This pull request makes the following changes:
* Adds a couple practical supports to patients created while seeding the test database for improved dev experience.

It relates to the following issue #s: 
* Fixes #[2193 ](https://github.com/DARIAEngineering/dcaf_case_management/issues/2193)

To test:
* Run `docker-compose run --rm web rails db:create db:migrate db:seed` to seed/re-seed your db.
* Then open a Rails console and run `PracticalSupport.all`. You should get back a non-empty array of practical supports.

PR #2513, opened 2022-02-14
add more practical supports

### LINKED ISSUE #2193: Add practical supports to seeds

We have a model, practical support, that CMs use pretty heavily, but when we seed the database with test data, none of the patients have practical support entries. We should adjust our test data to fix this!

Having test/experimental data that reflects patients accurately; improving the developer experience a bit.

The code changes will need to be made in `seeds.rb`. Practical Support is similar, relationship-wise, to Calls, so you can adopt some of that code.

Not every patient needs a PracticalSupport entry. I would suggest finding `if i.even?` in the seedfile (around L127?) and adding it there.

Try taking a look at the tests for an example of a filled out PracticalSupport entry.

* **Anything else?**
stems from a very good idea by @lomky in #2192

elimbaum:

* adds some more practical supports to seeds, now with money

* Fixes #2193
* Adds to #2447 

For reviewer:
* Tag this PR:

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -176,6 +176,10 @@ fund2 = Fund.create! name: 'CatFund',
         patient.practical_supports.create! support_type: 'Car rides', source: 'Neighbor'
       end
 
+      if i % 5 == 0
+        patient.practical_supports.create! support_type: 'Hotel', source: 'Donation', amount: 100
+      end
+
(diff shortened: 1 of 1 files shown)


colinxfleming (approved):

colinxfleming:
Thank you @elimbaum !

PR #2476, opened 2022-01-09
Generate generic pledges.

colinxfleming:
I rule and have completed some work on Case Manager that's ready for review!

Adds a generic pledge form generator.

TK address handling

This pull request makes the following changes:
* Make a generic pledge form

no view changes

It relates to the following issue #s: 
* Fixes #2341  

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #2518, opened 2022-02-21
Add generic pledges, lightly customized by fund

### LINKED ISSUE #2411: Graduate Pledge Generation into full Objecthood

Pull Pledge Generation out of in-code configuration and into its own object.

Stable application behavior

Right now this lives directly in the pledge geernation flow and is linked to a name set of the Fund object. Instead it should probably be a subobject of funds - needs to be optional!

* **Anything else?**
Do we need a frontend for these fields? Maybe just on one we create for the Fund object?

colinxfleming:

This graduates pledge generation config to full objecthood, as very wisely suggested by @lomky .

I don't think we need a frontend at this time - I imagine this will be almost entirely set and forget.

I have verified the pledges for the existing funds come out the same. Here's a snippet you can use to build a pledge config in local dev:

```
# Make a fake pledge config and build a pledge with it
Fund.last.build_pledge_config(logo_width: nil,
  logo_height: nil,
  contact_email: 'info@catfund.org',
  billing_email: 'billing@catfund.org',
  logo_url: nil,
  address1: '123 Fake Street',
  address2: 'Memphis, TN 12345',
  phone: '281-330-8004').save
# Now do the send pledge flow

# If there isn't an associated config object, don't show the pledge generator
PledgeConfig.destroy_all
# Now do the pledge send flow
```

My deployment plan is to immediately move over the two funds who actually have config, and not mess with the other funds at this time (keeping that work for #2476 ).

To test, make a PledgeConfig object and try to generate a pledge. (I've verified that they layout is identical for both, though height needs to drop down to 75 and width needs tobe nil for both)

* makes an optional pledge config object

no view 

* Fixes #2411

For reviewer:
* Tag this PR:


colinxfleming:
ready to go, finally! whew

lomky (approved):
Love it! :shipit:

PR #2524, opened 2022-03-02
Upgrade ruby version to 3.1.1 without upgrading rails

xmunoz:
A rework of #2440


Gemfile
```diff
@@ -9,6 +9,7 @@ gem 'nokogiri', '>= 1.13.2'
 gem 'tzinfo-data', require: false
 gem 'bootsnap', '>= 1.4.2', require: false
 gem 'rexml' # not a ruby default in 3, but a requirement of bootsnap
+gem 'net-smtp', require: false
```
colinxfleming, line 12:
```suggestion
gem 'net-smtp', require: false # For compat reasons, can remove after rails 7
```

colinxfleming (approved):
This worked for me! Successfully built in a macos environment, and started it up successfully. I think between us that's docker + local, so we're cool.

One snag I did notice was that puma no longer installs clean out of the box on mac os x, and it requires a little bit of config: `bundle config set --global build.puma --with-cflags=-fms-extensions`. I think we can just post that in slack for now, I'm including it here for posterity. 

I also googled the net/smtp thing and it looks like that's something that gets fixed in rails 7, so suggested a note to remind us to remove it when rails 7 hits.

Thanks for sticking on this! 🙏

PR #2537, opened 2022-03-29
Downgrade ruby and bundler because heroku doesn't support it yet

xmunoz:
Still this: https://github.com/heroku/heroku-buildpack-ruby/issues/1273

Also, perform minor and patch package upgrades.


Gemfile.lock
```diff
@@ -228,9 +226,6 @@ GEM
       net-protocol
       timeout
     nio4r (2.5.8)
-    nokogiri (1.13.3)
-      mini_portile2 (~> 2.8.0)
-      racc (~> 1.4)
     nokogiri (1.13.3-x86_64-linux)
```
colinxfleming, line 229: this fights back for me on a local env (mac) - I'm going to submit a commit that brings this back to the general instance

colinxfleming (approved):
looks good to me, other than nokogiri gems being bound to a system. non-native installs are annoying but kinda necessary here. hang tight, commit incoming

Gemfile.lock
xmunoz: nice catch!

Gemfile.lock
```diff
@@ -190,6 +190,7 @@ GEM
     launchy (2.5.0)
       addressable (~> 2.7)
     libsqreen (1.0.4.0.0)
+    libv8-node (16.10.0.0)
     libv8-node (16.10.0.0-x86_64-linux)
```
xmunoz, line 194: Maybe it's worth removing all linux-specific dependencies. I'll try to do that now.

xmunoz:
jk, that doesn't work, removed last commit

PR #2543, opened 2022-04-09
add matrix gem, which is required for builds and no longer in ruby stdlib

colinxfleming:

Heroku builds are failing with:

```
/tmp/codon/tmp/buildpacks/50d5eddf222a9b7326028041d4e6509f915ccf2c/lib/language_pack/helpers/rake_runner.rb:100:in `load_rake_tasks!': Could not detect rake tasks (LanguagePack::Helpers::RakeRunner::CannotLoadRakefileError)
ensure you can run `$ bundle exec rake -P` against your app
and using the production group of your Gemfile.
rake aborted!
LoadError: cannot load such file -- matrix
```

turns out `matrix` got separated from the ruby stdlib but our heroku builds depend on it still, so re-adding it to our gemfile explicitly.

see https://github.com/gdelugre/origami/issues/81 for example

* add matrix gem to gemfile

no view changes
no issues

For reviewer:
* Tag this PR:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'tzinfo-data', require: false
 gem 'bootsnap', '>= 1.4.2', require: false
 gem 'rexml' # not a ruby default in 3, but a requirement of bootsnap
 gem 'net-smtp', require: false # For compat reasons, can remove after rails 7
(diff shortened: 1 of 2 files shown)


colinxfleming:
godmode to get deployable

PR #2546, opened 2022-04-13
Bumps nokogiri past vulnerabilities

lomky:
I rule and have completed some work on Case Manager that's ready for review!

bundle-audit is mad about nokogiri! I bumped to the suggested version and everything looks to be loading clean for me.

This pull request makes the following changes:
* nokogiri >= 1.13.4

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #2547, opened 2022-04-13
Order lines by name instead of id and update corresponding test

### LINKED ISSUE #2433: Order lines alphabetically rather than by id

Lines currently order by id rather than alphabetically; we should swap that up so that it sorts accordingly in views.

Consistency, and CMs not being surprised by things that are out of order from what they're used to.

`default_scope` is usually bad news, so my preference is to avoid it. I think the main two places we can get away with changing this are the `@lines` instance variable in lines_controller, and the options for select builder in `PatientsHelper#line_options`.

* **Anything else?**
nah

alsokate:

Order lines by name instead of id and update corresponding test

* Sorts line instance by name in lines_controller
* Updates test in patients_helper_test to expect sorted line names

* Fixes #2433

For reviewer:
* Tag this PR:


xmunoz:
Once #2546 gets merged, you'll need to rebase main into this branch and then force push so that CI passes. Looks like there was a pretty high-profile vulnerability in nokogiri recently, which we've patched in the other PR.

colinxfleming (approved):
bless, thank you very much @alsokate !

PR #2551, opened 2022-04-18
split js into webpacker chunks

colinxfleming:

This takes advantage of webpack's chunking system to reduce the size of our builds. The short version is that it goes from a single import to this:

image

which utilizes code splitting as described here: https://webpack.js.org/guides/code-splitting/ (warning: your eyes are going to glaze over)

I think there's possibly some further optimization we could do here if we were committed to getting build times REALLY low, but I don't think we need to fret about that right now, so let's take the free basket.

* splits our js into chunks using webpacker's nice clean system

no view changes (I checked the main call list and did a few interactions and we're cool)
no issues

For reviewer:
* Tag this PR:


xmunoz (approved):
LGTM

PR #2554, opened 2022-04-19
remove js-routes gem

colinxfleming:

We were only really using js-routes in one place, and it was to set behavior after logging calls. It added a `Route` top level obj to the js namespace. We don't _really_ need it since this is the only place we do conditional stuff based on page, as far as I know.

You can double-check this work by `git grep "Route"`.

big s/o to past us for heavy system tests that made figuring this out much easier than I thought it was going tobe!

* refactor to not need js-routes gem
* remove js-routes gem

no view changes
no issues, but see https://github.com/DARIAEngineering/dcaf_case_management/pull/2392

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/calls/create.js.erb
+++ b/app/views/calls/create.js.erb
@@ -1,4 +1,4 @@
-var isOnEditPage = location.pathname.indexOf(substring) !== 1;
+var isOnEditPage = location.pathname.indexOf('edit') !== 1;
 var isOnDashboard = location.pathname === '/';
 
 if (isOnEditPage) {
--- a/app/views/calls/create.js.erb
+++ b/app/views/calls/create.js.erb
@@ -10,7 +10,7 @@ if (URL.indexOf(substring) !== 1) {
     });
 }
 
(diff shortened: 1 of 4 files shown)


app/views/calls/create.js.erb
```diff
@@ -1,7 +1,7 @@
-var URL = location.pathname,
-    substring = "edit";
+var isOnEditPage = location.pathname.indexOf('edit') !== 1;
+var isOnDashboard = location.pathname === '/';
```
lomky, line 2: Will swapping out a var for a hardcoded basepath give us any heartburn down the road? I don't think so?

lomky (approved):
LGTM! One thought that itched at my brain but not a blocker

app/views/calls/create.js.erb
colinxfleming: I don't _think_ so but the thought did cross my mind. That said our base routes haven't changed in several years at this point.

PR #2564, opened 2022-04-30
Cleanup Patient History

### LINKED ISSUE #1974: Delete history on patient archive and clean them out of db

While doing something unrelated in db I noticed that we have a bunch of spare ~Audit Trail~ PaperTrailVersion objects hanging out, including from patients that have been archived or deleted. We should remove these, since we don't need them after patient archive.

Keeping db memory usage down (AuditTrails take up almost half of our db usage). This will let us not

Here's what I'd suggest:
* On `ArchivedPatient#convert_patient`, destroy all audit trail objs associated with a given patient as part of that process
* In the `nightly_cleanup` task, establish a safety net by deleting any audit trail obj older than a year.

* **Anything else?**

nope!

elimbaum:

Some housekeeping on `PaperTrail` audit objects.

* When a patient is archived, also destroy all associated `PaperTrail` versions
* During `nightly_cleanup`, delete any `PaperTrail` objects > 1 year old. 

* Fixes #1974
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/app/lib/paper_trail_version.rb
+++ b/app/lib/paper_trail_version.rb
@@ -51,6 +51,10 @@ class PaperTrailVersion < PaperTrail::Version
     object_changes['urgent_flag']&.last == true
   end
 
+  def self.destroy_old
+    PaperTrailVersion.where("created_at < ?", 1.year.ago).destroy_all
+  end
+
   private
 
   def format_fieldchange(key, value)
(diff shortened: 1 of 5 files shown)


colinxfleming (approved):
perfectly done! thanks much @elimbaum !

PR #2568, opened 2022-05-01
Rails 7, take 4

colinxfleming:

Rails 7 upgrade: we can now run with this. I set the gems, ran bundle update, then rails app:update and took the diffs where appropriate.

A couple bullet requests were being rude so I killed those in a few spots I felt it appropriate (non-controller/system tests, specifically - bullet sometimes throws false flags when no dependent obj is present.

* rails 7!

no view changes

* Bumps #2462

For reviewer:
* Tag this PR:


config/environments/production.rb
```diff
@@ -50,11 +50,6 @@
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
 
-  # Prevent host header injection
```
colinxfleming, line 53: Kept, just moved to the bottom.

test/test_helper.rb
```diff
@@ -22,13 +22,10 @@ class ActiveSupport::TestCase
   include FactoryBot::Syntax::Methods
 
   before do
-    Bullet.start_request
```
colinxfleming, line 25: This shuts off bullet in non-http requests -- stuff that doesn't spoof a rails server. This was throwing all sorts of obnoxious false positives.

colinxfleming (commented):
couple of notes

xmunoz:
Ok, finally reviewing this now.

xmunoz:
Took it for a test spin locally, and everything looks good!

bin/rails
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env ruby
```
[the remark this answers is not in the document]
xmunoz: It's literally just style changes (e.g. changing single quotes to double quotes_

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_12_19_044530) do
+ActiveRecord::Schema[6.1].define(version: 2021_12_19_044530) do
```
xmunoz, line 13: My local version has some updates. We probably need to update this file. Here is the beginning of my diff:
```
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[6.1].define(version: 2021_12_19_044530) do
-
+ActiveRecord::Schema[7.0].define(version: 2021_12_19_044530) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
@@ -43,17 +42,17 @@ ActiveRecord::Schema[6.1].define(version: 2021_12_19_044530) do
     t.integer "patient_contribution"
     t.integer "naf_pledge"
     t.integer "fund_pledge"
-    t.datetime "fund_pledged_at"
+    t.datetime "fund_pledged_at", precision: nil
...
```

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,117 @@
+# Be sure to restart your server when you modify this file.
```
xmunoz, line 1: Do we need to check in this file we don't actually uncomment anything in it?

config/initializers/inflections.rb
```diff
@@ -4,13 +4,13 @@
 # are locale specific, and you may define rules for as many different
```
xmunoz, line 4: Same comment as above. Does this file need to be checked in if we don't uncomment anything?

config/initializers/filter_parameter_logging.rb
```diff
@@ -2,6 +2,10 @@
 
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += [
+  # Defaults
+  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn,
```
xmunoz, line 6: I don't think we use any of these params anywhere. Remove?

config/initializers/assets.rb
```diff
@@ -1,12 +1,10 @@
 # Be sure to restart your server when you modify this file.
 
 # Version of your assets, change this if you want to expire all your assets.
-Rails.application.config.assets.version = '1.0'
+Rails.application.config.assets.version = "1.0"
 
 # Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
-# Add Yarn node_modules folder to the asset load path.
-Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
xmunoz, line 9: Why don't we need to do this anymore?

config/environments/test.rb
```diff
@@ -60,8 +59,20 @@
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
 
-  config.log_level = :info
```
xmunoz, line 63: Is info the default log level?

config/environments/development.rb
```diff
@@ -22,15 +14,18 @@
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  # Enable server timing
```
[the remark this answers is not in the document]
xmunoz: nice! context: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing

config/environments/test.rb
xmunoz: Hmm....here it says that the default is debug: https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels

config/initializers/new_framework_defaults_7_0.rb
colinxfleming: yeah! rails app:upgrade generates you a thing that lets you turn the scary stuff on piece by piece, so you can enjoy new rails features and incrementally turn on some of the potentially breaking improvements. So as time goes on we'll turn on these defaults, then change the `use_framework_defaults` in `application.rb` to `7.0` from `6.1` or whatever it is now.

config/initializers/inflections.rb
colinxfleming: yeah, good to have this file here in the unlikely event we do need it. (We've found a lot of benefit in keeping the state of this kinda thing as close to the current rails state of the art as possible, so that if we do have to start using something we don't have to figure out what the initial shape should be.)

config/initializers/assets.rb
colinxfleming: not 100% sure, might have been my screwup. I'll look

config/environments/test.rb
colinxfleming: we can pretty harmlessly pop this back in if preferred!

config/initializers/filter_parameter_logging.rb
colinxfleming: eh, worth keeping em in for the 'start from baseline' reasons detailed above I think

colinxfleming:
summary: going to look into schema.rb changes and whether it matters that I removed node modules from assets

xmunoz:
Can we also deprecated webpacker in this PR? https://github.com/rails/webpacker#webpacker-has-been-retired-

colinxfleming:
bleh @ webpacker being deprecated, but that's almost certainly for the best. let's do that after we get rails 7 merged.

config/initializers/assets.rb
colinxfleming: I'm not sure if this works either way but the git blame is 5 years old, probably from the last time we did a rails upgrade, so I'm just gonna revert this and keep it in here.

xmunoz (approved):
Ok, opened an issue for removing webpacker. Doing that will probably also make #2580 a lot easier.

colinxfleming:
yay. thanks much for the detailed review and work here @xmunoz

PR #2569, opened 2022-05-02
Rework call list to remove part of last note and include Initial Call Date instead

colinxfleming:

This adds a sortable Initial Call Date column to the call list, and drops last note (which hasn't worked right for awhile; part of our justification for dropping it is that nobody has complained in that time). Last note will still show up on the new call flow.

* Add initial call date to call list
* Remove last note from call list

image

* Fixes https://github.com/DARIAEngineering/dcaf_case_management/issues/2501

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/notetakeable.rb
+++ b/app/models/concerns/notetakeable.rb
@@ -2,13 +2,6 @@
 module Notetakeable
   extend ActiveSupport::Concern
 
-  def most_recent_note_display_text
-    note_text = most_recent_note.try(:full_text).to_s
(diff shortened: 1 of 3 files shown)


lomky (approved):
LFTM! I am surprised we didn't have to update any system tests, do we not check anywhere for how the call list displays?

colinxfleming:
I think we didn't test that because (blast from the past) precision clicking was something I wasn't good at in capybara at the time, haha. Live and learn?

PR #2571, opened 2022-05-03
Commits schema updates for integer over bigint

lomky:

I keep seeing this change when I'm doing the build, so opening this PR to figure out if we want it, or want to figure out how to keep bigints

* runs the build & seed
* commits the schema changes

For reviewer:
* Tag this PR:

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -49,9 +49,9 @@ ActiveRecord::Schema.define(version: 2021_12_19_044530) do
     t.datetime "pledge_generated_at"
     t.datetime "pledge_sent_at"
     t.boolean "textable"
-    t.bigint "clinic_id"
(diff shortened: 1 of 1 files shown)


colinxfleming:
I have no earthly idea why this is happening and it is definitely driving me up the wall! but I think we do want FK refs to be bigints, even if strictly speaking we will never even get close to the point where a difference between int and bigint will matter.

lomky:
Running through a git bisect, I see the first time this pops up is on https://github.com/DARIAEngineering/dcaf_case_management/pull/2537, the PR downgrading back to 3.0.2 & doing a few dep bumps. Strange. more digging needed.

xmunoz:
I don't think we need this anymore

Issue #2501, opened 2022-02-02
Can we add initial call date to the call list panels?


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

A fund asked if we could add initial call date to the call list panels.

Maybe we sub this in for the (broken) show-last-note display?

* **What feature or behavior is this required for?**
Keeping track of patients without going into their individual patient records.

* **How could we solve this issue? (Not knowing is okay!)**

Interested in discussing this with Emilia!

* **Anything else?**

colinxfleming:
ex of what we'd be replacing:

![image](https://user-images.githubusercontent.com/3866868/157358047-5303b7cb-9c0f-48c2-b48f-4d41c5d9b0cb.png)

the notes on the right there, but if there's a better option I'm interested

colinxfleming:
Yes, removing notes in favor of initial call date is fine given that last note shows up on the call modal (and also the popup doesn't work still).

So changes here:
* new order of the table is: Phone, Name, Initial Call Date, Weeks Along, Status, Appt Date, Notes, icons
* Make Initial Call Date sortable (a la status, appt date)
* make the modal consistent with the table by adding initial call date, and weeks along

PR #2577, opened 2022-05-08
Adds back the Notes preview on table view

lomky:

Turns out a fund uses the Notes preview as a core part of their flow! Adding that back.

* Adds back the Notes column & related code removed in #2569 
  * No longer offers the (nonfunctional) `+`
  * Uses the unicode … instead of ...
  * Deletes dead code from removing the `+`
* Shrinks Table font from 17px->16px

image

* Related to #2501 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/helpers/notes_helper.rb
+++ b/app/helpers/notes_helper.rb
@@ -1,15 +1,5 @@
 # Functions related to displaying notes
 module NotesHelper
-  def plus_sign_icon(note)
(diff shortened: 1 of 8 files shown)


colinxfleming (approved):
This works for me (with x glyph intact). The test failure is from an i18n string disappearing. Thanks for hopping on this so quick, really appreciated.

image

PR #2582, opened 2022-05-13
scroll flashes; adjust css

### LINKED ISSUE #2429: Flashes don't float when you scroll

Get feedback from actions! Flashes (green/red boxes) appear at the top of the page, so if you are editing something at the _bottom_ of the page, you may not see the message. See attached image for a partially-hidden flash:

Screen Shot 2021-11-17 at 20 59 26

Flashes - success/errors from form submissions and actions.

We can probably make these have `absolute` positioning (I think that's the CSS?) so they always show up at the same place.

* **Anything else?**

elimbaum:

This changes makes feedback flashes (user input save confirmation, errors) always show up, even if you scroll the page. Before, flashes were stuck at the top of the page, so if you were on a very tall page and making changes at the bottom, you might miss feedback.

* Makes flashes `position: fixed` rather than `absolute`
* Fixes centering of flashes

Example success:

Screen Shot 2022-05-13 at 12 03 20

Example error:

Screen Shot 2022-05-13 at 12 03 52

* Fixes #2429
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/app/javascript/styles/flash.scss
+++ b/app/javascript/styles/flash.scss
@@ -4,10 +4,19 @@
 @import './_variables';
 
 #flash {
-  position: absolute;
+  position: fixed;
+  top: 0;
+  padding-top: 0.5em;
+  left: 0;
   width: 100%;
-  z-index: 999;
+  z-index: 1100;
   transition-duration: .5s;
+  pointer-events: none;
+}
+
+div.flash-message {
+  margin: auto;
+  pointer-events: auto;
 }
 
 #error_explanation {
--- a/test/integration_helper.rb
+++ b/test/integration_helper.rb
@@ -77,6 +77,8 @@ module IntegrationHelper
   end
 
   def click_away_from_field
+    # close the flash
+    find('#flash button.close')&.click
     find('nav').click
     wait_for_ajax
   end
--- a/test/integration_helper.rb
+++ b/test/integration_helper.rb
@@ -77,8 +77,12 @@ module IntegrationHelper
   end
 
   def click_away_from_field
-    # close the flash
-    find('#flash button.close')&.click
+    # close the flash if it's in the way
+    begin
+      find('#flash button.close')&.click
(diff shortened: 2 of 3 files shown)


app/javascript/styles/flash.scss
```diff
@@ -4,10 +4,19 @@
...
 
 #flash {
-  position: absolute;
+  position: fixed;
+  top: 0;
+  padding-top: 0.5em;
+  left: 0;
   width: 100%;
-  z-index: 999;
+  z-index: 1100;
```
elimbaum, line 12: this is so that the flash appears above the nav bar (z-index 1000 something)

app/javascript/styles/flash.scss
```diff
@@ -4,10 +4,19 @@
...
   transition-duration: .5s;
+  pointer-events: none;
```
elimbaum, line 14: this is so that we can click things behind the invisible alignment div

app/javascript/styles/flash.scss
```diff
@@ -4,10 +4,19 @@
...
+}
+
+div.flash-message {
+  margin: auto;
+  pointer-events: auto;
```
elimbaum, line 19: this is so we *can't* click things behind the flash, unless it is closed

test/integration_helper.rb
```diff
@@ -77,6 +77,12 @@ def go_to_dashboard
   end
 
   def click_away_from_field
+    # close the flash if it's in the way
+    begin
+      find('#flash button.close')&.click
```
elimbaum, line 82: is this ok? (versus *first* check if the element, if it does click it)

colinxfleming: I think this is a great approach!

colinxfleming (approved):
Excellent CSS work here. I admit I cannot provide a quality review of the specific stylesheet changes but this does work beautifully for me. Thanks a ton for tackling this.

PR #2583, opened 2022-05-13
Eli/2575 test trim urgent

elimbaum:
I rule and have completed some work on Case Manager that's ready for review!

This pull request makes the following changes:
* Implements a test for the trim urgent/flagged patients method
* (another change here)
* (etc)

(If there are changes to the views, please include a screenshot so we know what to look for!)

It relates to the following issue #s: 
* Fixes #2575 
* Blocked by #2576 
* Bumps #Y

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #2587, opened 2022-05-19
add test for previously-skipped patient method

### LINKED ISSUE #2575: trim_urgent_patients test

Test our methods!

We're currently skipping the `trim_urgent_patients` test! We probably shouldn't! (as far as I can tell this is the only test we skip.)

https://github.com/DARIAEngineering/dcaf_case_management/blob/ba8f7ecc12cf5680076e07290198c6d7920ccc31/test/models/patient_test.rb#L349-L353

* **Anything else?**

There may be a reason we didn't test it, may be worth a `git blame` archaeology dig.

elimbaum:

* Test the `trim_shared_patients` method, which we had been skipping.
* Replaces #2583 (now using the new `shared` rather than `flagged` language)

* Fixes #2575 
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/test/models/patient_test.rb
+++ b/test/models/patient_test.rb
@@ -348,7 +348,10 @@ class PatientTest < ActiveSupport::TestCase
 
       describe 'trim_shared_patients method' do
         it 'should trim shared patients after they have been inactive or resolved' do
-          skip "test not implemented for patient#trim_shared_patients"
+          @patient.update shared_flag: true
+          @patient.update resolved_without_fund: true
+          Patient.trim_shared_patients
+          assert_not @patient.shared_flag
         end
       end
 


elimbaum:
Seeing some weird system test failures (had to re-run test suite, and also seeing these happen with local system tests). I think we're good to merge but will take a look to see if we borked something with system test reliability.

colinxfleming (approved):
honestly system tests / anything that hits javascript through a browser engine is dicey and it's a small miracle it works at all. Our system tests _generally_ work; if they fail sporadically it's almost def not your fault - throw the retry module in  the class and see if that helps if things are giving you trouble.

ANYWAY thanks a ton for seeing something saying something and knocking this out!

PR #2610, opened 2022-06-18
R7 default: turn on key generator hash digest class default

colinxfleming:

Toggle on the key generator hash default class. This seems to primarily be for cookie generation and similar secret stuff, which we do relatively little of - probably just login stuff which expires after time anyway.

* turn on the new key generator hash class

no view changes

* Bumps #2462

For reviewer:
* Tag this PR:


xmunoz (approved):
I rebased main to run the docker build on this PR. Looks good!

PR #2613, opened 2022-06-18
R7: turn on the inverse_of scope setting

colinxfleming:

Automatically infer `inverse_of` for associations with a scope, whatever that means. Probably doesn't affect us much since we generally do not do scopes.

* turn on this new setting

no view changes

* Bumps #2462 

For reviewer:
* Tag this PR:


xmunoz (approved):
yolo

PR #2628, opened 2022-06-26
Urgent config adjust (2400)

### LINKED ISSUE #2400: New "Urgent Reset Options" config has a max value of 28, can that be noted in the config description?

It looks like there's actually a max value at 28 for the "Urgent Reset Options" config (https://github.com/DARIAEngineering/dcaf_case_management/issues/2338). I arrived at that conclusion after BAF attempted to set at 30, got an "invalid value" error, and then trial-and-erroring it from 10 upward. 

"Urgent Reset Options" config (https://github.com/DARIAEngineering/dcaf_case_management/issues/2338)

Immediate/small solution: It's not a huge priority, but updating the prompt (currently reads: "Number of idle days until a patient is removed from the urgent list. Defaults to 6 days") to include the max value would be helpful from a UX perspective. 

For more questions (not urgent), see below. 

* **Anything else?**

The 28 day max is not an issue for BAF (a huge help to lift the max from 6 to 28!), but our BAF CM lead did share wonderings about whether there are any concerns about funds not un-checking the Urgent flag when cases are resolved. 

Our BAF CM lead shared that we always un-check the "urgent" flag once cases have been resolved, as opposed to letting the "urgent" flag lapse on its own. I.e., it's less of a concern that people *aren't* closing out cases and more about worrying we've let something slip through the cracks. Is there any concern from DARIA developers' POV about cases that aren't being closed out in a real/practical sense from any of the funds?

elimbaum:

* Updates help text language to clarify maximum number of days for urgent timeout
* Makes that maximum timeout a bit longer, since it seems like some funds were bumping up against a (relatively artificial) upper bound

* Fixes #2400 
* Bumps #Y

For reviewer:
* Tag this PR:


colinxfleming (approved):

colinxfleming:
thank you @elimbaum !

PR #2630, opened 2022-07-02
Add email to clinics; display of a clinic's email/fax number when sending a pledge

### LINKED ISSUE #2528: Track and surface whether a clinic prefers pledges via fax or email

Clinics are slowly crawling into the year 2005 and accepting pledge documents in email, in some cases. A fund asked me whether it was possible to keep tabs on this, either via normalized fields (e.g. clinic.prefers_email and clinic.email field) or an open text field (e.g. clinic.notes = 'email pledges to this clinic at metallica@example.com').

One interesting curveball is that this kind of is a departure from how we display clinic data - so far it's only been clinic name / city / state and whatever's in clinic finder, and we don't make other fields bubble up. I don't think that's necessarily a problem! But it means we don't have a clear place to display this.

Open questions:
* Should we do this?
* If so, normalized data or not normalized data?
* If so, how and where do we surface that?

Case managers not having to cross reference an external spreadsheet a bunch of times a shift.

TBD

* **Anything else?**
nah

elimbaum:

* Adds an email field to clinics
* Makes that field editable in the clinic page
* Displays that email on the pledge screen
* If no email, displays the clinic's fax number. If no fax number, just displays the fund's fax service URL.

* Fixes #2528 
 
For reviewer:
* Tag this PR:

### DIFF
--- a/app/helpers/pledges_helper.rb
+++ b/app/helpers/pledges_helper.rb
@@ -15,4 +15,14 @@ module PledgesHelper
       t('patient.menu.cancel_pledge')
     end
   end
+
+  def clinic_pledge_email(patient)
+    email = patient.clinic&.email_for_pledges
+    link_to email, 'mailto:' + email, target: '_blank'
+  end
+
+  def clinic_pledge_fax(patient)
+    fax = patient.clinic&.fax
+    fax
+  end
 end
--- a/app/helpers/pledges_helper.rb
+++ b/app/helpers/pledges_helper.rb
@@ -18,6 +18,7 @@ module PledgesHelper
 
   def clinic_pledge_email(patient)
     email = patient.clinic&.email_for_pledges
+    return unless email
     link_to email, 'mailto:' + email, target: '_blank'
   end
 
--- a/app/services/pledge_form_generator.rb
+++ b/app/services/pledge_form_generator.rb
@@ -92,6 +92,18 @@ class PledgeFormGenerator
       contact_email: 'info@yellowhammerfund.org',
       billing_email: 'billing@yellowhammer.org',
       img: 'yhf_logo.png'
+    },
+    'Test Fund' => {
+      addr: [
+        'Test Fund',
+        '42 Wallaby Way',
+        'Sydney Australia NSW'
+      ],
+      width: 200,
+      height: nil,
+      img: 'dcaf_logo.png',
+      contact_email: 'info@testfund.net',
+      billing_email: 'billing@testfund.net'
     }
   }
 
--- a/app/services/pledge_form_generator.rb
+++ b/app/services/pledge_form_generator.rb
@@ -93,18 +93,6 @@ class PledgeFormGenerator
       billing_email: 'billing@yellowhammer.org',
       img: 'yhf_logo.png'
     },
-    'Test Fund' => {
-      addr: [
-        'Test Fund',
-        '42 Wallaby Way',
-        'Sydney Australia NSW'
-      ],
-      width: 200,
-      height: nil,
-      img: 'dcaf_logo.png',
-      contact_email: 'info@testfund.net',
-      billing_email: 'billing@testfund.net'
-    }
   }
 
   private
--- a/app/views/patients/_pledge.html.erb
+++ b/app/views/patients/_pledge.html.erb
@@ -37,21 +37,22 @@
         <p class="row">
           <%= t('patient.pledge.step_three.downloads') %>
         </p>
+      <% end %>
 
-        <% if patient.clinic&.email_for_pledges.present? %>
+      <p class="row"><%= t('patient.pledge.step_three.submit') %></p>
+      <% if patient.clinic&.email_for_pledges.present? %>
+        <p class="row">
+          <%= clinic_pledge_email(patient) %>
+        </p>
+      <% else %>
+        <% if patient.clinic&.fax.present? %>
           <p class="row">
-            <%= clinic_pledge_email(patient) %>
-          </p>
-        <% else %>
-          <% if patient.clinic&.fax.present? %>
-            <p class="row">
-              <%= t('patient.pledge.step_three.clinic_fax') %>:&nbsp;<b><%= clinic_pledge_fax(patient) %></b>
-            </p>
-          <% end %>
-          <p class="row">
-            <%= t('patient.pledge.step_three.fax_service') %>:&nbsp;<%= fax_service %>
+            <%= t('patient.pledge.step_three.clinic_fax') %>:&nbsp;<b><%= clinic_pledge_fax(patient) %></b>
           </p>
         <% end %>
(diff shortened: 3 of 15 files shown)


app/views/patients/_pledge.html.erb
```diff
@@ -35,7 +35,12 @@
       <% if current_tenant.pledge_generation_config.present? %>
         <p class="row"><%= t('patient.pledge.step_three.generated', fund: current_tenant.name) %></p>
         <p class="row">
-          <%= t('patient.pledge.step_three.downloads') %> <%= fax_service %>
+          <%= t('patient.pledge.step_three.downloads') %>
+          <% if patient.clinic&.email_for_pledges.present? %>
+              <%= clinic_pledge_email(patient) %>
+          <% elsif fax_service.present? %>
+              <%= fax_service %>
+          <% end %>
```
elimbaum, line 43: @colinxfleming is this what you had in mind? I think the logic is flipped around from what you had written in #2528

elimbaum: ok, nvm change this to

- display email if email
- otherwise display fax number (if exists) and fax service

test/system/submit_pledge_test.rb
```diff
@@ -69,6 +72,35 @@ class SubmitPledgeTest < ApplicationSystemTestCase
...
+      assert has_text? 'Generate your pledge form'
+      find('#pledge-next').click
+      wait_for_no_element 'Review this preview of your pledge'
+
+      assert has_text? 'Awesome, you generated a CATF'
+      # now we should see the email
+      assert has_text? "pledges@catfund.biz"
+      refute has_text? "Fax service"
+    end
+  end
```
elimbaum, line 102: why doesn't this test work!

elimbaum: to follow up – if i put a `byebug` breakpoint inside the test body, i can confirm that, indeed, `@patient.clinic.email_for_pledges` has the correct value. _but then for some reason_ the view isn't picking up that value! See [here](https://github.com/DARIAEngineering/dcaf_case_management/compare/eli/clinic-email#diff-e3fbc4b423f7b6afa5f19a9d23e9208f91d1740b7723fbd3f3a32635be9c11edR41-R54) for what the view is doing.

app/views/patients/_pledge.html.erb
```diff
@@ -35,8 +35,23 @@
...
+            <%= clinic_pledge_email(patient) %>
+          </p>
+        <% else %>
+          <% if patient.clinic&.fax.present? %>
+            <p class="row">
+              Clinic fax:&nbsp;<strong><%= clinic_pledge_fax(patient) %></strong>
+            </p>
+          <% end %>
+          <p class="row">
+            Fax service:&nbsp;<%= fax_service %>
```
elimbaum, line 52: these `&nbsp;` make me nervous. better way to handle? the issue is that leading/trailing spaces aren't observed when concatenating tags – in the case of the fax #, I'm `<strong>`ing the number itself, so the space gets collapsed; in the case of `fax_service`, we have a link, so the space between the words and the `<a>` is collapsed.

test/system/update_patient_info_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'application_system_test_case'
 
 class UpdatePatientInfoTest < ApplicationSystemTestCase
+  extend Minitest::OptionalRetry
```
elimbaum, line 4: this just kept popping up here and there with random failures. this seems to improve.

test/system/submit_pledge_test.rb
elimbaum: massive h/t @FeminismIsAwesome for reminding me that models have to be saved!!! drrrrrrr

elimbaum:
This is ready for review pending passing CI by PR #2631!

app/services/pledge_form_generator.rb
```diff
@@ -92,6 +92,18 @@ def generate_pledge_pdf
       contact_email: 'info@yellowhammerfund.org',
       billing_email: 'billing@yellowhammer.org',
       img: 'yhf_logo.png'
+    },
+    'Test Fund' => {
```
colinxfleming, line 96: Is it necessary to check this in? This doesn't actually change the pledge itself, right?

db/seeds.rb
```diff
@@ -35,16 +35,16 @@
                      subdomain: 'sandbox',
                      full_name: 'Sand Box Fund',
                      site_domain: 'www.petfinder.com',
-                     phone: '202-452-7464'
-
-
+                     phone: '202-452-7464',
+                     pledge_generation_config: 'Test Fund'
```
colinxfleming, line 39: preference for reverting, if we can

colinxfleming (commented):
this looks good to me, by and large! The approach is very sound and the diffs are clean.

Before I hit approve, want to check on the changes to seeds and the pledge generator - are those changes revertable, or am I misreading and they're required for the rest of these diffs? If they aren't, let's please revert; if they are please let me know why.

After we get that squared away I think this is ready to go. nicely done.

app/helpers/pledges_helper.rb
```diff
@@ -15,4 +15,14 @@ def cancel_pledge_button
       t('patient.menu.cancel_pledge')
     end
   end
+
+  def clinic_pledge_email(patient)
+    email = patient.clinic&.email_for_pledges
```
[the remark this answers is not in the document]
elimbaum: ooh yeah i love `unless`. underrated keyword.

app/services/pledge_form_generator.rb
elimbaum: ah, prob not. the issue is that without a `pledge_generation_config` none of the pledge stuff shows up, so i couldn't debug. I can

1. remove it entirely?
2. or set catfund to use one of DCAF or YHF.

app/views/patients/_pledge.html.erb
elimbaum: @colinxfleming this look ok?

app/services/pledge_form_generator.rb
colinxfleming: oh! I see what you mean here.

I would bet that this has utility outside of funds having pledge_generation_config. What do you think about moving the new stuff in the pledge partial it to step 3 above the 'I sent the pledge' checkbox and reverting this change? Seems like 'having the number to dial on hand' lends itself well to the sending step. Thoughts?

elimbaum: ah, sure. let me have a look - i think i know what you mean

elimbaum: ok, moved the contact info down, so it will show up regardless.

colinxfleming (approved):
beautiful work as always. Sorry it took me awhile to wheel back to this!

PR #2633, opened 2022-07-04
#2514 fix bug where it stopped reporting the sum of pledges

### LINKED ISSUE #2514: "External pledges sum" in csv export: field values are no longer all numerical
Can we fix the "External pledges sum" field in the .csv export so that all the values are numerical and are summing correctly? See attached screenshot from a BAF export. I think it's been going on for a few months - I hadn't had time to look more closely last November/December when I first ran into the issue. 
Screen Shot 2022-02-15 at 5 46 47 PM

It's helpful to have all of these values be numbers when we're reporting out. I have a little R script that I run and it's been breaking because the values in this column aren't all numerical. 

I feel like folks who are smarter than me might be able to pinpoint this as a small tweak, but I have no clear clues as to where this is broken :)

* **Anything else?**

FeminismIsAwesome:

This fixes a bug where it was returning a list of the external pledges instead of summing up the value. 

previous code before what i think is an oops is:

  sum = 0
    all_pledges = external_pledges.all
    
    all_pledges.each do |pledge|
      sum += pledge.try :amount
    end
    sum

i think right now, given the fact we have amount on external_pledges, we can just lean on the sum function in activerecord. Added tests to account for what i think are the cases, since i think we just deal with USD. 

* Fixes #2514

For reviewer:

### DIFF
--- a/app/models/concerns/exportable.rb
+++ b/app/models/concerns/exportable.rb
@@ -166,7 +166,7 @@ module Exportable
   end
 
   def external_pledge_sum
-    external_pledges.inject { |x, acc = 0| acc = x.try(:amount) || 0}
+    external_pledges.sum(:amount)
   end
 
   def all_external_pledges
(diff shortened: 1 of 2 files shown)


colinxfleming:
Hey Ian, awesome to hear from you! Hope you are doing well. (I'm on
vacation this wknd but wanted to say hi)

PR #2641, opened 2022-07-18
Upgrade rails to 7.0.3.1 because of CVE-2022-32224

xmunoz:
We need to upgrade rails because of a critical vulnerability in our current version of activerecord: https://github.com/advisories/GHSA-3hhc-qp5v-9p2j.

Failing `bundle audit`: https://github.com/DARIAEngineering/dcaf_case_management/runs/7373046758?check_suite_focus=true

Just ran `bundle update` and committed the diff.


colinxfleming (approved):
generally I think you want to run `bundle update rails` to target the library updates, but all of these are point or minor releases so fire away

PR #2659, opened 2022-08-04
reset locked timer after unlocking an account

colinxfleming:

A fund pointed out that, when unlocking CM accounts, they've been automagically relocking as part of the nightly process. This is because the automagic relock takes last login, and they don't always log in before the nightly job runs again. This solves for that case by resetting the timer when unlocking an account.

* When unlocking an account, reset the inactivity lock timer

no view changes

* Fixes https://github.com/DARIAEngineering/dcaf_case_management/issues/2657

For reviewer:
* Tag this PR:


app/models/user.rb
```diff
@@ -82,7 +82,12 @@ def self.from_omniauth(access_token)
 
   def toggle_disabled_by_fund
     # Since toggle skips callbacks...
-    update disabled_by_fund: !disabled_by_fund
+    attrs = { disabled_by_fund: !disabled_by_fund }
+    # also, if unlocking, count that as a sign in for locking purposes
+    if disabled_by_fund?
+      attrs.merge!({ current_sign_in_at: Time.zone.now })
+    end
```
lomky, line 89: I find this a bit hard to read - I thought we had this backwards at first! Relying on the current state of the thing we are in the middle of toggling is what we're doing, right? A quick suggestion for the explainer comment:
```suggestion
    # if we're currently locked, count this unlocking as a sign-in to avoid rapid relocking
    if disabled_by_fund?
      attrs.merge!({ current_sign_in_at: Time.zone.now })
    end
```

test/models/user_test.rb
```diff
@@ -255,6 +255,17 @@ class UserTest < ActiveSupport::TestCase
       assert @user.disabled_by_fund
     end
 
+    it 'resets the current sign in token' do
```
lomky, line 258:
```suggestion
    it 'resets the current sign in token only on reenable' do
```

lomky (approved):
Small working tweaks for the sake of future-us!

PR #2662, opened 2022-08-08
bump setup instructions for new fund

colinxfleming:

Not very much changed, but bump setup instructions (mostly to remind us to ask some particular questions). I don't think anything else has changed since we last did one of these.

no views
no issues

For reviewer:
* Tag this PR:


colinxfleming:
docs only so godmoding

PR #2668, opened 2022-08-12
add handling fund field

### LINKED ISSUE #2661: Add way to track which fund is taking the lead on a solidarity case

I was right, funds do want to know whomst running point on solidarity cases. Let's mark that down and increase the utility of this new field. 

better comms and gardens

probably a dropdown on abortion information with a list of funds in the external pledges rack. The current fund should be in there too. Might require a little tuning.

* **Anything else?**

Let's also affirm that there isn't a security concern here (or at least not one that's greater than the general external pledges tracker)

elimbaum:

* Add `Handling Fund` field for solidarity pledges
* If the `Handling Fund` is changed to the current fund (= `nil`), `solidarity` is unchecked
  * We do not automatically check `solidarity` if the handling fund is changed to an external fund.

new UI element:

Screen Shot 2022-08-12 at 17 02 33

* Fixes #2661 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -131,7 +131,7 @@ class PatientsController < ApplicationController
   ABORTION_INFORMATION_PARAMS = [
     :clinic_id, :resolved_without_fund, :referred_to_clinic, :completed_ultrasound,
     :procedure_cost, :patient_contribution, :naf_pledge, :fund_pledge,
-    :fund_pledged_at, :pledge_sent_at, :solidarity
+    :fund_pledged_at, :pledge_sent_at, :solidarity, :handling_fund
   ].freeze
 
   FULFILLMENT_PARAMS = [
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -131,7 +131,7 @@ class PatientsController < ApplicationController
   ABORTION_INFORMATION_PARAMS = [
     :clinic_id, :resolved_without_fund, :referred_to_clinic, :completed_ultrasound,
     :procedure_cost, :patient_contribution, :naf_pledge, :fund_pledge,
-    :fund_pledged_at, :pledge_sent_at, :solidarity, :handling_fund
+    :fund_pledged_at, :pledge_sent_at, :solidarity, :solidarity_lead
   ].freeze
 
   FULFILLMENT_PARAMS = [
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -17,7 +17,6 @@ class Patient < ApplicationRecord
 
   # Callbacks
   before_validation :clean_fields
-  before_save :update_handling_fund
   before_save :save_identifier
   before_save :update_pledge_sent_by_sent_at
   before_save :update_fund_pledged_at
@@ -226,10 +225,6 @@ class Patient < ApplicationRecord
 
   def initialize_fulfillment
     build_fulfillment.save
-  end
-
-  def update_handling_fund
-    self.solidarity = false if self.handling_fund.blank?
   end
 
   def update_pledge_sent_by_sent_at
--- a/app/models/patient.rb
+++ b/app/models/patient.rb
@@ -17,6 +17,7 @@ class Patient < ApplicationRecord
 
   # Callbacks
   before_validation :clean_fields
+  before_save :update_handling_fund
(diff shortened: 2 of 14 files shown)


app/views/patients/_abortion_information.html.erb
```diff
@@ -43,8 +43,11 @@
           <%= f.check_box :solidarity,
                           label: t('patient.abortion_information.clinic_section.solidarity'),
                           label_class: 'tooltip-header-checkbox',
-                          data: { 'tooltip-text': solidarity_help_text } %>
+                          data: { 'tooltip-text': solidarity_help_text } %>            
         <% end %>
+        <%= f.select :handling_fund,
```
colinxfleming, line 48: Let's get a tooltip on this too, text along the lines of:

> This is the fund serving as point-of-contact or coordinating funding for this patient.

app/models/patient.rb
```diff
@@ -227,6 +228,10 @@ def initialize_fulfillment
     build_fulfillment.save
   end
 
+  def update_handling_fund
+    self.solidarity = false if self.handling_fund.blank?
```
colinxfleming, line 232: I kinda think we can be a little lazier here and not have this autotoggle - I don't want to be too prescriptive about the use at first.

test/system/update_patient_info_test.rb
```diff
@@ -144,6 +144,7 @@ class UpdatePatientInfoTest < ApplicationSystemTestCase
       check 'Referred to clinic'
       check 'Ultrasound completed?'
       check 'Solidarity Pledge'
+      select 'Metallica Abortion Fund', from: 'patient_handling_fund'
```
colinxfleming, line 147: make sure to add an assert below too (around L181)

colinxfleming (commented):
very well done and clean, I think the code is pretty much there. I think changes I'd like to see before hitting go boil down to:

* adding an assert to the systemtest
* renaming to `Solidarity Lead`
* skipping the auto-update (though I definitely like where your head's at there)
* adding a tooltip so people know what the verbiage is


after that I think we can fire it!

elimbaum:
sounds good – will make these changes. fair on the auto update, makes me a bit skittish

app/controllers/patients_controller.rb
```diff
@@ -131,7 +131,7 @@ def find_patient_minimal
   ABORTION_INFORMATION_PARAMS = [
     :clinic_id, :resolved_without_fund, :referred_to_clinic, :completed_ultrasound,
     :procedure_cost, :patient_contribution, :naf_pledge, :fund_pledge,
-    :fund_pledged_at, :pledge_sent_at, :solidarity
+    :fund_pledged_at, :pledge_sent_at, :solidarity, :handling_fund
```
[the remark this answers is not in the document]
elimbaum: should current fund be an option for solidarity lead?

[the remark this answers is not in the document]
colinxfleming: I think yeah, probably a good idea

app/views/patients/_abortion_information.html.erb
```diff
@@ -43,8 +43,15 @@
...
-                          data: { 'tooltip-text': solidarity_help_text } %>
+                          data: { 'tooltip-text': solidarity_help_text } %>            
         <% end %>
+        <%= f.select :solidarity_lead,
+            solidarity_leads,
+            {
+              label: t('patient.abortion_information.clinic_section.solidarity_lead'),
+              label_class: 'tooltip-header-input',
+            },
+            data: {'tooltip-text': 'dogs' } %>
```
elimbaum, line 54: Slightly cursed because bootstrap select has slightly different behavior. ([github issue](https://github.com/bootstrap-ruby/bootstrap_form/issues/103), [code](https://github.com/bootstrap-ruby/bootstrap_form/blob/13af7bdf07bddd95685293a5e42d006ec0e0bad7/lib/bootstrap_form/inputs/select.rb)). rather than taking a single `options` hash, `select` takes an `options` hash _and_ an `html_options` hash. don't fully understand why, but the data attributes need to go in the latter.

elimbaum: Screen Shot 2022-08-16 at 18 57 21

colinxfleming (approved):
🙏  thank you very much

PR #2681, opened 2022-08-29
Removed budget bar. Added email to patients. Fixed docker errors on W…

thepiratebae:
…SL2.

I'm bad at RoR and have completed some work on Case Manager that's ready for review!

Working on feature requests for NH abortion fund. 

* Hid the budget bar
* Added email field to patients
* fixed css and js services commands to work with WSL2 and Windows Docker
* deleted web-variables.env to work with WSL2 and Windows Docker

edit_patient

add_new_patient


app/javascript/styles/budget_bar.scss
```diff
@@ -9,3 +9,7 @@ $budget-bar-font-size: 14px;
   color: $black !important;
   font-size: $budget-bar-font-size;
 }
+
+#budget_bar {
```
colinxfleming, line 13: I think taking this out is going to be a little more complicated - one of the tough things about maintaining this system is that we have every fund running off the same instance, so changes like this can't be universal or it would shut things off for others too. I would recommend solving this by changing the Config model so that this is toggle-able on and off, with a default of on -- see @lomky 's PR in https://github.com/DARIAEngineering/dcaf_case_management/pull/2343/files for an example of how to do this.

app/views/patients/_new_patient.html.erb
```diff
@@ -9,6 +9,11 @@
                        required: true,
                        pattern: "[0-9]{10}|[0-9]{3}\.[0-9]{3}\.[0-9]{4}|[0-9]{3}[-][0-9]{3}[-][0-9]{4}" %>
     </div>
+    <div class="col-2">
+      <%= f.text_field :email,
```
colinxfleming, line 13: Premise here makes sense, but email is pretty sensitive as far as data goes - if we're going to do this I want to make sure it's not available by default, and is opt-in per fund (which we can do by adding a Config along the lines of `Allow Patient Email Field`, default it to false, and let funds turn it to true if they really want it). This is kinda dangerous information to handle.

colinxfleming (commented):
hey @thepiratebae ! glad you were able to spin things up.

good news and bad news here -- good news is that I am super glad to have someone on windows contributing again so we keep that part of the codebase sharp.

bad news: daria is a multitenant application, which means we need to make these non-changes for most funds (who, for example, use the budget bar quite a bit). We fortunately have a way to do that via the Config model, where we can add certain things and then check for them. (For example, funds can completely turn off the practical support stuff if they want; look around the codebase for `hide_practical_support?` for the implementation.)

I'll follow up with you later in the week, but please post in slack if you want to discuss before then!

app/javascript/styles/budget_bar.scss
thepiratebae: AH, ok that makes sense. For some reason I thought different funds were running on different instances. I'll check out that PR, thank you.

app/views/patients/_new_patient.html.erb
thepiratebae: Sounds good, I'll ask specifically why they're looking for email to see if it's really necessary.

app/javascript/styles/budget_bar.scss
colinxfleming: I made an issue so we can track this element of work specifically - https://github.com/DARIAEngineering/dcaf_case_management/issues/2688

app/views/patients/_new_patient.html.erb
colinxfleming: Just to be clear this is plausible to me, but as a matter of approach whenever we add new fields we have a conversation about whether it's really necessary, what the risks are, whether we need to handle anything special etc. Email is particularly sensitive! so we'll definitely want to have a conversation about it. getting the reasoning/need on record from the fund is definitely appreciated.

colinxfleming: gotcha. let's target #2688 as your active unit of work for now (since it's reasonably straight forward and doesn't raise data privacy issues). I'll start up an email chain with NH to see if there are end runs around the other fields that don't involve tracking scary stuff.

PR #2694, opened 2022-09-06
Upgrade omniauth-google-oauth2 to fix #2693

xmunoz:
Going to deploy this branch to sandbox to test

PR #2695, opened 2022-09-06
Allow staging deploys from branches

xmunoz:
Right now, there is no way to test things on sandbox from a branch on Github. Sometimes, doing a sandbox deploy from a branch is required for stuff like #2694.

### DIFF
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   deploy-staging:


colinxfleming:
I'm not sure I totally understand what's going on here - can you help me understand?

fysa when I've had to do this I usually do the instructions here - https://devcenter.heroku.com/articles/git#deploy-from-a-branch-besides-main - basically boils down to `git remote add heroku (some url)` and then `git push heroku branch:main`, and then `git push heroku main` when I'm done (or wait until a bump of main overwrites it).

xmunoz:
This is adding the `workflow_dispatch` option to the Github Actions Deploy Staging workflow. This option adds a button to the workflow that allows you to select the branch to run the workflow on. I personally like this better than the git option, because it's more auditable so that there is no confusion about when and where staging releases are happening.

xmunoz:
Screenshots: https://docs.github.com/es/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow

colinxfleming (approved):

Issue #2688, opened 2022-09-01
Add a config for not displaying the budget bar


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

A fund requested that we add a config to shut off the budget bar, since they don't really use it and it just makes stuff more confusing.

* **What feature or behavior is this required for?**
fund workflow

* **How could we solve this issue? (Not knowing is okay!)**
Add a config to `app/models/config.rb` following the patterns there; add a form to the configs index so we can turn this on/off by fund. Make that config null out the budget bar display in `app/views/dashboards/index.html.erb` (probably by putting the `overview` partial in an `if Config.hide_budget_bar` block).

* **Anything else?**

reserved for @thepiratebae

colinxfleming:
this is resolved actually

PR #2696, opened 2022-09-06
Fixes "Add a config for not displaying the budget bar #2688"

thepiratebae:

Attempts to satisfy #2688 

* Adds a `hide_budget_bar` config in `config.rb`
* Renames `validate_hide_practical_support` to `validate_yes_or_no`
* Skips `overview` partial in `app/views/dashboards/index.html.erb` if config value is 'yes'

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -19,6 +19,7 @@ class Config < ApplicationRecord
     days_to_keep_fulfilled_patients: "Number of days (after initial entry) to keep identifying information for a patient whose pledge has been fulfilled and marked audited. Defaults to 90 days (3 months).",
     days_to_keep_all_patients: "Number of days (after initial entry) to keep identifying information for any patient, regardless of pledge fulfillment. Defaults to 365 days (1 year).",
     shared_reset: "Number of idle days until a patient is removed from the shared list. Defaults to 6 days, maximum 6 weeks.",
+    hide_budget_bar: 'Enter "yes" to hide the budget bar display.',
   }.freeze
 
   enum config_key: {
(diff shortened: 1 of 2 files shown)


app/models/config.rb
```diff
@@ -54,7 +56,7 @@ class Config < ApplicationRecord
       [:validate_singleton, :validate_start_of_week],
 
     hide_practical_support:
-      [:validate_singleton, :validate_hide_practical_support],
+      [:validate_singleton, :validate_yes_or_no],
```
colinxfleming, line 59: chefs kiss

app/models/config.rb
```diff
@@ -137,6 +142,10 @@ def self.shared_reset
     shared_reset_days.to_i
   end
 
+  def self.hide_budget_bar?
```
colinxfleming, line 145: If you wanted to add a test for that it would be nice, but not necessary - you can probably crib off the hide practical support test if it exists. should be in `tests/models/config_test.rb`.

app/views/dashboards/index.html.erb
```diff
@@ -1,6 +1,8 @@
 <div class="row">
   <div class="col-sm-12">
-    <%= render partial: 'dashboards/overview' %>
+    <% unless Config.hide_budget_bar? %>
```
colinxfleming, line 3: Again not necessary, but we tend to do browser selenium tests for stuff like this (rails calls em system tests). There's probably an example to crib off in `test/system/updating_configs_test.rb`. That said, I wouldn't bother with this unless you've written browser tests before, as they're kinda frustrating to troubleshoot.

colinxfleming (commented):
this is a great pr - narrow scope, generalizes some existing work, and very clean code.

It's in draft still so I assume there might be something else still coming? So holding on hitting approve until it moves to ready (and I didn't pull it down to test it). but from my observation this looks pretty good. Thanks for knocking it out!

colinxfleming:
gonna push this over the finish line since I'd like it out

colinxfleming (approved):

PR #2707, opened 2022-09-28
Don't allow CMs to change fulfillment attributes

krbullock:

Prevents CM users (all non-admins, to be accurate) from changing fulfillment attributes.

* Uses strong params to ignore nested `fulfillment_attributes` when creating or updating patients, except for admin users.
* Adds tests on `PatientsController#create` and `#update` to make sure we don't regress.

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -152,10 +152,12 @@ class PatientsController < ApplicationController
   OTHER_PARAMS = [:shared_flag, :initial_call_date, :pledge_sent].freeze
 
   def patient_params
-    params.require(:patient).permit(
-      [].concat(PATIENT_DASHBOARD_PARAMS, PATIENT_INFORMATION_PARAMS,
-                ABORTION_INFORMATION_PARAMS, OTHER_PARAMS, FULFILLMENT_PARAMS)
+    permitted_params = [].concat(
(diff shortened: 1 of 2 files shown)


app/controllers/patients_controller.rb
```diff
@@ -152,10 +152,12 @@ def find_patient_minimal
...
 
   def patient_params
-    params.require(:patient).permit(
-      [].concat(PATIENT_DASHBOARD_PARAMS, PATIENT_INFORMATION_PARAMS,
-                ABORTION_INFORMATION_PARAMS, OTHER_PARAMS, FULFILLMENT_PARAMS)
+    permitted_params = [].concat(
+      PATIENT_DASHBOARD_PARAMS, PATIENT_INFORMATION_PARAMS,
+      ABORTION_INFORMATION_PARAMS, OTHER_PARAMS
     )
+    permitted_params.concat(FULFILLMENT_PARAMS) if current_user.admin?
```
colinxfleming, line 159: very minor context fix here

```suggestion
    permitted_params.concat(FULFILLMENT_PARAMS) if current_user.allowed_data_access?
```

colinxfleming (approved):
spot on. thank you @krbullock (and extra thank you for expanding the tests to cover this case)!

PR #2709, opened 2022-10-03
Refine throttling on password and auth routes

djohnston32:

Sets dedicated rate limits on key authentication and password routes to protect against attacks or spamming. Also narrows window for checking the general ip rate limit.

* Add throttling rules for /users/auth routes and /users/password routes
* Check ip limit every 10 seconds instead of every 5 minutes. The 60rpm limit stays the same, it's just checked more often (10 requests / 10 seconds vs. 300 requests / 300 seconds).
* Clear rails cache between rack-attack tests so we don't have to worry about ips or emails conflicting across tests.

For reviewer:
* Tag this PR:

### DIFF
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -30,7 +30,7 @@ class Rack::Attack
   # Throttle all requests by IP (60rpm)
   #
   # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
-  throttle('req/ip', :limit => 300, :period => 5.minutes) do |req|
+  throttle('req/ip', :limit => 60, :period => 60.seconds) do |req|
     req.ip # unless req.path.start_with?('/assets')
   end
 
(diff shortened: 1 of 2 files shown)


djohnston32:
Even before my changes, I'm not able to trigger any of the throttling rules in my dev environment, and the tests for exceeding throttling limits always fail. Is there an option I can set somewhere so I can verify these changes locally?

krbullock:
> Even before my changes, I'm not able to trigger any of the throttling rules in my dev environment, and the tests for exceeding throttling limits always fail. Is there an option I can set somewhere so I can verify these changes locally?

Maybe it's hitting [the bit that excludes localhost](https://github.com/DARIAEngineering/dcaf_case_management/blob/main/config/initializers/rack-attack.rb#L16-L19) from the limits — can you trigger them if you comment that bit out?

djohnston32:
I had the same thought but it doesn't seem to trigger even with the localhost safelist rule commented out.

djohnston32:
Ah, figured it out. Rack-attack depends on caching which is disabled by default in development and test environments. I can see now that some of the new limits I set are too aggressive so I'll send out a revision after I make some improvements.

djohnston32:
Updated general rate limit by ip to 60 requests / 60 seconds in new revision. I as getting 503s during normal usage with 10/10, but 300/300 seemed too high. With 60/60 I only got throttled when simulating bad behavior, and normal operation didn't trigger it.

Also manually verified 5 requests / 20 seconds for password and auth routes were working properly, and verified all rack-attack tests were passing with caching enabled.

colinxfleming (approved):
nicely done! thank you for the dive through rack-attack configs!

colinxfleming:
(and through the very particular development environment configuration)

PR #2722, opened 2022-10-18
Sanitize strings before exporting csv

djohnston32:

Escape string fields containing formulas before exporting to prevent [CSV Injection](https://owasp.org/www-community/attacks/CSV_Injection).

* I opted for doing this directly in the export instead adding validation to the individual fields. This way there's no risk of forgetting this validation when a new field is added.
* I found formulas still executing with leading whitespace, so that's why it's included in the regex. Tested the regex with a number of examples with rubular. Included tab and carriage return because they were listed alongside the other formula characters in the OWASP documentation linked above.
* If we wanted to be even more aggressive, the CSV module has an option to force quotes on all fields. That would be more comprehensive but I opted for the more targeted approach because quotes everywhere seems potentially annoying for the CSV user.

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/exportable.rb
+++ b/app/models/concerns/exportable.rb
@@ -157,7 +157,16 @@ module Exportable
     value = public_send(field)
     if value.is_a?(Array)
       # Use simpler serialization for Array values than the default (`to_s`)
-      value.reject(&:blank?).join(', ')
+      sanitize(value.reject(&:blank?).join(', '))
+    else
+      sanitize(value)
+    end
+  end
+
+  def sanitize(value)
+    if value.is_a?(String) && value.start_with?(/\s*[=@+\-\t\r]/)
+      # Escape certain special characters to prevent formula injection.
+      "'#{value}"
     else
       value
     end
--- a/test/controllers/patients_controller_test.rb
+++ b/test/controllers/patients_controller_test.rb
@@ -14,7 +14,8 @@ class PatientsControllerTest < ActionDispatch::IntegrationTest
                       name: 'Susie Everyteen',
                       primary_phone: '123-456-7890',
                       other_phone: '333-444-5555',
-                      line: @line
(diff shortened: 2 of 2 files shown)


app/models/concerns/exportable.rb
```diff
@@ -157,7 +157,16 @@ def get_field_value_for_serialization(field)
...
       # Use simpler serialization for Array values than the default (`to_s`)
-      value.reject(&:blank?).join(', ')
+      sanitize(value.reject(&:blank?).join(', '))
+    else
+      sanitize(value)
+    end
+  end
+
+  def sanitize(value)
+    if value.is_a?(String) && value =~ /\A\s*[=@+\-\t\r]/
```
colinxfleming, line 167: do you think this makes this more readable? Could go either way
```suggestion
    if value.is_a?(String) && value.starts_with? /\s*[=@+\-\t\r]/
```

colinxfleming (approved):
nicely done. I think this is a pretty clean approach to a potentially complicated problem!

Seeking your thoughts on a ruby-ism, but will defer to you here I think. Optionally, a unit test would be nice but not required here.

I'll merge on Tuesday either way, since this is in good shape and fixes the injection problem. Thanks much @djohnston32 !

app/models/concerns/exportable.rb
djohnston32: Good call, I like `starts_with` more too now that you mention it. Changed in new revision. Was getting warnings about ambiguity without parentheses so added parentheses around the regex too.

djohnston32:
Changed to use `start_with` and added a unit test.

app/models/concerns/exportable.rb
colinxfleming: heroic

colinxfleming (approved):
outstanding

PR #2723, opened 2022-10-18
Add validation to ensure numeric fields are not negative

djohnston32:

Adds validation on numeric fields to ensure negative amounts of money aren't allowed.

Adds validation to the following fields:
* fulfillments
    * fund_payout
* patients
    * fund_pledge
    * naf_pledge
    * patient_contribution
    * procedure_cost
* archived_patients
    * fund_pledge
    * naf_pledge
    * patient_contribution
    * procedure_cost
* external_pledges
    * amount
* events
    * pledge_amount

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/archived_patient.rb
+++ b/app/models/archived_patient.rb
@@ -35,6 +35,11 @@ class ArchivedPatient < ApplicationRecord
             presence: true
   validates :appointment_date, format: /\A\d{4}-\d{1,2}-\d{1,2}\z/,
                                allow_blank: true
+  validates :procedure_cost, numericality: { only_integer: true, allow_nil: true, greater_than: 0 }
+  validates :fund_pledge,
+            :naf_pledge,
+            :patient_contribution,
+            numericality: { only_integer: true, allow_nil: true, greater_than_or_equal_to: 0 }
   validates_associated :fulfillment
 
   # Archive & delete audited patients who called a several months ago, or any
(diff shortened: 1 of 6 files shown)


app/models/archived_patient.rb
```diff
@@ -35,6 +35,11 @@ class ArchivedPatient < ApplicationRecord
             presence: true
   validates :appointment_date, format: /\A\d{4}-\d{1,2}-\d{1,2}\z/,
                                allow_blank: true
+  validates :procedure_cost, numericality: { only_integer: true, allow_nil: true, greater_than: 0 }
```
colinxfleming, line 38: my experience with accountants is that they looooove being able to zero things out as an indicator to themselves of something or other. I think for every field in all these models we can validate `only_integer: true, greater_than_or_equal_to: 0` like you are for the other fields here.

colinxfleming (changes requested):
This is right on. I think seeing this in action, I'd like to slightly expand the scope of this - there's not really a reason to not have most of our int fields be zero or up, right?

I think two changes here:

* validate `greater_than_or_equal_to: 0` on everything, rather than `greater_than: 0`
* expand this to the following integer fields on the following models:

```
Clinic: costs_5wks thru costs_30wks, gestational_limit
fulfillment: gestation_at_procedure
patient: last menstrual period weeks, last menstrual period days, age, household size children, household size adults
```

sorry for the change in direction!

app/models/archived_patient.rb
djohnston32: Sounds good! Updated that and added the rest of the fields in the latest revision.

colinxfleming (approved):
thank you very much, and thank you again for handling the change in direction so well!

PR #2733, opened 2022-10-27
Add unique index to fields with uniqueness constraints in model

djohnston32:

Several fields in `acts_as_tenant` models had uniqueness constraints in the model, but not enforced at the db level. This change adds unique indexes for those fields at the db level to fix that.

Added for the following fields:
* `Clinic :name`
* `Line :name`
* `CallListEntry  :patient` (this one was also scoped to :user)

A few other fields in tenant models already had unique indexes in the db.

For reviewer:
* Tag this PR:


djohnston32:
The changes from integer->bigint in `db/schema.rb` seem to be unrelated to this change by the way. Not sure where those are coming from.

colinxfleming (approved):
thank you!

the schema.rb changes are some weird rails bug - they've gone back and forth a couple of times. Not worth sweating it.

PR #2735, opened 2022-10-31
Change seed value for fax service

djohnston32:

Changes fax service seed value to the url of a real fax service. I used a fax service url instead of petsfinder because there's no other text in the UI to identify it as the fax service value. I saw efax elsewhere in the code, and I figured using something other than petfinder makes it easier to connect the link in the UI back to its source location in the code when developing/debugging.

Screenshot of development version of site after change:
image

For reviewer:
* Tag this PR:

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -92,7 +92,7 @@ fund2 = Fund.create! name: 'CatFund',
     Config.create config_key: :referred_by,
                   config_value: { options: ['Metal band'] }
     Config.create config_key: :fax_service,
-                  config_value: { options: ['www.yolofax.com'] }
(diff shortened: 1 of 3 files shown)


djohnston32:
For consistency's sake, I changed yolofax in the tests too even though I think that's not strictly necessary since it's just a test value that won't show up in the UI. For what it's worth, there's also a "metallicarules" URL in a few places in the tests that doesn't appear to be registered either. I left as is since it's just in tests but I'm happy to change that as well if we want to be extra cautious.

colinxfleming (approved):
thank you! works great, and thank you for updating the tests

PR #2736, opened 2022-10-31
Enable old password checking and session limiting

djohnston32:

Adds the [devise-security](https://github.com/devise-security/devise-security#session-limitable) devise extension and enables these modules:
* `:password_archivable` - saves used passwords in an old_passwords table for history checks (prevent reusing passwords)
* `:session_limitable` - ensures that there is only one session usable per account at once

The password functionality requires a new table `old_passwords`, and the session functionality requires a new field in the users table `unique_session_id`. The session field could be a bit confusing since it's separate from the existing `sessions` table, so I added a comment to the migration file adding it that should hopefully clarify why it's there for future readers.

Message after submitting password update form with previously used password:
image

Message in original browser after logging in as the same user in a different browser:
image

Fixes issues:
* https://github.com/DARIAEngineering/dcaf_case_management/issues/1460

For reviewer:
* Tag this PR:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem 'activerecord-session_store'
 
 # Our authentication library is devise, with oauth2 for google signin
 gem 'devise', '~> 4.8'
+gem 'devise-security'
 gem 'omniauth-google-oauth2', '~> 1.1.1'
 gem "omniauth-rails_csrf_protection", '~> 1.0'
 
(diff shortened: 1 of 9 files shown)


colinxfleming (changes requested):
great work here - tested and works as expected. 

A bit of business context: we're pretty committed to being english and spanish but don't support languages other than that. Would you mind removing the non-en and non-es i18n files? after that this is good to go!

colinxfleming (approved):
thank you! works great.

PR #2737, opened 2022-11-01
Validate field length on various fields

djohnston32:

Adds validation to various fields to protect against huge values being submitted.

* Adds basic max length validations on a number of fields
* Adds length checks on free-input config options
* Adds length checks on special circumstances

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -52,6 +52,19 @@ class Config < ApplicationRecord
   }.freeze
 
   VALIDATIONS = {
+    insurance:
+      [:validate_length],
+    external_pledge_source:
+      [:validate_length],
+    pledge_limit_help_text:
+      [:validate_length],
+    practical_support:
+      [:validate_length],
+    referred_by:
+      [:validate_length],
+    voicemail:
+      [:validate_length],
+
     start_of_week:
       [:validate_singleton, :validate_start_of_week],
 
@@ -60,7 +73,7 @@ class Config < ApplicationRecord
 
     budget_bar_max:
       [:validate_singleton, :validate_number],
-    
+
     resources_url:
       [:validate_singleton, :validate_url],
     fax_service:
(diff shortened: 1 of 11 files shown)


app/models/config.rb
```diff
@@ -252,4 +268,13 @@ def validate_patient_archive
     def validate_shared_reset
       validate_number && options.last.to_i.between?(SHARED_MIN_DAYS, SHARED_MAX_DAYS)
     end
+
+    def validate_length
+      total_length = 0
+      options.each do |option|
+        total_length += option.length
```
colinxfleming, line 275: I probably gave unclear directions here, so please forgive me. There are legit reasons for configs to have a total length of 1k+ - for example, at least two funds have a very long list of funds they work with that come out to 1500+ characters. So I think we'll need to adjust the strategy here in particular. We could:
* raise the cap to 4k, like notes are at, and keep this strategy as you have it written here (def fixes it, still keeps us flexible)
* check individual elements against the 150 char limit (mostly fixes the root problem, keeps us flexible)
* say to ourselves 'configs are an admin-only endpoint and so they don't really need validation' (a reasonable approach, would leave only this admin-only module vulnerable to this screwing-around-with)

I'll defer to you on the right path here.

colinxfleming (changes requested):
Excellent work here - I think two minor fixes in configs and then we can ship. Did some smoke testing of other fields and the controls worked perfectly.

app/models/config.rb
djohnston32: Makes sense! I opted for the first option, raising the cap to 4k. I don't feel strongly, but overall I feel like it gives sufficient protection while minimizing the likelihood of false positives compared to a per-element limit.

colinxfleming (approved):
thank you!

PR #2739, opened 2022-11-02
Logout user on email change

djohnston32:

Logout user when admin changes their email.

* Devise doesn't have options for configuring this out of the box, so this change implements the solution from [this stackoverflow answer](https://stackoverflow.com/a/29214116) instead. Adds a session validation token to the user table, and a method for removing the token on the user. Then calls the method when a user's email is updated to invalidate their session.

For reviewer:
* Tag this PR:


colinxfleming (approved):
works as expected! nicely done

PR #2740, opened 2022-11-02
Add host header checks in production

djohnston32:

Adds host header checks to prevent maliscious redirects.

* Adds host header check to production environment

For reviewer:
* Tag this PR:

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -87,6 +87,7 @@ Rails.application.configure do
   # CUSTOM STUFF STARTS HERE
   # Prevent host header injection
   config.action_controller.default_url_options = { host: ENV['SITE_URL'] }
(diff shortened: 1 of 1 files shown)


djohnston32:
Do we also want to have this in place for sandbox? I assume that's not considered a production environment? In the development environment, `config.hosts` is set to `nil` with the comment `# To allow usage of lvh.me, a tunnelling service, in development for multitenancy`, so I'm not sure if adding this protection there would cause other issues.

colinxfleming (approved):
nailed it!

fwiw: Sandbox is configured identically to our production environment, just with less beefy machines. So it does load stuff from `production.rb` and this should apply to it, since SITE_URL is set in there. Development environments as per that comment refer to, like, localhost or docker.

PR #2750, opened 2022-11-15
Prevent locked accounts from accessing user profile

djohnston32:

Adds the locked account check to the `/users/edit` and `/users` routes to prevent locked accounts from viewing or updating their user profile.

* used devise controller generator to generate registration controller.
* Added locked account check to newly generated controller

For reviewer:
* Tag this PR:


colinxfleming (approved):
Yep, this does the trick! Excellent work here, very impressed you managed to keep it as surgical as you did.

PR #2758, opened 2022-11-23
Budget bar statistics

### LINKED ISSUE #2658: Should we optionally display the weekly/monthly budget on the budget bar?

More than one fund has reached out to me recently to ask for this. We've generally discouraged doing Real Budgeting in DARIA (excel is way better and more consistent than we'll ever be) but people are making pretty good arguments that some signals would be helpful.

@lwaldsc as one of the original intellectual architect of the budget bar interested in your thoughts - my recollection is that we didn't do this originally because we thought it would bum out DCAF CMs (one late case and there you go). but wonder if this is similarly a bummer for smaller funds.

not doing math

if we _were_ to do this, it would look like:
* config to toggle display on/off
* add some math to the budget bar like `(soft pledged: X) + (sent pledged: Y) = X+Y / (weekly budget) (Z%)`

* **Anything else?**

nah

elimbaum:

* adds statistics to budget bar (spent, remaining, percentages)
* refactors code to make adding more aggregate stats easier (and prevent code duplicate!)

* Fixes #2658 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/dashboards/_budget_bar.html.erb
+++ b/app/views/dashboards/_budget_bar.html.erb
@@ -16,13 +16,19 @@
 
 <div class="row" id='budget-report'>
   <% expenditures.each_pair do |type, patient_hashes| %>
-    <div class="col-sm-4">
-      <%= t("dashboard.budget_bar.#{type}_report",
-            amount: number_to_currency(expenditures[type].map { |h| h[:fund_pledge] }.inject(:+) || 0,
-                                       precision: 0,
-                                       unit: '$',
-                                       format: '%u%n'),
-            count: expenditures[type].count) %>
+    <div class="col-sm-3">
+      <%= budget_bar_statistic(t("dashboard.budget_bar.#{type}"), expenditures[type], limit) %>
     </div>
   <% end %>
+
+  <div class="col-sm-3">
+    <%= budget_bar_statistic(t('dashboard.budget_bar.spent'), expenditures.values.flatten, limit) %>
+  </div>
+
+  <div class="col-sm-3">
+    <%= budget_bar_statistic_builder(t('dashboard.budget_bar.remaining'),
+                                     budget_bar_remaining(expenditures, limit),
+                                     nil,
+                                     limit) %>
+  </div>
 </div>
--- a/app/views/dashboards/_budget_bar.html.erb
+++ b/app/views/dashboards/_budget_bar.html.erb
@@ -26,15 +26,19 @@
 
   <% if Config.show_aggregate_statistics? %>
     <div class="col-sm-3">
-      <%= budget_bar_statistic(t('dashboard.budget_bar.spent'),
-                               expenditures.values.flatten,
-                               limit: limit) %>
+      <strong>
+        <%= budget_bar_statistic(t('dashboard.budget_bar.spent'),
+                                 expenditures.values.flatten,
+                                 limit: limit) %>
+      </strong>
     </div>
 
     <div class="col-sm-3">
-      <%= budget_bar_statistic_builder(name: t('dashboard.budget_bar.remaining'),
-                                       amount: budget_bar_remaining(expenditures, limit),
-                                       limit: limit) %>
+      <strong>
+        <%= budget_bar_statistic_builder(name: t('dashboard.budget_bar.remaining'),
+                                         amount: budget_bar_remaining(expenditures, limit),
+                                         limit: limit) %>
+      </strong>
     </div>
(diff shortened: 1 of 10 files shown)


elimbaum:
new view
Screen Shot 2022-11-22 at 22 49 41

elimbaum:
with config turned off

Screen Shot 2022-11-25 at 18 27 03

q's here:
- do we still want to show percents on sent/pledged when stats are turned off?
- do we want percents on those fields to begin with? (regardless of config value)
- should we unit test or is system test sufficient?

elimbaum:
@colinxfleming rereading the issue noticed you originally wanted to cut the %s. let me know what you think about this ^ both in terms of language and data presentation.

colinxfleming:
Answer to qs:
* I think if aggregate statistics is turned off, let's keep it exactly the same as it is now
* Seeing the percents in action I kinda like the percents (and you've done a great job of handling the edge cases/overflow cases here), so if you're comfortable with them let's keep them
* a lot of this hinges on `sum_fund_pledges` so a unit test for that might be nice, but if we're pressed then I think it's skippable since, as you noted, there are system tests for this

colinxfleming:
I don't really have any suggested changes, this is great work. Happy to hit approve when it comes out of draft

elimbaum:
thanks for comments!

> I think if aggregate statistics is turned off, let's keep it exactly the same as it is now

by "same as it is now", do you mean `main` or what I currently have?

& i'll make sure to exercise `sum_fund_pledges` a bit better in unit tests.

colinxfleming:
by same as it is now, I mean behavior on main - I'd like this to essentially function as a feature swap, where you can opt in to this fancy thing by config, but otherwise it's indistinguishable from how it works now.

elimbaum:
ok! i'll adjust that

elimbaum:
when stats are turned off, here's what it looks like

Screen Shot 2022-11-27 at 21 53 56

besides some minor layout changes, identical to before.

app/views/dashboards/_budget_bar.html.erb
```diff
@@ -16,13 +16,28 @@
...
-            count: expenditures[type].count) %>
+    <div class="col-sm-3">
+      <%= budget_bar_statistic(t("dashboard.budget_bar.#{type}"),
+                               expenditures[type],
+                               limit: limit,
+                               show_aggregate_statistics: Config.show_aggregate_statistics?) %>
+    </div>
+  <% end %>
+
+  <% if Config.show_aggregate_statistics? %>
```
elimbaum, line 27: checking the config twice here. looks like it gets cached, but is there a better way? perhaps more generally, does this architecture (statistic-string builder + options) seem ok to you?

app/views/dashboards/_budget_bar.html.erb
```diff
@@ -16,13 +16,28 @@
...
+                               limit: limit,
+                               show_aggregate_statistics: true) %>
+    </div>
+
+    <div class="col-sm-3">
+      <%= budget_bar_statistic_builder(name: t('dashboard.budget_bar.remaining'),
+                                       amount: budget_bar_remaining(expenditures, limit),
+                                       count: nil,
+                                       limit: limit,
+                                       show_aggregate_statistics: true) %>
```
elimbaum, line 40: in these cases, we have already checked the config — we know it's `true`.

elimbaum:
...actually, maybe we should be using [keyword arguments](https://stackoverflow.com/questions/15062570/when-to-use-keyword-arguments-aka-named-parameters-in-ruby) here.

elimbaum:
@colinxfleming ready to go!

app/views/dashboards/_budget_bar.html.erb
colinxfleming: yeah, this is a very railsy pattern for html generation and I think it serves us nicely here. I don't think double-checking is problematic here (as you noted, query gets cached) but the way around this for a heavier query that didn't get cached would be to declare this as an instance variable in the controller, e.g. `@show_stats = Config.show_aggregate_statistics?` and then check that variable directly. (Again, to be clear, shouldn't do that here thanks to db cache magic!)

colinxfleming (approved):
I always forget ruby kwargs are a thing. They're well used here.

I'm going to make a very soft style change and put spent/remaining in bold so they're easy to distinguish, and then I'm going to merge. No real suggestions from my corner on the logic, this is very well done. Thanks much @elimbaum !

PR #2762, opened 2022-12-02
Do not sign in on password reset

djohnston32:

Similar to https://github.com/DARIAEngineering/dcaf_case_management/pull/2708 which signed out the user after a password _change_, this changes to not sign user in after password _reset_.

 Flips devise's config.sign_in_after_reset_password option to false. Displays a message set by devise (changeable [here](https://github.com/DARIAEngineering/dcaf_case_management/blob/96277ff58f9e54395288bcb1548c565ae164b8ca/config/locales/devise.en.yml#L38)): "Your password has been changed successfully."

For reviewer:
* Tag this PR:

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -236,7 +236,7 @@ Devise.setup do |config|
 
   # When set to false, does not sign a user in automatically after their password is
   # reset. Defaults to true, so a user is signed in automatically after a reset.
-  # config.sign_in_after_reset_password = true
+  config.sign_in_after_reset_password = false
 
   # ==> Configuration for :encryptable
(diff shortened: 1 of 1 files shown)


djohnston32:
Not sure of a good way to verify this locally since the test users' emails are fake. If there's a good way to verify I can go back and do so.

colinxfleming (approved):
Confirmed working on my end! I tested by pulling this code down, setting `SITE_URL` in my local env vars and adjusting development.rb a bit to have `config.action_mailer.default_url_options` and then doing the password reset, which dumps the email contents to the logger.

colinxfleming:
thank you @djohnston32 !

PR #2790, opened 2022-12-23
Allow multiple emails in email_for_pledges

### LINKED ISSUE #2660: Allow multiple clinic emails

No good deed goes unpunished! A few clinics have multiple emails they use, so we should represent that.

> Several clinics ask us to send the pledge form to more than one email. Can you make it possible to add multiple email addresses separated by a coma? Or a separate cc box? That would be perfect.

tough to put into words

add `multiple` attribute to the email field; keep it simple, right? 

* **Anything else?**

colinxfleming:

This loosens the email_for_pledges field to allow arbitrary text, to let people punch in multiple emails.

We're giving up email formatting in the frontend in exchange for not doing a lot of complicated multi-element form parsing, which is a pain in the behind. But I think arbitrary text is probably fine here.

* Allow arbitrary text input (max length 500) for Clinic#email_for_pledges
* gentle adjustments around that (i18n)

The change here is that this sorta thing is now allowed, and we aren't doing any enforcement on it besides length. I think that's probably fine, since we aren't actively emailing people out of this system and so it's not Required to be Right.

image

* Fixes #2660

For reviewer:
* Tag this PR:

### DIFF
--- a/app/controllers/clinics_controller.rb
+++ b/app/controllers/clinics_controller.rb
@@ -49,8 +49,9 @@ class ClinicsController < ApplicationController
 
   def clinic_params
     clinic_params = [:name, :street_address, :city, :state, :zip,
-                     :phone, :fax, :active, :email_for_pledges,
-                     :accepts_naf, :accepts_medicaid, :gestational_limit]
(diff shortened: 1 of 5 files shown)


lomky (approved):
:shipit:

PR #2791, opened 2022-12-24
allow sorting users table by role and by status

### LINKED ISSUE #2656: Make user index table sortable by status

Request from a fund:

> Can the User Account Management page be updated so that it's sortable? Specifically we are interested in being able to sort by Status, so that the users with Active accounts are up at the top of the list rather than mixed together. 

Makes sense!

User management

I think this is mostly a matter of applying the stupidtable data attribute? So hopefully a quick one-liner. See `app/views/users/_users_table` for both the location of the change and, I think, an example of stupidtable being used.

* **Anything else?**

nope!

colinxfleming:

long overdue simple change here for user management - add tablesort by role and by status. Hate to grab a beginner friendly but want to reduce this pain point for funds with large vol bases.

* Add tablesort on user panel by role and status

image

sort looks good!

* Fixes #2656 

For reviewer:
* Tag this PR:


app/views/users/_users_table.html.erb
```diff
@@ -3,10 +3,10 @@
     <tr class="user-table-header">
       <th><!-- spacer --></th>
       <th><%= t('user.user_table.full_name') %></th>
```
lomky, line 5:
```suggestion
      <%= th_autosortable t('user.user_table.full_name'), 'string', local_assigns % %>
```

why not?

app/views/users/_users_table.html.erb
```diff
@@ -3,10 +3,10 @@
...
-      <th><%= t('user.common.role') %></th>
+      <%= th_autosortable t('user.common.role'), 'string', local_assigns %>
       <th><%= t('user.common.last_login') %></th>
```
lomky, line 7:
```suggestion
      <%= th_autosortable t('user.common.last_login'), 'string', local_assigns %>
```

ditto

lomky (approved):
🎉

PR #2794, opened 2022-12-26
Generate pledges via service

colinxfleming:

DARIA side of remote pledge generation. This sets up the option to make remote calls to a service we control to generate a pdf based on some inputs. We can turn that on for a fund with `PledgeConfig.remote_pledge` (to turn on the flow) and `PledgeConfig.remote_pledge_extras` (to add fund-specific inputs if necessary). By and large this new endpoint follows the same general path as the existing flow, and is mostly reliant on different db fields. 

(Please take my word for it that it calls out to the ext service properly.)

* Adds an endpoint to call out to a yet-to-deploy service for pledge generation
* model changes, library installs etc in support of same

The new flow with a required field, built with `Fund.last.build_pledge_config(remote_pledge: true, remote_pledge_extra: {big_cat: true}).save`:

image

* Bumps https://github.com/DARIAEngineering/dcaf_case_management/issues/2773

For reviewer:
* Tag this PR:


app/controllers/patients_controller.rb
```diff
@@ -55,6 +55,31 @@ def download
     end
   end
 
+  def fetch_pledge
+    endpoint = ENV.fetch('PLEDGE_GENERATOR_ENDPOINT', 'http://localhost:3001/generate')
+    basic_auth = { username: ENV.fetch('PLEDGE_GENERATOR_USER', 'apiuser'), password: ENV.fetch('PLEDGE_GENERATOR_TOKEN', 'PLEDGETOKEN') }
+    extra_params = params.permit(*current_tenant.pledge_config.remote_pledge_extras.map { |x, y| x.to_sym })
+    payload = {
+      fund: Rails.env.development? ? 'test' : current_tenant.name.downcase,
+      base: {
```
colinxfleming, line 64: the idea here is that some of these every fund is going to need (patient name, who's receiving the $, etc) and some are very fund specific. This is trying to set this up to be as dynamic as possible, and my hope is that when it's situated it won't require code changes in this codebase. I do think it might take some time for what's in `base` to settle though, so please don't think about that too hard while we're reviewing.

app/models/pledge_config.rb
```diff
@@ -1,11 +1,6 @@
 class PledgeConfig < ApplicationRecord
   acts_as_tenant :fund
 
-  validates_presence_of :contact_email,
```
colinxfleming, line 4: different fields now do this so there's a legit reason for these not to be filled out

app/views/patients/_pledge.html.erb
```diff
@@ -16,6 +16,21 @@
...
+      <% elsif current_tenant.pledge_config.remote_pledge? %>
+        <div class="alert alert-danger" style="display:none;"> <%= t('patient.pledge.step_two.blank_name_error') %></div>
+        <p class="row"><strong><%= t('patient.pledge.step_two.generate_note') %></strong></p>
+
+        <div class="row">
+          <%= bootstrap_form_with url: fetch_pledge_patient_path(patient),
+                                  method: 'post',
+                                  local: true do |f| %>
+            <% current_tenant.pledge_config.remote_pledge_extras.each_pair do |field, required| %>
+              <%= f.text_field field.to_sym, label: field.humanize, required: required %>
```
colinxfleming, line 28: Rather than enumerated error messages, we're leaning a lot more on `required` html attributes here. (The shape of `remote_pledge_extras` hashes is `field: required_bool`, e.g. `{must_have: true, nice_to_have: false}`.)

colinxfleming (commented):
couple of notes

app/views/patients/_pledge.html.erb
lomky: this expected structure would be good to put in the DB comment field

lomky (approved):
Ty for humoring my paranoia, lgtm!

PR #2815, opened 2023-01-18
Upgrade ruby to 3.2.0

xmunoz:

Basically this PR, but without renovate: #2792


xmunoz:
JK, looks like there are breaking changes in this upgrade.

xmunoz:
Down to only 2 errors.

xmunoz:
I give up

xmunoz:
@colinxfleming @lomky can one of you take a look?

colinxfleming:
it looks like you have it in a good state here and the failures are bullet related, which is probably a bs error. I'm gonna try downgrading bullet

PR #2823, opened 2023-01-26
county field dropdown

### LINKED ISSUE #2771: Optionally restrict county field to a dropdown

Some funds work only in service areas, and so geo matters a lot to them. We enumerate states since there aren't that many valid ones, but sometimes there are only so many counties funds care about, so it might make sense to enumerate these as well with a dropdown rather than just the free text input field.

cleaner patient info tracking

This is probably a config of comma separated values where each value is a county, and if the config is set this means that you get a dropdown something like `[nil, values, ..., other]` instead of the free text field. (My hunch is that we don't need to support free text Other values, and in fact that would be a bit of a hassle.)

* **Anything else?**
nah

elimbaum:

* adds a config for county field
* if config is defined, display a dropdown
* otherwise, display text field (existing behavior)

* Fixes #2771 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -117,6 +117,14 @@ module PatientsHelper
     ]
   end
 
+  def county_options(current_value = nil)
+    county_options = Config.find_or_create_by(config_key: 'county').options
+
+    return [] if county_options.blank?
+
+    options_plus_current([nil] + county_options, current_value)
+  end
+
   def household_size_options
     (0..10).map { |i| i }
            .unshift([t('common.prefer_not_to_answer'), -1])
@@ -146,6 +154,8 @@ module PatientsHelper
 
     active_clinics | inactive_clinics
   end
+
+  
 
   def disable_continue?(patient)
     patient.pledge_info_present? ? 'disabled="disabled"' : ''
--- a/app/helpers/patients_helper.rb
+++ b/app/helpers/patients_helper.rb
@@ -155,8 +155,6 @@ module PatientsHelper
     active_clinics | inactive_clinics
   end
 
-  
-
   def disable_continue?(patient)
     patient.pledge_info_present? ? 'disabled="disabled"' : ''
   end
--- a/app/views/patients/_patient_information.html.erb
+++ b/app/views/patients/_patient_information.html.erb
@@ -40,7 +40,13 @@
           </div>
         </div>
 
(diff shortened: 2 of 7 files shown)


elimbaum:
Screen Shot 2023-01-25 at 20 50 00

example: prior value was `Test`; dropdown options `A B C` populate.

app/views/patients/_patient_information.html.erb
```diff
@@ -40,7 +40,13 @@
...
         </div>
 
-        <%= f.text_field :county, autocomplete: 'off' %>
+        <% if county_options.present? %>
+          <%= f.select :county, options_for_select(county_options(patient.county),
+                                        patient.county), autocomplete: 'off' %>
+        <% else %>
+          <%= f.text_field :county, autocomplete: 'off' %>
+        <% end %>
+
```
elimbaum, line 49: --> new view code

colinxfleming: well handled I think! maybe a system test confirming this works, but I don't think that's completely necessary and defer to you.

app/helpers/patients_helper.rb
```diff
@@ -117,6 +117,14 @@ def income_options
     ]
   end
 
+  def county_options(current_value = nil)
```
colinxfleming, line 120: could we get a quick unit test for this?

colinxfleming (approved):
great work as usual! I think some additional quick tests (unit test for county_options dropdown, maybe a system test testing that it swaps to the dropdown if the config is set) would be good here but I don't think they're critical - interested in whether you think they're worth it.

Once we have that square we can merge, but this is in mergeable shape as is, so marking as such!

db/schema.rb
```diff
@@ -48,9 +48,9 @@
     t.datetime "pledge_generated_at", precision: nil
     t.datetime "pledge_sent_at", precision: nil
     t.boolean "textable"
-    t.bigint "clinic_id"
-    t.bigint "pledge_generated_by_id"
-    t.bigint "pledge_sent_by_id"
+    t.integer "clinic_id"
```
colinxfleming, line 51: these have gone back and forth several times and I'm choosing not to think about it very hard anymore

colinxfleming (approved):
outstanding work as per usual! nothing to spruce up here, thanks much @elimbaum !

PR #2841, opened 2023-02-11
Remove Turbolinks

### LINKED ISSUE #2667: help tooltips being displayed twice on navigation

on navigation away from the patient information page, the help tooltips (`(?)`) are momentarily doubled up. weird! see screenshot below.

good user interface

may need to run a debugger in browser to figure out what's causing this! could be something with turbolinks, but I don't know why the help tooltips are the _only_ UI element being doubled.

* **Anything else?**

screenshot:

Screen Shot 2022-07-26 at 22 03 22

elimbaum:

* Removes turbolinks - use normal links, etc.
* Replaces load events with `DOMContentLoaded`
* Compare with #2842 

* Fixes #2667

For reviewer:
* Tag this PR:


colinxfleming:
fwiw I'm pretty okay with the removal option over the replace option - if turbolinks weren't a rails standard for so long we would have ditched it awhile ago, since it requires so many 'you gotta replace this regular thing with this specific incantation' usage. Do let me know what you think when you're satisfied with your poking around!

colinxfleming (approved):
the only difference I see here is that the blue loading bar doesn't happen (which is the Turbolinks Signal); other than that, this looks fantastic. I think if you want to remove the `data-turbolinks-track` usage in some of the views to round this out, that'd be pretty good but not required. I think fire when ready!

elimbaum:
waiting on #2880 due to security advisory

colinxfleming (approved):

PR #2863, opened 2023-03-08
Add note in docs about runtime of deletion script

xmunoz:
We learned this after off-boarding our first fund.

### DIFF
--- a/docs/administering/OFFBOARDING_A_FUND.md
+++ b/docs/administering/OFFBOARDING_A_FUND.md
@@ -27,7 +27,7 @@ I suggest doing these console commands by connecting to the console via the Hero
    ```bash
    rails "fund_management:delete_fund[CatFund]"
    ```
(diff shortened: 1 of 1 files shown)


docs/administering/OFFBOARDING_A_FUND.md
```diff
@@ -27,6 +27,7 @@ I suggest doing these console commands by connecting to the console via the Hero
    ```bash
    rails "fund_management:delete_fund[CatFund]"
    ```
+   NOTE: This step takes a long time to run, on the order of 10 minutes. Be patient if the script appears to be hanging. It's not hanging.
```
lomky, line 30:
```suggestion
   NOTE: This step takes a long time to run, on the order of 10 minutes. Be patient if the script appears to be hanging. It's likely not hanging. If you are still unsure, the script can be safely interrupted and restarted after you have investigated the console.
```

lomky (approved):

PR #2880, opened 2023-03-16
updated rack and activesupport

dfelsent:

* (your change here)
* (another change here)
* (etc)

* Fixes #X
* Bumps #Y

For reviewer:
* Tag this PR:


Gemfile
```diff
@@ -116,4 +116,6 @@ group :test do
   gem 'pdf-inspector', require: 'pdf/inspector' # test pdf contents
   gem 'minitest-stub-const'
   gem 'rack-test', '>= 0.6.3', require: 'rack/test' # needed to test rack-attack
+  gem 'activesupport',  '>= 7.0.4.3'
+  gem 'rack', '>= 2.2.6.4'
```
colinxfleming, line 120: sorry for the bad directions from me here! Generally rails is a bundle of gems that includes activesupport, activemodel, active-whatever, etc and if those aren't in version sync with each other some weird stuff happens. (I think `rack` might also be in there?) long story short if we wanted to pin activesupport, we'd want to pin `rails` instead, but a `bundle update rails` might also do the trick.

Since we gotta get the build green again I'm gonna take this and do the `bundle update`s over the gemfile pins.

Really appreciate your doing the work of proving that this gets the build green again and I am very grateful for that - sometimes version pins like this unnest a whole slew of weird junk. thank you!

colinxfleming (approved):
thank you @dfelsent !

PR #2893, opened 2023-03-27
Practical support: add support_date field

### LINKED ISSUE #2882: Practical support: add support_date field

Complex abortions can be a multi-day process, and patients may need rides etc on multiple days. We should adjust our practical support flow to allow this case. 

Coordinating practical support across multiple days

Probably a few steps here:
* right now we enforce practical support type uniqueness. We'll probably have to drop that, ~or scope it to day in addition. I don't know if I have a strong opinion here so whatever's easiest~ (CF note: drop the constraint here I think) 
* add a support_date db field to the practical support model
* add a support_date input to the practical support partial. (You can probably crib off `Patient.appointment_date` and its input)

* **Anything else?**

mercedesb:
Complex abortions can be a multi-day process, and patients may need rides etc on multiple days. We should adjust our practical support flow to allow this case.

* Remove support type uniqueness constraint
* Add support_date to practical support model
* Update practical support UI to allow editing support_date

image

* Fixes #2882 

Open questions
* I didn't add any validations for support_date. I didn't think it made sense to require presence b/c there will be a whole bunch of existing records without a date. Let me know if you want any validation checking that the date is in the future or something like that.
* Let me know if you prefer something different for the UI changes.
* I'm not great at Spanish, I think my translation makes sense but probably good to have someone fluent make sure it's correct.

For reviewer:
* Tag this PR:

### DIFF
--- a/app/views/practical_supports/_entries.html.erb
+++ b/app/views/practical_supports/_entries.html.erb
@@ -1,20 +1,20 @@
 <% if patient.practical_supports.present? %>
-<div class="row practical-support-headers">
-  <div class="col-10">
-    <div class="row">
-      <h5 class="col-3"><%= t 'patient.practical_support.support_needed' %></h5>
-      <h5 class="col-4"><%= t 'patient.practical_support.point_of_contact' %></h5>
-      <h5 class="col-3">Amount</h5>
-      <h5 class="col-2"><%= t 'patient.practical_support.support_date' %></h5>
+  <div class="row practical-support-headers">
+    <div class="col-12">
+      <div class="row">
+        <h5 class="col-3"><%= t 'patient.practical_support.support_needed' %></h5>
+        <h5 class="col-5"><%= t 'patient.practical_support.point_of_contact' %></h5>
+        <h5 class="col-4">Amount</h5>
+      </div>
     </div>
+    <div class="col"><!-- Where button would go --></div>
   </div>
-  <div class="col"><!-- Where button would go --></div>
-</div>
-
-<% patient.practical_supports.each do |entry| %>
   <hr/>
-  <%= render 'practical_supports/edit',
-             patient: patient,
-             practical_support: entry %>
+
+  <% patient.practical_supports.each do |entry| %>
+    <%= render 'practical_supports/edit',
+              patient: patient,
+              practical_support: entry %>
+    <hr/>
   <% end %>
 <% end %>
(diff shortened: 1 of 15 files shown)


colinxfleming (commented):
looking great on first glance! I think I would say let's move `Remove` and the new date field onto a second row and add some spacer indicator -- that's gonna get cramped on small screens. So I think something like:

* Row 1: Support Needed, providing Fund, Amount
* Row 2: Date, confirmed, Remove button
* some kinda line break/hr/spacer (I'm not sure we have a good existing example of this, so follow your heart)

other than that I think this is extremely looking good.

re: the other questions, 
* I don't think you need to add a validation for support date and we're generally in the habit of 'collect less data, not more, unless it's crucial' -- I don't think support date is generally crucial.
* Don't worry about the spanish translation; our funds understand we're in the business of 'close enough' when it comes to spanish mode (and tend to reach out if we pitch something that doesn't make any sense)

I'll give this a deeper review in a few days!

mercedesb:
Okee dokee, I went ahead and updated the UI. The first screenshot is on a normal screen (and it'll just grow as the screen gets bigger) and the second screenshot is on a bit smaller screen.

image

image

app/views/practical_supports/_entries.html.erb
```diff
@@ -1,19 +1,20 @@
 <% if patient.practical_supports.present? %>
-<div class="row practical-support-headers">
-  <div class="col-10">
-    <div class="row">
-      <h5 class="col-3"><%= t 'patient.practical_support.support_needed' %></h5>
-      <h5 class="col-5"><%= t 'patient.practical_support.point_of_contact' %></h5>
-      <h5 class="col-4">Amount</h5>
+  <div class="row practical-support-headers">
```
colinxfleming, line 2: I'm not quite sure what's happening here, but it looks like when I update an attribute on one of these entries it deletes it from db. To recreate, start with something like this:

image

after unchecking the confirm box, it disappears (and deletes from db):

image

colinxfleming (changes requested):
The design and flow of this is spot on! Marking requested changes because of the form deleting bug noted inline. I think this is probably a result of the button moving inside some form elements or something like that, such that the autocomplete of the form makes it think it wants to hit delete instead of patch. But not totally clear.

I think the system tests are correctly catching this bug, so I think when those are addressed we're good to go here.

Let me know if you've got any questions or need an assist here. Sorry for the late review!

mercedesb:
@colinxfleming sorry about the delay in getting this fixed. I started a new job 2 weeks ago and that's been taking up all my brain space. I think this should be good to go now. Pesky `<% end %>` in the wrong place.

colinxfleming (approved):
working great for me. sorry it took so long to review this! thank you very much!

Issue #1559, opened 2018-12-17
UI cleanup work on Config index page


colinxfleming:
Config improvements via an excellent flag by @lomky in https://github.com/DCAFEngineering/dcaf_case_management/pull/1503#pullrequestreview-185419523 --

> As I was poking around in the GUI, I found myself wanting some visual clarity in the Current Options list for hard coded base options vs custom options. There's also some weirdness in the languages display. Ultimately I think those are nice-to-have and probably things to put to a general 'cleanup the configs page' issue. Thoughts?

this is an excellent flag and we should do it!

colinxfleming:
marking future features since there isn't an obvious Next Step to this / there's a little hashing out to do yet

colinxfleming:
@emtot22 is taking a design look at this; @colinxfleming owes notes about how configs work with written examples on sandbox

colinxfleming:
How configs work:

The /configs page is a series of global configurations a fund's case management team director can apply. These generally fall into three categories:

* More dropdown options for case managers: Some, like Voicemail Options and Practical Support Options, expand the list of dropdown options when case managers are logging data. For example, at sandbox.dcabortionfund.org/configs , make a change to Language Options - add some language that isn't there and hit the Update Options for Language button right below it, and then head to sandbox.dcabortionfund.org/patients/42/edit and see your new language in the Preferred Language dropdown. These are also previewed on the Config page itself.

* Resources for case managers: Some, like Fax Service Options and Resources Url Options, add a link to a fund's case manager guidance (usually a googly drive link). For example, you'll see the Resources Url is a link to petfinder, and the 'CM Resources' link in the top nav panel reflects that. (You can edit and play around with this too, and see your changes reflected in the navbar after you save them.)

* Actual backend config: Days to keep all patients and days to keep fulfilled patients let funds determine when patient information gets destroyed; some, like Budget Bar Max Options, remind case managers about money.

There are a few other one-off things but generally things fall into these buckets.

Team Directors look at these relatively rarely and edit them even less. Still, it's important to have this be a relatively safe, validated page, because bad configuration values can potentially break parts of DARIA.



Things I don't like about this page:

* It's an ugly list of form inputs that aren't really grouped by rhyme or reason. This was fine when it was like, three, but now it's 15-16 inputs and we'll probably have to add more as time goes on. I would bet we can do better UX than 'eye bleedingly long list'.

[...]

PR #2921, opened 2023-04-13
Improve the configuration model

Nubbify:
What we're doing here is both adding a new dictionary for default values (in anticipation of pre-populating these fields in the new config designs and also because it's a bit cleaner) and also changing the way we do validation so that we return useful error messages in our errors array whenever validation fails.

* New DEFAULTS dictionary inside of the config model
* Return validation failure reasons from config controllers

* #1559


lomky:
Moved to #3372 & matched to the updated config. Thanks for this @Nubbify!

PR #2923, opened 2023-04-13
Autodeletion

### LINKED ISSUE #2847: Create a configurable, automatic data deletion cleanup process

When we initially created Daria for DCAF back in ~2015 we prioritized anonymized data retention to meet their needs and protect privacy. Now that we serve many funds and times have changed, we want to make sure funds have the final call on whether their patient data is archiving or just fully deleted after a certain time.

Reflecting potentially changed risk decision making over the past few years. Empowering users. Relates to #2682 

- Add a configuration toggle for 'delete after X days' for patients.
- Add that deletion processing
  - This should look a lot like how the Archive logic is configured and flows
- make sure it applies to Archived Patients as well

* **Anything else?**
This should definitely be an opt in - we don't want folks surprised by this behavior.

elimbaum:

* creates autodeletion system
* TODO testing
* TODO nightly

* Fixes #2847 
* Bumps #Y

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/concerns/autodeletable.rb
+++ b/app/models/concerns/autodeletable.rb
@@ -3,11 +3,8 @@ module AutoDeletable
     extend ActiveSupport::Concern
 
     def delete_date
-        if Config.days_until_delete.blank?
-            initial_call_date + Config.days_until_delete
-        else
-            nil
-        end
+        # if config is blank, we return nil here
+        initial_call_date + Config.days_until_delete if Config.days_until_delete.blank?
     end
 
     def autodelete!
--- a/app/models/concerns/autodeletable.rb
+++ b/app/models/concerns/autodeletable.rb
@@ -1,0 +1,21 @@
+# Allows models to be deleted after some period of time
+module AutoDeletable
+    extend ActiveSupport::Concern
+
+    def delete_date
+        if Config.days_until_delete.blank?
+            initial_call_date + Config.days_until_delete
+        else
+            nil
+        end
+    end
+
+    def autodelete!
+        return if delete_date.blank?
+
+        if delete_date < Date.today
+            destroy!
+        end
+    end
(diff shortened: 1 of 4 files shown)


app/models/concerns/autodeletable.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    # This funny business allows us to call both Patient.autodelete! and
+    # ArchivedPatient.autodelete! without any extra code
+    # 
+    # (Any modules that include this Concern will get all methods in the
+    # ClassMethods module, below, as new class methods of their own!)
+    def self.included(base)
+        base.extend(ClassMethods)
+    end
```
elimbaum, line 20: if this is to gnarly i can make a change ;)

test/models/autodeletable_test.rb
```diff
@@ -0,0 +1,112 @@
...
+                end
+
+                # present day - doesn't delete because <, not <=
+                Patient.autodelete!
+                assert_includes Patient.all, @pt_a
+
+                Timecop.freeze(1.day.after) do
+                    Patient.autodelete!
+                    refute_includes Patient.all, @pt_a
+                end
```
elimbaum, line 108: ok this is a lil weird, because i'm using `<` with `today`, so it does a _date_-, not _time_-based comparison. I don't particularly have an issue with this -- I don't think funds will care _exactly_ which day a patient deletes. but let me know and I can make it a bit more intuitive.

elimbaum:
While this draft works, given our discussions yesterday, I think we should take a bit of time to consider exactly what we want this PR to do (and not do)

elimbaum:
@lomky let me know what you think of this — and particularly what other kinds of tests would help! have some time to work on things this week

PR #2946, opened 2023-05-03
Add a11y testing

### LINKED ISSUE #1691: Perform an a11y/WCAG review

We've got a decent userbase and haven't holistically looked at accessibility for our case management corps. Disability and accommodation are a reproductive justice issue so we should take the time to step back and make sure our software matches our values on this.

Making sure anyone who wants to case manage can do so without fighting with our system.

We should use this ticket to go thru the a11y and WCAG lists and see what they recommend doing, what linters we can set up, etc. I suspect there are easy wins to be had.

* **Anything else?**

Hold on implementing or breaking tickets out until the case management team has had a chance to add an accessibility question to their own case management materials.

xmunoz:
Fixes #1691 .


xmunoz:
We're going to take a manual testing approach for now.

colinxfleming:
let's keep this open for one more day - want to try something with this as a base

colinxfleming:
@xmunoz had a sec to take a pass at this and here's the output - the `routing_test` (the more straightforward capybara test we have) now has an `assert_accessible page` in it. (Goofily, since rails 6 or so, integration tests have been more 'post a payload to this endpoint', and system tests have been the 'load up your browser' tests.) Definitely a gotcha, but `assert_accessible` should cooperate now.

sorry for some of the gemfile regressions. I think probably drop and redo the merge commit there.

xmunoz:
One of the a11y violations was fixed by upgrading `bootstrap_form`. Unfortunately, this causes some of the integration tests to start failing. There is only one other violation remaining, and the fix looks pretty straight-forward.

xmunoz:
After more investigation, I'm not sure how to fix the other a11y concern. It specifically concerns missing table header text for the following scenarios.
header

header

We are using those header spaces for layout. The elements in these column are icons which are hidden from aria, meaning they are invisible to screen readers.

Issue #413, opened 2016-06-24
Use attr_encrypted for protecting sensitive data at rest?


colinxfleming:
Leaving this here for now -- https://github.com/attr-encrypted/attr_encrypted

colinxfleming:
Marking this hacktoberfest to increase visibility. 

As you can imagine, there's a lot of sensitive information in this system -- I'm interested in whether or not this is an effective strategy for safely storing information in such a way that some jerk can't just run the mongodb equivalent of `select * from patients`. Unfortunately, the PII we most want to keep under wraps (names and contact info) are also what we search by, which means we can't just plug and play. 

If it's not an effective strategy, I'm pretty interested in what other people's experiences have been.

colinxfleming:
@Kevin-Wei I've been thinking about this a lot lately. Would love to pick your brain on this.

Kevin-Wei:
http://rocketjob.github.io/symmetric-encryption/

Kevin-Wei:
even with heroku configuration http://rocketjob.github.io/symmetric-encryption/heroku.html

colinxfleming:
@Kevin-Wei this looks awesome. 

I think we have two requirements here: 

* encrypt our existing data (if encrypting the entire document is the path of least resistance I am cool as hell with that)
* make sure new data gets encrypted as it's saved and updated etc

If you have any code in that direction can you whip up a WIP PR with it, so we have it handy?

colinxfleming:
Quick dump: https://www.compose.com/articles/encrypting-sensitive-data-in-your-mongodb-database/
https://github.com/rocketjob/symmetric-encryption
https://github.com/attr-encrypted/attr_encrypted/issues/130

colinxfleming:
* `symmetric-encryption` seems straightforward to run but entails some additional setup
* `attr_encrypted` also seems straightforward, 
* last ditch, this gem which the maintainer is looking to shove off (RIP mongodb): https://github.com/KoanHealth/mongoid-encrypted-fields

colinxfleming:
I'm promoting this to active, but it's definitely complicated and tough, so please.

`symmetric_encryption` seems to be slightly better documented, but migrating is going to be difficult. Let's start with a proof of concept of adding an encrypted field on the model and go from there.

colinxfleming:
Tagging this as hacktoberfest -- if you have set this up in MongoDB before I would love to talk to you!

colinxfleming:
[...]

PR #3048, opened 2023-08-18
Issue 413: Encrypt sensitive data in the clinics table

mercedesb:

This PR encrypts sensitive clinic data at-rest.

* Configures deterministic encryption to support cases where that's required (like when we want to have uniqueness validation)
    * You'll want to set an env var in production for `ACTIVE_RECORD_ENCRYPTION_DETERMINISTIC_KEY` prior to running the post-deploy rake task
* Encrypts sensitive clinic columns
* Adds post-deploy rake task to encrypt existing clinic data
    * run `rails clinic:encrypt_sensitive_columns` after you deploy this branch
    * if you run a select query on the clinics table in production, you should see both a bunch of the columns are encrypted

* Bumps #413 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/clinic.rb
+++ b/app/models/clinic.rb
@@ -9,6 +9,15 @@ class Clinic < ApplicationRecord
   # e.g. so a fund can have an 'OTHER CLINIC' catchall.
   EXCLUDED_ZIP = '99999'
 
+  encrypts :name
+  encrypts :street_address
+  encrypts :city
+  encrypts :state
+  encrypts :zip
+  encrypts :phone
+  encrypts :fax
+  encrypts :email_for_pledges
+
   # Scopes
   # Is gestational_limit either nil or above x?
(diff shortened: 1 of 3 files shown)


mercedesb:
`coordinates` is probably a sensitive column but it's a decimal array field so out-of-the-box ActiveRecord encryption doesn't totally work. Let me know if you'd like me to spend some time on a small custom solution so we can encrypt non-string fields (I'd do that work in a separate PR)

colinxfleming (approved):
Does what it says on the tin! Nice work again. I've also added the new env var to our staging and prod environments. 

I think encrypting coordinates is good if we can swing it. I'm willing to change the datatype for that column if it's helpful though that's obviously a bit more involved. If you want to tackle it, whip up an issue for it I think? 

thanks much @mercedesb !

Issue #3017, opened 2023-07-11
Practical support: When mode on, don't display budget bar or fulfillment page


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

Once the config is set up, begin implementing practical support mode with the following two items:
* Don't display the budget bar on the main dashboard
* Don't display the fulfillment tab, even if a pledge is submitted 

* **What feature or behavior is this required for?**
Implementing practical support only mode

* **How could we solve this issue? (Not knowing is okay!)**
Add some conditionals to the views.

* **Anything else?**
depends on https://github.com/DARIAEngineering/dcaf_case_management/issues/2876

Issue #2174, opened 2021-04-21
Adjustments for practical support only funds?


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

We got our first completely practical support based fund. Since they aren't doing pledges, the pledge of the app is not super helpful.

Generally CMs consider a patient good to go when they fire off the pledge, and the fund figures a patient is good to go after the fulfillment process. But neither of these will happen with our practical support model. How to accommodate that? 

* **What feature or behavior is this required for?**

not confusing CMs who work at practical support funds

* **How could we solve this issue? (Not knowing is okay!)**

not clear. this is a ticket for discussion.

* **Anything else?**

Stems from an email conversation with NMRCRC

Issue #2142, opened 2021-03-05
How to mark patients complete for a practical support fund?


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

NMRCRC is pushing us forward in this respect. They need a way to mark someone as all clear without pledges. How do we do that? I have no idea, but we should figure something out for them.

* **What feature or behavior is this required for?**

NMRCRC support

* **How could we solve this issue? (Not knowing is okay!)**

no idea, but highly suggestible!

* **Anything else?**

n

colinxfleming:
a quick idea: practical support mode, which hides some fields and instead shows a 'all set' checkbox or something to mark stuff donezo?

emtot22:
@colinxfleming Let me know what you think! I'm open to wording/language suggestions 

https://www.figma.com/file/Zry1ECvlVlSylFFKc6AxVv/DCAF-2?node-id=2%3A141

colinxfleming:
Coming back to this, I think this design does crystalize for me what we should do under a practical support mode:

* new Config to turn on 'Practical Support Mode'
* Submit Pledge button -> Mark Complete button (with a new modal flow to match) when that config is on. (unclear what the requirements should be to mark someone complete)
* new patient field along the lines of `marked_complete` for this flow (and include in status, exports, etc)
* Don't display Pledge Fulfillment if the config is on under any circumstances
* DO display in the accountant flow if marked complete

Open question (repeated from inline):
* what should be required to mark someone complete? (I think we can get started without this but good to know)

colinxfleming:
(Probably also hide the abortion information and budget bar sections)

PR #3050, opened 2023-08-25
Add Practical Support Mode!

### LINKED ISSUE #2876: Add a Practical Support Mode config

Some funds do not deal in pledges and instead just deal in practical support. The pledge stuff doesn't make sense for them, along with some other utilities that are abortion cost specific. To meet their needs we're going to build out a flow. To start this, let's add a Practical Support Mode config.

paring the app down to just useful stuff for practical support funds

there's a good approach here for this in https://github.com/DARIAEngineering/dcaf_case_management/pull/2838 that you can use as a guide!

* **Anything else?**
stems from https://github.com/DARIAEngineering/dcaf_case_management/issues/2174

elimbaum:

* Adds a config for practical support mode (#2876) and `marked_complete` field in the Patient model.
* Updates views when practical support mode is on; specifically, hides budget bar and fulfillment options, and changes language in table (#3017)
* Changes the `Submit Pledge` button to `Mark Complete`, and updates the confirmation modals:

Screenshot 2023-08-25 at 13 32 46

Screenshot 2023-08-25 at 13 42 06

Currently, there is no valid for marking a patient complete, as there is for fulifilling a pledge; this could be changed.

DRAFT until tests are added.

* Fixes #2876,  #3017, #2174, #2142

For reviewer:
* Tag this PR:

### DIFF
--- a/app/helpers/pledges_helper.rb
+++ b/app/helpers/pledges_helper.rb
@@ -16,6 +16,22 @@ module PledgesHelper
     end
   end
 
+  def mark_complete_button
+    content_tag :span, class: 'btn btn-primary btn-lg submit-btn btn-block',
+                       aria: { hidden: true },
+                       id: 'submit-pledge-button' do
+      t('patient.menu.mark_complete')
+    end 
+  end
+
+  def mark_incomplete_button
+    content_tag :span, class: 'btn btn-warning btn-lg cancel-btn btn-block',
+                       aria: { hidden: true },
+                       id: 'submit-pledge-button' do
+      t('patient.menu.mark_incomplete')
+    end 
+  end
+
   def clinic_pledge_email(patient)
     email = patient.clinic&.email_for_pledges
     return unless email
--- a/app/models/concerns/statusable.rb
+++ b/app/models/concerns/statusable.rb
@@ -20,10 +20,13 @@ module Statusable
     dropoff: { key: I18n.t('patient.status.key.dropoff'),
                help_text: I18n.t('patient.status.help.dropoff')},
     resolved: { key: I18n.t('patient.status.key.resolved', fund: ActsAsTenant.current_tenant&.name),
-                help_text: I18n.t('patient.status.help.resolved')}
+                help_text: I18n.t('patient.status.help.resolved')},
+    completed: { key: I18n.t('patient.status.key.completed'),
+                 help_text: I18n.t('patient.status.help.completed') },
   }.freeze
 
   def status
+    return STATUSES[:completed][:key] if practical_support_completed?
     return STATUSES[:fulfilled][:key] if fulfillment.fulfilled?
     return STATUSES[:resolved][:key] if resolved_without_fund?
     return STATUSES[:pledge_unfulfilled][:key] if days_since_pledge_sent > 150
@@ -35,6 +38,10 @@ module Statusable
   end
 
   private
+
+  def practical_support_completed?
+    Config.practical_support_mode? && marked_complete?
+  end
 
   def contact_made?
     calls.each do |call|
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -22,7 +22,8 @@ class Config < ApplicationRecord
     hide_budget_bar: 'Enter "yes" to hide the budget bar display.',
     aggregate_statistics: 'Enter "yes" to show aggregate statistics on the budget bar.',
     hide_standard_dropdown_values: 'Enter "yes" to hide standard dropdown values. Only custom options (specified on this page) will be used.',
-    time_zone: "Time zone to use for displaying dates. Default is Eastern. Valid options are Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, Indiana (East), or Puerto Rico."
+    time_zone: "Time zone to use for displaying dates. Default is Eastern. Valid options are Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, Indiana (East), or Puerto Rico.",
+    practical_support_mode: 'Enter "yes" to enable practical-support-only mode. Pledges will be disabled.'
   }.freeze
 
   enum config_key: {
@@ -46,10 +47,9 @@ class Config < ApplicationRecord
     aggregate_statistics: 17,
     hide_standard_dropdown_values: 18,
     county: 19,
-    time_zone: 20
+    time_zone: 20,
+    practical_support_mode: 21,
   }
-
-  # which fields are URLs (run special validation only on those)
 
   # symbols are required here because functions are not objects in rails :)
   CLEAN_PRE_VALIDATION = {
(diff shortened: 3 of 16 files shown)


app/models/config.rb
```diff
@@ -308,7 +315,7 @@ def validate_time_zone
 
     def validate_yes_or_no
       # allow yes or no, to be nice (technically only yes is considered)
-      options.last =~ /\A(yes|no)\z/i
+      validate_singleton and options.last =~ /\A(yes|no)\z/i
```
colinxfleming, line 318: great rails

app/models/config.rb
```diff
@@ -22,7 +22,8 @@ class Config < ApplicationRecord
     hide_budget_bar: 'Enter "yes" to hide the budget bar display.',
     aggregate_statistics: 'Enter "yes" to show aggregate statistics on the budget bar.',
     hide_standard_dropdown_values: 'Enter "yes" to hide standard dropdown values. Only custom options (specified on this page) will be used.',
-    time_zone: "Time zone to use for displaying dates. Default is Eastern. Valid options are Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, Indiana (East), or Puerto Rico."
+    time_zone: "Time zone to use for displaying dates. Default is Eastern. Valid options are Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, Indiana (East), or Puerto Rico.",
+    practical_support_mode: 'Enter "yes" to enable practical-support-only mode. Pledges will be disabled.'
```
colinxfleming, line 26:
```suggestion
    practical_support_mode: 'Enter "yes" to enable practical-support-only mode. Pledges will be disabled, in in favor of marking people complete/incomplete.'
```

app/views/dashboards/index.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-sm-12">
-    <% unless Config.hide_budget_bar? %>
+    <% unless Config.hide_budget_bar? or Config.practical_support_mode? %>
```
colinxfleming, line 3: 👨‍🍳

colinxfleming (commented):
this is fabulous stuff, thank you @elimbaum ! Very clean effective rails and you've hit the vast majority of the spots where we'd need to do this. I think we'll also want to add this to exports and the accounting panel but both of those can be done in separate PRs if you think this one is too busy. I'll pull down and acid test more extensively once tests are in place if that's cool with you, let me know whether there's anything in particular you'd like my opinion on.

elimbaum:
Thanks for review @colinxfleming! am moving this week (and starting school...) but once I get a dev environment spun back up I'll close this out!

app/helpers/pledges_helper.rb
```diff
@@ -16,6 +16,22 @@ def cancel_pledge_button
...
+  end
+
+  def mark_incomplete_button
+    content_tag :span, class: 'btn btn-warning btn-lg cancel-btn btn-block',
+                       aria: { hidden: true },
+                       id: 'submit-pledge-button' do
+      t('patient.menu.mark_incomplete')
+    end 
+  end
+
```
elimbaum, line 34: These two functions look _extremely_ similar to the submit pledge button pair (above). Thoughts on condensing, or fine to keep separate?

app/models/concerns/statusable.rb
```diff
@@ -20,10 +20,13 @@ module Statusable
     dropoff: { key: I18n.t('patient.status.key.dropoff'),
                help_text: I18n.t('patient.status.help.dropoff')},
     resolved: { key: I18n.t('patient.status.key.resolved', fund: ActsAsTenant.current_tenant&.name),
-                help_text: I18n.t('patient.status.help.resolved')}
+                help_text: I18n.t('patient.status.help.resolved')},
+    completed: { key: I18n.t('patient.status.key.completed'),
+                 help_text: I18n.t('patient.status.help.completed') },
```
elimbaum, line 25: What do we use `help_text` for? I wasn't able to find a reference

app/models/config.rb
```diff
@@ -335,4 +342,10 @@ def validate_length
       end
       true
     end
+
+    def validate_only_one_practical_support
+      # TODO - Should not be able to use both practical support configs at the
+      # same time.
+      true
+    end
```
elimbaum, line 350: After this is done, I'm thinking about opening a separate issue to combine "practical support mode" and "hide practical supports" into a single option — practical supports only, on, off. in the interim, should we have some validation? Weird things will happen if we turn on both options.

How would that best work? We can't compare the actual saved values because on validation, the submitted value has not yet been committed. We could have a conditional check based on the option key?

app/views/patients/_complete.html.erb
elimbaum, line 1: what do you think about this modal? i don't know if we need any confirmation steps, so maybe we don't even _need_ the modal - just the button?

app/views/patients/_menu_pledge_button.html.erb
```diff
@@ -1,13 +1,29 @@
...
+                aria: { label: t('patient.menu.mark_complete') },
+                remote: true %>
+  <% else %>
+    <%= link_to mark_incomplete_button,
+                mark_complete_path(patient),
+                class: 'submit-pledge-button',
+                aria: { label: t('patient.menu.mark_incomplete') },
+                remote: true %>
+  <% end %>
 <% else %>
```
elimbaum, line 15: same question here as with the button controller function - very similar code between in/complete and submit/cancel.

app/views/patients/complete.js.erb
elimbaum, line 1: ergh i hate this but is it ok?

app/helpers/pledges_helper.rb
colinxfleming: I'd say keep separate - reasonable for them to diverge a bit here

app/models/concerns/statusable.rb
colinxfleming: These show up in the Status tooltip on patient records!

image

app/models/config.rb
colinxfleming: Let me say a bit more about the thinking here, and let's see if that makes sense with how you're thinking about it. I think there are three possible states:

* fund is not doing practical support at all (in which case we'd want to hide these options) - unusual but it happens
* fund is doing pledges AND practical support - most of our clientele
* fund is doing zero pledges, ONLY practical support - unusual but it happens

So I think you're def right that 'hide practical support on' and 'practical support only on' doesn't make sense, but I think the other cases (hide off, practical only on; hide on, practical only off;  hide off, practical only off) are cases I think we'd like to support. Does that match your understanding?

app/views/patients/_complete.html.erb
colinxfleming: I think yes to the modal -- an extra UX step to signify 'this is a significant action' is helpful (and helps prevent problems from misclicks)

app/views/patients/_menu_pledge_button.html.erb
colinxfleming: I think reasonable to collapse this one if you like. Up to you!

app/views/patients/complete.js.erb
colinxfleming: this reads as ok to me, what's the issue?

Issue #3078, opened 2023-10-02
upgrade `acts_as_tenant`


xmunoz:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
We can not currently upgrade acts_as_tenant even to the next patch version, because it appears that the maintainer does not follow semver guidelines and has introduced breaking changes in 0.5.2. We are currently on 0.5.1.

* **What feature or behavior is this required for?**
Keeping things up to date.

* **How could we solve this issue? (Not knowing is okay!)**
Probably need to refactor our use of `acts_as_tenant`. Changelog: https://github.com/ErwinM/acts_as_tenant/blob/master/CHANGELOG.md#052

* **Anything else?**
Nope!

PR #3086, opened 2023-10-17
Move to dependabot

lomky:

We tried and failed to use renovate's bundled version bumps. It would still be very nice to have the grouped updates, and dependabot has long ago caught upto supporting our version of yarn, so swapping back gets us the grouped updates.

* Removes Renovate
* Adds dependabot
* Avoids the one-know-bad case we'll handle in #3078 

For reviewer:
* Tag this PR:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,40 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/githubDARIAEngineering/dcaf_case_management//administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      minor-patch:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+      - "xmunoz"
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "yarn"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      minor-patch:
```
xmunoz, line 29: I think if you exclude this group it will automatically match all.

.github/dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+        patterns:
+          - "*"
```
xmunoz, line 31:
```suggestion
```

xmunoz (approved):
Just one small optional suggestion

.github/dependabot.yml
```diff
@@ -0,0 +1,38 @@
...
+          - "@acts_as_tenant*" # DARIAEngineering/dcaf_case_management#3078
+        update-types:
+          - "minor"
+          - "patch"
+    reviewers:
+      - "xmunoz"
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "yarn"
```
lomky, line 24:
```suggestion
  - package-ecosystem: "npm"
```

PR #3102, opened 2023-11-04
update the security doc

colinxfleming:
(stems from a conversation I'm having, plus we should spruce this doc up anyway)


colinxfleming:
godmode since it's just a doc change

PR #3103, opened 2023-11-05
Upgrade acts_as_tenant

### LINKED ISSUE #3078: upgrade `acts_as_tenant`

colinxfleming:

Not my greatest idea, but this end-runs around some changes to how AAT handles state in integration-y tests by hardsetting it in controller. I tried a lot of other stuff that didn't work (and learned a lot about integration tests spawning threads in the process, unfortunately) and I think this is probably the way to go.

The nut of this is that when you do something like this very common integration test or system test thing:

```
# log_in @user
visit root_path
fill_in ...
click_button ...
```

it's doing the visit in its own process or fork or thread or whatever which doesn't have access to `Thread.current ` in the test, or `ActsAsTenant` in the test, and so it thinks no tenant is set and it eats mud. It DOES have access to db, though, so the end run here is saying 'I know this object exists, just go get it and set it in db before making the request'.

I tested this by confirming that I could log in to both in different browsers (incognito mode), and it still looks like it's scoping things properly, so I think we're set on that front. 

image

* upgrades `acts_as_tenant` gem
* fights with test suite to get things passing again

no view changes
* Fixes #3078 

For reviewer:
* Tag this PR:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'omniauth-google-oauth2', '~> 1.1.1'
 gem "omniauth-rails_csrf_protection", '~> 1.0'
 
 # Run multiple funds on one server
-gem 'acts_as_tenant', '~> 0.5.0'
+gem 'acts_as_tenant', '~> 0.5'
 
 # Strong Password for user password validation for folks not on oauth
 gem 'strong_password', '~> 0.0.10'
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem 'omniauth-google-oauth2', '~> 1.1.1'
 gem "omniauth-rails_csrf_protection", '~> 1.0'
 
 # Run multiple funds on one server
-gem 'acts_as_tenant', '~> 0.5'
+gem 'acts_as_tenant', '~> 0.6'
 
 # Strong Password for user password validation for folks not on oauth
 gem 'strong_password', '~> 0.0.10'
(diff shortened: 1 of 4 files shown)


colinxfleming:
Leaving a couple notes from hell here:
* this is ultimately a thread/process/parallelization problem safety problem; something about the upgrade here is causing `ActsAsTenant.current_tenant` to nil out when the systemtest process gets actually spawned
* bafflingly, `ActsAsTenant.current_tenant` is there for the setup, but when the endpoints get called by the browser it chokes because no ActsAsTenant in that process/thread/parallelization
* so if we can get capybara to spawn with ActsAsTenant preserved we should be okay

colinxfleming:
increasingly I think the path to victory here involves setting `config.middleware.use ActsAsTenant::TestTenantMiddleware` and getting that to work, which unfortunately requires some pretty stupid surgery in our test helper. see also https://github.com/ErwinM/acts_as_tenant/pull/279 for ref

xmunoz:
Going to merge this now so it goes out with today's deploy

PR #3112, opened 2023-11-26
In dev mode, recompile scss on change

colinxfleming:

One annoying thing about our current frontend posture, that made development extremely difficult, was that we didn't have `watch` working, so to see your css changes you'd need to stop and restart `bin/dev`, a mondo pain in the butt. This gets `watch` working again, to save a lot of time and headache in dev.

To try it out, crack open app/assets/stylesheets/styles/globals.scss and change the `background-color` on L18 to `red`, then save it and reload to see it change automagically. I also tested autosaving and a few other odds and ends.

* moves styles out of `app/javascript` and into `app/assets`, much to my chagrin

no view changes (thankfully)

* sourced heavily from #2795 though this is more narrow in scope - it doesn't improve how we use variables or organize the vendored stuff better, but we'll live

For reviewer:
* Tag this PR:


lomky (approved):
oh _thats_ what we were missing?! 🥳

PR #3163, opened 2024-03-07
 added support_date_key under patient in ES and EN.yml

### LINKED ISSUE #2706: Add i18n key for pledge_generated_by_id

The `pledge_generated_by` field shows up in the changelog like so:
```
<span class="translation_missing" title="translation missing: en.activerecord.attributes.patient.pledge_generated_by_id">Pledge Generated By</span>: (empty)
```
ew.

Clean changelog
I think we handled this for Clinic? Do the same thing for the key. This will entail opening up `en.yml` and `es.yml` and adding keys. (Look for `clinic_id`.)

* **Anything else?**
nah

mbichoffe:
I have completed some work on Case Manager that's ready for review!

* This pull request adds the 'pledge_generated_by' key and string to the en.yml and es.yml files.

@colinxfleming I could not find the issue to screenshot with the fix. Could you verify on your end that the error is not showing up?

* Fixes #2706 

For reviewer:
* Tag this PR:


lomky:
alas, wrong colin! @colinxfleming is the one who wrote the ticket

mbichoffe:
> alas, wrong colin! @colinxfleming is the one who wrote the ticket

Ooops! Thanks for catching that. Updated the PR message tagging the right Colin

colinxfleming (changes requested):
you found the right spots here! Marking as requesting changes to fix up a rails-ism and then we're good to go. @ me or @lomky on slack if you have any questions.

Thank you @mbichoffe and welcome, glad to have you!

mbichoffe:
> you found the right spots here! Marking as requesting changes to fix up a rails-ism and then we're good to go. @ me or @lomky on slack if you have any questions.
> 
> Thank you @mbichoffe and welcome, glad to have you!

@colinxfleming thank you for the feedback and the warm welcome! Could you please provide more specific details about the rails-ism issue you've mentioned? I don't see any specific changes requested, but I'm certainly missing something and I want to make sure I fully understand the changes required to move forward with this PR. Thanks again!

config/locales/en.yml
```diff
@@ -80,6 +80,7 @@ en:
         other_phone: Other phone
         patient_contribution: Patient contribution
         pledge_generated_at: Pledge generated at
+        pledge_generated_by: Pledge generated by
```
colinxfleming, line 83: Sorry, looks like github ate my comment here! Basically the rails-ism is that in database this is `pledge_generated_by_id` -- `pledge_generated_by` is a rails shorthand that signals that you want the User object, which we almost always do, but not here because we're dealing with changes to the id value. See below -- note that the `translation missing` error refers 

So basically, turn this and the spanish one into `pledge_generated_by_id` and you're all set, I think.

image

colinxfleming (commented):
sorry, looks like github ate the comment I wrote up! trying again ha

config/locales/en.yml
mbichoffe: done - lmk if there's any other changes you'd like to see!

colinxfleming (approved):
perfect, thank you @mbichoffe !

image

PR #3182, opened 2024-04-02
Upgrade docker compose to v2

xmunoz:
v1 is now deprecated on the default github actions runners https://github.com/actions/runner-images/issues/9557

All subsequent PRs are likely to have failing CI until this is merged to main.

Also, update docs.

https://docs.docker.com/compose/migrate/


colinxfleming (approved):
oh yeah this is the good stuff

Issue #3171, opened 2024-03-20
New fields for practical support?


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

This is a discussion ticket for some practical support work requested by a fund. Their model of practical support has adjusted quite a bit and I'd like us to consider adding some fields on the Practical Support object to support that work.  (Practical support is our representation of logistics concerns like bus tickets, airfare, a hotel stay, driving someone to and from their appointment, etc. Some funds use a different name for it that I can't remember off the top of my head.)

Specifically, these fields:
- [x] A detail/notes field -- scrambleable/nonexportable -- up to this point we've talked about just using notes, but I think they've made a compelling case for why this is important to be associated with particular elements of care. They're doing a lot of logistical information and coordinating and looking at it JUST against a particular record is helpful. At the very least, could add a `practical_support_notes` field to the patient model - something severed from the main notes field is helpful.
- [ ] date of purchase -- exportable -- I'm lukewarm on this being crucial (might be better served as a note, or with the attachment field) but I expect this would make the existing date field `date of service` and this would be date of purchase. (My inclination is to pass and tell people to use notes for this, or be insinuated from a receipt upload or something.)
- [x] attachment_url -- idk if exportable or not but probably -- a text field to store receipts uploaded to a fund's google drive or something. I explicitly passed on managing uploads for them which felt out of scope, but we handle google drive links elsewhere in configs and it's fine. I FEEL like this is safe to keep on archive but interested in @lomky 's thoughts.
- [x] fulfilled/paid out -- exportable -- a boolean field, similar to confirmed, but that indicates whether or not they've done whatever, for fund logistics reasons. 

I think this entails a bit of frontend work to get all the information in, and would make it the most complicated subobject we have. I think I'm pretty okay with that as this represents (often) the most complicated part of abortion care, so fine to have that granular level of detail here. 

* **What feature or behavior is this required for?**
Better practical support utilities

* **How could we solve this issue? (Not knowing is okay!)**
Let's get some thoughts on those issues and their safety, figure out scope, and if so we'll do a quick frontend design and then ticket for hacknite.

* **Anything else?**

lomky:
[...]

PR #3206, opened 2024-05-27
Practical support improvements: Add attachment_url field + fulfilled checkbox, config to display attachment_url

colinxfleming:

* adds practical support fields `attachment_url` and `fulfillment`
* adds a config to display `attachment_url` (defaulting to not showing it)

A little cramped, but looks chill:
image
image

* Bumps #3171 

For reviewer:
* Tag this PR:


app/models/config.rb
```diff
@@ -24,6 +24,7 @@ class Config < ApplicationRecord
     time_zone: "Time zone to use for displaying dates. Default is Eastern. Valid options are Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, Indiana (East), or Puerto Rico.",
     procedure_type: "Any kind of distinction in procedure your fund would like to track. Field hides if no options " \
                     "are added here. Please separate with commas.",
+    display_practical_support_attachment_url: 'CAUTION: Whether or not to allow people to enter attachment URLs for practical support entries; for example, a link to a file in Google Drive. Please ensure that any system storing these is properly secured by your fund!',
```
lomky, line 27: i def still feel wary about this, but I think this is the right balance of empowering vs security

lomky (approved):
Can we update teh [data dictionary](https://github.com/DARIAEngineering/dcaf_case_management/blob/main/docs/DATA_DICTIONARY.md) doc too? I think we've drifted a bit

app/models/config.rb
colinxfleming: I also feel ways about this but yeah I think this is an okay balance

colinxfleming:
bumped the data dict -- merging

PR #3241, opened 2024-07-29
Revamp of PracticalSupport entry editing

colinxfleming:

There's a need for significantly more data with practical support entries; my plan here is to treat them like patients in the call log and have a summary, option to see a note, and option to click in for more detail. To start with that we're going to need to revamp the approach we're taking here to support a lot more data tracking; our two row thing isn't gonna cut it anymore.

Next up  we're adding practical support notes (text field, not for archive), date of purchase (archivable datestamp), and waiver present (boolean, archivable). 

Here we basically change up the display of practical support entries into a string, then move the editing to a modal. 

* Changes practical support display to match call list style patterns

image
image

no issues, but sets up bumps to #3214 and #3171

For reviewer:
* Tag this PR:


app/helpers/practical_supports_helper.rb
```diff
@@ -47,4 +47,13 @@ def practical_support_guidance_link
...
   end
+
+  def practical_support_display_text(practical_support)
+    content = [practical_support.support_type]
+    content.push "#{t('common.from')} #{practical_support.source}"
+    content.push "#{t('common.for')} #{number_to_currency(practical_support.amount)}" if practical_support.amount.present?
+    content.push '(confirmed)' if practical_support.confirmed?
+    content.push "#{t('common.on_')} #{practical_support.support_date.display_date}" if practical_support.support_date.present?
+    content.join(' ')
+  end
```
lomky, line 58: This is pretty hardcoded to English grammar, but its probably good enough for now?

The more robust alternative I can think of would be to encode distinct strings for each variation that take variables, allowing each locale to arrange them as it sees fit. See the `guidance_link` entry for a simpler example.

app/views/practical_supports/_edit.html.erb
```diff
@@ -1,65 +1,47 @@
...
-                           step: '0.01' %>
-      </div>
-
-      <div class="col-3">
-        <%= f.date_field :support_date, autocomplete: 'off', hide_label: true %>
-      </div>
-
+<div class="row">
+  <div class="col">
+    <%= bootstrap_form_with model: support,
```
lomky, line 3: any particular reason we renamed the model here? Just shorter and easier?

app/views/practical_supports/_entries.html.erb
```diff
@@ -1,19 +1,23 @@
...
-  </div>
-  <hr/>
 
-  <% patient.practical_supports.each do |entry| %>
-    <%= render 'practical_supports/edit',
-              patient: patient,
-              practical_support: entry %>
+      <div class="col-1 practical-support-note-preview">
+        <%#= entry.most_recent_note_display_text %>
+      </div>
```
lomky, line 11: Is this setup for a future ticket?

app/helpers/practical_supports_helper.rb
lomky: Do we want to display the fulfilled or nah? If we're worried about space, I think (C) and (F) are sufficient for this quick-view.
image

app/views/practical_supports/_edit.html.erb
```diff
@@ -1,65 +1,47 @@
...
+                        label_class: 'tooltip-header-checkbox',
+                        data: { 'tooltip_text': practical_support_fulfilled_help_text } %>
+      <% end %>
+    <% end %>
   </div>
-  <hr/>
 </div>
+
+<div class="row">
+  <div class="col">
```
lomky, line 39: I realize that we don't actually _need_ a `Close` button here, but I would really like one.Right now my muscle memory instinct is to click the button when I'm done, so better it be another way to exit the modal.

app/helpers/practical_supports_helper.rb
lomky: On last thing - this ordering implies that the Confirmed is on the amount, rather than the whole Support. What do you think of moving it to the end?

lomky (commented):
This is so cool. A couple questions and a plea for a modal Close button

app/helpers/practical_supports_helper.rb
colinxfleming: I think I'm in favor of the guidance_link strategy if this winds up needing to expand or gets awkward. I'd rather solidify it before we put the time into anything particularly robust, and the spanish is close enough I think with the `de`s and `por`s mostly being clean subs for the froms and fors. 

I think reordering a little bit is a good idea; I'm less worried about space though, I think this frees us up to a lot more by delegating updating out! As a general rule of thumb I don't think we need _every_ unit of data displayed and can iterate a little on that, but I'd believe that this should be rearranged at least a little bit.

app/views/practical_supports/_edit.html.erb
colinxfleming: pretty much yeah. Less to type!

app/views/practical_supports/_edit.html.erb
colinxfleming: I think super reasonable

app/views/practical_supports/_entries.html.erb
colinxfleming: yeah, next up is pretty much 'put notes in here'.

colinxfleming (commented):
So three things to do:
* add a close button
* display fulfilled
* slightly reorder the display text
thx @lomky !

app/helpers/practical_supports_helper.rb
```diff
@@ -47,4 +47,15 @@ def practical_support_guidance_link
     link_to t('patient.practical_support.guidance_link', fund: ActsAsTenant.current_tenant.name), url.to_s, target: '_blank'
     # "For guidance on practical support, view the #{link_content}."
   end
+
+  def practical_support_display_text(practical_support)
+    content = []
+    content.push "(#{t('activerecord.attributes.practical_support.confirmed')})" if practical_support.confirmed?
+    content.push "(#{t('activerecord.attributes.practical_support.fulfilled')})" if practical_support.confirmed?
```
lomky, line 54:
```suggestion
    content.push "(#{t('activerecord.attributes.practical_support.fulfilled')})" if practical_support.fulfilled?
```

app/views/practical_supports/_edit.html.erb
```diff
@@ -1,65 +1,51 @@
...
+        local: false do |f| %>
+      <%= f.select :support_type, options_for_select(practical_support_options, support.support_type) %>
+      <%= f.select :source,
+                    options_for_select(practical_support_source_options(support.source),
+                                      support.source) %>
+      <%= f.number_field :amount,
+                          value: number_to_currency(f.object.amount.to_f, unit: '', delimiter: ''),
+                          autocomplete: 'off',
+                          prepend: '$',
+                          step: '0.01' %>
```
lomky, line 15: A strange behavior, the edit page converts a blank value to `0.00`, and then if you save an unrelated change, it saves that as explicit `0.00`. That's going to be nonsensical for some support types, but I don't immediately know how to avoid this.

lomky (commented):
Two bugs! But I think nearly there!

app/helpers/practical_supports_helper.rb
colinxfleming: dorf

lomky (approved):
gooooo!

PR #3285, opened 2024-10-04
Fix staging deploy github action

xmunoz:

Staging is no longer being autodeployed from `main`. This is the error message:
```
/bin/sh: 1: heroku: not found
Created and wrote to ~/.netrc
Successfully logged into heroku
/bin/sh: 1: heroku: not found
```
This is because `heroku` no longer comes packaged with the default image of the [`ubuntu-latest`](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) Github-hosted runner (24.04). It was pre-installed on [22.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) however, hence the breakage. They claim to have removed it for ["maintenance reasons"](https://github.com/actions/runner-images/issues/9848) 🙄 

This PR explicitly installs heroku cli in the staging deploy workflow, and updates the [heroku-deploy Github Action](https://github.com/AkhileshNS/heroku-deploy/tree/v3.13.15) to the latest version.

https://devcenter.heroku.com/articles/heroku-cli

### DIFF
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Deploy sandbox environment
-        uses: akhileshns/heroku-deploy@v3.12.14
+        uses: akhileshns/heroku-deploy@v3.13.15
         with:
           heroku_api_key: ${{secrets.XMUNOZ_HEROKU_API_KEY}}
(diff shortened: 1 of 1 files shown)


xmunoz:
Successful workflow run: https://github.com/DARIAEngineering/dcaf_case_management/actions/runs/11187413238/job/31104408138

lomky (approved):
🎉

PR #3308, opened 2024-11-21
disable form field

### LINKED ISSUE #3233: Minor js bug - Status on patient dashboard now editable

This should be a display field, and looks like one, but appears editable now - 

image

To recreate, open a patient record and edit the status in the patient dashboard. (Note that it doesn't actually save db state!)

keeping uneditable data from appearing editable 

change the input element 

* **Anything else?**

rkaufman13:

Fixes #3233. Literally just adds the "disabled" attribute to a formfield that we don't want to be editable. I had some Docker issues (still not fixed, but I set up the app to run direct on my machine) so this is all I got to do tonight :)

For reviewer:
* Tag this PR:


colinxfleming (approved):
love it. thank you!

Issue #3240, opened 2024-07-29
Js bug: changes in patient dashboard getting eaten 


colinxfleming:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**

* **What feature or behavior is this required for?**

* **How could we solve this issue? (Not knowing is okay!)**

* **Anything else?**

lomky:
No longer true in production after #3237, but we should debug and rerelease!

rkaufman13:
Investigating this but don't let me having 'claimed' this stop anyone else from looking into it.

rkaufman13:
Couldn't replicate locally on the referenced commit. @lomky suggested spinning up a separate Heroku instance to try to debug in the wild.

F00bis:
Suspect the issue has something with JS implicit semicolon insertion in the autosave function. We need to make a test branch that reverts #3237 and inserts semicolons to run on a live heroku instance to see if this hunch pays off

PR #3216, opened 2024-06-06
React autosave: patient dashboard form

mercedesb:
I rule and have completed some work on Case Manager that's ready for review!

This PR rewrites the patient dashboard form in React with safe autosave.

The commits might be the easiest way to review the changes.

- Allow patients#update to respond to json 
- Add a few base reusable component types for creating forms (Input, Select, and Tooltip)
- Add a `useFetch` hook for easy async data fetching
- Add a `useFlash` hook for rendering flash messages
- Rebuild the form in React

### Before rewriting in React (notice console error)

https://github.com/mercedesb/dcaf_case_management/assets/9774690/ddafdf75-36c3-486f-b524-5bd996f6f4b9

### After rewriting in React (no console error)

https://github.com/mercedesb/dcaf_case_management/assets/9774690/5efd762b-d968-4c7a-8a91-76af37b27425

Note: there's a small little jump that happens sometimes but I haven't figured out where it's coming from yet

For reviewer:
* Adjust the title to explain what it does for the notification email to the listserv.
* Tag this PR:
  * `feature` if it contains a feature, fix, or similar. This is anything that contains a user-facing fix in some way, such as frontend changes, alterations to backend behavior, or bug fixes.
  * `dependencies` if it contains library upgrades or similar. This is anything that upgrades any dependency, such as a Gemfile update or npm package upgrade.
* If it contains neither, no need to tag this PR.

PR #3337, opened 2025-01-10
React autosave: patient dashboard form (bugs fixed 🙂 )

mercedesb:

This PR rewrites the patient dashboard form in React with safe autosave. This is mostly the same as #3216 but with the identified bugs fixed.

Let me know if anything else pops up. I'm sorry for the long delay on this 🫣 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/javascript/src/components/PatientDashboardForm.jsx
+++ b/app/javascript/src/components/PatientDashboardForm.jsx
@@ -130,6 +130,7 @@ export default PatientDashboardForm = ({
           label={i18n.t('patient.shared.status')}
           value={patientData.status}
           className="form-control-plaintext"
+          disabled="true"
           tooltip={statusTooltip}
           onChange={e => debouncedAutosave({ status: e.target.value })}
         />
--- a/app/javascript/src/components/PatientDashboardForm.jsx
+++ b/app/javascript/src/components/PatientDashboardForm.jsx
@@ -26,7 +26,6 @@ export default PatientDashboardForm = ({
 
   const autosave = async (updatedData) => {
     const updatedPatientData = { ...patientData, ...updatedData }
-    setPatientData(updatedPatientData)
 
     const putData = {
       name: updatedPatientData.name,
@@ -44,9 +43,9 @@ export default PatientDashboardForm = ({
     }
   }
 
-  const debouncedAutosave = useMemo((params) => {
-    return debounce(autosave, 300)
-  }, []);
+  const debouncedAutosave = (params) => {
+    return debounce(autosave(params), 300)
+  };
 
   // Stop the invocation of the debounced function after unmounting
   useEffect(() => {
--- a/app/views/patients/_patient_dashboard.html.erb
+++ b/app/views/patients/_patient_dashboard.html.erb
@@ -1,77 +1,12 @@
 <div id="patient_dashboard_content">
-  <%= bootstrap_form_with model: patient,
-                         html: { id: 'patient_dashboard_form' },
-                         local: false,
-                         method: 'patch',
-                         class: 'edit_patient' do |f| %>
-    <div class="row">
-
-      <div class="col-4">
-        <%= f.text_field :name,
-                         label: t('patient.shared.name'),
-                         autocomplete: 'off' %>
-      </div>
-      
-      <div class="col">
-        <%= f.select :last_menstrual_period_weeks,
-                     options_for_select(weeks_options, patient.last_menstrual_period_weeks ),
-                     label: t('patient.dashboard.weeks_along'),
-                     autocomplete: 'off',
-                     help: t('patient.dashboard.currently', weeks: patient.last_menstrual_period_now_weeks, days: patient.last_menstrual_period_now_days) %>
-      </div>
-
-      <div class="col mt-4">
-        <%= f.select :last_menstrual_period_days,
-                     options_for_select(days_options, patient.last_menstrual_period_days),
-                     autocomplete: 'off',
-                     skip_label: true,
-                     help: t('patient.dashboard.called_on', date: patient.initial_call_date.strftime("%m/%d/%Y")) %>
-        <%= f.label :last_menstrual_period_days, t('common.days_along'), class: "sr-only" %>
-      </div>          
-
-      <div class="col-3">
-        <%= f.date_field :appointment_date,
-                         label: t('patient.shared.appt_date'),
-                         autocomplete: 'off',
-                         help: t('patient.dashboard.approx_gestation', weeks: patient.last_menstrual_period_at_appt_weeks, days: patient.last_menstrual_period_at_appt_days) %>
-      </div>
-    </div>
-
-    <div class="row">
-      <div class="col-4">
-        <%= f.text_field :primary_phone,
-                         value: patient.primary_phone_display,
-                         label: t('patient.dashboard.phone'),
-                         autocomplete: 'off' %>
-      </div>
-
-      <div class="col">
-        <%= f.text_field :pronouns,
-                         value: patient.pronouns,
-                         autocomplete: 'off' %>
-      </div>
-
-      <div class="col">
-        <div class="form-group">
-          <label for="status"><%= t 'patient.shared.status' %> <%= tooltip_shell status_help_text(patient) %></label>
-          <input type="text" value="<%= patient.status %>" class="form-control form-control-plaintext" id="patient_status_display" autocomplete="off" disabled>
-        </div>
-      </div>
-
-      <div class="col-3">
-        <% if current_user.admin? %>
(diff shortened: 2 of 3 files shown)


lomky:
I will get this into sandbox and tested shortly!

Fixes #3240

test/system/update_patient_info_test.rb
```diff
@@ -28,7 +28,6 @@ class UpdatePatientInfoTest < ApplicationSystemTestCase
     describe 'updating name' do
       before do
         fill_in 'First and last name', with: 'Susie Everyteen 2'
-        click_away_from_field
```
colinxfleming, line 31: honestly dope

app/javascript/src/components/PatientDashboardForm.jsx
```diff
@@ -44,9 +43,9 @@ export default PatientDashboardForm = ({
     }
   }
 
-  const debouncedAutosave = useMemo((params) => {
-    return debounce(autosave, 300)
-  }, []);
+  const debouncedAutosave = (params) => {
+    return debounce(autosave(params), 300)
```
colinxfleming, line 47: this works like a charm for me fysa

app/views/patients/_patient_dashboard.html.erb
```diff
@@ -1,77 +1,12 @@
...
-      </div>
-    </div>
-  <% end %>
-</div>
+  <%= render ReactComponent.new("PatientDashboardForm", raw_props: { 
+    patient: patient.as_json, 
+    weeksOptions: weeks_options.map { |opt| { label: opt[0], value: opt[1] } },
+    daysOptions: days_options.map { |opt| { label: opt[0], value: opt[1] } },
+    initialCallDate: patient.initial_call_date.strftime("%m/%d/%Y"),
+    statusHelpText: status_help_text(patient),
```
colinxfleming, line 7: at some point we'll need to find a way to re-rack this from changes (easiest example on main: setting an appointment date makes the status `Fundraising`, removing appointment dates makes the status `Needs Appointment`. That said I'm perfectly fine letting this regress a bit and merging this, and tackling this later, because I'd like to start running this react code in prod again. (Please make an issue if we do that so we don't forget.)

app/javascript/src/components/PatientDashboardForm.jsx
```diff
@@ -26,7 +26,6 @@ export default PatientDashboardForm = ({
 
   const autosave = async (updatedData) => {
     const updatedPatientData = { ...patientData, ...updatedData }
-    setPatientData(updatedPatientData)
```
colinxfleming, line 29: this is the main culprit right?

colinxfleming (commented):
this is looking pretty good to me. I want to have @lomky take a look at it real quick but then I'm comfortable rolling with it. One minor regression noted in a comment but (as noted) I'm good to roll with this.

Sorry again for the review delay!

app/javascript/src/components/PatientDashboardForm.jsx
mercedesb: actually the change below with the debounce was the main culprit. a combo of the useMemo and not passing the params through to the autosave function

app/views/patients/_patient_dashboard.html.erb
mercedesb: I'm not totally following your comment but I think that's b/c I'm not as familiar with all the parts of the app as you are. Can you clarify by what you mean by "this"? The status help text? Or the status input itself?

colinxfleming: yeah, the status help text -- the content changes based on the state of the patient (and I believe that's a model field right now). On prod now it refreshes after a change without a page reload.

app/javascript/src/components/PatientDashboardForm.jsx
colinxfleming: ah, good to know! makes sense

mercedesb: I really need to learn to just stop reaching for useMemo. I mess it up nearly every time and it has cost me hours of my life 😅

app/views/patients/_patient_dashboard.html.erb
mercedesb: cool cool, I went ahead and fixed that in this branch 🤗 hopefully no regressions now

mercedesb:
Looks like I'm running into test failures due to the debounced save and ajax stuff with Capybara. I'll continue looking into this this week.

lomky:
I'm seeing a small bug when I try to set "0 days" in the weeks along settings, it blanks it out. I suspect its something to do with the javascript -> ruby translation with "whats a false", but I haven't dug in more

mercedesb:
> I'm seeing a small bug when I try to set "0 days" in the weeks along settings, it blanks it out. I suspect its something to do with the javascript -> ruby translation with "whats a false", but I haven't dug in more

Great, I'll add that to my list when I'm resolving the specs issue

mercedesb:
> I'm seeing a small bug when I try to set "0 days" in the weeks along settings, it blanks it out

Looks like this is what currently happens on main and is not a regression introduced by this branch

mercedesb:
😮‍💨 ok I think we're good! finally. sorry for the avalanche of notifications. I've addressed all feedback
- dynamic status text regression
- blank '0 days' regression
- spec failures
- a host of mini bugs that we're not gonna talk about

have a great week!

lomky:
All the functionality is looking great! I think we have one last visual to fix where the Days Along field is shifting up, but otherwise it seems to be working great for me and i'm so excited to get it merged!

in the old component, it had an invisible label pushing it down to be level with Weeks Along.

main:
image

branch:
image

mercedesb:
Visual regression fixed.

I've rebased this branch on latest main but am getting a security vulnerability that needs to be fixed (the `json` gem) as well as specs related to auth which shouldn't be from anything this branch touches.

I'm going to let this be and hope that those issues are resolved in the base branch. Let me know if you need me to rebase or retrigger workflows this week.

colinxfleming:
Popping in here to say re: build, there's currently SOMETHING up with [github actions and capybara](https://github.com/teamcapybara/capybara/issues/2800), this results in about 10 system test failures and we can ignore those for now. I also bumped the json gem on main, so if the regression is solved for I think we can smash merge even if build's not passing.

mercedesb:
@lomky 

Re: "Field Changes sometimes 'saved' but then lost", I currently have a 300 ms debounce on the input. I'm not able to reproduce what you've described. But I can sometimes force a page reload faster than the 300 ms if I'm realllllllyyyyy trying. Could that be something you're running into? Seeing the green banner from a previous save and reloading the page before the new change can be committed? Want me to decrease the 300 ms debounce on the autosave?

Re: Days Along off-by-one
Can you clarify your screenshot? I'm trying to guess what it's supposed to say based on the timestamp of when you left your comment but I still don't know which part of the screenshot is correct and which isn't. Could you let me know what you expect to see in which box and where it's wrong? That would save me a lot of time tracking down whatever could be happening.

app/javascript/src/components/Input.jsx
```diff
@@ -15,7 +15,15 @@ export default Input = ({
...
 }) => {
-  const [value, setValue] = useState(initialValue || "")
+  // if we don't have an onChange handler, this component is an uncontrolled component
+  // if we have an onChange handler, we'll manage state like a traditional controlled component
+  let value, setValue;
+  if (!onChange) {
+    value = initialValue
+  } else {
+    ([value, setValue] = useState(initialValue || ""))
+  }
```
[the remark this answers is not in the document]
mercedesb: This is a good point and something I went back and forth on a lot. I don't want to mutate the state variable directly either since that also breaks React conventions.

I've pushed up a more complex but more complete solution in today's commits.

lomky:
Re: "Field Changes" - I'm not trying especially hard, just typing and tabbing through fields! but that is pretty fast.

Re: Days Along off-by-one - The value selected is 2, but the display (see the weeks along below) says 3. That's true of the other places the value showed, too. But when I checked the console, it was saved properly as 2. If I changed the value, the display _also_ changed, but still off by one. it was a weird bug and I couldn't reproduce it.

For overall context, I spent about 5-10 minutes just naviating around, typing, and changing things, and saw the off-by-one once and field change loss two or three times. So it wasn't even the most common occurence, but too often for my comfort, if that makes sense?

mercedesb:
I _think_ bugs are fixed? 

I'm getting `Selenium::WebDriver::Error::SessionNotCreatedError` errors in the specs which seem to be affecting all the recent PRs so those failures aren't introduced by these changes.

mercedesb:
Going to close this stale PR

PR #3356, opened 2025-02-24
Move rails secrets to rails credentials, per deprecation

### LINKED ISSUE #3256: Migrate from `secrets` to `credentials`

* 
Right now we use the (quite old) style [secrets](https://github.com/DARIAEngineering/dcaf_case_management/blob/main/config/secrets.yml) file. Let's move over to using the new-style `configs/credentials/` files
 
Keeping up with Rails Norms

After a bit of research, let's follow the per-env key migration route. That best matches our current setup.

[This article has a good step by step](https://www.stefanwienert.de/blog/2023/10/24/rails-7-1-migrate-from-secrets-yml-to-credentials-enc-yml-short/).

## Development & Test
- create a `development.key` & `test.key`
- run the rails commands to set up each of their credential files
- remove the secrets file & update the .gitignore
- replace any reference to `Rails.application.secrets` with `Rails.application.credentials`
- commit these keys, creds, and code changes to the repo
- confirm the app still runs successfully locally
  - via docker
  - via bare-metal install
  - in test runs 

## Production
Either @colinxfleming or @lomky should handle this.

- generate the `production.key`
- run the rails commands to set up its corresponding credential file
- commit _only_ the credential file to the repo
- in Heroku, set RAILS_MASTER_KEY to the production key
- confirm functionality in Sandbox
- release to production

## Final cleanup
- delete the old secrets file

 
* **Anything else?**

S/o to the work done in #3128 & #3220 for kicking this off!

colinxfleming:

Minimal steps necessary to get us onto credentials and off of secrets. (No opinion here on whether we should use credentials full on - this just makes it so we can ditch the secrets file.)

* remove secrets file
* add a bunch of credentials files

no view changes

* Fixes #3256 

For reviewer:
* Tag this PR:


colinxfleming:
I've tested this on staging and I believe it works (my smoke test strategy was logging in, creating a note, and checking that clinic data was still readable). I think we can punch this when we're ready but @lomky if you wouldn't mind taking a gander I'd appreciate that.

lomky (approved):
Looks perfect!

PR #3372, opened 2025-03-19
Updated - Improve the configuration model

lomky:

Pulls in the work done in #2921. Thank you @Nubbify!

- Bumps #1559 

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -383,7 +383,7 @@ class Config < ApplicationRecord
 
     def validate_yes_or_no
       # allow yes or no, to be nice (technically only yes is considered)
-      if !options.last =~ /\A(yes|no)\z/i
+      if !(options.last =~ /\A(yes|no)\z/i)
         "Field must be either 'yes' or 'no'"
       end
     end
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -43,6 +43,30 @@ class Config < ApplicationRecord
     :voicemail,
   ]
 
+  DEFAULTS = {
+    insurance: nil,
+    external_pledge_source: nil,
+    pledge_limit_help_text: nil,
+    language: nil,
+    resources_url: nil,
+    practical_support_guidance_url: nil,
+    fax_service: nil,
+    referred_by: nil,
+    practical_support: nil,
+    hide_practical_support: false,
+    start_of_week: 'monday',
+    budget_bar_max: 1_000,
+    voicemail: 12,
+    days_to_keep_fulfilled_patients: 90,
+    days_to_keep_all_patients: 365,
+    shared_reset_days: 6,
+    hide_budget_bar: false,
+    aggregate_statistics: false,
+    hide_standard_dropdown_values: false,
+    county: nil,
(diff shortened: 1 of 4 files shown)


app/models/config.rb
```diff
@@ -343,38 +374,45 @@ def validate_start_of_week
...
+        "Timezone provided is not supported"
+      end
     end
 
     ### Practical support
 
     def validate_yes_or_no
       # allow yes or no, to be nice (technically only yes is considered)
-      options.last =~ /\A(yes|no)\z/i
+      if !options.last =~ /\A(yes|no)\z/i
```
colinxfleming: This one doesn't quite work perfectly -- entering in a weird value like `yeah` or `meow` into `Hide Practical Support Options` will yield a 500 error as follows

```
12:00:53 web.1  | [INFO ] Completed 500 Internal Server Error in 9ms (ActiveRecord: 1.1ms (3 queries, 0 cached) | GC: 0.0ms)
12:00:53 web.1  | [FATAL]   
12:00:53 web.1  | NoMethodError (undefined method `=~' for false:FalseClass):
12:00:53 web.1  |   
12:00:53 web.1  | app/models/config.rb:386:in `validate_yes_or_no'
12:00:53 web.1  | app/models/config.rb:294:in `call'
12:00:53 web.1  | app/models/config.rb:294:in `block in validate_config'
12:00:53 web.1  | app/models/config.rb:294:in `map'
12:00:53 web.1  | app/models/config.rb:294:in `validate_config'
12:00:53 web.1  | app/controllers/configs_controller.rb:14:in `update'
```

app/models/config.rb
```diff
@@ -43,6 +43,30 @@ class Config < ApplicationRecord
...
   ]
 
+  DEFAULTS = {
+    insurance: nil,
+    external_pledge_source: nil,
+    pledge_limit_help_text: nil,
+    language: nil,
+    resources_url: nil,
+    practical_support_guidance_url: nil,
+    fax_service: nil,
```
colinxfleming, line 53: ruby's hash accessor will return nil if it doesn't find a key, so if we wanted to take out the nils here we could. I'm fine either way

app/models/config.rb
```diff
@@ -343,38 +374,45 @@ def validate_start_of_week
...
+        "Timezone provided is not supported"
+      end
     end
 
     ### Practical support
 
     def validate_yes_or_no
       # allow yes or no, to be nice (technically only yes is considered)
-      options.last =~ /\A(yes|no)\z/i
+      if !(options.last =~ /\A(yes|no)\z/i)
```
colinxfleming, line 386: made a very slight adjustment to this to solve for an order-of-operations problem -- `!options.last =~ regex` was casting options.last to a boolean and then trying to string match that, so it was causing failures. Wrapping in parens solves for it

colinxfleming (commented):
this looks good from my perspective -- made a very slight adjustment to solve for a bug but good otherwise. I'm gonna merge!

colinxfleming:
godmode -- I exercised these options and they're all looking good

PR #3376, opened 2025-03-26
Implement workaround for teamcapybara/capybara#2800

xmunoz:
It's not registering the new driver:
```
ArgumentError: unknown driver: :local_selenium_chrome_headless
```


xmunoz:
I'm officially giving up on this. We either have a really non-standard capybara implementation, or I have no idea what I'm doing.

PR #3391, opened 2025-05-06
Split build tests into unit/sys

colinxfleming:

This started as a run at capybara lockstep but I think we'd really benefit from adjusting our build to split out unit tests (fast, still working) from system tests (slow, not really working because of chrome changes).

* Split build into system and unit tests
* Bump libraries to get static scanners back to green

no prod changes other than library bumps
no issues really 

For reviewer:
* Tag this PR:


colinxfleming:
godmode since this is mostly build related

PR #3393, opened 2025-05-07
Bugfix: Standardize shared_reset_days key

lomky:

* Move all the keys to a single name, `shared_reset_days`

For reviewer:
* Tag this PR:


app/models/config.rb
```diff
@@ -230,9 +230,9 @@ def self.archive_all_patients
   end
 
   def self.shared_reset_days
-    shared_reset_days = Config.find_or_create_by(config_key: 'shared_reset').options.try :last
+    shared_reset_days = Config.find_or_create_by(config_key: 'shared_reset_days').options.try :last
     # default 6 days
-    shared_reset_days ||= DEFAULTS[:shared_reset]
```
lomky, line 235: This was the bug. The key in the defaults was already `:shared_reset_days`

xmunoz (approved):

PR #3403, opened 2025-05-14
Fix Number and URL config validations

lomky:

* fixes the Numeric check for configs
* fixes the URL check for configs
* updates other information for configs in comments

Testing:
- Checked that every type of config validation is working properly in the GUI

For reviewer:
* Tag this PR:

### DIFF
--- a/.github/workflows/test_dev_env.yml
+++ b/.github/workflows/test_dev_env.yml
@@ -24,7 +24,7 @@ jobs:
           docker compose logs -t
           docker ps -a | (! grep Exited ) # Return a non-zero exit code if any of the containers are stopped
 
-  build:
+  unit_tests:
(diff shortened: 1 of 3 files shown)


app/models/config.rb
```diff
@@ -408,6 +408,12 @@ def validate_shared_reset_days
       end
     end
 
+    def validate_budget_bar_max
+      if !validate_number
```
colinxfleming, line 412: When I put `nah` in here it lets me, though it seems to evaluate to zero -- I think this is because `validate_number` produces a string if something's up. That said it doesn't break anything on the frontend thankfully

colinxfleming (approved):
one thing to check but then go for it , thx kat

PR #3406, opened 2025-05-26
Fix validators and unit test failures

colinxfleming:

We've had a number of unit tests failing in addition to the system tests, and a recent PR split out the unit tests and system tests to more easily identify what was failing. This brings things back to green by the following methods enumerated below. I'm slightly wary of changes to archive patient and so I intend to test on staging but should be fine.

* Fixes some bs test failures caused by openstruct not being available
* fixes `validate_number` such that it follows the new 'return error string on invalid` approach
* fixes `validate_url` such that it follows the new `return error string on invalid` approach
* fixes a few sub-validators to accommodate changes to `validate_number` and `validate_url`

no view changes
no issues

For reviewer:
* Tag this PR:

### DIFF
--- a/app/models/config.rb
+++ b/app/models/config.rb
@@ -278,8 +278,9 @@ class Config < ApplicationRecord
       cleaners.each { |cleaner| method(cleaner).call }
     end
 
-    # parent function. will handle errors; child validators should return true
-    # if value is valid for key, and false otherwise.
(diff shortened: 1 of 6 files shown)


colinxfleming:
fairly confident that this won't do anything destructive -- firing

Issue #2682, opened 2022-08-29
Encryption/deletion/protection of data in case of emergencies


lwaldsc:
Thanks for creating an issue! Please fill out this form so we can be sure to have all the information we need, and to minimize back and forth.

* **What are we trying to do?**
- Encrypt data (probably related to #413) and/or ensure regularly occurring deletion of data 
- Ensure funds are confident in DARIA's ability to protect their patients' data in case of legal action AND limit DARIA's liability/ability to retrieve patient data in case of legal action
- Issue is just documenting that this has been brought up in discussions with Funds during demo and onboarding

* **What feature or behavior is this required for?**
- Protecting patient data and funds from legal actions

* **How could we solve this issue? (Not knowing is okay!)**
- No idea. Regularly nuking from orbit?

* **Anything else?**

mercedesb:
@colinxfleming Is this related to #413? Just looking for a good next ticket and found some threads that seem similar

PR #3516, opened 2026-04-03
Encrypt Patient PII fields at the application level with Active Record Encryption

### LINKED ISSUE #413: Use attr_encrypted for protecting sensitive data at rest?

Rolling2405:

This encrypts sensitive patient information (names, phone numbers, addresses) so that personal data is protected at rest in the database. Anyone with raw database access can no longer read patient PII in plain text.

* Encrypt 9 Patient PII fields (name, phone, email, city, state, etc.) using Rails Active Record Encryption
* Rewrite patient search to work with encrypted fields (in-memory Ruby filtering instead of SQL `ilike`/`like`)
* Drop database indexes on encrypted non-deterministic columns
* Configure PaperTrail to skip PII fields on Patient so audit logs don't store plain-text copies
* Add `encryption:backfill_patients` rake task (tenant-scoped) to encrypt existing patient records

**Notes:**
* **After merging**, run `rake encryption:backfill_patients` to encrypt existing patient data in each fund.
* **Merge order:** This PR is independent, but must merge before #3517 (PaperTrail scrub depends on encryption being in place). Merge order: `#3516` → `#3517`.
* **Breaking change:** Patient search switches from SQL to in-memory filtering, which may affect search performance on very large datasets.

* Fixes #413
* Bumps #2682

For reviewer:
* Tag this PR:

### DIFF
--- a/app/lib/paper_trail_version.rb
+++ b/app/lib/paper_trail_version.rb
@@ -21,16 +21,22 @@ class PaperTrailVersion < PaperTrail::Version
     can_fulfill_id
     can_support_type
     can_support_id
+  ].freeze
+
+  # Patient PII fields whose VALUES are stripped from version records but
+  # whose KEYS are still recorded so the audit trail shows that the field
+  # changed (just without storing plaintext PII in the versions table).
+  REDACTED_PATIENT_FIELDS = %w[
     name
     primary_phone
     other_phone
     other_contact
     other_contact_relationship
     city
-    state
-    county
-    zipcode
   ].freeze
+  REDACTED_PLACEHOLDER = '[REDACTED]'.freeze
+
+  before_save :redact_patient_pii
   DATE_FIELDS = %w[
     appointment_date
     initial_call_date
@@ -73,6 +79,29 @@ class PaperTrailVersion < PaperTrail::Version
   end
 
   private
+
+  # Replace plaintext PII values with [REDACTED] before persisting the
+  # version. Keeps the audit signal (which field changed, when, by whom)
+  # while keeping ciphertext-equivalent privacy in the versions table.
+  def redact_patient_pii
+    return unless item_type == 'Patient'
+
+    if object.is_a?(Hash)
+      REDACTED_PATIENT_FIELDS.each do |field|
+        object[field] = REDACTED_PLACEHOLDER if object[field].present?
+      end
+    end
+
+    return unless object_changes.is_a?(Hash)
+
+    REDACTED_PATIENT_FIELDS.each do |field|
+      next unless object_changes.key?(field)
+
+      object_changes[field] = Array(object_changes[field]).map do |value|
+        value.present? ? REDACTED_PLACEHOLDER : value
+      end
+    end
+  end
 
   def format_fieldchange(key, value)
     shaped_value = if value.blank?
--- a/app/lib/paper_trail_version.rb
+++ b/app/lib/paper_trail_version.rb
@@ -21,6 +21,15 @@ class PaperTrailVersion < PaperTrail::Version
     can_fulfill_id
     can_support_type
     can_support_id
+    name
+    primary_phone
+    other_phone
+    other_contact
+    other_contact_relationship
+    city
+    state
+    county
+    zipcode
   ].freeze
   DATE_FIELDS = %w[
     appointment_date
--- a/app/models/concerns/patient_searchable.rb
+++ b/app/models/concerns/patient_searchable.rb
@@ -20,10 +20,11 @@ module PatientSearchable
       base = Patient
       base = base.where(line: lines) if lines
 
+      # find_each ignores .order() and iterates by primary key, so we
+      # collect all matches and sort in Ruby to preserve the original
+      # "most recently updated first" contract.
       matches = []
-      base.order(updated_at: :desc).find_each(batch_size: 100) do |patient|
-        break if search_limit.present? && matches.size >= search_limit
-
+      base.find_each(batch_size: 100) do |patient|
         name_match = patient.name&.downcase&.include?(search_term) ||
                      patient.other_contact&.downcase&.include?(search_term) ||
                      patient.identifier&.downcase&.include?(search_term)
@@ -36,6 +37,8 @@ module PatientSearchable
         matches << patient if name_match || phone_match
       end
 
+      matches.sort_by { |p| p.updated_at || Time.at(0) }.reverse!
+      matches = matches.first(search_limit) if search_limit.present?
       matches
     end
   end
--- a/app/models/concerns/patient_searchable.rb
+++ b/app/models/concerns/patient_searchable.rb
@@ -8,11 +8,11 @@ module PatientSearchable
     # Case insensitive and phone number format agnostic!
     # pass `search_limit: nil` for no limit.
     #
-    # Search is performed in-memory after loading records because
+    # Search is performed in-memory after decrypting records because
     # Patient PII fields are encrypted with ActiveRecord Encryption.
     # Encrypted fields cannot be searched with SQL LIKE/ILIKE, so we
-    # decrypt and filter in Ruby. This is performant for DARIA's
-    # dataset size (typically a few hundred active patients per line).
+    # decrypt and filter in Ruby. Records are loaded in batches via
+    # find_each to avoid pulling the entire table into memory at once.
     def search(name_or_phone_str, lines: nil, search_limit: DEFAULT_SEARCH_LIMIT)
       search_term = name_or_phone_str.downcase
       clean_phone = name_or_phone_str.gsub(/\D/, '')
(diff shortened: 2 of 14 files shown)


Rolling2405:
@colinxfleming Can you please take a look at this PR and merge it if it looks good when you get a chance?

Rolling2405:
> @colinxfleming Can you please take a look at this PR and merge it if it looks good when you get a chance?

Or @lomky If you could please review this so that it can be merged that would be great also!

colinxfleming:
Hey, thanks much for the PR. Wanted to let you know that I saw it and it’s
on my to do list!

Rolling2405:
@colinxfleming Just wanted to give you a gentle reminder to please review and merge the PR if it looks good.

app/models/concerns/patient_searchable.rb
```diff
@@ -7,24 +7,38 @@ module PatientSearchable
...
-        clean_phone = "%#{clean_phone}%"
-        matches = matches.or(base.where('primary_phone like ?', clean_phone))
-                         .or(base.where('other_phone like ?', clean_phone))
+      base = base.where(line: lines) if lines
+
+      # find_each ignores .order() and iterates by primary key, so we
+      # collect all matches and sort in Ruby to preserve the original
+      # "most recently updated first" contract.
+      matches = []
+      base.find_each(batch_size: 100) do |patient|
```
colinxfleming, line 27: This is an interesting idea. My operating assumption is that a ruby in-memory sort wasn't going to be a scaleable solution, but it _is_ true that most of our funds don't have super high volumes such that an in memory sort seems plausible to me. Given that we're running on fairly small hardware (both in terms of servers and database size) is there a point at which you think the approach of iterating thru the full set of a fund's patients records stops being viable? (assuming I'm reading this code right)

app/models/patient.rb
```diff
@@ -2,7 +2,20 @@
...
+  encrypts :name
+  encrypts :primary_phone, deterministic: true
+  encrypts :other_phone
+  encrypts :other_contact
+  encrypts :other_contact_relationship
+  encrypts :city
+  encrypts :state
+  encrypts :county
+  encrypts :zipcode
+
```
colinxfleming, line 15: I'm not quite ready to say that we should definitely encrypt these -- we don't search by these and I'm not sure the tradeoff is worth it to lose the audit trail. I'll give this some more thought but for now I'd prefer taking it out to limit the blast radius of this pr. 

My other question is why only primary_phone is deterministic -- should we be thinking about `name`, `other_phone`, and `other_contact` also being good candidates for this?

app/lib/paper_trail_version.rb
```diff
@@ -21,6 +21,15 @@ class PaperTrailVersion < PaperTrail::Version
...
     can_support_id
+    name
+    primary_phone
+    other_phone
+    other_contact
+    other_contact_relationship
+    city
+    state
+    county
+    zipcode
```
colinxfleming, line 32: I think I'd strongly prefer to know that these got changed even if we don't have exact data on from what to what -- can we keep these in paper trail?

colinxfleming (changes requested):
thanks much for thinking through this, your thoughtfulness (and patience) is appreciated. I have one major question about the scalability of this approach, and I have some requests about limiting the blast a bit. but I think once I have an answer to the scalability questions + a smaller blast radius I think we can give this a shot in staging. 

Separately, I'm curious about how we got on your radar and your methodology for these PRs, so if you wouldn't mind saying more about that (or emailing me @ c3flemin @ gmail) when you get a sec I'd appreciate it.

app/models/concerns/patient_searchable.rb
Rolling2405: Great question — this was a deliberate tradeoff and I did stress-test the assumption.

The reasoning: DARIA is multi-tenant via `acts_as_tenant :fund`, search is further scoped to the current line, and the archival pipeline (90 days post-audit, 365 days total) keeps active patients low. Realistic working set per query is roughly 100–1000 patients per fund's active line, and results are capped at 15.

Estimated performance: ~200–600ms vs ~5–20ms for SQL `ilike` — slower but well within acceptable UX for a search box, and the user already waits for autocomplete debounce.

Where it stops being viable: I'd say comfortably up to ~5,000 active patients per line. Above ~10–15k it'd start feeling sluggish and I'd want to revisit (probably with `blind_index` for exact-match fields, or accept partial-match limitations).

I considered the alternatives (`blind_index`, `pg_enquo`, n-gram tokenization, CipherStash) but they each had a tradeoff: exact-match only, no substring support, or paid SaaS. Pure in-memory filtering preserves the existing fuzzy search behavior with zero functional change for users — that felt like the right call for DARIA's current scale.

Happy to add a comment in the code documenting the scale assumption, and/or add a TODO for revisiting if any fund grows past ~5k active patients.

Rolling2405:
> thanks much for thinking through this, your thoughtfulness (and patience) is appreciated. I have one major question about the scalability of this approach, and I have some requests about limiting the blast a bit. but I think once I have an answer to the scalability questions + a smaller blast radius I think we can give this a shot in staging.
> 
> Separately, I'm curious about how we got on your radar and your methodology for these PRs, so if you wouldn't mind saying more about that (or emailing me @ c3flemin @ gmail) when you get a sec I'd appreciate it.

I'm passionate about abortion rights and reproductive rights in general and so I was just searching for open source projects that I could contribute to that align with this and that's how I found this project. For the PRs, I used Github Copilot to help me with them.

app/models/patient.rb
Rolling2405: Good points on both — happy to scope this down.

**On limiting blast radius**: I've dropped encryption from `state`, `county`, and `zipcode` in this PR (commit 1b431d31). They're geographic/locality fields that are less sensitive on their own than direct identifiers, and dropping them preserves PaperTrail visibility for those changes too. Encryption for those can come back in a follow-up PR if/when we want it. Keeping encryption on the direct identifiers: `name`, `primary_phone`, `other_phone`, `other_contact`, `other_contact_relationship`, and `city`.

**On why only `primary_phone` is deterministic**: this was driven by the `confirm_unique_phone_number` validation in `patient.rb` (~line 156), which runs `Patient.where(primary_phone: primary_phone).first`. That SQL equality query (and the unique DB index on `[primary_phone, fund_id]`) only works with deterministic encryption. I followed the same pattern as the existing `Clinic` model, where `name` is deterministic but the address fields are non-deterministic.

For `name`, `other_phone`, `other_contact` — non-deterministic was a deliberate security choice. Non-deterministic encryption is cryptographically stronger (no frequency analysis on ciphertext), and since search is in-memory now, I don't need SQL equality on those columns. Making them deterministic would weaken security with no functional benefit. That said, if you'd like exact-match SQL on any of them for some other reason, I'm open to flipping them.

app/lib/paper_trail_version.rb
Rolling2405: Fair point — totally agree the audit signal of "this field changed" is valuable even when the value itself shouldn't be stored. I went with the simplest "skip entirely" approach to minimize plaintext PII in the `versions` table, but you're right that loses information.

Switched to redacting the values instead of skipping the fields (commit 1b431d31). PaperTrail still records that the field was changed (with timestamp + user), but the `object` and `object_changes` JSON store `[REDACTED]` in place of the actual values. Implementation is a `before_save` callback on `PaperTrailVersion` that walks the JSON columns and replaces values for the 6 PII fields whenever `item_type == 'Patient'`. That keeps the audit trail intact and still ensures no plaintext PII is stored in version history. Tests assert both that the change is recorded AND that no plaintext leaks into the version row.

lomky:
> > thanks much for thinking through this, your thoughtfulness (and patience) is appreciated. I have one major question about the scalability of this approach, and I have some requests about limiting the blast a bit. but I think once I have an answer to the scalability questions + a smaller blast radius I think we can give this a shot in staging.
> > Separately, I'm curious about how we got on your radar and your methodology for these PRs, so if you wouldn't mind saying more about that (or emailing me @ c3flemin @ gmail) when you get a sec I'd appreciate it.
> 
> I'm passionate about abortion rights and reproductive rights in general and so I was just searching for open source projects that I could contribute to that align with this and that's how I found this project. For the PRs, I used Github Copilot to help me with them.

Hi @Rolling2405! We really appreciate your desire to help out with ReproTech. As you probably can tell, we can't really handle this level of PR influx. It's been a good exercise in reassessing how we want to approach things as a project.

If you're still interested in engaging at a speed we can match, please grab one issue in particular and start a conversation there. Once we are all on the same page with how we want to solve it, then we can move to PRs. For our project in particular, some features are implemented, or not implemented, in very intentional ways.

Either way, thank you for your time and thank you for working towards reproductive justice!

PR #3518, opened 2026-04-06
Clean up incomplete MFA registrations in nightly task (#2848)

### LINKED ISSUE #2848: Add a nightly cleanup task for incomplete multifactor registrations

Keep a clean database by removing defunct, incomplete MFA registrations

Finalizing #2819 

Add a line to our [Nightly Cleanup Task](https://github.com/DARIAEngineering/dcaf_case_management/blob/main/lib/tasks/nightly_cleanup.rake) Querying the AuthFactor's that are `registration_complete` `false` and old enough to be past becoming validated (an hour old or more, maybe?), and deleting them.

* **Anything else?**
Has to wait for #2819 to be merged!

Rolling2405:

This adds a nightly background job that automatically cleans up MFA registrations that were started but never completed, preventing stale security records from piling up in the database.

* Add `CleanupIncompleteMfaJob` that deletes incomplete MFA registrations older than 24 hours
* Register the job in `config/recurring.yml` to run nightly at 3am
* Job is tenant-scoped (runs per fund via `Fund.all.each`)

* Fixes #2848

For reviewer:
* Tag this PR:

### DIFF
--- a/lib/tasks/nightly_cleanup.rake
+++ b/lib/tasks/nightly_cleanup.rake
@@ -28,7 +28,7 @@ task nightly_cleanup: :environment do
       User.find_each do |user|
         user.auth_factors
             .where(registration_complete: false)
-            .where('created_at < ?', 1.hour.ago)
+            .where('created_at < ?', 48.hours.ago)
             .destroy_all
       end
       puts "#{Time.now} -- cleaned up incomplete MFA registrations for fund #{fund.name}"
--- a/lib/tasks/nightly_cleanup.rake
+++ b/lib/tasks/nightly_cleanup.rake
@@ -24,6 +24,11 @@ task nightly_cleanup: :environment do
       User.disable_inactive_users
       puts "#{Time.now} -- locked accounts of users who have not logged in since #{User::TIME_BEFORE_DISABLED_BY_FUND.ago} for fund #{fund.name}"
 
+      AuthFactor.where(registration_complete: false)
+                .where('created_at < ?', 1.hour.ago)
+                .destroy_all
+      puts "#{Time.now} -- cleaned up incomplete MFA registrations for fund #{fund.name}"
+
       Patient.trim_shared_patients
       puts "#{Time.now} -- trimmed shared patients for fund #{fund.name}"
 
--- a/lib/tasks/nightly_cleanup.rake
+++ b/lib/tasks/nightly_cleanup.rake
@@ -24,9 +24,13 @@ task nightly_cleanup: :environment do
       User.disable_inactive_users
       puts "#{Time.now} -- locked accounts of users who have not logged in since #{User::TIME_BEFORE_DISABLED_BY_FUND.ago} for fund #{fund.name}"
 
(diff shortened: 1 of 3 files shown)


lib/tasks/nightly_cleanup.rake
```diff
@@ -18,12 +18,21 @@ task nightly_cleanup: :environment do
...
       puts "#{Time.now} -- cleared all recently reached patients from call lists for fund #{fund.name}"
 
       User.disable_inactive_users
       puts "#{Time.now} -- locked accounts of users who have not logged in since #{User::TIME_BEFORE_DISABLED_BY_FUND.ago} for fund #{fund.name}"
 
+      # AuthFactor is not acts_as_tenant scoped, so query through users
+      User.find_each do |user|
+        user.auth_factors
+            .where(registration_complete: false)
+            .where('created_at < ?', 1.hour.ago)
```
colinxfleming, line 31: probably not much of a difference in general, but
```suggestion
            .where('created_at < ?', 48.hours.ago)
```

lib/tasks/nightly_cleanup.rake
```diff
@@ -18,12 +18,21 @@ task nightly_cleanup: :environment do
 
   Fund.all.each do |fund|
     ActsAsTenant.with_tenant(fund) do
-      User.all.each { |user| user.clean_call_list_between_shifts }
+      User.find_each { |user| user.clean_call_list_between_shifts }
```
colinxfleming, line 21: why this diff?

lib/tasks/nightly_cleanup.rake
```diff
@@ -18,12 +18,21 @@ task nightly_cleanup: :environment do
...
       puts "#{Time.now} -- cleared all recently reached patients from call lists for fund #{fund.name}"
 
       User.disable_inactive_users
       puts "#{Time.now} -- locked accounts of users who have not logged in since #{User::TIME_BEFORE_DISABLED_BY_FUND.ago} for fund #{fund.name}"
 
+      # AuthFactor is not acts_as_tenant scoped, so query through users
+      User.find_each do |user|
+        user.auth_factors
```
colinxfleming, line 29: to get out of a tenant, you can set a block `ActsAsTenant.without_tenant` if I recall correctly -- that would save you another run through the users table.

I also have a slim preference for this kind of thing as a model method to make it easier to run adhoc - unlikely here given the nature of it but if it's not a meaningful lift I'd appreciate that. Something like

```
class AuthFactor
  def self.clean_incomplete
    where(...)
  end
end
```

colinxfleming (changes requested):
Premise here looks good, but let's please use that ActsAsTenant.without_tenant trick and see if that saves us from having to iterate thru the full users table again. A few other notes but nothing too substantial

lib/tasks/nightly_cleanup.rake
Rolling2405: User.all.each loads all users into memory at once; find_each batches them
 in groups of 1000, which is safer as the user base grows.

PR #3529, opened 2026-04-08
Enable YJIT in production for 15-30% Rails performance improvement

Rolling2405:

This explicitly enables the YJIT just-in-time compiler in production, which typically delivers a 15–30% performance improvement for Rails applications with no code changes required.

* Add `config.yjit = true` to `config/environments/production.rb`
* Note: Rails 8.1+ enables YJIT by default; this is an explicit opt-in for clarity

For reviewer:
* Tag this PR:


colinxfleming:
nice to know that they're doing this by default, but my preference is that we follow rails defaults pretty tight so going to close this and follow unstated best practices the rails version du jour has (since by my read this is a noop)

PR #3613, opened 2026-09-07
Bump Essential Dependencies to green

lomky:

Green is the goal.

* Bump Ruby to 3.4.10
* `bundle update` on many deps
* `yarn up` on many deps


colinxfleming (approved):
doesn't look like any interesting diffs here except for the change in development.rb which obviously is not scary -- fire when ready

PR #3627, opened 2026-09-13
[Docker] Update and reorganize Dockerfile with caching in mind

lomky:

* Make sure the Dockerfile build is as cache-able as I know how to make it
* Update ENV to ARGS and MAINTAINER to LABELS
* Removed Ahab integration (RIP)


colinxfleming (approved):
premise makes sense to me fire when ready

